PR #215, opened 2014-09-04
Added autoClose to Navbar

pelonpelon:
## Problem:
Bootstrap allows for a class ( **in** ) to be added to **Nav**, causing **Nav** to collapse. Can be used with props.onToggle(). react-bootstrap introduces a new state ( **navOpen** ) that is hidden from props.onToggle() and prevents the optional bootstrap functionality.

## Current functionality:
With **Nav** open,
click on **NavItem**,
must click on **toggleButton** to close.

## Change:
With **Nav** open,
click on **NavItem**,
Nav closes automatically.

### default:
autoClose: true

#### Notes:
In Bootstrap, adding class ( **in** ) to Nav causes an abrupt closure which must be dealt with in further code. This solution elegantly causes Nav to close with the same animation that it opened with.

### DIFF
--- a/src/Navbar.jsx
+++ b/src/Navbar.jsx
@@ -26,7 +26,7 @@ var Navbar = React.createClass({
     onToggle: React.PropTypes.func,
     navExpanded: React.PropTypes.bool,
     defaultNavExpanded: React.PropTypes.bool,
-    autoClose: React.PropTypes.func
+    autoClose: React.PropTypes.bool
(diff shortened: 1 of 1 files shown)


mtscout6:
Can you add tests for this?

pelonpelon:
I'm sorry. That was months ago and I finished that react project. I won't have the time to review and write tests.

PR #239, opened 2014-09-27
Added getOverlayDOMNode option to OverlayTrigger

jsdir:
`getOverlayDOMNode` is an override prop for `OverlayTrigger` that allows one to optionally choose what DOM element to use when making positioning calculations.


mtscout6:
I can't say that I entirely understand what it is you are needing here. Could you include some tests and perhaps some updated documentation with an example of it's usage on the components page of the docs?

mtscout6:
Closing, will re-open when/if there's more input later on.

PR #293, opened 2014-11-27
Add a few additional checks to checkPosition in Affix

pbt:
I was using the affix component and I found that my component was uncontrollably flickering once I had scrolled past the affix's `offsetBottom` value. 

I looked at Bootstrap's JS for the affix plugin and found that there were a few additional checks that weren't present in this Affix component. Including them in `checkPosition` fixed the flickering problem for me.


mtscout6:
Your changes could also use some unit tests.

mtscout6:
Closing, will re-open when/if we get more input.

PR #304, opened 2014-12-09
Allow for setting container className on DropdownButton

wemcdonald:
Add <DropdownButton parentClassName="xxx"> to set the class on the containing <li> or <ButtonGroup>

### DIFF
--- a/src/DropdownButton.jsx
+++ b/src/DropdownButton.jsx
@@ -62,10 +62,15 @@ var DropdownButton = React.createClass({
         'dropup': this.props.dropup
       };
 
+    var className = classSet(groupClasses);
+    if (this.props.parentClassName) {
+      className += (' ' + this.props.parentClassName);
+    }
+
     return (
       <ButtonGroup
         bsSize={this.props.bsSize}
-        className={classSet(groupClasses)}>
+        className={className}>
         {children}
       </ButtonGroup>
     );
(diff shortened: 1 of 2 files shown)


src/DropdownButton.jsx
```diff
@@ -62,10 +62,15 @@ var DropdownButton = React.createClass({
         'dropup': this.props.dropup
       };
```
mtscout6: add `groupClasses[this.props.parentClassName] = true;` here instead of concatenating strings below.

src/DropdownButton.jsx
```diff
@@ -62,10 +62,15 @@ var DropdownButton = React.createClass({
...
 
+    var className = classSet(groupClasses);
+    if (this.props.parentClassName) {
+      className += (' ' + this.props.parentClassName);
```
mtscout6: See my comment above, to avoid concatenating the string.

src/DropdownButton.jsx
```diff
@@ -78,8 +83,13 @@ var DropdownButton = React.createClass({
         'dropup': this.props.dropup
       };
 
+    var className = classSet(classes);
+    if (this.props.parentClassName) {
+      className += (' ' + this.props.parentClassName);
```
mtscout6: Similar to my above comment on concatenating.

mtscout6:
`parentClassName` does not clearly indicate what it means, since `parent` indicates looking up from the current position. This is setting something in a child, so why not give a meaningful name to that child. How does `innerContainerClassName` sound?

mtscout6:
Closing, will reopen if/when we get more input.

PR #308, opened 2014-12-11
Update OverlayMixin to allow for a null overlay

mfunkie:
Use case:

Large table in which there is an item using OverlayMixin on every row.
If a `<span />` is returned by renderOverlay(), using React Dev Tools can be frustrating as the tree is littered with `<span />`s


mtscout6:
:+1:

mtscout6:
Thanks!

Issue #325, opened 2015-01-02
npm release v0.13.0 is raw source code


Dashed:
npm release v0.13.0 is raw source code. Thus, I can't `require('react-bootstrap')`.

At the moment only `v0.12` is viable.

mtscout6:
:+1: 

This could be cleaned up and automated with a pre-publish script as I proposed in #292.

mtscout6:
@Dashed 0.13.1 should be good to go :)

Issue #292, opened 2014-11-27
How do you manage this repo along with react-bootstrap-npm?


atulrpandey:
Do you manually copy over the src folder from here to react-bootstrap-npm project?

Thanks in advance. :)

mtscout6:
Have you considered using a pre-publish script defined in the package.json and forget the second repo?

See: https://docs.npmjs.com/misc/scripts

mtscout6:
I've been looking into this a bit this morning and there's a few considerations to think about here.

These types of issues keep cropping up (#325 #218), and they will likely continue to crop up.

However, copying everything in `cjs/` to the root of the repo is a non-starter since cleaning that up would be a nightmare.

What I'd like to do is rename `cjs/` to `lib/` which is more in line with other modules out there, including React. Then include a pre-publish script command.

This would be a breaking change though since you would need to change `require('react-bootstrap/...')` to `require('react-bootstrap/lib/...'`. Although `require('react-bootstrap').NavItem` would still work.

The benefit is that issues of the raw repo getting published to npm would go away, and you could remove the second repo.

mtscout6:
Fixed with #326

Issue #218, opened 2014-09-05
v0.11.1 npm common.js doesn't work


bovard:
## v0.11.0 works swell

```
LS2-00133:test bovardtiberi$ npm install react-bootstrap@0.11.0
npm WARN package.json react-bootstrap@0.11.1 No repository field.
npm http GET https://registry.npmjs.org/react-bootstrap/0.11.0
npm http 304 https://registry.npmjs.org/react-bootstrap/0.11.0
npm http GET https://registry.npmjs.org/react
npm http 304 https://registry.npmjs.org/react
npm http GET https://registry.npmjs.org/envify
npm http 304 https://registry.npmjs.org/envify
npm http GET https://registry.npmjs.org/esprima-fb
npm http GET https://registry.npmjs.org/xtend
npm http GET https://registry.npmjs.org/through
npm http GET https://registry.npmjs.org/jstransform
npm http 304 https://registry.npmjs.org/through
npm http 304 https://registry.npmjs.org/esprima-fb
npm http 304 https://registry.npmjs.org/xtend
npm http 304 https://registry.npmjs.org/jstransform
npm http GET https://registry.npmjs.org/object-keys
npm http GET https://registry.npmjs.org/base62/0.1.1
npm http GET https://registry.npmjs.org/source-map/0.1.31
npm http 304 https://registry.npmjs.org/source-map/0.1.31
npm http 304 https://registry.npmjs.org/object-keys
npm http 304 https://registry.npmjs.org/base62/0.1.1
npm http GET https://registry.npmjs.org/amdefine
npm http 304 https://registry.npmjs.org/amdefine
react@0.11.1 node_modules/react
└── envify@2.0.1 (through@2.3.4, esprima-fb@3001.1.0-dev-harmony-fb, xtend@2.1.2, jstransform@3.0.0)

react-bootstrap@0.11.0 node_modules/react-bootstrap
LS2-00133:test bovardtiberi$ node
> var Nav = require('react-bootstrap').Nav;
undefined
>
(^C again to quit)
>
```
## v 0.11.1 does not work :(

```
LS2-00133:test bovardtiberi$ rm -rf node_modules/
LS2-00133:test bovardtiberi$ npm install react-bootstrap@0.11.1
npm WARN package.json react-bootstrap@0.11.1 No repository field.
npm http GET https://registry.npmjs.org/react-bootstrap/0.11.1
npm http 304 https://registry.npmjs.org/react-bootstrap/0.11.1
[...]

PR #326, opened 2015-01-02
Add pre-publish script call, rename cjs/ -> lib/

mtscout6:
Addresses concerns like #325, #218, and any others like them that I have
missed. In particular this is the fix for an automated setup as
mentioned in #292.

__NOTE THIS IS A BREAKING CHANGE__, however the long term benefits outweigh the cost of the change IMHO.

Adds `grunt release:<major|minor|patch|custom-version>` command which will update the package version, commit, git tag, npm publish, bower repo update, and docs repo update.


trentgrover-wf:
What's the rationale for the folder rename to lib?  Can we get the same npm deploy advantage with the existing cjs dir?

mtscout6:
I renamed it to lib for two main reasons. First, it's the more common convention for node modules, trivial i agree. The second is for consistency between the cjs and amd approaches. Keeping the two approaches similar will allow downstream users an easier path to switch between amd and commonjs should they decide to do so. Take note that I updated the amd approach and documentation to follow suite.

While yes we can forego the name change and still gain the benefit of an npm deploy, but I don't feel that appropriately sets up this project for the future.

mtscout6:
What's the sentiment on all this? react-bootstrap 0.13.0 is still broken on npm. Also I didn't even realize that 0.12.0 is also broken on npm. This is making it difficult to upgrade and get the latest bug fixes.

mtscout6:
I also failed to mention earlier that this change will allow you to work with [npm link](https://docs.npmjs.com/cli/link)

trentgrover-wf:
I hesitate to change this but for wholly selfish reasons:

I have a private fork of this library that includes a Dart wrapper so these components can be used from Dart, but Dart requires that all of its exposed code live in the `lib` directory.  If this react-bootstrap started using the `lib` directly, I'd have to find some workaround on my fork for managing the Dart stuff (but, of course, that's my problem, not yours).

Anybody else going to weigh in?

mathieumg:
I don't really mind in this case, but I've also noticed that most of the projects go `/lib`, hence it is the approach I've also taken with my projects. That being said, if there is to be only one folder for the core files (and `/lib` isn't retained) I think it should be something that is agnostic to the module format, as opposed to `/cjs`. Just my two cents!

mtscout6:
The question I have now is what to do about bower? I don't like having the separate repo, but I also don't like committing a distributable. Thoughts?

mtscout6:
From what I can tell about bower there really is no better way than to commit the compiled output. So, keeping that additional repo is still necessary. I added a release task that will do the work of tagging, building and pushing to npm, bower repo, and docs repo in one swoop.

mtscout6:
@trentgrover-wf Regarding your concern about using with Dart, could we do something similar to how we are handling bower now?

mtscout6:
@mathieumg There's some changelog generation for you :smile: I'm still testing it locally a little bit, but I should wrap it up by tomorrow.

mathieumg:
Awesome! I like the approach too. :+1:

trentgrover-wf:
@mtscout6 you're doing all the heavy lifting for maintaining this library at the moment so don't worry about the dart issue when moving forward with this.  It's perfectly reasonable to expose the dart wrapper via a separate repo.  Maybe I'll get around to providing that code at some point for open-source use.

PR #330, opened 2015-01-07
Update for Bootstrap 3.3: render a .modal-backdrop inside of its .modal.

alloy-d:
In Bootstrap 3.3.0, the `.modal-backdrop` was moved inside its corresponding `.modal`. See: twbs/bootstrap#14724

With this, the `z-index` was removed from the `.modal-backdrop`, meaning that modal backdrops in React Bootstrap aren't layered properly when used with CSS from Bootstrap >= 3.3.0.

Note: this *breaks* Modals in Bootstrap < 3.3.0, because the backdrop winds up *above* the modal content.

What is React Bootstrap's policy for dealing with backwards-incompatible changes like this in Bootstrap itself? If you have a standard-ish way of handling this, I'm happy to update the PR accordingly!

### DIFF
--- a/src/Modal.jsx
+++ b/src/Modal.jsx
@@ -47,7 +47,7 @@ var Modal = React.createClass({
       'in': !this.props.animation || !document.querySelectorAll
     };
 
-    var modal = (
+    return (
       <div
         {...this.props}
         title={null}
@@ -57,6 +57,7 @@ var Modal = React.createClass({
         className={joinClasses(this.props.className, classSet(classes))}
         onClick={this.props.backdrop === true ? this.handleBackdropClick : null}
         ref="modal">
+        {this.props.backdrop ? this.renderBackdrop() : null}
         <div className={classSet(dialogClasses)}>
           <div className="modal-content">
             {this.props.title ? this.renderHeader() : null}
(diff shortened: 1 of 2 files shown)


mtscout6:
IMHO This is where [SemVer](http://semver.org/) comes in, react-bootstrap should try to stay with the current release of bootstrap and react. If users want backwards compatibility then they should use an older version, with the understanding that only bug fixes may or may not get published. I may note that although Bootstrap failed to bump the major version for a breaking change, that should not mean that we don't.

mtscout6:
Looks like this breaks the docs:



Although this is not entirely your fault. I have another change I'd like to bring in, then I'll look at getting that css from npm (I just created #348 for that issue)

cvrebert:
Please be advised that Bootstrap v3.3.4 is going to move the backdrop back outside of the modal again; see https://github.com/twbs/bootstrap/pull/15881

mtscout6:
Thanks for the heads up. We'll keep an eye out for its release.

vinnymac:
Looks like 3.3.4 is out, I have tried it in my project and it looks like it is working fine now.

mtscout6:
Wohoo ... darn

So, with the latest changes I've been making it's really easy to try out the new version of bootstrap. Just `npm install bootstrap@3.3.4 --save-dev` then run `npm run docs`. Unfortunately the newer version has removed their docs css from the npm module, which I just recently started using...

All said it should be a quick upgrade so long as we figure out that docs css issue.

PR #331, opened 2015-01-08
basic use of brand on Navbars

reiniergs:


mtscout6:
:+1:

mtscout6:
I appreciate more documentation, but your changes are breaking the build.

src/Button.jsx
```diff
@@ -70,13 +70,14 @@ var Button = React.createClass({
...
   renderNavItem: function (classes) {
+    classes['navbar-btn'] = true;
     var liClasses = {
       active: this.props.active
     };
 
     return (
       <li className={classSet(liClasses)}>
-        {this.renderAnchor(classes)}
+        {this.renderButton(classes)}
```
mtscout6: Instead of changing this to blindly render a button we should extract the logic on when to use a button vs an anchor and do the right one. That logic is here: https://github.com/react-bootstrap/react-bootstrap/blob/v0.15.0/src/Button.jsx#L39-L40

src/Button.jsx
```diff
@@ -70,13 +70,14 @@ var Button = React.createClass({
   },
 
   renderNavItem: function (classes) {
+    classes['navbar-btn'] = true;
```
mtscout6: Can you add a unit test to ensure we don't break this in the future?

mtscout6:
I appreciate the added documentation, I'm sure that will help a lot of folks out. Can you look into the build failures and my two line comments?

mtscout6:
I believe this has now been addressed in #392, since that demonstrates usage of the brand.

PR #333, opened 2015-01-13
Allow tabs to show after tab content with TabbedArea prop tabsLast

meg2208:
My use case required me to show the tabs themselves below the tab content. With this change, you can simply do something like this:

```
      <TabbedArea tabsLast={true} defaultActiveKey={2} animation={false}>
        <TabPane tab="Tab 1" eventKey={1} ref="pane1">Tab 1 content</TabPane>
      </TabbedArea>
```


meg2208:
Closing this for now because I had to commit the compiled code to that branch to get it working with requirejs.

PR #336, opened 2015-01-19
getValue() should return all selected options for multiple select

heydiplo:


mtscout6:
LGTM

trentgrover-wf:
+1
might be nice to have example usage in the docs site

mtscout6:
Somewhat related to #342

heydiplo:
@trentgrover-wf, example of multiple select is already in place, and the fact getValue() returns an array seems anticipated. Do you think we should explicitly state it?

@mtscout6, I like the idea of distinct inputs a lot, I think it would be much more natural. But it seems to be a big change that would take some time to implement. This fix, in turn, is simple, and someone can use it right now. And it can be just transferred to <Select /> later.

mtscout6:
@heydiplo I agree that it can come in here and let #342 move it later. Can you add something to the docs to reflect this before we bring it in?

heydiplo:
@mtscout6, sure, added it to the Input Types section.

mtscout6:
Gotcha, thanks for the heads up.

Issue #228, opened 2014-09-12
Tables in Panels are forced into Panel body


philwills:
Currently if you pass a Table in as the content of a Panel, it gets put into a `<div class="panel-body">`.

This leaves the table with padding around it, as opposed to the examples in the Bootstrap docs which show the `<table>` as a direct child of `<div class="panel">` whereby it gets matches the `.panel>.table` selector and loses the unsightly padding.

I would be happy to open a pull request, but I'm not sure what a good solution to this would look like.

mtscout6:
I'm in favor of the `flush` and `fill` props, with the caveat that @nickcolley's suggestion to ensure the same html order is retained. I don't know the ramifications of having two `panel-body`s, but if it doesn't hurt anything then it sounds good to me. Go ahead and submit a pull request and we'll get it refined from there.

taion:
The PR maintains the HTML ordering. I don't see how to make collapsing work, though.

mtscout6:
@philwills Does #346 resolve everything for you or is there more that I'm missing?

philwills:
Sadly, it doesn't look like it.

taion:
@philwills What is the issue still?

taion:
BTW, the updated example in the docs depends on #348 as well to render correctly.

philwills:
Sorry, this does actually fix things, I just wasn't applying the fill attribute to *all* the things I needed to in my use case. Having done so, it works a charm.

taion:
Cool - let me know if any other bugs crop up here. Otherwise, we should close this issue (:

davidnorth:
In the current version this doesn't work if the panel only has a single child. The table still gets wrapped in a panel body. As a workaround, I had to add <div fill/> after my table.

mtscout6:
@davidnorth Thanks for finding that

mtscout6:
@taion Do you see a quick solution for that?

taion:
My bad. Got too fancy with my test cases and ignored the obvious one.

mwhite:
@taion This doesn't work for `<Panel><TableWrapper /></Panel>` where TableWrapper renders `<Table fill />'.  I couldn't quite figure out how to fix it, not sure it's even possible without pre-rendering the children.  Not sure if worth re-opening or documenting.

taion:
Put `fill` on the `TableWrapper`. We may rewrite the `Panel` API to be more semantic and make better use of children anyway, to remove these issues.

PR #346, opened 2015-01-22
Support panel-filling tables and list groups

taion:
For react-bootstrap/react-bootstrap#228

Having multiple body elements in the panel is going to unavoidably screw up the collapse behavior, unfortunately, but it works otherwise.

### DIFF
--- a/docs/examples/PanelListGroupFill.js
+++ b/docs/examples/PanelListGroupFill.js
@@ -1,0 +1,13 @@
+var panelInstance = (
+    <Panel header="Panel heading">
+      Some default panel content here.
+      <ListGroup fill>
+        <ListGroupItem>Item 1</ListGroupItem>
+        <ListGroupItem>Item 2</ListGroupItem>
+        <ListGroupItem>&hellip;</ListGroupItem>
+      </ListGroup>
+      Some more panel content here.
+    </Panel>
+  );
+
+React.render(panelInstance, mountNode);
--- a/docs/examples/PanelListGroupFill.js
+++ b/docs/examples/PanelListGroupFill.js
@@ -1,5 +1,5 @@
 var panelInstance = (
-    <Panel header="Panel heading">
+    <Panel collapsable defaultExpanded header="Panel heading">
       Some default panel content here.
       <ListGroup fill>
         <ListGroupItem>Item 1</ListGroupItem>
--- a/test/PanelSpec.jsx
+++ b/test/PanelSpec.jsx
@@ -3,6 +3,7 @@
 var React          = require('react');
 var ReactTestUtils = require('react/lib/ReactTestUtils');
 var Panel          = require('../cjs/Panel');
+var Table          = require('../cjs/Table');
 
 describe('Panel', function () {
   it('Should have class and body', function () {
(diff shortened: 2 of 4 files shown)


docs/examples/PanelListGroupFill.js
```diff
@@ -0,0 +1,13 @@
...
+      <ListGroup fill>
+        <ListGroupItem>Item 1</ListGroupItem>
+        <ListGroupItem>Item 2</ListGroupItem>
+        <ListGroupItem>&hellip;</ListGroupItem>
+      </ListGroup>
+      Some more panel content here.
+    </Panel>
+  );
+
+React.render(panelInstance, mountNode);
```
mtscout6: This example does not do anything see with and without fill.

With Fill

Without Fill

test/PanelSpec.jsx
```diff
@@ -137,4 +138,19 @@ describe('Panel', function () {
...
+        Panel content
+        <Table fill />
+        More panel content
+      </Panel>
+    );
+
+    var children = instance.getDOMNode().children;
+    assert.equal(children[0].nodeName, 'DIV');
+    assert.equal(children[1].nodeName, 'TABLE');
+    assert.equal(children[2].nodeName, 'DIV');
```
mtscout6: Add assertion that the `panel-body` class was added.

test/PanelSpec.jsx
```diff
@@ -137,4 +138,19 @@ describe('Panel', function () {
...
+    var instance = ReactTestUtils.renderIntoDocument(
+      <Panel>
+        Panel content
+        <Table fill />
+        More panel content
+      </Panel>
+    );
+
+    var children = instance.getDOMNode().children;
+    assert.equal(children[0].nodeName, 'DIV');
```
mtscout6: Add assertion that the `panel-body` class was added.

docs/examples/PanelListGroupFill.js
taion: I spot checked on Safari and Chrome that it looks like this on my side:

With `fill`:
image

Without `fill`:
image

taion:
Added the asserts for `.panel-body`. Let me know if you'd prefer that I squash this commit (I'd prefer to squash it).

docs/examples/PanelListGroupFill.js
mtscout6: Something must have been really out of wack with my local repo. I am seeing the same thing now.

taion: I wouldn't send in a PR that just completely doesn't work at all. :innocent:

mtscout6:
So the only problem now is getting them to collapse?

taion:
Yes. We can solve it in this way. Get rid of the `body` ref altogether, and instead just use `scrollHeight` on the `panel` ref for the collapsable dimension value.

What do you think? It doesn't break any of the tests and it seems to keep the same collapse behavior.

taion:
One caveat - the styling for making list groups fill collapsable panels properly was only added in v3.3.0. The react-bootstrap docs are currently using v3.2.0, so the styling for panel-filling list groups in collapsable panels is not correct.

taion:
The last commit implements what I was talking about. It makes the code a lot simpler. Let me know if you have any concerns.

mtscout6:
Looks great, thanks!

PR #370, opened 2015-02-09
ListGroup now inherits classes from props

jingchan:


mtscout6:
Looks great! Can you add some tests to ensure we don't break this in the future? Also, please read the new CONTRIBUTING doc at the root of the repo, and amend your commit with the format outlined so the changelog will update when this is published.

PR #376, opened 2015-02-11
Move className prop to outer-most element in dropdown

### LINKED ISSUE #375: DropdownButton className not on the outer-most element
If you set a `className` property on a `DropdownButton`. It adds the class to button, which is not the outer-most element in the rendered markup. This makes the component a little awkward to work with. For example, if you want to add the dropdown to an `input-group`, you cannot.

defrex:
fixes #375

### DIFF
--- a/src/DropdownButton.jsx
+++ b/src/DropdownButton.jsx
@@ -27,8 +27,6 @@ var DropdownButton = React.createClass({
   },
 
   render: function () {
-    var className = 'dropdown-toggle';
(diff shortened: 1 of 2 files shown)


mtscout6:
I like where this, but there are failing unit tests, please fix those.

defrex:
Nice. The tests are now fixed.

mtscout6:
Looks good, thanks!

PR #382, opened 2015-02-15
Adds Property to Modal for Large Size

Enigmatical:
Pretty simple alteration I found myself needing and couldn't find the best route to insert.

It looks like div.modal-dialog classes are not accessible from the Modal component.

In order to create large modals in Bootstrap, the class .modal-lg needs to be added to the div.modal-dialog element (http://getbootstrap.com/javascript/#modals-sizes).

This adds a property to Modal "large" (Boolean) that will add that class to dialogClasses.

```
<Modal title="Modal Title" large />
```

### DIFF
--- a/src/Modal.jsx
+++ b/src/Modal.jsx
@@ -40,6 +40,7 @@ var Modal = React.createClass({
     var dialogClasses = this.getBsClassSet();
     delete dialogClasses.modal;
     dialogClasses['modal-dialog'] = true;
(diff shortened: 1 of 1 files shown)


mtscout6:
You can accomplish the same thing already with `<Modal title='Modal Title' bsSize='large' />`. This is the second time in less than a month that this issue has come up; guess it's time to update the docs :smile:.

PR #390, opened 2015-02-17
Add js code tab

bryonmarks:
I find the documentation for react-bootstrap to be really useful but don't generally use JSX.  I'd like to propose adding a read-only tab to the examples that would show the compiled JS.

### DIFF
--- a/docs/src/ReactPlayground.js
+++ b/docs/src/ReactPlayground.js
@@ -205,8 +205,8 @@ var ReactPlayground = React.createClass({
           <div ref="mount" />
         </div>
         {editor}
-        <a className={classSet(toggleClasses)} onClick={this.handleToggleJS} href="#">{this.state.mode !== this.MODES.JS ? 'show js' : 'hide js'}</a>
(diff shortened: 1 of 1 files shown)


mtscout6:
I like the concept, but I don't care too much for having two buttons that can activate the dropdown. How do you feel about doing something like on travis-ci for example where there is a button inside that can toggle the behavior?

Travis-Ci [example](https://travis-ci.org/react-bootstrap/react-bootstrap/builds/51084130):

Default View:



Mouse Over:



Ideally I'd like to see that we change the color of that inner dot to reflect an active toggle.

Some added sugar on top would be if that setting applied to all code examples so an end user would only have to toggle the behavior once for all examples.

AlexKVal:
@bryonmarks If this PR still has relevance, please let us know. 
We'll reopen this PR in that case.

As for the code:
- it needs to resolve merge conflicts with current version at least,
- and concider @mtscout6's comment.

Thank you. :sparkles:

PR #391, opened 2015-02-17
Changing NavItem title prop to node

EricWarnke:
Current NavItem titles are restricted to strings, which is not how Bootstrap was designed. Changing this to node to allow for more complicated functions.

### DIFF
--- a/src/NavItem.jsx
+++ b/src/NavItem.jsx
@@ -11,7 +11,7 @@ var NavItem = React.createClass({
     active: React.PropTypes.bool,
     disabled: React.PropTypes.bool,
     href: React.PropTypes.string,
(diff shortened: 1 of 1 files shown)


mtscout6:
:+1: Good idea, thanks!

Can you add some tests so we don't regress on this?

Also, when you've got some tests in place, squash the commits and change your commit message following the guidelines from the Contributing doc at the root of this repo. This way the change will be reflected in the release notes. I think the proper tag for this change is `[fixed]` since this addresses a bug.

EricWarnke:
I would write tests, but I'm seriously not that kind of developer. I dabble in the front-end arts only when I need to and I've never written a test in my life. Sorry!

PR #394, opened 2015-02-18
Added check if we don't need to calc calculate the overlay position on mount

### LINKED ISSUE #353: Empty nodes for tooltips and performance
Currently when using a tooltip react-bootstrap creates a div with an empty span for every tooltip in the app. When the number of tooltips is small this is not a problem but when using tooltips a lot it really makes a huge performance issue. Isn't there a way to not create those empty elements? I assume the problem here is the way react works but maybe there is a workaround?

mtetlow:
I created this pull request because I was seeing a significant amount of layout thrashing when I had several hundred tooltips being rendered at once. It seems the layout was invalidated by the overlay position calculation fired within `componentDidMount`. By checking `this.props.defaultOverlayShown`, we can avoid this layout reflow when mounting an OverlayTrigger which is not shown by default (should be the only time this calc needs to happen in `componentDidMount`??). I am not sure this will fully resolve #353 , but it helped my implementation quite a bit.

Chrome dev tools with several hundred tooltips:
Before: http://i.imgur.com/e4UZ5l6.png , http://i.imgur.com/Tw39F9t.png 
After: http://i.imgur.com/bU0f7VY.png .

### DIFF
--- a/src/OverlayTrigger.jsx
+++ b/src/OverlayTrigger.jsx
@@ -119,7 +119,9 @@ var OverlayTrigger = React.createClass({
   },
 
   componentDidMount: function() {
-    this.updateOverlayPosition();
+    if (this.props.defaultOverlayShown) {
+      this.updateOverlayPosition();
+    }
   },
 
   handleDelayedShow: function () {


mtscout6:
What exactly is responsible for setting `defaultOverlayShown`? Looks like the caller is responsible for providing it because I don't see any other reference to it in the rest of the Library.

Can you add some documentation on this so it does not trip up others in the future?

Can you amend your commit message to the format prescribed in the Contributing guide at the root of the repo? I think the proper tag for this is `[fixed]`, so if you can include the issue number #353 in the commit message that will be easy to track the issue that was fixed from the release notes.

mtetlow:
I don't believe this commit fixes #353  in it's entirety. I do think it could improve the performance in the interim though. I assume the point of https://github.com/react-bootstrap/react-bootstrap/commit/120882b595a46042164055b356d1d9d382465424 was to get the prop `defaultOverlayShown` mentioned in propTypes to work. Unfortunately by triggering `updateOverlayPosition` on each componentDidMount, browser reflows were occurring each time an element mounted. I think we'd still find more performance gains by figuring out a way to only render the overlayTrigger on demand, based upon `props.trigger`. 

Should I just mention #353 in the commit message instead of also mentioning #353 and tagging it as fixed?

I'd be happy to document this prop for `OverlayTrigger` but I am unsure of where the best place would be. The only spots I see `OverlayTrigger` mentioned in docs are for components like `Tooltip`. Do you think it would be useful to add an example with a `defaultOverlayShown` tooltip right around here: https://github.com/react-bootstrap/react-bootstrap/blob/master/docs/src/ComponentsPage.js#L254 ?

mtscout6:
Ya, mentioning #353 is fine as opposed to marking it as fixed. This commit message should still be tagged with `[fixed]` since it does at least fix some performance concerns.

I agree with you that finding a better long term solution to only render once triggered is desired. I think doing something along the lines described by Ryan Florence in his React conf presentation [Hype (Starting at 15:21)](https://www.youtube.com/watch?v=z5e7kWSHWTg) about Portals.

I'm not even really sure what `defaultOverlayShown` does or looks like, so I'll lean on your experience with it to identify a decent place to document it. I guess we could skimp on the docs here until we get the longer term solution in there as well.

mtetlow:
Ok, commit message updated with [fixed] tag as well as mentioning the issue.

All `defaultOverlayShown` seems to do is show the overlay component on render. It looks like you would use with something like `<OverlayTrigger defaultOverlayShown={true} ... />`. I vote skimp on docs until there's a section for complete docs on the `OverlayTrigger` component and all of it's props.

mtetlow:
Anything else I can do to get this merged?

mtscout6:
Oversight on my part that this was ready.

PR #398, opened 2015-02-19
DropdownStateMixin: return early if the target has been detached

sdemjanenko:
This happens if there is a dropdown inside a dropdown and the inner
dropdown closes (removing the inner dropdown DOM).

### DIFF
--- a/src/DropdownStateMixin.js
+++ b/src/DropdownStateMixin.js
@@ -48,7 +48,7 @@ var DropdownStateMixin = {
   handleDocumentClick: function (e) {
     // If the click originated from within this component
     // don't do anything.
-    if (isNodeInRoot(e.target, this.getDOMNode())) {
+    if (isNodeInRoot(e.target, this.getDOMNode()) || !isNodeInRoot(e.target, document)) {
       return;
     }
 


mtscout6:
Can you add some tests for this?

Also, please update your commit message according to the Contributing guide at the root of this repo.

sdemjanenko:
I can add a test.  This commit is more of a feature request than it is a bug fix.  I don't think any of the commit message heading tags apply.

mtscout6:
Are you doing something like this?

```
<DropdownButton title="Dropdown">
        <MenuItem eventKey="1">Dropdown link</MenuItem>
        <MenuItem eventKey="2">Dropdown link</MenuItem>
        <MenuItem eventKey="3">
          <DropdownButton>
            <MenuItem eventKey="1">Dropdown link</MenuItem>
            <MenuItem eventKey="2">Dropdown link</MenuItem>
          </DropdownButton>
        </MenuItem>
      </DropdownButton>
```

sdemjanenko:
I made my own `<DropdownButton>` component (I call it DropdownWidget) because I wanted to be able to put more than just `MenuItems` .  Within this I add a `<Chosen>` component.

```javascript
<DropdownWidget>
  <h1>...</h1>
  <Chosen ... />
</DropdownWidget>
```

When I click on an option in the `Chosen` the dropdown ends up closing.

mtscout6:
Nested dropdowns are not supported in Bootstrap 3.x. See: https://github.com/twbs/bootstrap/pull/6342

sdemjanenko:
I don't think this issue should be closed.  I am mixing in `DropdownStateMixin` in to my own components.  These do not play well with being nested inside a `BootstrapDropdown` but there is no reason why they can't.  I current have my own mixin which overrides a method in the `DropdownStateMixin` to make it do the right thing.  I would like this override to live in the community library.

mtscout6:
I can't honestly say at this point in time that that mixin will still even be there when I'm done with the Dropdown overhaul I'm doing in #526. Since I'm moving things to ES6 classes, I'm not using mixins.

sdemjanenko:
That makes more sense.  You should have mentioned that in your reason for closing the issue.

PR #392, opened 2015-02-17
[added] Example for collapsable Navbar in docs.

dozoisch:
I added a bit of doc for the navbar section and added an example of how to use the toggleNavKey!

PR #400, opened 2015-02-20
Add Collapsable class that can handle multiple Nav objects

jamesakers:
These proposed changes will allow for a cleaner implementation of the bootstrap collapse component. Most of the functionality was moved up from the Nav element. This handles the correct sizing and doesn't offset the navbar elements where there are multiple `<Nav>` elements as was the original issue that led me down this path.

```html
<Navbar toggleNavKey={1}>
  <Collapsable eventKey={1}>
     <Nav navbar={true}>
       <NavItem>Item 1</NavItem>
       <NavItem>Item 2</NavItem>
     </Nav>
     <Nav right={true} navbar={true}>
       <NavItem>Right Item 1</NavItem>
       <NavItem>Right Item 2</NavItem>
     </Nav>
  </Collapsable>
</Navbar>
```
> Results:
```html
<nav role="navigation" class="navbar navbar-default">
  <div class="container">
    <div class="navbar-header">
      <button class="navbar-toggle" type="button">
        <span class="sr-only">Toggle navigation</span>
        <span class="icon-bar"></span>
        <span class="icon-bar"></span>
        <span class="icon-bar"></span>
      </button>
    </div>
    <div class="collapse navbar-collapse" style="height: 0px;">
      <ul class="nav navbar-nav">
        <li><a href="item1">Item 1</a></li>
        <li><a href="item2">Item 2</a></li>
      </ul>
      <ul class="nav navbar-nav navbar-right">
        <li><a href="rightItem1">Right Item 1</a></li>
        <li><a href="rightItem2">Rith Item 2</a></li>
      </ul>
    </div>
  </div>
</nav>
```

### DIFF
--- a/docs/examples/CollapsableNav.js
+++ b/docs/examples/CollapsableNav.js
@@ -1,0 +1,23 @@
+var navbarInstance = (
+    <Navbar brand="React-Bootstrap" toggleNavKey={0}>
+      <CollapsableNav eventKey={0}> {/* This is the eventKey referenced */}
+        <Nav navbar>
+          <NavItem eventKey={1} href="#">Link</NavItem>
+          <NavItem eventKey={2} href="#">Link</NavItem>
+          <DropdownButton eventKey={3} title="Dropdown">
+            <MenuItem eventKey="1">Action</MenuItem>
+            <MenuItem eventKey="2">Another action</MenuItem>
+            <MenuItem eventKey="3">Something else here</MenuItem>
+            <MenuItem divider />
+            <MenuItem eventKey="4">Separated link</MenuItem>
+          </DropdownButton>
+        </Nav>
+        <Nav navbar right>
+          <NavItem eventKey={1} href="#">Link Right</NavItem>
+          <NavItem eventKey={2} href="#">Link Right</NavItem>
+        </Nav>
+      </CollapsableNav>
+    </Navbar>
+  );
+
+React.render(navbarInstance, mountNode);
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -357,6 +357,16 @@ var ComponentsPage = React.createClass({
                     </pre>
                   </div>
                   <ReactPlayground codeText={fs.readFileSync(__dirname + '/../examples/NavbarCollapsable.js', 'utf8')} />
+
+                  <h3>Mobile Friendly (Multiple Nav Components)</h3>
+                  <p>To have a mobile friendly Navbar that handles multiple <code>Nav</code> components use <code>CollapsableNav</code>. The <code>toggleNavKey</code> must still be set, however, the corresponding <code>eventKey</code> must now be on the <code>CollapsableNav</code> component.</p>
+                  <div className="bs-callout bs-callout-info">
+                    <h4>Div collapse</h4>
+                    <p>The <code>navbar-collapse</code> div gets created as the collapsable element which follows the <a href="http://getbootstrap.com/components/#navbar-default">bootstrap</a> collapsable navbar documentation.</p>
+                    <pre>&lt;div class="collapse navbar-collapse"&gt;&lt;/div&gt;</pre>
+                  </div>
+
+                  <ReactPlayground codeText={fs.readFileSync(__dirname + '/../examples/CollapsableNav.js', 'utf8')} />
                 </div>
 
                 {/* Tabbed Areas */}
--- a/docs/src/ReactPlayground.js
+++ b/docs/src/ReactPlayground.js
@@ -9,6 +9,7 @@ var Button = require('../../lib/Button');
 var ButtonGroup = require('../../lib/ButtonGroup');
 var ButtonToolbar = require('../../lib/ButtonToolbar');
 var CollapsableMixin = require('../../lib/CollapsableMixin');
+var CollapsableNav = require('../../lib/CollapsableNav');
 var Carousel = require('../../lib/Carousel');
 var CarouselItem = require('../../lib/CarouselItem');
 var Col = require('../../lib/Col');
--- a/src/CollapsableNav.jsx
+++ b/src/CollapsableNav.jsx
@@ -1,0 +1,110 @@
+var React = require('react');
+var joinClasses = require('./utils/joinClasses');
+var BootstrapMixin = require('./BootstrapMixin');
+var CollapsableMixin = require('./CollapsableMixin');
+var classSet = require('./utils/classSet');
+var domUtils = require('./utils/domUtils');
+var cloneWithProps = require('./utils/cloneWithProps');
+
+var ValidComponentChildren = require('./utils/ValidComponentChildren');
+var createChainedFunction = require('./utils/createChainedFunction');
+
+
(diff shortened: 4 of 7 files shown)


src/Collapsable.jsx
```diff
@@ -0,0 +1,99 @@
...
+var CollapsableMixin = require('./CollapsableMixin');
+var classSet = require('./utils/classSet');
+var domUtils = require('./utils/domUtils');
+var cloneWithProps = require('./utils/cloneWithProps');
+
+var ValidComponentChildren = require('./utils/ValidComponentChildren');
+var createChainedFunction = require('./utils/createChainedFunction');
+
+
+var Collapsable = React.createClass({
```
mtscout6: You misspelled Collapsible, alsothis is specific to `Nav`s, so can we name it `CollapsibleNav`?

mtscout6: Nevermind on the spelling, I guess both are valid: http://www.thefreedictionary.com/collapsable You learn something new everyday :smile:

src/Collapsable.jsx
```diff
@@ -0,0 +1,99 @@
...
+        var n = nodes[key].getDOMNode(), h = n.offsetHeight, computedStyles = domUtils.getComputedStyles(n);
+        height += (h + parseInt(computedStyles.marginTop, 10) + parseInt(computedStyles.marginBottom, 10));
+      }
+    }
+    return height;
+  },
+
+  render: function () {
+
+    var classes = this.props.collapsable ? this.getCollapsableClassSet() : {};
```
mtscout6: Note that [`this.getCollapsableClassSet()`](https://github.com/react-bootstrap/react-bootstrap/blob/v0.15.1/src/CollapsableMixin.js#L101-L117) Will break apart classes passed in via props as well, so can you change the false case to be the `classSet` representation of `this.props.className`?

src/Collapsable.jsx
```diff
@@ -0,0 +1,99 @@
...
+  getCollapsableDOMNode: function () {
+    return this.getDOMNode();
+  },
+
+  getCollapsableDimensionValue: function () {
+    var height = 0;
+    var nodes = this.refs;
+    for (var key in nodes) {
+      if (nodes.hasOwnProperty(key)) {
+        var n = nodes[key].getDOMNode(), h = n.offsetHeight, computedStyles = domUtils.getComputedStyles(n);
```
mtscout6: Minor style tweek, can you break up each assignment to a separate line? The way I look at it is that as devs we read faster from top to bottom. When logic trails too far to the right then it requires a context shift to wrap your head around what's going on.

src/Collapsable.jsx
jamesakers: The mixin, `CollapsableMixin` was spelled Collapsable and I rolled with it. So I would propose renaming that as well. The name `CollapsibleNav` makes sense to me. Would you like another PR with the proposed changes or are you taking it from here?

src/Collapsable.jsx
```diff
@@ -0,0 +1,99 @@
...
+    return height;
+  },
+
+  render: function () {
+
+    var classes = this.props.collapsable ? this.getCollapsableClassSet() : {};
+    classes['navbar-collapse'] = this.props.collapsable;
+
+    return (
+      <div eventKey={this.props.eventKey} className={joinClasses(this.props.className, classSet(classes))} >
```
mtscout6: Here you could be doubling up on classes as my above comment addresses.

src/Collapsable.jsx
```diff
@@ -0,0 +1,99 @@
...
+        return true;
+      }
+    }
+    if (this.props.activeHref != null) {
+      if (child.props.href === this.props.activeHref) {
+        return true;
+      }
+    }
+
+    return child.props.active;
```
mtscout6: Since you already check if this is true above and return there, wouldn't it be safe to return false here?

mtscout6:
Thanks for putting this together!

Can you include some docs on how this is used? Also, please see the Contributing guide at the root of this repo on how to format your commit message. We use a script to generate release notes now, based on how commit messages are formed.

mtscout6:
I didn't realize that PR #392 was already available, do those updated docs explain/satisfy what you need?

src/Collapsable.jsx
jamesakers: Not sure on this one. Let's say a dev was using `this.props.active` for another purpose we would always be setting it to false when cloning the child. The implementation above won't override anything. Side note: `getChildActiveProp()` is present in `Nav.jsx`, `SubNav.jsx` and now `CollapsableNav.jsx` shouldn't this be split out into a util?

jamesakers:
I will make the requested changes, squash commits and update the the message to comply with the Contributing guide. PR #392 doesn't satisfy or explain features of this PR. It was a great doc update though, I had to do some code spelunking to figure the event key requirements prior to realizing it didn't handle multiple `Nav`s. I will extend that documentation.

src/Collapsable.jsx
jamesakers: I will revise.

src/Collapsable.jsx
jamesakers: for sure

src/Collapsable.jsx
jamesakers: :+1:

jamesakers:
Looks like I picked up a change from upstream. Fixing.

jamesakers:
Requested changes made... what is the status on moving toward Collapsible?

elliott-davis:
I just bumped into this issue while trying to add multiple navs.  It would be awesome to have this merged in.

mtscout6:
Still missing some docs demonstrating its usage.

jamesakers:
I will take care of the docs demonstrating usage. I assume we are going to stick with Collapsable vs. Collapsible?

mtscout6:
For right now yes, though I would take a pull request updating it separate from this.

jamesakers:
This is ready but would like feedback on what was added to the documentation.

elliott-davis:
The docs examples look good to me.  Code_Samples++.  

Also, the parts about how to use the toggleNavKey are helpful.

docs/examples/CollapsableNav.js
```diff
@@ -0,0 +1,23 @@
...
+          <NavItem eventKey={2} href="#">Link</NavItem>
+          <DropdownButton eventKey={3} title="Dropdown">
+            <MenuItem eventKey="1">Action</MenuItem>
+            <MenuItem eventKey="2">Another action</MenuItem>
+            <MenuItem eventKey="3">Something else here</MenuItem>
+            <MenuItem divider />
+            <MenuItem eventKey="4">Separated link</MenuItem>
+          </DropdownButton>
+        </Nav>
+        <Nav navbar={true} right={true}>
```
mtscout6: When using booleans the mere presence of a props is presumed to be true. The following is also valid and cleaner, IMHO: `<Nav navbar right>`

jamesakers: Agree. It also is more consistent with the other examples. Making change.

mtscout6:
Thanks!

PR #401, opened 2015-02-20
Added utils to module interface

kirill-konshin:
Some of the utility classes/functions may be used externally, which means they must be exposed in module interface.

### DIFF
--- a/tools/amd/index.js
+++ b/tools/amd/index.js
@@ -51,6 +51,18 @@ define(function (require) {
     Table: require('./lib/Table'),
     TabPane: require('./lib/TabPane'),
     Tooltip: require('./lib/Tooltip'),
-    Well: require('./lib/Well')
+    Well: require('./lib/Well'),
+    utils: {
+        classSet: require('./lib/utils/classSet'),
+        cloneWithProps: require('./lib/utils/cloneWithProps'),
+        createChainedFunction: require('./lib/utils/createChainedFunction'),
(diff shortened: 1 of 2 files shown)


kirill-konshin:
See this comment for more details https://github.com/mtscout6/react-router-bootstrap/pull/26#issuecomment-75324068

mtscout6:
@kirill-konshin I'd like to remove the utilities that are accessible from React in lieu of including them through our public api. When you wrap that up, please change your commit message according to the [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) guidelines, this way the change will be reflected in the release notes.

kirill-konshin:
Yes, sure, I totally vote for not to copy-paste React's features. The only thing that is used by ReactRouterBootstrap that is not available as public API in browser is joinClasses.

mtscout6:
ReactRouterBootstrap has or at least should have a direct dependency on React, and it uses CommonJS as well. So it can pull that dependency from React.

kirill-konshin:
True, but unfortunately using CommonJS and Webpack we only can embed the `react/lib/joinClasses` into the build since it is not publicly available in browser build. I am now preparing updated version of PR.

kirill-konshin:
I have updated PR for ReactRouterBootstrap https://github.com/mtscout6/react-router-bootstrap/pull/26#issuecomment-75658820. This PR is no longer needed.

PR #403, opened 2015-02-21
[added] bsStyle prop support for Modal to set the header color

stefanfisk:


mtscout6:
I like the idea of this!

The styling is off though, note the hard corners on the top:



I've bee using [twbs-contextual-modals](https://github.com/binlabs/twbs-contextual-modals) for my project which applies more style to the entire modal, and you can still activate it with `bsStyle='primary'`.

stefanfisk:
Adding overflow: hidden to .modal-content seems to solve it (commited the change), but I don’t have enough time to test it properly right now.

Personally I thing that twbs-contextual-modals goes a bit too far, but to each his own ;)

mtscout6:
Looks good, can you squash your commits for this?

stefanfisk:
Done!

stefanfisk:
Just a sec, the commit message is wrong...

stefanfisk:
There we go.

mtscout6:
Thanks!

mtscout6:
Looks like this caused this bug: #448  @stefanfisk Can you look into this?

davidhouweling:
@mtscout6 @stefanfisk  I believe the correct way to resolve this would be to add a border-top-radius of modal-content + 1 to modal-title. However, there isn't really a clean way to do this across all browsers.

e.g. if modal-content border-top-radius = 6px (default value), the modal-title border-top-radius should be 7px.

PR #405, opened 2015-02-23
Update Carousel.jsx

lilaja:
Carousel wrapping. Render carousel controls on this.props.wrap = true


mtscout6:
Ah ok, I see it now. Can you amend you commit message as per the Contributing guidelines at the root of the repo? This way this bug fix will be documented in the release notes.

lo1tuma:
I’ve run into this problem as well.
@lilaja do you continue your work on this fix?

PR #409, opened 2015-02-25
Use grunt-karma for npm run test

amarant:
Use grunt-karma to run the "npm run" karma tasks instead of shell
commands, for windows users.


mtscout6:
I like it, thanks!

Issue #368, opened 2015-02-06
dropdown menus


jvillasante:
I'm using React.bootstrap in combination with react-router and react-router-bootstrap. The problem I'm facing is that when I put a Dropdown on a Navbar, the actions works as expected, but the dropdown keeps open, I would like that when the user clicks a menu from the dropdown, the route kicks and the dropdown goes away. Here's my code, the return clause of the render method:

```javascript
    return (
      <div>
        <Navbar ref="header" className="navbar navbar-default" brand={brand} fixedTop={true} fluid={true} toggleNavKey={1}>
          <Nav className="navbar-left" eventKey={1}>
            <NavItemLink to="dashboard" eventKey={2}>Dashboard</NavItemLink>
            <DropdownButton eventKey={3} title="Admin">
              <MenuItemLink to="users" eventKey="1">Users</MenuItemLink>
              <MenuItemLink to="projects" eventKey="2">Projects</MenuItemLink>
            </DropdownButton>
          </Nav>
        </Navbar>

        <RouteHandler />

        <div className="navbar-footer">
          <div className="container">
            <p className="text-muted">
              <span>some footer text</span>
            </p>
          </div>
        </div>
      </div>
    );
```

tjwebb:
>  but the dropdown keeps open,

+1

z-ax:
+1.

Also, with a `react-mini-router` route is reset to default action when DD is opened - and `onClick` or `onSelect` aren't triggered at all in this case. Not sure yet what's happening.

z-ax:
Seen 15th release, any updates here?

wiredprairie:
@jvillasante - Have you tried to use `onSelect` with the `DropdownButton`? I don't have a working example that is exactly like yours, but I just noticed that the Dropdown closes when I've attached an event handler to `OnSelect`:

    <DropdownButton eventKey={3} title="Admin" onSelect={ this.OnSelected } >

    onSelected: function(e) {
        // doesn't need to have functionality (necessarily) ... just wired up
    }

Internally, a call to `DropdownButton.setDropDownState(false)` is made which will hide the dropdown menu. So, you should be able to call that directly on the DropdownButton instance as well if needed.

rapilabs:
@wiredprairie that solution will close the menu, however the link is not followed due to the default action being prevented: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/MenuItem.jsx#L24

rapilabs:
[...]

PR #411, opened 2015-02-25
Enable hiding of drop down menu when a MenuItem with a href is clicked

rapilabs:
This patch hides the drop down menu by adding an onSelect if a href prop is present and ignoring the e.preventDefault().

This may not be the optimal approach, but it satisfies my requirements.

Ref: Issue #368


mtscout6:
The preferable approach is getting drafted out in #419

mtscout6:
Closing in favor of work being done in #526.

Issue #342, opened 2015-01-22
Input refactoring. Split select, radio, textarea, text-input


ivan-kleshnin:
Currently they are in one codebase and the code is contaminated with multiple **if** statements...
I think it's worth splitting onto two, three or more distinct components. **react-bootstrap** is, perhaps, the only one that merged them together. My anti-favorite is `<Input type="select"/>` :( 

__-- EDIT by @mtscout6 --__

I originally wrote down these thoughts against #205, that issue has been resolved, but these notes apply better here.

I think we should extract smaller components, where some are not publicly exposed.

We could add these components publicly:

- [x] `RadioGroup` - Addressed in #962 
- [x] `RadioButton` - Addressed in #962 
- [ ] `CheckboxGroup`
- [ ] `Checkbox`
- [ ] `TextArea`
- [ ] `TextInput` - With with `type` prop to support password, email, etc.
- [ ] `FileInput` - This one doesn't even work with our current implementation, but
  we should figure this one out at some point.
- [ ] `Select`
- [x] `Static` - Addressed in PR #674 
- [x] `Submit` - Addressed in PR #626 

@ivan-kleshnin Correct me if I missed one.

The public components would make use of these private components:

- [x] `FormGroup` - Generic `form-group` and validation wrapper - Completed in #205
- [ ] `InputGroup` - Generic label, help text, and addon wrapper

Each control than can should implement a `getValue()` function.

Since checkboxes do not correlate with each other the way radios do with the name property, I'm thinking they'd be different group types. This gets especially difficult when figuring out how to implement a `getValue()` function for the two different groups.

Example form when all is said and done:

```
class MyForm extends React.Component {
  render() {
    return (
      <form>
        {/* type prop defaults to 'text' */}
        <Input label='Text' placeholder='Enter text' />
        <Input type='email' label='Email' placeholder='Enter email' />
        <TextArea label='Text Area' placeholder='textarea' />
        {/* inline prop optional */}
        <CheckboxGroup inline>
          <Checkbox label='Package 1' checked />
          <Checkbox label='Package 2' />
          <Checkbox label='Package 3' checked readonly />
        </CheckboxGroup>
        {/* inline prop optional */}
        <RadioGroup name='radio-name' checked='option1' inline>
          <Radio label='Option 1' value='option1' />
          <Radio label='Option 2' value='option2' />
          <Radio label='Option 3' value='option3' />
        </RadioGroup>
        {/* multiple prop optional */}
[...]

PR #413, opened 2015-02-26
[added] inputGroupClassName prop to Input

gldnspud:
This allows for specifying a size, such as input-group-sm and input-group-lg, as in:

    <Input
      type="text"
      placeholder="Search"
      inputGroupClassName="input-group-sm"
      buttonAfter={<Button><Glyphicon glyph="search" /></Button>}
      />

Before:

before

After:

after

### DIFF
--- a/src/Input.jsx
+++ b/src/Input.jsx
@@ -4,6 +4,7 @@ var classSet = require('./utils/classSet');
 var Button = require('./Button');
 
 var Input = React.createClass({
+
   propTypes: {
     type: React.PropTypes.string,
     label: React.PropTypes.node,
@@ -12,6 +13,7 @@ var Input = React.createClass({
     addonAfter: React.PropTypes.node,
     buttonBefore: React.PropTypes.node,
     buttonAfter: React.PropTypes.node,
+    bsSize: React.PropTypes.oneOf(['small', 'medium', 'large']),
     bsStyle: function(props) {
       if (props.type === 'submit') {
         // Return early if `type=submit` as the `Button` component
@@ -140,8 +142,14 @@ var Input = React.createClass({
       </span>
     ) : null;
 
+    var inputGroupClassName;
+    switch (this.props.bsSize) {
+      case 'small': inputGroupClassName = 'input-group-sm'; break;
+      case 'large': inputGroupClassName = 'input-group-lg'; break;
+    }
+
     return addonBefore || addonAfter || buttonBefore || buttonAfter ? (
-      <div className="input-group" key="input-group">
+      <div className={joinClasses(inputGroupClassName, 'input-group')} key="input-group">
         {addonBefore}
         {buttonBefore}
         {children}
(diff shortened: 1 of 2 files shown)


mtscout6:
I'd rather solve this using `bsSize` similar to how it works with buttons, wells, modals, etc. The [`BootstrapMixin` is responsible for this`](https://github.com/react-bootstrap/react-bootstrap/blob/v0.15.1/src/BootstrapMixin.js#L20-L23). So ideally you'd only have to do this:

```
<Input
  type="text"
  placeholder="Search"
  bsSize="small"
  buttonAfter={<Button><Glyphicon glyph="search" /></Button>}
  />
```

gldnspud:
OK, I'll give that a try.

gldnspud:
@mtscout6 I didn't end up using the `BootstrapMixin` because `Input` is more complex than, say, `Button`, and has a different range of sizing options.

There may be more paths in `Input` that could also make use of the `bsSize` prop, but  I figured I'd start with the intent of that prop as applied to my specific problem of correcting the `div.input-group` size.

src/Input.jsx
```diff
@@ -12,6 +13,9 @@ var Input = React.createClass({
     addonAfter: React.PropTypes.node,
     buttonBefore: React.PropTypes.node,
     buttonAfter: React.PropTypes.node,
+    bsSize: function(props) {
+      return React.PropTypes.oneOf(['small', 'medium', 'large']).apply(null, arguments);
+    },
```
mtscout6: You can do that with:

```
bsSize: React.PropTypes.oneOf(['small', 'medium', 'large'])
```

No need to wrap in a function.

test/InputSpec.jsx
```diff
@@ -123,6 +123,22 @@ describe('Input', function () {
...
 
+  it('renders input-group with sm or lg class name when bsSize is small or large', function () {
+    var instance = ReactTestUtils.renderIntoDocument(
+      <Input addonBefore="$" bsSize="small" />
+    );
+
+    assert.ok(ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'input-group'));
+    assert.ok(ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'input-group-sm'));
+
+    var instance = ReactTestUtils.renderIntoDocument(
```
mtscout6: Drop `var` here, you don't need to redeclare the variable.

mtscout6:
There is some desire to split those up later, see #342. Don't worry I'm not asking you to do that, but when we do that will be a good opportunity to clean this up further.

gldnspud:
@mtscout6 sounds great, looking forward to it. If I warm up to this project over the next few weeks, I'd be glad to help here and there as I can. :)

mtscout6:
Thanks!

mtscout6:
Released in 0.16.1!

Issue #418, opened 2015-02-26
Replace usage of classSet with classnames module from npm


mtscout6:
React will be [deprecating](http://facebook.github.io/react/blog/#react-with-add-ons) classSet soon and we shouldn't be doing this ourselves anyways.

https://www.npmjs.com/package/classnames

mtscout6:
Resolved by #424

Issue #420, opened 2015-02-26
Support React 0.13


pilwon:
```
npm ERR! Darwin 14.1.0
npm ERR! argv "node" "npm" "install" "react-bootstrap"
npm ERR! node v0.12.0
npm ERR! npm  v2.5.1
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package react does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer react-bootstrap@0.15.1 wants react@>=0.12

npm ERR! Please include the following file with any support request:
npm ERR!     npm-debug.log
```

[React 0.13 has already reached RC1](http://facebook.github.io/react/blog/2015/02/24/react-v0.13-rc1.html).
[The popular react-router library is also working on 0.13 support](https://github.com/rackt/react-router/milestones/React%200.13).
It'd be great if `react-bootstrap` officially supports React 0.13 as well.

steos:
Hey. React 0.13.0-rc2 works fine for me with latest react-bootstrap (0.16.1). What's blocking this issue? I need React 0.13 for seamless ImmutableJS support.

meric:
The npm package could use an update to the peerDependencies row.

mtscout6:
We need to ensure we don't break the docs.

bastiao:
+1 for update.

pedroteixeira:
It seems to be working fine with 0.13

wmertens:
@steos seamless immutablejs support? Tell me more!

steos:
@wmertens well depends on how you define seamless I guess... As of 0.13 React supports iterators and immutable-js sequences as children.

mtscout6:
Resolved by #424

PR #424, opened 2015-02-28
0.13 Compatibility and Migration

jquense:
in response to #420. This is an initial stab at migration in preparation for `0.13`. As far as I can tell the only migration warning was for the use of keyed objects in ListItem, which was easily fixed. There was very little that needed to be changed to get the tests to pass, but since react-async and react-router-component both do not have `0.13.x` compatible releases or branches I am a bit limited in my ability to run the docs and give it a visual once over. Consider this PR a starting point not a finished product.


mtscout6:
Thanks for getting this started!

jquense:
I think you are good to go, expect for the react-async and react-router-component deps for docs (I manually installed them locally and they worked fine).

One point of discussion might be, I removed `classSet` and `joinClasses` and just installed `classnames` which is one the of recommended modules for replacing the internal `cx()`. It is super tiny module, but it is an external dependency. It would be no problem to just inline it as well if that is more to your liking. Overall tho the module simplifies a lot of logic and saves a bunch of bytes.

I also replaced CloneWithProps with the new `cloneElement` which, in most cases is a better fit anyway.

docs/src/ReactPlayground.js
```diff
@@ -1,5 +1,5 @@
 var React = require('react');
-var classSet = require('react/lib/cx');
+var classSet = require('classNames');
```
wmertens: You have a Mac and thus a case-insensitive filesystem :grin:. This should be 'classnames' right?

ie8/index.html
```diff
@@ -7,7 +7,7 @@
     <title>React Bootstrap IE8 test page</title>
 
     <!-- Bootstrap -->
-    <link href="//maxcdn.bootstrapcdn.com/bootstrap/3.1.1/css/bootstrap.min.css" rel="stylesheet">
+    <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.1.1/css/bootstrap.min.css" rel="stylesheet">
```
wmertens: Why? "//" works fine?

ie8/package.json
```diff
@@ -3,12 +3,12 @@
...
     "envify": "^1.2.1",
-    "react": "^0.10.0",
+    "react": "^0.13.0",
     "http-server": "^0.6.1"
   },
   "scripts": {
     "postinstall": "npm run build",
     "start": "http-server -c-1",
-    "build": "grunt --gruntfile ../Gruntfile.js && browserify -t reactify src.js > bundle.js"
+    "build": "grunt --gruntfile ../Gruntfile.js && browserify -t reactify src.js > bundle.js --harmony"
```
wmertens: Does this work? That `--harmony` needs to go in the commandline somewhere, not after the redirect, right?

test/DropdownButtonSpec.jsx
```diff
@@ -12,7 +12,7 @@ describe('DropdownButton', function () {
 
   afterEach(function() {
     if (instance && ReactTestUtils.isCompositeComponent(instance) && instance.isMounted()) {
-      React.unmountComponentAtNode(instance.getDOMNode().parent);
+      React.unmountComponentAtNode(instance.getDOMNode());
```
wmertens: shouldn't this use React.findDOMNode too?

test/FadeMixinSpec.jsx
```diff
@@ -36,24 +36,27 @@ describe('FadeMixin', function () {
...
-      setTimeout(function(){
-        assert.ok(!element.className.match(/\bin\b/));
-        assert.ok(element.className.match(/\bfade\b/));
-        assert.ok(!child.className.match(/\bin\b/));
-        assert.ok(child.className.match(/\bfade\b/));
-        done()
-      }, 25)
-    }, 25)
-  });
+  // it.only('Should remove the in class for all elements', function (done) {
```
wmertens: Is this test commented out on purpose?

wmertens:
Looking forward to bumping my app to 0.13 :smile:

ie8/index.html
jquense: Nope,  opening it locally chrome will assume the protocol is `file:` instead of http when unspecified at least on windows. This whole section though doesn't seem to work at anyway hasn't been updated since react 10. Maybe just an old artifact

docs/src/ReactPlayground.js
jquense: Good catch thanks

test/DropdownButtonSpec.jsx
jquense: `getDOMNode` isn't depreciated in createClass, as far as I know, so there is no warning. I can be changed but it isn't necessary

test/FadeMixinSpec.jsx
jquense: yup, its brittle (I should know i wrote it :P) and fails incorrectly. I need to write a new one

mtscout6: I'd prefer to see this fixed before we merge this. I'm not a fan if intentionally bringing in commented out code.

mtscout6:
Running `npm run build`results in failure. I don't know why travis-ci didn't kick off a build....

Changing to the `classnames` module will resolve #418! :+1: 

I don't want to take this unless we are sure that the docs will work. So, we will need to either wait for an updated react-router-component or switch to react-router. Unless the docs just work, which I have not been able to verify yet.

Once everything is all said and done can you squash these commits and format your commit message according to the [contributing guidelines](../blob/master/CONTRIBUTING.md). Please ensure that the content of the commit message outlines any upgrade guide type information that downstream users should be aware of.

jquense:
I am working on trying to get all the build stuff up and working. Sorry about that, I was focused on getting the tests to pass and neglected the other stuff :P 

The issue with the docs right now is just that I literally can't get npm to install the routing components because their peer dep range is pinned to `0.12`. I did get the docs to work by manually installing the components and everything worked great, so that's encouraging.

I will be sure to clean up the commits, and messages as well as soon as it looks like its safe.

jquense:
@mtscout6 I've updated (amoung other things) build task for the AMD and CommonJs Packages, as well as the `package.json` and `bower.json` files. I am unsure how one would test that the changes work, as I am not particularly familiar with the grunt task. I would appreciate a second set of eyes on [those changes](https://github.com/jquense/react-bootstrap/commit/e5dfdc0a237f5fe73bf6cf735b9532ee708a2496) if you get a moment

ie8/index.html
mrlamroger: Is this page only for IE8 testing?

mtscout6:
Still can't spin up the docs locally.

jquense:
The docs are beyond my ability to fix. The only way you can get them working without the components updating is to go download the components from github, extract them to `node_modules` and run `npm install` on each of them. I'm not suggesting we do that, just that if you WANT to get them running before they are updated, that would be what you'd need to do.

jquense:
@mtscout6 I think we are good to go...I ended up using react-router to get the docs up and running. I have what I can to make sure its all working, but I am not the most familiar with the setup, so you will need to eyeball it I think. I squashed the commits as well, let me know if that's everything, I don't have a lot of time in the near future, but if its simple I am sure I can do it.

mtscout6:
Split Drop down buttons don't quite look right

mtscout6:
Also missing `react-router` in package.json.

ie8/index.html
mtscout6: Yes, it's only for testing IE8

mtscout6:
Also note the code toggle is exhibiting some odd behavior when expanded:

jquense:
I will fix the router dependency, oops. 

@mtscout6 I am sorry, I cannot for the life of me figure out what is wrong with either the CodeMirror styles or the buttons (it is broken with any button with just a caret). The styles are _exactly_ the same as the current production react-bootstrap, and original bootstrap docs. It's not React specific because adding a button directly to the DOM (outside of react) results in the same issue.

I can "fix" the code mirror style issue by setting the scroll div to height auto, but I am not sure if that is a reasonable fix.

mtscout6:
This is how I ended up getting around the split button problem: https://github.com/react-bootstrap/react-bootstrap/commit/e7ccd4700c249f0cdc3ae64f02010712dd86696b. For some reason that button does not have any content which is affecting it's rendered height. This feels really hacky, but I don't know of a better solution at the moment. I have pushed a `react-0.13` branch with a few changes. Next up I'll see if I can figure out what's going on with that CodeMirror pane.

mtscout6:
Released in v0.18.0, @jquense Huge thanks for doing most of the leg work on this!

jquense:
Woohoo! Glad to help. thanks for the project!

Issue #419, opened 2015-02-26
Change calls to e.preventDefault() to allow onClick and onSelect calls to override


mtscout6:
There has been a number of concerns raised recently regarding `e.preventDefault();` getting applied without regard for what a developer would actually want to occur. This occurs regularly when attempting to do client side routing. [This search](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue+is%3Aopen+preventDefault) will highlight a few such cases.

I'd like to move to a more consistent model similar to [this](https://github.com/rackt/react-router/blob/v0.12.4/modules/components/Link.js#L59-L76). Where the user's `onClick` or `onSelect` handler can return true or false to explicitly control it. This is also similar to how [jQuery](http://stackoverflow.com/questions/1551389/how-to-enable-default-after-event-preventdefault) has handled things.

List of components to address: (Not conclusive yet)
- [ ] Dropdown
- [ ] Panel (Collapse)
- [ ] PanelGroup
- [ ] Button
- [ ] Nav
- [ ] Navbar
- [ ] NavItem
- [ ] Carousel
- [ ] CollapsableNav
- [ ] PageItem
- [ ] Pager
- [ ] Split Button
- [ ] TabbedArea

refs about documenting it:
#432
#320

stefanfisk:
Per http://facebook.github.io/react/docs/events.html, React no longer checks the return value of event handlers to decide how to proceed. If we do this, the behaviour will probably be unexpected by a lot of people, since one wouldn't expect to have to read the docs to figure out how onClick and other standard handlers work.

In es6 it is also easy to accidentally write event handlers that return values when using fat arrows without curly braces, potentially leading to hard to identify bugs.

stefanfisk:
My point is that if the behaviour is to be implemented then the handlers should use non-standard names to make it clear that they implement specific behaviours.

mtscout6:
Good points, thanks! I can agree with your points and go without a return value for those event handlers. 

We can still check the event value `event.defaultPrevented` which may be all we need in most cases. Although we'll have to take it on a case by case basis, but I think that should take care of most of the scenarios. 

With Dropdown MenuItem with an href this gets tricky. In a client side routing scenario such as described in #368 we want to call `event.preventDefault()` to stop the browser from following a link, but we still want the dropdown to close.

I see three options:

[...]

PR #427, opened 2015-03-03
[changed] CollapsableMixin state tracking

### LINKED ISSUE #399: collapsibleMixin sets dimension of an element in pixels
This approach works without problems if a content of the collapsible element changes only on `render` calls. But it also can change without React doing anything. 

For example, when you have an image of unknown dimensions, an `` is rendered, `_updateDimensionAfterRender()` is called, after that browser gets a response with image dimensions and updates layout. And, because the height of my panel is set in pixels in element style, page looks like this (it's an accordion):

Broken accordion

If I reload the page, it renders correctly, because the image is cached by my browser and is immediately rendered with proper height and width.

I managed to work around this by setting `ref`: `` and binding to `onload`:

```javascript
    componentDidMount() {
        var image = this.refs.mainImage;
        if (image !== undefined && this.props.onDOMUpdate !== undefined) {
            image.getDOMNode().onload = () => {
                this.props.onDOMUpdate();
            }
        }
    },

    /* parent element that contains collapsible element */
    onDOMUpdate() {
        this.setState({_rerenderHack: !this.state._rerenderHack});
    },
```

It works but I don't like such hacks. Do you have any ideas how to make `collapsibleMixin` more robust to changes of its children elements?

joemcbride:
* Added internal 'is expanded' tracking (see comments in mixin)
* Added basic `CollapsablePanel` example
* Added aria-expanded attributes to Panel
* Fixes expand/collapse animation for stand-alone panel
* Fixes #399, where the panel would stay a fixed size

Fixes #399

### DIFF
--- a/docs/examples/CollapsableParagraph.js
+++ b/docs/examples/CollapsableParagraph.js
@@ -1,0 +1,37 @@
+var CollapsableParagraph = React.createClass({
+  mixins: [CollapsableMixin],
+
+  getCollapsableDOMNode: function(){
+    return this.refs.panel.getDOMNode();
+  },
+
+  getCollapsableDimensionValue: function(){
+    return this.refs.panel.getDOMNode().scrollHeight;
+  },
+
+  onHandleToggle: function(e){
+    e.preventDefault();
+    this.setState({expanded:!this.state.expanded});
+  },
+
+  render: function(){
+    var styles = this.getCollapsableClassSet();
+    var text = this.isExpanded() ? 'Hide' : 'Show';
+    return (
+      <div>
+        <Button onClick={this.onHandleToggle}>{text} Content</Button>
+        <div ref="panel" className={classSet(styles)}>
+          {this.props.children}
+        </div>
+      </div>
+    );
+  }
+});
+
+var panelInstance = (
+  <CollapsableParagraph>
+    Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid. 3 wolf moon officia aute, non cupidatat skateboard dolor brunch. Food truck quinoa nesciunt laborum eiusmod. Brunch 3 wolf moon tempor, sunt aliqua put a bird on it squid single-origin coffee nulla assumenda shoreditch et. Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident. Ad vegan excepteur butcher vice lomo. Leggings occaecat craft beer farm-to-table, raw denim aesthetic synth nesciunt you probably haven't heard of them accusamus labore sustainable VHS.
+  </CollapsableParagraph>
+);
+
+React.render(panelInstance, mountNode);
--- a/src/CollapsableMixin.js
+++ b/src/CollapsableMixin.js
@@ -1,101 +1,149 @@
 var React = require('react');
-var TransitionEvents = require('./utils/TransitionEvents');
+var TransitionEvents = require('react/lib/ReactTransitionEvents');
 
 var CollapsableMixin = {
 
   propTypes: {
-    collapsable: React.PropTypes.bool,
     defaultExpanded: React.PropTypes.bool,
     expanded: React.PropTypes.bool
   },
 
-  getInitialState: function () {
+  getInitialState: function(){
+    var defaultExpanded = this.props.defaultExpanded != null ?
+      this.props.defaultExpanded :
+        this.props.expanded != null ?
+        this.props.expanded :
+        false;
+
     return {
-      expanded: this.props.defaultExpanded != null ? this.props.defaultExpanded : null,
+      expanded: defaultExpanded,
       collapsing: false
     };
   },
 
-  handleTransitionEnd: function () {
-    this._collapseEnd = true;
-    this.setState({
-      collapsing: false
-    });
-  },
-
-  componentWillReceiveProps: function (newProps) {
-    if (this.props.collapsable && newProps.expanded !== this.props.expanded) {
-      this._collapseEnd = false;
-      this.setState({
-        collapsing: true
-      });
-    }
-  },
-
-  _addEndTransitionListener: function () {
-    var node = this.getCollapsableDOMNode();
-
-    if (node) {
-      TransitionEvents.addEndEventListener(
-        node,
-        this.handleTransitionEnd
-      );
-    }
-  },
-
-  _removeEndTransitionListener: function () {
-    var node = this.getCollapsableDOMNode();
-
-    if (node) {
-      TransitionEvents.removeEndEventListener(
-        node,
-        this.handleTransitionEnd
-      );
-    }
-  },
-
-  componentDidMount: function () {
-    this._afterRender();
-  },
-
-  componentWillUnmount: function () {
-    this._removeEndTransitionListener();
-  },
-
-  componentWillUpdate: function (nextProps) {
-    var dimension = (typeof this.getCollapsableDimension === 'function') ?
-      this.getCollapsableDimension() : 'height';
-    var node = this.getCollapsableDOMNode();
-
-    this._removeEndTransitionListener();
-  },
-
-  componentDidUpdate: function (prevProps, prevState) {
-    this._afterRender();
-  },
-
-  _afterRender: function () {
-    if (!this.props.collapsable) {
+  componentWillUpdate: function(nextProps, nextState){
+    var willExpanded = nextProps.expanded != null ? nextProps.expanded : nextState.expanded;
+    if (willExpanded === this.isExpanded()) {
       return;
     }
 
-    this._addEndTransitionListener();
-    setTimeout(this._updateDimensionAfterRender, 0);
+    // if the expanded state is being toggled, ensure node has a dimension value
+    // this is needed for the animation to work and needs to be set before
+    // the collapsing class is applied (after collapsing is applied the in class
+    // is removed and the node's dimension will be wrong)
+
+    var node = this.getCollapsableDOMNode();
+    var dimension = this.dimension();
+    var value = '0';
+
+    if(!willExpanded){
+      value = this.getCollapsableDimensionValue();
+    }
+
+    node.style[dimension] = value + 'px';
+
+    this._afterWillUpdate();
   },
 
-  _updateDimensionAfterRender: function () {
+  componentDidUpdate: function(prevProps, prevState){
+    // check if expanded is being toggled; if so, set collapsing
+    this._checkToggleCollapsing(prevProps, prevState);
+
+    // check if collapsing was turned on; if so, start animation
+    this._checkStartAnimation(); 
+  },
+
+  // helps enable test stubs
+  _afterWillUpdate: function(){
+  },
+
+  _checkStartAnimation: function(){
+    if(!this.state.collapsing) {
+      return;
+    }
+
     var node = this.getCollapsableDOMNode();
-    if (node) {
-        var dimension = (typeof this.getCollapsableDimension === 'function') ?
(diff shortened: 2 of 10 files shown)


src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
 var React = require('react');
 var TransitionEvents = require('./utils/TransitionEvents');
 
+/* This mixin keeps track of its own 'internal' expanded state.  This is to
+ * help enable running css animations on the collapsable component.  The
+ * expanded state /cannot/ bet set /before/ the mixin has had a chance to
```
mtscout6: misspelled "be" Change "bet" -> "be"

src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
...
+ * ensure that dimensions have been set on the collaspable component.  This
```
mtscout6: change "collaspable" -> "collapsable"

src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
...
+ * is due to how browers interact with css transitions.  If the mixin does
```
mtscout6: change "browers" -> "browsers"

src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
 var React = require('react');
 var TransitionEvents = require('./utils/TransitionEvents');
```
mtscout6: Can we pull this directly from React at `react/lib/ReactTransitionEvents`?

mtscout6: Or would this work: http://facebook.github.io/react/docs/animation.html#low-level-api-reacttransitiongroup

src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
...
-  },
+    var complete = (function (){
+      this._removeEndEventListener(node, complete);
+      // remove dimension value - this ensures the collapsable item can grow
+      // in dimension after initial display (such as an image loading)
+      node.style[dimension] = '';
+      this.setState({
+        collapsing:false
+      });
+    }).bind(this);
```
mtscout6: Man we really need to change this lib to use babel...

src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
...
+ * internalToggle() function in componentDidUpdate().  Ex:
+
+   componentDidUpdate: function(prevProps, prevState){
+     var wasExpanded = prevState.expanded;
+     var isExpanded = this.state.expanded;
+     if(wasExpanded != isExpanded){
+       this.internalToggle();
+     }
+   }
+*/
```
mtscout6: I can't say that I completely understand why we can't continue to allow `expanded` to be used on the components that use this mixin. The point of this mixin is to isolate that logic away from the components that use it. If the mixin needs to retain it's own internal representation of `expanded` then use something internal such as `__collapsableMixinExpanded__`.

src/CollapsableMixin.js
joemcbride: >When children are declaratively added or removed from it (as in the example above) special lifecycle hooks are called on them

We're not adding/removing elements so I don't think we can use `ReactTransitionGroup`.

I did a diff on the core `ReactTransitionEvents` and it looks like we should be able to use that.  The note in our utils says our version is modified, but the diff says otherwise.

src/CollapsableMixin.js
```diff
@@ -1,103 +1,133 @@
...
+/* This mixin keeps track of its own 'internal' expanded state.  This is to
+ * help enable running css animations on the collapsable component.  The
+ * expanded state /cannot/ bet set /before/ the mixin has had a chance to
+ * ensure that dimensions have been set on the collaspable component.  This
+ * is due to how browers interact with css transitions.  If the mixin does 
+ * not share an 'expanded' state then it makes it a lot easier to handle the
+ * order in which the dimension and `expanded` information is set.
+
+ * This pattern requires the owner of this mixin to manually call the
+ * internalToggle() function to toggle the expanded/collapsed state instead
```
mtscout6: Having a public api method named internal feels wrong.

src/CollapsableMixin.js
mtscout6: Cool!

docs/examples/CollapsablePanel.js
```diff
@@ -0,0 +1,72 @@
+var CollapsablePanel = React.createClass({
```
mtscout6: For this example the name `Panel` feels like the Bootstrap Panel. Can we rename this to `CollapsableParagraph`?

docs/examples/CollapsablePanel.js
```diff
@@ -0,0 +1,72 @@
...
+      this.internalToggle();
+    }
+  },
+
+  render: function(){
+    var styles = this.getCollapsableClassSet();
+    var text = this.isExpanded() ? 'Hide' : 'Show';
+    return (
+      <div>
+        <div><Button onClick={this.onHandleToggle}>{text} Content</Button></div>
```
mtscout6: We can drop this div wrapper

src/Panel.jsx
```diff
@@ -23,22 +35,29 @@ var Panel = React.createClass({
...
-      this._isChanging = true;
-      this.props.onSelect(this.props.eventKey);
-      this._isChanging = false;
+  handleSelect: function(e){
+    if(this.props.onSelect) {
+      this.props.onSelect(e, this.props.eventKey)
+    } else {
+      this.handleClickEvent(e);
+      this.handleToggle();
     }
```
mtscout6: Following the discussion on #419 instead of an else call here we can deal with the toggle if `e.selected` is true, and leave the e.preventDefault decision to the dev handling the `onSelect` event. I'm thinking something like this:

```
handleSelect: function(e) {
  e.selected = true;

  if (this.props.onSelect) {
    this.props.onSelect(e, this.props.eventKey);
  }

  if (e.selected) {
    this.handleToggle();
  }
}
```

src/Panel.jsx
```diff
@@ -23,22 +35,29 @@ var Panel = React.createClass({
...
+  },
 
+  handleClickEvent: function(e){
     e.preventDefault();
+  },
 
-    this.setState({
-      expanded: !this.state.expanded
-    });
+  handleToggle: function(){
```
mtscout6: Thanks makes sense to be the method name in the mixin. Ideally i'd like to figure out how to have the mixin basically handle this toggling itself.

src/CollapsableMixin.js
joemcbride: >I can't say that I completely understand why we can't continue to allow expanded to be used on the components that use this mixin.

They can use their own `expanded`.  The difference is previously this mixin was relying on that `external expanded` property instead of its own.

I know this feels a little hacky, but if you look at the [bootstrap collapse](https://github.com/twbs/bootstrap/blob/master/js/collapse.js) component this is exactly how they do it.  It is just a bit less code since they're not working within a structured component framework and not keeping track of state.

If the component doesn't keep its own representation of expanded then there is no way to set `node.style['height'] = '150px'` *before* the `collapsing=true` state is set.  If that fixed height style isn't set *before* the `collapsing=true` state is set, Safari will *never* run the css animation.  The css transition is defined on the `collapsing` style.

Trying to set `node.style['height'] = '150px'` in `componentWillUpdate` doesn't work, I believe because the `collapsing` class and the dimension is set essentially at the same time.  Safari just won't run the animation.  Chrome, and I think Firefox, handle this fine just not Safari.

This wasn't an issue previously since the dimension was not being removed.  Clearing out the dimension with `node.style['height'] = ''` makes all this unfortunately necessary (at least this is the only way I could get it working).

We can change the internal expanded to so some crazy name, no problem.

src/Panel.jsx
joemcbride: I guess it could be possible to move the `componentDidUpdate` code to the mixin.  The `Panel` would still be responsible for setting `expanded=true` and any component that used the mixin would be stuck with using the `expanded` property.  That's probably not a huge deal though.

src/Panel.jsx
joemcbride: `onSelect` is optional, so if its not provided wouldn't we need to call `e.prevendDefault()`?

src/Panel.jsx
joemcbride: I think part of the reason why I removed the `expanded` tracking from the mixin is that it started to get confusing when you should use the `internal expanded` vs. the `external expanded`.  With only one set of state tracking in the mixin then it's very clear.

docs/examples/CollapsablePanel.js
joemcbride: That was the intent  - a simplified version of the existing `Panel`.  I can change the name if you think that would be confusing.

joemcbride: Simplified this a bunch.

joemcbride:
Moved the previous `componentDidMount` code into the mixin and added `_`'s to all 'internal' functions.  This simplifies the api quite a bit.  The mixin now only relies upon its owner to toggle the expanded state/prop value.

Added `aria-expanded` attributes.

joemcbride:
So - spent a few more hours on this (whoohoo?).  Good news is I *was* able to remove the extra state tracking.  Setting the dimension in `componentWillUpdate` is working - perhaps I was trying it in `componentWillReceiveProps` before?  I don't know - too many hours at the keyboard.  Anyway - I added a second commit so the changes could be seen.  The tests cleaned up nice.  I can squash these and swap to `[fixed]` or `[added]` after review.

docs/examples/CollapsablePanel.js
mtscout6: Since it does not look like a `Panel` I'd prefer to rename it.

src/Panel.jsx
mtscout6: If the underlying component is a button and not an anchor we should be fine.

mtscout6:
Can you also update that example in the docs with the latest changes?

docs/examples/CollapsableParagraph.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  getCollapsableDimensionValue: function(){
+    return this.refs.panel.getDOMNode().scrollHeight;
+  },
+
+  onHandleToggle: function(e){
+    e.preventDefault();
+    this.setState({expanded:!this.state.expanded});
+  },
```
mtscout6: I think this method should reside in the mixin as well. I'd rename it to `handleToggle` though since it is in response to an event and not an event itself.

src/CollapsableMixin.js
```diff
@@ -1,101 +1,149 @@
...
-var TransitionEvents = require('./utils/TransitionEvents');
+var TransitionEvents = require('react/lib/ReactTransitionEvents');
 
 var CollapsableMixin = {
 
   propTypes: {
-    collapsable: React.PropTypes.bool,
     defaultExpanded: React.PropTypes.bool,
     expanded: React.PropTypes.bool
   },
```
mtscout6: We should add documentation for both these props.

src/CollapsableMixin.js
```diff
@@ -1,101 +1,149 @@
...
-  },
-
-  componentWillReceiveProps: function (newProps) {
-    if (this.props.collapsable && newProps.expanded !== this.props.expanded) {
-      this._collapseEnd = false;
-      this.setState({
-        collapsing: true
-      });
+  componentWillUpdate: function(nextProps, nextState){
+    var willExpanded = nextProps.expanded != null ? nextProps.expanded : nextState.expanded;
```
mtscout6: What happens when we have been getting props and the updated props is null? Should that case be false and continue not to use the internal state of this mixin? Should we care?

Issue #406, opened 2015-02-23
condition makes it impossible to reset offsets, pushes and pulls in the grid


leroix:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Col.jsx#L49

A reset of an offset, push or pull requires the property value to be `0`. This if statement will cause something like `mdPush={0}` to be ignored.

See http://getbootstrap.com/css/#grid-responsive-resets

PR #438, opened 2015-03-05
Col Push/Pull/Offset value zero support

aforty:
According to the Bootstrap documentation (and my own experience), column offsets must be reset to zero for other sizes if not used.

"In addition to column clearing at responsive breakpoints, you may need to reset offsets, pushes, or pulls. See this in action in the grid example."

For example `<Col xsOffset={2} xs={8} sm={4} md={4} lg={4}>...</Col>` when viewed in 'sm/md/lg' does not produce the expected result of having a column width of '4'. It looks only correct when viewed in 'xs'.

The example would have to be rewritten as `<Col xs={8} xsOffset={2} sm={4} smOffset={0} md={4} mdOffset={0} lg={4} lgOffset={0}>...</Col>` to render correctly in all views.

However, Offset/Push/Pull property values were ran through a "truthy" check in Col.jsx. So when the property value was `0` the column would not render. This has been fixed by doing a `>= 0` check instead which ensures that the property value is simply a non-negative number.

### DIFF
--- a/src/Col.jsx
+++ b/src/Col.jsx
@@ -46,19 +46,19 @@ var Col = React.createClass({
 
       prop = size + 'Offset';
       classPart = size + '-offset-';
-      if (this.props[prop]) {
+      if (this.props[prop] >= 0) {
         classes['col-' + classPart + this.props[prop]] = true;
       }
 
       prop = size + 'Push';
       classPart = size + '-push-';
-      if (this.props[prop]) {
+      if (this.props[prop] >= 0) {
         classes['col-' + classPart + this.props[prop]] = true;
       }
 
       prop = size + 'Pull';
       classPart = size + '-pull-';
-      if (this.props[prop]) {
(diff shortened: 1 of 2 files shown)


joemcbride:
@aforty This looks good.  Can you add some tests to lock down this behavior?  Also, can you alter your commit message to meet our contributing guidelines?  Basically start the commit message with `[fixed]` so the changlog can auto-track your change.

https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md

This should fix https://github.com/react-bootstrap/react-bootstrap/issues/406

joemcbride:
I should also say, please add `Fixes #406` to the commit as well, that way the bug will close when this gets pulled in.

aforty:
Alright, take a look and accept if all is well.

mtscout6:
Looks great, Thanks!

PR #457, opened 2015-03-24
Fixed behaviour of carousel controls if wrap=true

lo1tuma:
This PR is based on #405. I’ve changed the commit message and added a test.

Closes #405.


mtscout6:
Thanks!

PR #458, opened 2015-03-25
Keep active onclick behaviour of DropdownButton

tangrammer:


mtscout6:
Can you describe the problem you are trying to solve here?

tangrammer:
Hi Matt,
I need that my dropdownbutton behaves as normal button so I could attach
onclick handlers
In my case I had a dropdownbutton without children but I need to use it as
menu button
Thanks
Juan

src/DropdownButton.jsx
```diff
@@ -111,6 +111,11 @@ var DropdownButton = React.createClass({
     e.preventDefault();
 
     this.setDropdownState(!this.state.open);
+    if(this.props.onClick && this.props.noCaret ){
```
lo1tuma: Why should `noCaret` be set to `true` in order to execute the `onClick` handler?

tangrammer: I found a scenario where my DropDownButton didn't have any items (maybe
related to user permissions) then I had to use noCaret=true but I needed to
keep working with onClick handler

2015-04-13 16:30 GMT+02:00 Mathias Schreck <notifications@github.com>:

mtscout6: I'm not following the `noCaret` check for this either. To me the `noCaret` should only be used as presentation logic to toggle whether the caret is displayed or not. It should not have an effect on clicking. If it is causing a problem with clicking can you supply a test to prove so.

mtscout6:
Can you please add some tests for this behavior?

mtscout6:
Re-reading this today I'm wondering if a better solution would be for you to render a `Button` instead of `DropdownButton` in your parent component. That parent component should be aware if children are present or not and render the appropriate button.

mtscout6:
I'm going to close this, @tangrammer I'm actively working to fix numerous problems with the DropdownButon in #526.

PR #463, opened 2015-03-30
Use .ownerDocument instead of root document

martijnvermaat:
With this change, event listeners are attached to the component DOM node's owner document instead of to the window root document. The two are not the same when the component is mounted inside an IFrame, for example using [react-frame-component](https://github.com/ryanseddon/react-frame-component).

In my case, this fixes incorrect positioning of tooltips and broken close handlers of dropdown buttons.

### DIFF
--- a/src/AffixMixin.js
+++ b/src/AffixMixin.js
@@ -104,7 +104,7 @@ const AffixMixin = {
     this._onWindowScrollListener =
       EventListener.listen(window, 'scroll', this.checkPosition);
     this._onDocumentClickListener =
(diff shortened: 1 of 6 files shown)


mtscout6:
Interesting, I was not aware of that.

Can you amend your commit message header to include `[fixed]` as per the contributing guidelines. This will ensure that this is documented in the generated release notes. I say fixed because this feels like it was a silent bug that this fixes.

martijnvermaat:
@mtscout6 Sure, done!

PR #476, opened 2015-04-03
Fix Windows build problems

### LINKED ISSUE #471: '.' is bit recognized as an internal or external command while trying to build on Windows
Hello all,

I'm trying to use react-bootstrap. I just downloaded it and I'm trying to build it on Windows.

I just ran:

    npm run build

 Here's the result:

image

Any idea of how to build it on Windows?

mtscout6:
Closes #471

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 [Bootstrap 3][bootstrap] components built with [React][react]
 
-[![Build Status][build-badge]][build]
+[![Travis Build Status][build-badge]][build]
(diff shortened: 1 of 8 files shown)


mtscout6:
So... Windows sucks AppVeyor is not passing because of joyent/node#7374...

mtscout6:
I'd like to get AppVeyor working at some point, but for now I cannot get it to pass simply because npm install will not work. I have tested that `npm run build` does work on Windows 7, though I cannot vouch for any other version greater than that right now.

PR #480, opened 2015-04-04
[fixed] Don't try to access .ownerDocument on null

martijnvermaat:
As noted by @lukemcgregor on ee0382e that code breaks when trying to access `.ownerDocument` on `null`.

I attempt to fix that here by using the root document as a fallback and abstracting that to a domUtils function.

I also added one test case for this fix, but I'm not sure it's the best way to test it.

### DIFF
--- a/src/DropdownStateMixin.js
+++ b/src/DropdownStateMixin.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import domUtils from './utils/domUtils';
 import EventListener from './utils/EventListener';
 
 /**
@@ -56,7 +57,7 @@ const DropdownStateMixin = {
   },
 
   bindRootCloseHandlers() {
-    let doc = React.findDOMNode(this).ownerDocument;
+    let doc = domUtils.ownerDocument(this);
 
     this._onDocumentClickListener =
       EventListener.listen(doc, 'click', this.handleDocumentClick);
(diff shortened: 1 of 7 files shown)


lukemcgregor:
Yeah that should sort it out :)

src/DropdownStateMixin.js
```diff
@@ -56,7 +57,7 @@ const DropdownStateMixin = {
   },
 
   bindRootCloseHandlers() {
-    let doc = React.findDOMNode(this).ownerDocument;
+    let doc = domUtils.ownerDocument(React.findDOMNode(this));
```
mtscout6: You are repeating `domUtils.ownerDocument(React.findDOMNode(this))` a lot. Have you considered moving `React.findDOMNode(this)` inside the `domUtils.ownerDocument` function? So, you'd only have to call `domUtils.ownerDocument(this)`.

martijnvermaat: That was my first thought too, but decided against it because `domUtils` (currently) doesn't deal with anything React specific, just with the browser dom. If you think it's okay to introduce React there I can change it of course.

mtscout6: I don't see a problem with adding React there. The amount of code it DRYs up is enough of an argument. This whole library also depends on React, so it's not introducing anything too crazy.

martijnvermaat:
@mtscout6 Updated as per your suggestion and rebased on current master. Please let me know if there's anything else you'd want me to change.

mtscout6:
Looks great, thanks!

PR #489, opened 2015-04-08
Update CollapsableParagraph.js

cesarvelandia:
using classNames

### DIFF
--- a/docs/examples/CollapsableParagraph.js
+++ b/docs/examples/CollapsableParagraph.js
@@ -1,3 +1,5 @@
+var classNames = require('classNames');
+
 const CollapsableParagraph = React.createClass({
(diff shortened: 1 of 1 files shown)


mtscout6:
The examples are evaled at runtime which is what allows you to change them in the code tab and see live updates. This also means that we have some variables pre-defined for the examples. `classSet` is in fact the classnames module. See: https://github.com/react-bootstrap/react-bootstrap/blob/v0.20.2/docs/src/ReactPlayground.js

cesarvelandia:
Noted. It'd be nice that this was clearer (classSet > className) as it seems a bit a misleading in the example. Thanks.

mtscout6:
I understand, that's why I followed up with 6dfcf36 :smile:

Issue #440, opened 2015-03-05
Revisit constants and configurability


mtscout6:
Issue #306 highlights the ability to modify the constants used for prop validation and mapping. The name constants is deceiving since they are not truly constant values. We should give some thought to this practice give it a proper name. When we do this we should add some documentation for those wishing to work with custom classes. Issue #404 is a fine example of such a case.

trevorr:
Perhaps the right fix is to add `isKeyOf` to `CustomPropTypes` and use that within the `propTypes` definition of `BootstrapMixin`, so that additions to `STYLES` are used? I think it would also be better to add a function to `BootstrapMixin` to define new styles, rather than modifying `constants`. That would provide more control (e.g. add but not remove), clarity (`STYLES` being a map is strange from an external perspective), and debuggability (a place to set a breakpoint when styles are added), and it's slightly less of an exposed implementation detail. It might be reasonable to then make `constants` private again. Thoughts?

AlexKVal:
Superceded by https://github.com/react-bootstrap/react-bootstrap/issues/646

Issue #306, opened 2014-12-10
Allow custom bsStyle


janmarek:
Now you can't add classes like btn-custom or alert-custom to your stylesheet and use them with react-bootstrap components. bsStyle property should be validated less strictly.

ivan-kleshnin:
@trentgrover-wf you can't rely on bootstrap draft semantics and call it "rules". 
It were built for user extension. Bootstrap `info`, `warning`, etc. were built more for demo purpose than for real use. Thats BS developers opinion, not mine. End users may and should create their own sets of statuses etc. The current way of handling this is very uncomfortable, I 100% agree here with @janmarek. You don't need to validate class names against predefined sets at all. That simple.

mtscout6:
I can see the value in doing some kind of validation against a known set of styles you support, yet I can also see the value in allowing for custom styles. I think you can add your own custom styles and still keep the existing validation by doing:

```
var constants = require('react-boostrap/constants');
constants.STYLES.custom = 'custom';
```

The name constants is deceiving here since nothing has been done to ensure that those values have been modified. If you did this as one of your initial tasks prior to rendering anything, then you'd have the custom options you want with the validation to support it.

mathieumg:
Until the day these are enforced to be constant, or the filename changes.

mtscout6:
We can cross that bridge when that day comes :smile:. I'm not planning to change that anytime soon.

ivan-kleshnin:
@mtscout6 thank your for this trick, it may help.

mtscout6:
No problem!

mtscout6:
@janmarek Does the above suggestion help? Do we need to keep this issue open for any reason?

janmarek:
OK, I'm closing it.

nemothekid:
If you set `bsStyle=""`, you get a ton of console errors where
```
Warning: Invalid prop `bsStyle` of value `` supplied to `Button`, expected one of ["default","primary","success","info","warning","danger","link","inline","tabs","pills"]. Check the render method of `DropdownButton`.
```

[...]

Issue #404, opened 2015-02-22
All Modal classes should be options


hallister:
`modal-dialog`, `modal-content` and `modal-header` all seem to be hard-coded. This prevents reusing the modal for other components, such as a [sidebar](http://mgcrea.github.io/angular-strap/#/asides#asides). Having a `modalDialogClass`, `modalContentClass`, etc prop would be really simple. I can PR it, but wanted to gauge interest/naming conventions.

hallister:
So after playing with this for a while now, I'm starting to see the problem. It seems that react-bootstrap is *heavily* tied to the CSS markup of the original Bootstrap. This seems odd to me, as it only allows modification of the original components via CSS. So if I want to make a slightly modified modal `Aside`, I have to write it from the ground up because the bsClass for a modal has to be `modal` and can't be `aside`. 

I'm rewriting the components I plan on using using a more "options-oriented" approach, but I can't really PR it against this project since it would break how things are currently done.

mtscout6:
I was not aware of AngularStrap, pretty cool! I like the Aside that it provides. One of the original goals of this library was to target default Bootstrap CSS behavior. With that said though the goal was to allow it to be overridden: https://github.com/react-bootstrap/react-bootstrap/blob/v0.15.1/src/Modal.jsx#L30.

Which was quickly lost: https://github.com/react-bootstrap/react-bootstrap/blob/v0.15.1/src/Modal.jsx#L42

It should not be too difficult to switch out each of the hard ties to modal with the supplied `bsClass` prop, which I'd happily take a PR for.

hallister:
I actually did this two ways.

First, I did took out the hard ties to the class names and passed in configuration options for each class used in the modal. This turned into to some ugly markup, as the custom modal looked like this:

````HTML
<Modal
        key="sidebar"
        title="Navigation"
        className={baseClasses}
        animation={false}
        backdrop={true}
        onRequestHide={this.handleToggle}
        backdropClass="aside-backdrop"
        dialogClass="aside-dialog"
        contentClass="aside-content"
        titleClass="aside-title"
        headerClass="aside-header"
        openClass="aside-open"
        key="sidebar"
        modalStyle={style}
      >
````

That just looks... icky to me. So I ended up making another modification where I broke Modal into several sub-components:

````HTML
<ModalCustom className="aside" backdropClassName="aside-backdrop">
    <ModalDialog className="aside-dialog" contentClassName="aside-content">
[...]

PR #496, opened 2015-04-08
[added] styleMaps.addStyle and CustomPropTypes.keyOf

trevorr:
Proposed fix for #306 and partial fix for #440.

### DIFF
--- a/src/BootstrapMixin.js
+++ b/src/BootstrapMixin.js
@@ -1,28 +1,28 @@
-import React from 'react';
-import constants from './constants';
+import styleMaps from './styleMaps';
+import CustomPropTypes from './utils/CustomPropTypes';
 
 const BootstrapMixin = {
   propTypes: {
-    bsClass: React.PropTypes.oneOf(Object.keys(constants.CLASSES)),
-    bsStyle: React.PropTypes.oneOf(Object.keys(constants.STYLES)),
-    bsSize: React.PropTypes.oneOf(Object.keys(constants.SIZES))
+    bsClass: CustomPropTypes.keyOf(styleMaps.CLASSES),
+    bsStyle: CustomPropTypes.keyOf(styleMaps.STYLES),
+    bsSize: CustomPropTypes.keyOf(styleMaps.SIZES)
   },
 
   getBsClassSet() {
     let classes = {};
 
-    let bsClass = this.props.bsClass && constants.CLASSES[this.props.bsClass];
+    let bsClass = this.props.bsClass && styleMaps.CLASSES[this.props.bsClass];
     if (bsClass) {
       classes[bsClass] = true;
 
       let prefix = bsClass + '-';
 
-      let bsSize = this.props.bsSize && constants.SIZES[this.props.bsSize];
+      let bsSize = this.props.bsSize && styleMaps.SIZES[this.props.bsSize];
       if (bsSize) {
         classes[prefix + bsSize] = true;
       }
 
-      let bsStyle = this.props.bsStyle && constants.STYLES[this.props.bsStyle];
+      let bsStyle = this.props.bsStyle && styleMaps.STYLES[this.props.bsStyle];
       if (this.props.bsStyle) {
         classes[prefix + bsStyle] = true;
       }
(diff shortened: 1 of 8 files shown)


src/utils/CustomPropTypes.js
```diff
@@ -57,4 +66,18 @@ function createMountableChecker() {
...
 }
 
+function createKeyOfChecker(obj) {
+  function validate(props, propName, componentName) {
+    let propValue = props[propName];
+    if (!obj.hasOwnProperty(propValue)) {
+      let valuesString = JSON.stringify(Object.keys(obj));
+      return new Error(
+        ('Invalid prop `' + propName + '` of value `' + propValue + '` ') +
+        ('supplied to `' + componentName + '`, expected one of ' + valuesString + '.')
```
mtscout6: We're using ES6 you can use string interpolation here to clean this up: http://babeljs.io/docs/learn-es6/#template-strings

src/BootstrapMixin.js
```diff
@@ -1,11 +1,16 @@
...
-    bsClass: React.PropTypes.oneOf(Object.keys(constants.CLASSES)),
-    bsStyle: React.PropTypes.oneOf(Object.keys(constants.STYLES)),
-    bsSize: React.PropTypes.oneOf(Object.keys(constants.SIZES))
+    bsClass: CustomPropTypes.keyOf(constants.CLASSES),
+    bsStyle: CustomPropTypes.keyOf(constants.STYLES),
+    bsSize: CustomPropTypes.keyOf(constants.SIZES)
+  },
+
+  addStyle(name) {
+    constants.STYLES[name] = name;
```
mtscout6: I'm thinking we should rename `constants` to `styleMaps`? (I'm open to suggestions)

Then move this method there. It feels odd to configure this through the `BootstrapMixin`.

mtscout6:
This looks really great I like the simplicity of it. We should really add some documentation for this though so folks are aware that this is possible without combing the issues to find it. The Components page does not feel right for this though. I'm wondering if we should add a "Customize" page to the docs. This is partly why I feel that the `addStyles` function feels wrong on the `BootstrapMixin` mixins are generally tied directly to components and this configuration would be modifiable outside of components.

src/utils/CustomPropTypes.js
trevorr: Would that be preferable even with needing to escape the backticks within the string? I basically copied the code from `ReactPropTypes` `createOneOfChecker`. The surrounding code also doesn't use interpolation.

mtscout6: Is there any major advantage to backticks in the error message? We could just change them to quotes.

src/BootstrapMixin.js
trevorr: Agreed, particularly since mixins aren't supported for ES6 classes, so I figure `BootstrapMixin` might disappear someday.

src/utils/CustomPropTypes.js
mtscout6: I didn't change everything to ES6 templates when I switched everything to ES6, I'm just trying to keep things as clean as possible going forward.

trevorr: That's fine with me. I think they're a bit odd anyway. The only issue is consistency with React error messages, but that seems minor.

src/BootstrapMixin.js
mtscout6: That's been my general thought, but I haven't drafted an official statement about it yet.

trevorr:
Perhaps splitting off the styles from the other constants would be better, since they're the only ones likely to be extended by users? They could be in a module called `Styles` and the rest could remain (unexported) in `constants`.

mtscout6:
#404 Is one such example where the Classes could be modified. But you're right that the Sizes and Glyphs should need to be exported.

mtscout6:
The code looks great! Can you squash the commits with a commit message following the [Contributing Guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)? A commit message of that formate will reflect the change in the automated changelog when we release.

trevorr:
Squashed and also fixed a warning by removing unused `import React` from `BootstrapMixin`.

mtscout6:
Thanks!

Issue #516, opened 2015-04-14
Cannot specify class for TabPane rendered element


thehuey:
Using a TabbedArea with TabPane(s), I wanted to add a class to some TabPane elements so they can be "pull-right".

    <TabbedArea >
    <TabPane className="pull-right" >
    </TabPane>
    </TabbedArea>

Does not apply the specified className to the rendered <LI> element.

PR #517, opened 2015-04-14
[added] TabbedArea NavItem renderTab() className [fixes] #516

thehuey:
TabbedArea TabPane rendering doesn't carry across the className
property.  This means it is not easy for users to customize
the look/feel of the tabs.

Added className to the properties copied in the renderTab function
and also pass additional properties down via ...other

### DIFF
--- a/src/TabbedArea.js
+++ b/src/TabbedArea.js
@@ -108,12 +108,13 @@ const TabbedArea = React.createClass({
   },
 
   renderTab(child) {
-    let key = child.props.eventKey;
+    let {eventKey, className, tab } = child.props;
     return (
       <NavItem
-        ref={'tab' + key}
-        eventKey={key}>
-        {child.props.tab}
+        ref={'tab' + eventKey}
+        eventKey={eventKey}
+        className={className} >
+        {tab}
       </NavItem>
     );
   },
--- a/test/TabbedAreaSpec.js
+++ b/test/TabbedAreaSpec.js
@@ -182,4 +182,20 @@ describe('TabbedArea', function () {
 
     assert.ok(ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'nav-pills'));
   });
+
+  it('Should pass className to rendered Tab NavItem', function () {
+    let instance = ReactTestUtils.renderIntoDocument(
+      <TabbedArea activeKey={3}>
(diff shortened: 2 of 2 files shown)


src/TabbedArea.js
```diff
@@ -108,12 +108,14 @@ const TabbedArea = React.createClass({
   },
 
   renderTab(child) {
-    let key = child.props.eventKey;
+    let {eventKey, className, tab, ...other} = child.props;
```
AlexKVal: For what reason all `...other` props could be applied from child to parent ?

thehuey: The variable 'child' is not accurate.  It is the existing element being
rendered.  In the 'render' function, it calls:

ValidComponentChildren.map(this.props.children, renderTabIfSet, this)

I think you are right and that I should remove this.  It would copy the
TabPane properties to the corresponding Nav header.

mtscout6:
Can you add some tests to ensure that classNames are getting applied as you expect? This way this feature is not accidentally broken in the future. Also, after you add some tests can you squash the commits. There's no need for multiple commits for this change.

mtscout6:
Closes #516

thehuey:
@mtscout6 Test added to TabbedAreaSpec and squashed commits.

Once Travis CI finishes, I think this should be good.

mtscout6:
LGTM

/cc @react-bootstrap/collaborators How does this look?

dozoisch:
Just one quick question, this will apply the className passed to `Tab` on the `NavItem`. Not sure if it's the obvious behavior.

Should it be another prop like `navClassName` so that you can send different classNames to the NavItem and TabPane?

mtscout6:
@dozoisch Good call!

thehuey:
Currently there is no className being passed back to the renderPane function.  @dozoisch If we want to support custom classes on either one, we should agree on a naming scheme and open a different ticket.

How do the following properties look:
className => Tab Pane, used in renderPane.
navClassName => NavItem class used in the Nav for tabs.

I'll open a new issue?

dozoisch:
@thehuey That looks good to me!

mtscout6:
The issue of the classNames is different, and will require some more discussion. Let's continue that in #560. I also updated the 1.0.0 roadmap documentation to address this issue.

Issue #396, opened 2015-02-19
Progress Bar Percentages Getting Rounded Oddly


rtucker88:
So I have a Progressbar on a page that gets incremented by 5 percent each time it is advanced. When it gets to 55%, 56% is displayed.

This is happening because of Math.ceil being used in the getPercentage function. Is there a reason for using ceil over round?

If we consider the common use case to be a progress bar being from 0 to 100, then we can look for inconsistencies in the output. I quickly tested the cases from 0 to 100 using Math.ceil, Math.round, and Math.floor. The results are below: 

CEIL
IN, OUT
7, 8
14, 15
28, 29
55, 56
56, 57

ROUND
IN, OUT

FLOOR
IN, OUT
29, 28
57, 56
58, 57

I'm happy to open a PR if you would like to change to Math.round.

mathieumg:
> Is there a reason for using ceil over round?

To be optimistic I imagine? :stuck_out_tongue: 

That being said you can always leverage the `label` property if you need more fine-tuned display customization.

mtscout6:
I think rounding down should be the default, and perhaps add a `roundUp` prop for those that care to round up. I'd be open to a pull request that supported both.

YannPl:
The fact that percentage is rounded up causes a bug in the use of stacked bar if the max value is more than 100.

In example, try this code in the documentation example:
``` JavaScript
const progressInstance = (
  <ProgressBar>
    <ProgressBar bsStyle='success' max={1000} now={755} key={1} />
    <ProgressBar bsStyle='warning' max={1000} now={245} key={2} />
  </ProgressBar>
);

React.render(progressInstance, mountNode);
```
The first bar gets 25% and the second gets 76% so it's displayed in a second line and is invisible in the bar.

Can you please fix this either by using math.round or allow percentages to be decimals ?

Thanks a lot =)

mtscout6:
Fixed by #532

PR #520, opened 2015-04-15
[Fix] #396 - use of math.round in ProgressBar

YannPl:
Use math.round in getPercentage of ProgressBar to avoid the bug described in #396 with stacked bars

### DIFF
--- a/src/ProgressBar.js
+++ b/src/ProgressBar.js
@@ -27,7 +27,7 @@ const ProgressBar = React.createClass({
   },
 
   getPercentage(now, min, max) {
-    return Math.round((now - min) / (max - min) * 100);
+    return Math.round(((now - min) / (max - min) * 100)*1000)/1000;
   },
 
   render() {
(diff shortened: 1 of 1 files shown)


src/ProgressBar.js
```diff
@@ -27,7 +27,7 @@ const ProgressBar = React.createClass({
   },
 
   getPercentage(now, min, max) {
-    return Math.ceil((now - min) / (max - min) * 100);
+    return Math.round(((now - min) / (max - min) * 100)*1000)/1000;
```
mtscout6: Can you assign 100 and 1000 to a variable name to explain their purpose. This will ensure ease of maintainability in the future.

mtscout6:
Looks like you have some linting failures which are failing Travis CI. You should be able to see the problem locally by running `npm run lint` or `npm test`. Both those commands should pass on your machine. 

After extract those names to meaningful variable names can you squash your commits and apply a commit message as prescribed in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) at the root of this repo. Applying your commit message in that format will ensure this change is described in the automated changelog.

PR #531, opened 2015-04-17
[fixed] ProgressBar percentage issue when stacked

YannPl:
Using stacked ProgressBar with a maximum superior to 100 could lead to
rounding issues that made the bar incomplete.

See issue #396 for mor details on the error

### DIFF
--- a/src/ProgressBar.js
+++ b/src/ProgressBar.js
@@ -27,7 +27,8 @@ const ProgressBar = React.createClass({
   },
 
   getPercentage(now, min, max) {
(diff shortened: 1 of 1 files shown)


YannPl:
Sorry for the duplicate PR with #520 but I haven't found a way to do it otherwise.
I ran the commands as you said and everything should be OK now.

mtscout6:
Looks great, for future reference you can use a force push to get around replacing commits on a branch. It's generally not a good thing against something like a master branch, but can be really useful when cleaning up pull requests. No worries though, I really appreciate this fix.

Issue #423, opened 2015-02-27
Add factories like React.DOM


wmertens:
For those who prefer plain coffeescript, having to individually create factories or elements is a bit of a pain. It would be nice if there were an equivalent of `React.DOM` so you could write

```coffeescript
React = require 'react'
N = null
B = require 'react-bootstrap/factories'

App = React.createClass
  render: ->
    B.Grid N,
      B.Row N,
        B.Col md:6,
          "Hi"
        B.Col md:6,
          "there"

React.render (React.createElement App), document.body
```

or even

```coffeescript
React = require 'react'
N = null
{Grid, Row, Col} = require 'react-bootstrap/factories'

App = React.createClass
  render: ->
    Grid N,
      Row N,
        Col md:6,
          "Hi"
        Col md:6,
          "there"

React.render (React.createElement App), document.body
```

mtscout6:
Have you seen [coffee-react](https://github.com/jsdf/coffee-react)? Although not a direct port of what you are looking for it's close...

mtscout6:
Perhaps I'm missing something, but doesn't React require you to use the namespace to `React.DOM` as well? Even then since React v0.12 you would have to wrap each DOM function with factories as well: http://facebook.github.io/react/blog/2014/10/28/react-v0.12.html

wmertens:
@mtscout6 I really prefer not writing end tags and weird curly braces...
Yes, React.DOM provides factories, not elements. Check out http://codepen.io/anon/pen/PwRggP?editors=001 to play with a live example.

wmertens:
Here's how React does it, basically call createFactory on all tags when the module is imported... https://github.com/facebook/react/blob/2aeb8a2a6beb00617a4217f7f8284924fa2ad819/src/browser/ReactDOM.js

jquense:
exporting factories or elements is considered [bad practice](http://facebook.github.io/react/blog/2014/10/14/introducing-react-elements.html#anti-pattern-exporting-factories) in the React ecosystem. It would break the module for anyone using JSX witch as part of the compile step, wraps the Component in a `createElement()` (jsx doesn't use React.DOM). If you were going to do this it needs to be as a seperate export from the normal Component Types.

wmertens:
Yes, I'm asking for e.g. 'bootstrap/factories' just like you can require
'react/addons'.

tacomanator:
[...]

PR #535, opened 2015-04-18
adds ability to require components already wrapped in factories

tacomanator:
Allows import of components pre-wrapped in factories with 'react-bootstrap/lib/factories'.

I had same problem as this guy: https://github.com/react-bootstrap/react-bootstrap/issues/423.


tacomanator:
Redoing to fix CI warnings

PR #540, opened 2015-04-19
Handle multiple children in <Badge>. Fixes #532

### LINKED ISSUE #532: <Badge> doesn't accept multiple text segments
This is possibly a slightly unusual use-case, but the behaviour is still surprising.

```jsx
<Badge>£10</Badge>
```

Probably similar to #296

glenjamin:
The new test fails without the code change.


mtscout6:
Thanks for the fix!

mtscout6:
Fixes #540

PR #558, opened 2015-04-23
[fixed] ListGroup work with single child contained in array

### LINKED ISSUE #548: ListGroup fails if list contains a single entry
It looks like this bug was introduced by f52380584f842b9db5efae4a96d77369234d5828. I suspect that

    let child = this.props.children;

should be 

    let child = this.props.children[0];

cdock1029:
Check if child is array before accessing props

fixes #548

### DIFF
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -13,7 +13,7 @@ class ListGroup extends React.Component {
 
     if (!this.props.children) {
       return this.renderDiv(items);
-    } else if (React.Children.count(this.props.children) === 1) {
+    } else if (React.Children.count(this.props.children) === 1 && !Array.isArray(this.props.children)) {
       let child = this.props.children;
 
       childrenAnchors = child.props.href ? true : false;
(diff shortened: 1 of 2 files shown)


mtscout6:
This looks great! Can you squash the commits and prefix your commit message with [fixed] as per the [contributing guidelines (Commit Subjects for Public API Changes section)](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) at the root of the repo?

mtscout6:
:+1: 

Needs one more approval from another contributor @react-bootstrap/collaborators

dozoisch:
LGTM

mtscout6:
@cdock1029 Did you see @leth's additonal issues in #548? Does that seem reasonable to also tackle with this PR?

cdock1029:
Yes I can work on that 



—

mtscout6:
Thanks!

AlexKVal:
Comfirm. :+1: 

Before this patch console in browser:
```
Uncaught TypeError: Cannot read property 'href' of undefined
```
error links to the line of code
```js
childrenAnchors = child.props.href ? true : false
```
https://github.com/react-bootstrap/react-bootstrap/blob/v0.21.0/src/ListGroup.js#L19

After this patch both test cases are working OK.

mikechau:
Also running into this issue, patch resolves the problem for me. :+1:

mtscout6:
Thanks for the fix!

PR #565, opened 2015-04-24
Dry run and verbose flags for release process

### LINKED ISSUE #563: Add --dry-run flag to release tooling
The current release tooling is currently sketchy to test, especially since it's usage is very public in nature. Adding a `--dry-run` flag that simply skips any steps that tag, push, commit, or publish. An additional thing to skip would be the final cleanup that removes temporary folders. This way you can verify that the contents of these temporary folders are as you expect.

Original conversation for this from #547

mtscout6:
Closes #563


mtscout6:
@dozoisch How does this look to you?

mtscout6:
@AlexKVal If you're still trying to tackle the docs version behind by one bug, than this change should make that testing easier.

dozoisch:
Looks good :+1: ! Really clean!

I like the added progress info!

PR #571, opened 2015-04-26
[fixed] ListGroup rendering a ul when ListGroupItem has an onClick handler

yeojz:
Currently 
```html
<ListGroup>
    <ListGroupItem onClick={() => null}>1st Child</ListGroupItem>
    <ListGroupItem>2nd Child</ListGroupItem>
</ListGroup>
```
will result in

```html
<ul class="list-group">
   <a class="list-group-item active">1st Child</a>
   <span class="list-group-item">2nd Child</span>
</ul>
```

This P/R fixes the above to render a `<div>` instead.

A test case has also been added to `ListGroupSpec.js`

### DIFF
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -16,14 +16,15 @@ class ListGroup extends React.Component {
     } else if (React.Children.count(this.props.children) === 1 && !Array.isArray(this.props.children)) {
       let child = this.props.children;
 
-      childrenAnchors = child.props.href ? true : false;
+      childrenAnchors = this.isAnchor(child.props);
 
     } else {
 
       childrenAnchors = Array.prototype.some.call(this.props.children, (child) => {
-        return !Array.isArray(child) ? child.props.href : Array.prototype.some.call(child, (subChild) => {
-            return subChild.props.href;
+        return !Array.isArray(child) ? this.isAnchor(child.props) : Array.prototype.some.call(child, (subChild) => {
+            return this.isAnchor(subChild.props);
         });
+
       });
 
     }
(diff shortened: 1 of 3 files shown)


apkiernan:
LGTM

src/ListGroup.js
```diff
@@ -35,6 +35,10 @@ class ListGroup extends React.Component {
     }
   }
 
+  isAnchor(props){
+    return (props.href || props.target || props.onClick);
```
mtscout6: `target` by itself wouldn't do anything. Probably best to exclude from this check?

yeojz: If that's the case, should the exclusion apply to `ListGroupItem.js` as well? 
Line 34 checks for target to render anchor tags.

mtscout6: I'd say yes.

mtscout6:
When you wrap this up can you rebase your changes on the latest master instead of merging master in? That way we can avoid multiple merge commits.

src/ListGroup.js
yeojz: Fixed.

mtscout6:
Looks like your rebase contains other commits, can you run an interactive rebase  only apply your commits and squash them.

AlexKVal:
I've merged the latest version of the patch locally.

`npm test`-ed.
`npm run docs` checked console in Chrome and Firefox.
`npm run build`-ed it.
`npm link`-ed to my pet project for additional checks.

All is green. :+1:

mtscout6:
Thanks for the fix!

Issue #487, opened 2015-04-07
Document recommended way to include glyphicons fonts.


ronjouch:
Hello.

I couldn't find documentation on the recommended way to include the `glyphicons` eot/svg/ttf/woff/woff2 font set (in a npm/browserify/gulp project).

I thought I'd just have a gulp task copying the fonts from where it lives in `node_modules` to my `dist/fonts` folder (right?),

* ... but `bootstrap` is a `devDependency` of `react-bootstrap`, so my project depending on just `react-bootstrap` doesn't pull it, and the fonts are nowhere to be found in my `node_modules` folder.
* ... and I could make my app depend on `bootstrap`, but then I have to ensure the version of my direct `bootstrap` dep. matches the version of `react-bootstrap`'s `bootstrap` dependency.

Am I missing something? Should I do something entirely different? I'll be glad to submit a PR on the documentation to clarify this.

mtscout6:
We leave it up to downstream users to determine how they get the bootstrap css and fonts. This is because many folks use custom bootstrap themes. We could state that clearer on the Getting Started page of the docs.

dozoisch:
What I personally do is the second one. React-Bootstrap doesn't really depend on any bootstrap version, unless there are big breaking change in the main lib (which are pretty rare). So I have bootstrap in my deps and I include bootstrap.less in my build.

ronjouch:
Thanks @mtscout6 @dozoisch, that helped. Here's a doc PR hopefully clarifying that.

ronjouch:
My initial pull request was funnily done on my repo (doh), here's a proper one.

PR #585, opened 2015-04-27
Getting Started: document concerns/recommendations on including Bootstrap

ronjouch:
As discussed in #487 and commented by @mtscout6 @dozoisch

### DIFF
--- a/docs/src/GettingStartedPage.js
+++ b/docs/src/GettingStartedPage.js
@@ -21,7 +21,12 @@ const Page = React.createClass({
                   <h2 id="setup" className="page-header">Setup</h2>
                   <p className="lead">You can import the lib as AMD modules, CommonJS modules, or as a global JS script.</p>
 
-                  <p>First add the Bootstrap CSS to your project; check <a href="http://getbootstrap.com/getting-started/" name="Bootstrap Docs">here</a> if you have not already done that. Then:</p>
+                  <p>First add the Bootstrap CSS to your project; check <a href="http://getbootstrap.com/getting-started/" name="Bootstrap Docs">here</a> if you have not already done that. Note that:</p>
+                  <ul>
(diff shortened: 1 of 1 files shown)


docs/src/GettingStartedPage.js
```diff
@@ -21,7 +21,12 @@ const Page = React.createClass({
                   <h2 id="setup" className="page-header">Setup</h2>
                   <p className="lead">You can import the lib as AMD modules, CommonJS modules, or as a global JS script.</p>
 
-                  <p>First add the Bootstrap CSS to your project; check <a href="http://getbootstrap.com/getting-started/" name="Bootstrap Docs">here</a> if you have not already done that. Then:</p>
+                  <p>First add the Bootstrap CSS to your project; check <a href="http://getbootstrap.com/getting-started/" name="Bootstrap Docs">here</a> if you have not already done that. Note that:</p>
+                  <ul>
+                    <li>Because many folks use custom bootstrap themes, we do not directly depend on Bootstrap. It is up to you to to determine how you get and link to the bootstrap css and fonts.</li>
+                    <li>React-Bootstrap doesn't really depend on any bootstrap version (unless there are big breaking changes in the main lib, which are pretty rare). Just have bootstrap in your dependencies and ensure your build can read your less/sass/scss entry point.</li>
```
mtscout6: This isn't necessarily true, we have built this library with Bootstrap v3.x in mind. I've had problems getting it to work with a legacy site we are migrating at work that was built against Bootstrap 2.x. It would probably be best to pull the current version of bootstrap we are running against and note that that is the version these docs are using. That way the version would stay dynamic with when we update our bootstrap dev dependency. It may be possible that this library will work with versions of bootstrap that are not too distant from the current one.

ronjouch: @mtscout6 okay, was just paraphrasing @dozoisch from https://github.com/react-bootstrap/react-bootstrap/issues/487#issuecomment-96770300 . Here's an updated PR that hopefully works better.

AlexKVal:
LGTM

dozoisch:
looks good

Issue #526, opened 2015-04-17
DropdownButtons need an overhaul. 


mtscout6:
I know there are a lot of problems with the DropdownButton, which I hope to remedy soon. I know I'll have some time to work on them next week, but I can't guarantee a delivery date. I'm mainly creating this issue to aggregate all the dropdown issues in one place.

- #195 
- #230 
- #243 
- #250 
- #253 
- #299 
- #303 
- #368 
- #398 
- #411 
- #421 
- #456 
- #458 
- #470 
- #100 
- #486
- #522
- #629

Notes as per W3C WAI-ARIA Spec:

- [ ] When presenting the menu, ensure that it is completely visible on screen.
- Keyboard navigation. Should resolve #588
  - [x] Space or Enter - With focus on the button pressing Space or Enter will toggle the display of the drop-down menu. Focus remains on the button.
  - Down Arrow
    - [x] With focus on the button and no drop-down menu displayed, pressing Down Arrow will open the drop-down menu and move focus into the menu and onto the first menu item.
    - [x] With focus on the button and the drop-down menu open, pressing Down Arrow will move focus into the menu onto the first menu item.
  - Up and Down Arrow
    - [x] With focus on the drop-down menu, the Up and Down Arrow keys move focus within the menu items, "wrapping" at the top and bottom.
  - Escape
    - [x] With focus on the drop-down menu, pressing Escape closes the menu and returns focus to the button. Resolves #486 
  - Tab
    - [x] With focus on the button pressing the Tab key will take the user to the next tab focusable item on the page.
    - [x] With focus on the drop-down menu, pressing the Tab key will take the user to the next tab focusable item on the page. Note that this may be difficult to achieve on a web page.
    - [ ] Typing a letter (printable character) key moves focus to the next instance of a visible node whose title begins with that printable letter.
- [x] ~~The menu button itself has a role of button.~~ It is a button so not needed.
- [x] The menu button has an aria-haspopup property, set to true.
- [ ] Auto-generate an ID if one is not provided.
  - [ ] Ensure that id does not break server -> client rendering (See: facebook/react#4000)

Notes per reported issues:

- [x] Honor event handling api discussed in #419. Should resolve problem with:
  - #195
  - #243
  - #250
  - #368
  - #411
  - #456
  - #522 
  - #541 
  - #573 
  - #678
- [x] Allow custom className to DropdownMenu. Should resolve #230, #253, related to #595 
[...]

PR #588, opened 2015-04-28
[fixed] Keyboard traversing on DropdownMenu

christianalfoni:
The original bootstrap allows for keyboard traversing on dropdowns. This is an important feature as many solutions allow for keyboard traversing in the interface.

React is not able to use state to handle what element to focus, so this has to be handled "outside of React". This is the strategy:

1. Make the dropdown menu focusable, but not tabable, by using `tabIndex="-1"`. This will enable the ability to listen for keydown events

2. When a dropdown menu is opened focus the menu, waiting for keydown events

3. When a 38/40 key event is registered jump to previous or next possible focusable element, which is hyperlinks (a tag). This has to be checked on each keydown as there can be dividers in between the focusables and the list itself might change

Created a test which has to be written a bit differently due to testing focus. But it should all work :-)

### DIFF
--- a/src/DropdownMenu.js
+++ b/src/DropdownMenu.js
@@ -4,10 +4,64 @@ import classNames from 'classnames';
 import createChainedFunction from './utils/createChainedFunction';
 import ValidComponentChildren from './utils/ValidComponentChildren';
 
+const traverseKeys = [38, 40];
+
 const DropdownMenu = React.createClass({
   propTypes: {
     pullRight: React.PropTypes.bool,
-    onSelect: React.PropTypes.func
+    onSelect: React.PropTypes.func,
+    style: React.PropTypes.object
+  },
+
+  handleMenuKeyDown(event) {
+
+    if (event.keyCode === 13) {
+      return;
+    }
+
+    event.preventDefault();
+
+    // Go to previous item if possible
+    if (event.keyCode === traverseKeys[0]) {
+      this.focusPrevious();
+    }
+
+    // Go to next item if possible
+    if (event.keyCode === traverseKeys[1]) {
+      this.focusNext();
+    }
+  },
+
+  getFocusableMenuItems() {
+
+    // Grab hyperlinks and convert to array
+    return [].slice.call(this.getDOMNode().querySelectorAll('a'), 0);
+  },
+
+  focusNext() {
+
+    let activeElement = document.activeElement;
+    let items = this.getFocusableMenuItems();
+    let activeItemIndex = items.indexOf(activeElement);
+
+    if (activeItemIndex !== items.length - 1) {
+      items[activeItemIndex + 1].focus();
+    }
+
+  },
+
+  focusPrevious() {
+
+    let activeElement = document.activeElement;
+    let items = this.getFocusableMenuItems();
+    let activeItemIndex = items.indexOf(activeElement);
+
+    if (activeItemIndex === -1) {
+      items[items.length - 1].focus();
+    } else if (activeItemIndex !== 0) {
+      items[activeItemIndex - 1].focus();
+    }
+
   },
 
   render() {
@@ -16,11 +70,19 @@ const DropdownMenu = React.createClass({
         'dropdown-menu-right': this.props.pullRight
       };
 
+    // Remove outline on focus by default
+    let style = this.props.style || {};
+    style.outline = style.outline || 'none';
+
+    // Adding tabIndex -1 to allow for focus, but not tabable
     return (
         <ul
+          tabIndex="-1"
+          onKeyDown={this.handleMenuKeyDown}
           {...this.props}
           className={classNames(this.props.className, classes)}
-          role="menu">
+          role="menu"
+          style={style}>
           {ValidComponentChildren.map(this.props.children, this.renderMenuItem)}
         </ul>
       );
@@ -30,11 +92,13 @@ const DropdownMenu = React.createClass({
     return cloneElement(
       child,
       {
+
         // Capture onSelect events
         onSelect: createChainedFunction(child.props.onSelect, this.props.onSelect),
 
         // Force special props to be transferred
         key: child.key ? child.key : index
+
       }
     );
   }
--- a/src/DropdownStateMixin.js
+++ b/src/DropdownStateMixin.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import domUtils from './utils/domUtils';
 import EventListener from './utils/EventListener';
+import createChainedFunction from './utils/createChainedFunction';
 
 /**
  * Checks whether a node is within
@@ -37,7 +38,11 @@ const DropdownStateMixin = {
 
     this.setState({
       open: newState
-    }, onStateChangeComplete);
+    }, newState && !this.state.open ? createChainedFunction(this.focusMenu, onStateChangeComplete) : onStateChangeComplete);
+  },
+
+  focusMenu() {
+    this.refs.menu.getDOMNode().focus();
   },
 
   handleDocumentKeyUp(e) {
--- a/test/DropdownButtonSpec.js
+++ b/test/DropdownButtonSpec.js
@@ -201,4 +201,37 @@ describe('DropdownButton', function () {
     let carets = button.getElementsByClassName('caret');
     assert.equal(carets.length, 0);
   });
+
+  it('should focus dropdown items when pressing cursor up and down keys', function () {
+
+    // Have to run this in actual DOM to verify focus
+    let testContainer = document.createElement('div');
+    document.body.appendChild(testContainer);
+
+    // Create instance as normal, only in DOM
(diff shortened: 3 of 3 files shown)


src/DropdownMenu.js
```diff
@@ -16,11 +70,19 @@ const DropdownMenu = React.createClass({
...
       };
 
+    // Remove outline on focus by default
+    let style = this.props.style || {};
+    style.outline = style.outline || 'none';
+
+    // Adding tabIndex -1 to allow for focus, but not tabable
     return (
         <ul
+          tabIndex="-1"
```
mtscout6: Shouldn't the individual `li`s need this?

src/DropdownMenu.js
```diff
@@ -16,11 +70,19 @@ const DropdownMenu = React.createClass({
         'dropdown-menu-right': this.props.pullRight
       };
 
+    // Remove outline on focus by default
+    let style = this.props.style || {};
+    style.outline = style.outline || 'none';
```
mtscout6: We shouldn't need this if the the `tabIndex='-1'` is applied to the `li`s.

src/DropdownMenu.js
christianalfoni: Ah, no, that tabIndex allows the list itself to be focusable so that it can listen to keyDown events. The items in the list are focused when those keyDown events occur. The items do not need "tabIndex" as they are "a"-tags. Should be very careful messing around with tabIndexes too as those are global, only "-1" is safe as it does not interfere with tabbing, only makes the element focusable.

The original bootstrap listens tot he keyDown events on document and each dropdown instance checks if it is open and then reacts to that. But since the react-bootstrap dropdowns themselves do not know if they are open or not, only their parent does, I chose to focus the menu itself to listen for those keyDown events.

An alternative would be to pass the "open" state of the parent to the menu, but that is less flexible I think.

mtscout6:
I was right in the middle of working on this very thing :wink:. Related to #526.

christianalfoni:
Oh, hehe, hopefully it has some value :-)

src/DropdownMenu.js
mtscout6: Ok, that makes sense. In that case, i'd rather not expose the style prop. We don't really expose that anywhere else.

src/DropdownMenu.js
christianalfoni: You need to be able to listen for the keyDown events for traversing though. Not sure how you can do that without focusing the list element itself? Maybe you have some global stuff to handle that?

src/DropdownMenu.js
mtscout6: Actually you know what each of those anchors should have `tabIndex='-1'` as per the W3C WAI-ARIA Spec. See: http://www.w3.org/WAI/PF/aria-practices/#menubutton

christianalfoni: Ah, yeah, I was a bit unsure about the style prop.

Hm hm... this is a tricky one. 

*Authors can use either the JavaScript focus() method to move focus to the appropriate element in the widget, or they can use a WAI-ARIA property called aria-activedescendant on the actual widget container to indicate which descendant element in the widget has focus. Use the following procedure when focus is completely dependent on the use of tabindex to provide focus in a widget:

Set tabindex="0" on the current active descendant in the container widget while setting tabindex="-1" on all the other descendant elements of the widget.

As the user navigates (e.g., arrows) away from an item, set tabindex="-1" on the old item, and tabindex="0" on the new.

Use the JavaScript focus() method to put focus on the new item, whose tabindex="0".*

Hm, if I were to build a dropdown from scratch I would indeed use tabIndex='-1' on each `li` element, but bootstrap relies on focusing the `a` elements in the list to style correctly. 

This implementation follows the practice of original bootstrap at least, so yeah... going away from `a` focus to `li` with tabIndex="-1" to focus those will not behave like Boostrap regarding styles.

christianalfoni: Or maybe that was something else? Soooo much information... ;-)

mtscout6:
If you don't mind I can continue this from my end, I'm in the middle of overhauling everything about the DropdownButton since it's been having all sorts of problems. We really appreciate the effort though, and you helped me figure out how to test the focused state, that's what I was struggling with.

christianalfoni:
No worries, just glad to help :-)

mtscout6:
There are more components that could use keyboard improvements should you care to help with those. This is part of our [1.0.0 Roadmap](https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap) to ensure each component adheres to the W3C WAI-ARIA spec. If you want to create an issue for the one you're working on so we don't step on each others toes through all this.

christianalfoni:
Ah, cool! To be perfectly honest I would love to contribute, but very unsure I have the time. Got this one through due to it being a fix for our current project, but I will of course contribute if I meet similar situations :-)

you guys are doing fantastic job with this, especially setting up the repo with the tests, linting etc., great stuff! so looking very much forward to 1.0.0.

Again, I will gladly help if I find the time!

src/DropdownMenu.js
mtscout6: No worries, I appreciate the reference.

mtscout6:
@christianalfoni It's still a work in progress but you can see what I've got so far here: https://github.com/react-bootstrap/react-bootstrap/blob/dropdown-revisited/src/revisited/DropdownButton.js

I'm re-writing that component from the ground up, so it's nowhere near what it will look like in the end, but the keyboard navigation is working as I like. You had some significant influence in the arrow navigation logic.

christianalfoni:
Hi @mtscout6,

dang... I should have patented the approach ;-)

I think it is looking good! As I understand the code the "MenuItem" Component has the "tabIndex=-1" attribute?

I have one question though. Is there any reason why DropdownButton and DropdownMenu is so tightly coupled? I think a better approach might be to move everything that has to do with the DropdownMenu into DropdownMenu and just do:

```js
render() {
    let id = this.props.id || this.state.id;

    const rootClasses = {
      dropdown: true,
      open: this.state.open
    };

    return (
      <div className={classNames(rootClasses)}>
        <button
          ref='toggle-btn'
          className='btn btn-default dropdown-toggle'
          onClick={this.toggleOpen}
          onKeyDown={this.handleKeyDown}
          onBlur={this.handleBlur}
          type='button'
          id={id}
          tabIndex='0'
          aria-expanded={this.state.open}>
          <span>Dropdown </span>
          <span className='caret'></span>
        </button>
       {this.state.open ? <DropdownMenu>{this.props.children}</DropdownMenu> : null}
      </div>
    );
  }
```

That way you can use the DropdownMenu wherever you want and toggling its "open" is a matter of rendering it or not. On the DropdownMenu `componentDidMount` you should then be able to natively focus on the `ul` listening to keypresses for traversal.

Just putting in my two cents on it, hope it made sense. Could make an example if some parts it unclear :-)

mtscout6:
What you're saying does make sense. The issue I'm trying to deal with is that the toggle button needs to respond most of the same keys. For example, when open and focus is still on the button and the tab key is pressed we are supposed to close it. The same applies if focus is on one of the selected items. I still have a lot of cleanup to do on this, and many more tests to write so it's in no way complete yet. I do want to keep as much of it DRY as possible though as I go along.

Yes the menu item is the one with `tabIndex='-1'` which only means that focus can only be achieved programatically. 

I will get to having a DropdownMenu, I'm just taking things a little bit at a time. I'm completely re-writing that component from the ground up.

christianalfoni:
Hi Matt!

Yeah, I can imagine it is a challenge nailing that DRY-ness of a library like react-bootstrap. Especially behaviour across components, like the dropdown with the tabbing and click to close all etc. :-)

Let me know if you meet any specific challenges and want some input!

mtscout6:
@christianalfoni I know you need this, and I don't know exactly when I'll be finishing up the DropdownButton re-write. How do you feel about us merging this in knowing that it may slightly change when I finish up my work? How do the other @react-bootstrap/collaborators feel about such a proposal?

mtscout6:
@christianalfoni If you're up for that, there is a merge conflict with master if you don't mind rebasing on the latest master.

christianalfoni:
Hi again @mtscout6 , thanks for considering it, but I actually implemented a bit of custom stuff into our project today, so we are set :-) Maybe I could share that with you, for reference? Maybe it has something you can use in your rewrite?

Thanks again!

mtscout6:
By all means I'd love to see what you've got

christianalfoni:
I see this was fixed? Sorry, I got occupied with lots of other stuff. Let me know if you still would like some input ;-)

mtscout6:
Unfortunately no it's not fixed yet, just closed the PR. I'm really trying to push myself to have the first round of the DropdownButton rewrite out within the next week.

christianalfoni:
Cool, just let me know if you want input on some stuff!

PR #590, opened 2015-04-28
Add Footer component

nax83:
In order to use the Sticky feature, you need to add the following css to your index.html

http://getbootstrap.com/examples/sticky-footer/sticky-footer.css


mtscout6:
http://getbootstrap.com/getting-started/#examples

Bootstrap mentions that those are examples of things you can do, but are not part of the core library. Since we are seeking parity with that which provided by the core Bootstrap library, I don't feel that this is the right place for this.

dozoisch:
Should we add that to documentation?

PR #595, opened 2015-04-29
Support adding button class to DropdownButton

santiagoaguiar:
As mentioned on #376, making rounded buttons using classes becomes harder unless we can add custom classes to the wrapped Button element. 

Before fixing #376 this worked OK, but the change broke existing code (ie. code that assumed the classes went to the Button element) that cannot be fixed unless this property is introduced.

### DIFF
--- a/src/DropdownButton.js
+++ b/src/DropdownButton.js
@@ -20,7 +20,8 @@ const DropdownButton = React.createClass({
     onClick:   React.PropTypes.func,
     onSelect:  React.PropTypes.func,
     navItem:   React.PropTypes.bool,
-    noCaret:   React.PropTypes.bool
+    noCaret:   React.PropTypes.bool,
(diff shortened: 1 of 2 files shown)


dozoisch:
LGTM

mtscout6:
LGTM

For personal reference when revisiting DropdownButtons I'm going to link this to #526

mtscout6:
One last thing for the release notes can you amend your commit message to prefix it with [added] so the change is reflected in the changelog. For reference see the contributing guide about this: https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes

santiagoaguiar:
@mtscout6 done, hope I did it right, first time doing a PR of more than 1 line of code ;).

mtscout6:
:+1: You did it right, thanks!

PR #599, opened 2015-05-01
[IE9] Check for console before using console.warn

cymen:
I realize the check for `NODE_ENV === production` fixes the issue for production builds but I regularly test with IE9 locally while developing.


cymen:
@AlexKVal Are you onboard with this?

AlexKVal:
Good catch! Thank you :smile: :+1:

mtscout6:
LGTM

Issue #597, opened 2015-04-30
can't set id of ListGroup


randrianov:
React.createElement(ListGroupItem, {id: '1'}) doesn't set 'id' html attr

mtscout6:
@apkiernan Looks like your commit message missed the last piece of information of the change.

apkiernan:
Yeah will clean that up before submitting the PR.

mtscout6:
:+1:

apkiernan:
Closed #601

PR #601, opened 2015-05-01
[fixed] #597 able to set ID on ListGroup

apkiernan:
Fixed `ListGroup` render to allow setting id on element through spread `{...this.props}` on element.


dozoisch:
LGTM

mtscout6:
LGTM

PR #602, opened 2015-05-01
[added] Optionally allow Modal content styles to be set

cymen:
I would like to be able to set the .modal-content style in some cases. For example, I am using react-select in a modal and want the autocomplete pane to visibly extend beyond the bottom of the modal.

I updated the documentation but am unsure if I should have for an optional thing like this.

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -217,7 +217,7 @@ const ComponentsPage = React.createClass({
                   <h1 id='modals' className='page-header'>Modals <small>Modal</small></h1>
 
                   <h3 id='modals-static'>A static example</h3>
-                  <p>A rendered modal with header, body, and set of actions in the footer.</p>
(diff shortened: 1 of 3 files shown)


cymen:
I can of course use something like:
```css
.modal-content {
  overflow: visible !important;
}
```
But I'd like to avoid using `!important` if possible.

mtscout6:
The bigger question I have is why we are even using style so much in that component, and not relying on the bootstrap css? Can you look into that and see if it's possible to just remove them?

cymen:
@mtscout6 Good point -- I looked at the commit on react-bootstrap that added the class and it didn't explain why (and was doing a lot of other things). So I assumed it was for a good reason and at least made it configurable...

However, on vanilla bootstrap, I'm not seeing `overflow: hidden` applied to `.modal-content` in normal usage. I tried removing it from react-bootstrap and it seems to be fine without it. So I will close this PR and open another to just remove the hidden class.

cymen:
Opened https://github.com/react-bootstrap/react-bootstrap/pull/605 for removal of the hidden class.

PR #604, opened 2015-05-03
[fixed] failing test of deprecation

dozoisch:
I think the __deprecated__ applied to the Collapsible also once Collapsable was being imported. So instead of setting a prop I made wrapper component that exposes the two functions that can be called. I also re-added these two functions with warning to the Collapsible component.

### DIFF
--- a/src/CollapsableNav.js
+++ b/src/CollapsableNav.js
@@ -1,7 +1,33 @@
+import React from 'react';
 import CollapsibleNav from './CollapsibleNav';
+import deprecationWarning from './utils/deprecationWarning';
 
-let CollapsableNav = CollapsibleNav;
+const deprecationLink = 'https://github.com/react-bootstrap/react-bootstrap/issues/425#issuecomment-97110963';
 
-CollapsableNav.__deprecated__ = true;
+const CollapsableNav = React.createClass({
+  getCollapsableDOMNode() {
+    return this.refs.collapsible.getCollapsableDOMNode();
+  },
+
+  getCollapsableDimensionValue() {
+    return this.refs.collapsible.getCollapsableDimensionValue();
+  },
+
+  componentDidMount() {
+    deprecationWarning(
+      'CollapsableNav',
+      'CollapsibleNav',
+      deprecationLink
+    );
+  },
+
+  render() {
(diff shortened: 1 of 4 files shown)


AlexKVal:
@dozoisch I propose another solution. The simpler one ;)
https://github.com/react-bootstrap/react-bootstrap/pull/608

dozoisch:
There's one thing missing, though we have changed  2 functions name that were exposed on the component without deprecating them!

https://github.com/react-bootstrap/react-bootstrap/pull/604/files#diff-c5a88cb9e192df4b2a4e2ed9a0a18884R57

AlexKVal:
They are methods of `CollapsibleMixin`.
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/CollapsibleNav.js#L12

And they warns about deprecation ;)
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/CollapsableMixin.js#L8
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/CollapsableMixin.js#L17
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/CollapsableMixin.js#L26

And their deprecation is also tested
https://github.com/react-bootstrap/react-bootstrap/blob/master/test/CollapsibleMixinSpec.js#L221
https://github.com/react-bootstrap/react-bootstrap/blob/master/test/CollapsibleMixinSpec.js#L240
and by such line in every test-case in `CollapsableMixinSpec`
https://github.com/react-bootstrap/react-bootstrap/blob/master/test/CollapsableMixinSpec.js#L62
https://github.com/react-bootstrap/react-bootstrap/blob/master/test/CollapsableMixinSpec.js#L159+L164
and so on ;)

By this patch commit  ;)
https://github.com/react-bootstrap/react-bootstrap/commit/f1564eedb03#diff-a6edac1fa9696d775c913464da4c8065R43

And you have approved it :stuck_out_tongue_winking_eye: 
https://github.com/react-bootstrap/react-bootstrap/pull/591#issuecomment-97302782

dozoisch:
Oh alright! I didn't check the Mixin! LG then

dozoisch:
Closing as 608 fixes it.

Issue #448, opened 2015-03-11
Reason for (overflow:hidden) change on Modal 


kRITZCREEK:
When updating react-bootstrap in my Application this change broke my UI.

[Modal.jsx, line-61](https://github.com/react-bootstrap/react-bootstrap/commit/13f395d2304727d65ee22242423150c22b73addb#diff-3f8ea8f4138955e64a0c26a6e9f8ac24R61)

What's the particular reason for this change (also the commit message doesn't reflect it)?

![modal-overflow](https://cloud.githubusercontent.com/assets/6189397/6595154/e204f2a4-c7e7-11e4-99cb-a4a34dac2291.png)

This was the behaviour before the change.

![modal-overflow-hidden](https://cloud.githubusercontent.com/assets/6189397/6595157/f49e4db6-c7e7-11e4-9ccc-ab5a008270fa.png)

Afterwards the Timepicker Component gets chopped. It's mainly a Question since i don't really mind using an older version.

mtscout6:
Looks like an oversight, that should be removed.

stefanfisk:
@mtscout6 that was a major oversight form my part. my PR set overflow: hidden for the whole modal. a solution would be to set it only if bsStyle is passed, and document accordingly. Will get a PR for it when I have the time, hopefully this weekend. if you don't feel like waiting you can revert my code in the meantime, likely not too many people have even noticed it.

kRITZCREEK:
@stefanfisk any updates on the matter?

stefanfisk:
I’m afraid not, as I’m currently tied up in a bunch of work.

But as I said, just reverting the commit would at least remove the regression it introduced.

taion:
This was resolved by #605.

PR #605, opened 2015-05-03
[fixed] Modal div.modal-content should not have hidden class

cymen:
Vanilla Bootstrap Modal does not have the CSS class `hidden` by default on the `div.modal-content`. It was added to react-bootstrap in commit https://github.com/react-bootstrap/react-bootstrap/commit/13f395d2304727d65ee22242423150c22b73addb that added support for bsStyle property to Modal for setting the header color. There were a number of changes and why `hidden` was added to `div.modal-content` isn't clear.

My reason for not wanting this is when using say react-select inside a modal, the dropdown should be visible in the event it overflows the modal.

### DIFF
--- a/docs/examples/ModalContained.js
+++ b/docs/examples/ModalContained.js
@@ -12,7 +12,7 @@
 const ContainedModal = React.createClass({
   render() {
     return (
-      <Modal {...this.props} bsStyle='primary' title='Contained Modal' animation>
+      <Modal {...this.props} title='Contained Modal' animation>
         <div className='modal-body'>
           Elit est explicabo ipsum eaque dolorem blanditiis doloribus sed id ipsam, beatae, rem fuga id earum? Inventore et facilis obcaecati.
         </div>
--- a/docs/examples/ModalOverlayMixin.js
+++ b/docs/examples/ModalOverlayMixin.js
@@ -28,7 +28,7 @@ const CustomModalTrigger = React.createClass({
     }
 
     return (
-      <Modal bsStyle='primary' title='Modal heading' onRequestHide={this.handleToggle}>
+      <Modal title='Modal heading' onRequestHide={this.handleToggle}>
         <div className='modal-body'>
           This modal is controlled by our custom trigger component.
         </div>
--- a/docs/examples/ModalStatic.js
+++ b/docs/examples/ModalStatic.js
@@ -5,7 +5,6 @@ function handleHide() {
 const modalInstance = (
   <div className='static-modal'>
     <Modal title='Modal title'
-      bsStyle='primary'
       backdrop={false}
       animation={false}
       container={mountNode}
--- a/docs/examples/ModalTrigger.js
+++ b/docs/examples/ModalTrigger.js
@@ -1,7 +1,7 @@
 const MyModal = React.createClass({
   render() {
     return (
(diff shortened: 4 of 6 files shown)


dozoisch:
don't worry about failing tests, I will rerun the tests once #604 is merged

AlexKVal:
I've re-run TravisCI tests. They are green now.

AlexKVal:
to @cymen
> ... why `hidden` was added to `div.modal-content` isn't clear.

The reason is https://github.com/react-bootstrap/react-bootstrap/pull/403#issuecomment-75965860

And then "Reason for (overflow:hidden) change on Modal #448" issue with the bug has been created.

Author has acknowledged the bug https://github.com/react-bootstrap/react-bootstrap/issues/448#issuecomment-78350077 and he just have no time to fix it https://github.com/react-bootstrap/react-bootstrap/issues/448#issuecomment-86364657.

Therefore it seems you need to make sure that your solution doesn't introduce regression for https://github.com/react-bootstrap/react-bootstrap/pull/403#issuecomment-75964220

And if you address all of it, chances are, you also solve #448 :sparkles:

mtscout6:
@stefanfisk This looks like the beginning of your bug fix. Do you have any input on this?

cymen:
@AlexKVal Thanks for digging those links up -- I didn't think to check in the right place and now it makes sense.

AlexKVal:
@cymen You are welcome :sparkles:

cymen:
This PR does cause a regression on as illustrated in #403. I'll keep working on it -- I think bootstrap modal does indeed to the workaround of only applying `overflow: hidden` if a bsStyle is applied. I'm confirming that and will update this PR.

cymen:
The fix for the regression is to apply to `.modal-header` if `bsStyle` is a property of Modal this CSS:

```css
    border-top-left-radius: inherit; 
    border-top-right-radius: inherit;
```

This is IE 9+.

After researching this more, I would advocate for react-bootstrap not supporting bsStyle on the modals (vanilla bootstrap does not and I think that is likely for good reasons). It causes a headache that is best pushed down to those consuming react-bootstrap. But maybe the above solution is acceptable.

mtscout6:
@cymen I created twbs/bootstrap#16431 let's see where that goes.

AlexKVal:
> I would advocate for react-bootstrap not supporting bsStyle on the modals (vanilla bootstrap does not and I think that is likely for good reasons). It causes a headache that is best pushed down to those consuming react-bootstrap.

:+1:

mtscout6:
I would agree with that, @stefanfisk I'm sorry but we will need to take this out.

stefanfisk:
No worries, it was a quick and dirty idea that turned out to be terrible.

mtscout6:
@cymen can you change the docs to not be demonstrating use of bsStyle, and remove these two lines: https://github.com/react-bootstrap/react-bootstrap/blob/v0.21.2/src/Modal.js#L102-L103. Then this should be ready to merge in.

cymen:
@mtscout6 Sure -- there were a few more changes but I think I got everything in 379ea3ecbcf0cb2838a6cf85fc8fd9a3d07240ea

AlexKVal:
LGTM :+1:

PR #606, opened 2015-05-03
[added] property disabled on MenuItem

### LINKED ISSUE #246: Disable MenuItem
I have hacked my MenuItem code to include a `propTypes { disabled: React.PropTypes.bool }`

In recent browsers, (Chrome, IE11) the `transferPropsTo` in the render method produces <li disabled> which renders them as disabled with a "no-entry" icon, but earlier browsers don't respond to the disabled attribute. I have therefore added disabled to the class map:

    var classes = {
        'dropdown-header': this.props.header,
        'divider': this.props.divider,
        'disabled': this.props.disabled
      };

and I have CSS:

```
.disabled {
	color: lightgray;
	cursor: not-allowed;
}
```

But in general, it might be better to allow arbitrary classes to be passed down to the `<li>` - what do you think?

dozoisch:
The test will currently fail, but I'll rerun them when https://github.com/react-bootstrap/react-bootstrap/pull/604 gets merged.

closes #246

### DIFF
--- a/docs/examples/MenuItem.js
+++ b/docs/examples/MenuItem.js
@@ -1,0 +1,26 @@
+function onSelectAlert(eventKey, href, target) {
+  alert('Alert from menu item.\neventKey: "' + eventKey + '"\nhref: "' + href + '"');
+}
+
+function preventDefault() {}
+
+const MenuItems = (
+  <div className="clearfix">
+    <ul className="dropdown-menu open">
+      <MenuItem header>Header</MenuItem>
+      <MenuItem onSelect={preventDefault}>link</MenuItem>
+      <MenuItem divider/>
+      <MenuItem header>Header</MenuItem>
+      <MenuItem onSelect={preventDefault}>link</MenuItem>
+      <MenuItem disabled>disabled</MenuItem>
+      <MenuItem title="See? I have a title." onSelect={preventDefault}>
+        link with title
+      </MenuItem>
+      <MenuItem eventKey={1} href="#someHref" onSelect={onSelectAlert}>
+        link that alerts
+      </MenuItem>
+    </ul>
+  </div>
+);
+
+React.render(MenuItems, mountNode);
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -170,6 +170,25 @@ const ComponentsPage = React.createClass({
                   <ReactPlayground codeText={Samples.SplitButtonRight} />
                 </div>
 
+                {/* Menu Item */}
+                <div className='bs-docs-section'>
+                  <h1 id='menu-item' className='page-header'>Menu Item <small> MenudItem</small></h1>
+                  <p>This is a component used in other components (see <a href="buttons">Buttons</a>, <a href="#navbars">Navbars</a>).</p>
+                  <p>It supports the basic anchor properties <code>href</code>, <code>target</code>, <code>title</code>.</p>
+                  <p>It also supports different properties of the normal Bootstrap MenuItem.
+                    <ul>
+                      <li><code>header</code>: To add a header label to sections</li>
+                      <li><code>divider</code>: Adds an horizontal divider between sections</li>
+                      <li><code>disabled</code>: shows the item as disabled, and prevents the onclick</li>
+                      <li><code>eventKey</code>: passed to the callback</li>
+                      <li><code>onSelect</code>: a callback that is called when the user clicks the item.</li>
+                    </ul>
+                  <p>The callback is called with the following arguments: <code>eventKey</code>, <code>href</code> and <code>target</code></p>
+                  </p>
+                  <ReactPlayground codeText={Samples.MenuItem} />
+                </div>
+
+              {/* Panels */}
                 <div className='bs-docs-section'>
                   <h1 id='panels' className='page-header'>Panels <small>Panel, PanelGroup, Accordion</small></h1>
 
@@ -607,6 +626,7 @@ const ComponentsPage = React.createClass({
                   <code>getValue()</code> will not work when used this way.</p>
                   <ReactPlayground codeText={Samples.InputWrapper} />
                 </div>
+
               </div>
 
               <div className='col-md-3'>
@@ -623,6 +643,7 @@ const ComponentsPage = React.createClass({
                     <SubNav href='#buttons' key={1} text='Buttons'>
                       <NavItem href='#btn-groups' key={2}>Button groups</NavItem>
                       <NavItem href='#btn-dropdowns' key={3}>Button dropdowns</NavItem>
+                      <NavItem href='#menu-item' key={25}>Menu Item</NavItem>
                     </SubNav>
                     <NavItem href='#panels' key={4}>Panels</NavItem>
                     <NavItem href='#modals' key={5}>Modals</NavItem>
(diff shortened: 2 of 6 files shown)


docs/src/ComponentsPage.js
```diff
@@ -574,6 +574,24 @@ const ComponentsPage = React.createClass({
                   <code>getValue()</code> will not work when used this way.</p>
                   <ReactPlayground codeText={Samples.InputWrapper} />
                 </div>
+
+                {/* Menuitem */}
+                <div className='bs-docs-section'>
+                  <h1 id='menu-item' className='page-header'>MenuItem</h1>
```
mtscout6: For consistency this title should probably follow the same format as [List group](http://localhost:4000/components.html#listgroup). Note an English friendly spaced variant with `small` component names to its side.

docs/src/ComponentsPage.js
```diff
@@ -574,6 +574,24 @@ const ComponentsPage = React.createClass({
...
+                      <li><code>header</code>: To add a header label to sections</li>
+                      <li><code>divider</code>: Adds an horizontal divider between sections</li>
+                      <li><code>disabled</code>: shows the item as disabled, and prevents the onclick</li>
+                      <li><code>eventKey</code>: passed to the callback</li>
+                      <li><code>onSelect</code>: a callback that is called when the user clicks the item.</li>
+                    </ul>
+                  <p>The callback is called with the following arguments: <code>eventKey</code>, <code>href</code> and <code>target</code></p>
+                  </p>
+                  <ReactPlayground codeText={Samples.MenuItem} />
+                </div>
```
mtscout6: I wonder about positioning of this section. Should we put this next to either the dropdown buttons or nav sections?

dozoisch: Yeah I was wondering where to put it. I guess we should put it next to dropdown like bootstrap does.

docs/src/ComponentsPage.js
dozoisch: Alright I will fix it.

docs/src/ComponentsPage.js
mtscout6: :+1:

dozoisch:
alright Ill try to address the issue when I get some time today

dozoisch:
@mtscout6  with the revisited MenuItem is this still needed? Or should I implemented in the dropdown-reviited branch?

mtscout6:
I think this can come into master now, You can add it to the dropdown-revisited branch as well if you like. I've been rebasing and force pushing that branch regularly to keep it up to date. If you don't push it there then I'll be sure to account for it.

mtscout6:
I added a check to make sure I account for it.

dozoisch:
Alright, I'll clean it tonight so we can merge it

dozoisch:
@mtscout6 alright, it's updated now!

docs/src/ComponentsPage.js
```diff
@@ -170,6 +170,25 @@ const ComponentsPage = React.createClass({
                   <ReactPlayground codeText={Samples.SplitButtonRight} />
                 </div>
 
+                {/* Menu Item */}
+                <div className='bs-docs-section'>
+                  <h1 id='menu-item' className='page-header'>Menu Item <small> MenudItem</small></h1>
+                  <p>This is a component used in other components (see <a href="buttons">Buttons</a>, <a href="#navbars">Navbar</a>).</p>
```
AlexKVal: `#navbars">Navbar<a`
Navbar => Navbars

AlexKVal:
From the English language point of view
`active` vs `disabled` :smile: two properties.

And both are legitimate from `lib` point of view.

I'm not sure does it need a documenting note about them.
In theory `Bootstrap` users have to know about this `active` and `disabled` as classes
and therefore they should not have questions about it.

AlexKVal:
`This is a component used in other components (see Buttons, Navbar).`
It is added into `Buttons` submenu.
Should we add it into `Navbars`-submenu too ?

mtscout6:
If you add a prop for active as well, then we should add prop type validations to ensure that only one of them is used at a time. I'm in favor of the idea, though not required in this PR as it could be followed up in another.

dozoisch:
I didn't add the active prop, it was already there!

taion:
@mtscout6

We don't have that validation for e.g. `Button`: https://github.com/react-bootstrap/react-bootstrap/blob/1b98730d16a128566ad48159ecab419720988992/src/Button.js#L9-L10

I don't think we need to enforce that invariant - it's potentially not logically meaningless, and it doesn't break anything internally for us (I think?).

mtscout6:
Ah, I guess they are semantically different anyways my bad.

dozoisch:
Alright, I can correct the Navbar to Navbars tonight if needed, but will have to wait the whole day!

AlexKVal:
It's possible to click on `disabled` and it gains focus.
I've looked at disabled `Button`, it doesn't allow focusing by these css:
```css
cursor: not-allowed;
pointer-events: none;
```
Though I'm not sure does the component need this.

screen shot 2015-05-28 at 7 59 45 pm

dozoisch:
@mtscout6 @AlexKVal  alright navbar => navbars has been corrected. Sorry for the delay

AlexKVal:
It is added into Buttons submenu.
Should we add it into Navbars-submenu too ?

If we should it and you just have no time for it, then I can add it after merge.
Code LGTM.

docs/examples/MenuItem.js
```diff
@@ -0,0 +1,26 @@
...
+  alert('Alert from menu item.\neventKey: "' + eventKey + '"\nhref: "' + href + '"');
+}
+
+function preventDefault() {}
+
+const MenuItems = (
+  <div className="clearfix">
+    <ul className="dropdown-menu open">
+      <MenuItem header>Header</MenuItem>
+      <MenuItem onSelect={preventDefault}>link</MenuItem>
```
mtscout6: I'm not a fan of this, though I'm sure I'll have this cleaned up in the dropdown revisited branch so I guess we can live with it in the time being.

test/MenuItemSpec.js
```diff
@@ -107,4 +107,13 @@ describe('MenuItem', function () {
...
   });
+
+  it('Should be `disabled` link', function () {
+    let instance = ReactTestUtils.renderIntoDocument(
+      <MenuItem disabled>
+        Title
+      </MenuItem>
+    );
+    assert.ok(instance.getDOMNode().className.match(/\bdisabled\b/));
+  });
```
mtscout6: Perhaps adding a test for the click event to ensure that `e.preventDefault()` is called when disabled.

docs/examples/MenuItem.js
mtscout6: I have this cleaned up in the dropdown-revisited branch now.

test/MenuItemSpec.js
mtscout6: This is covered in the dropdown-revisited branch now.

mtscout6:
LGTM

dozoisch:
@mtscout6 Thx for cleaning these up in dropdown-revisited... I'm completely struggling with time these days :/

mtscout6:
No worries friend, that's why the Maintaining guide says do what you can when you can. We understand when other priorities take precedence.

PR #610, opened 2015-05-03
[added] favicon

dozoisch:
I also removed the window.React = React because Facebook changed the way it works in React 0.13


AlexKVal:
```sh
$ npm run docs
$ open http://localhost:4000/components.html
```
There are no rendered components and in dev-console:
`> ReferenceError: React is not defined`

As for the Icon. I think it needs more contrast. 
`B` and the lines around have to be brighter.
And it seems that letter `B` needs to be bigger, more visible.
win
macosx

AlexKVal:
I would split up the `Icon` part and `remove global React` into different PRs.
Or I just would remove `remove global React` part from this PR :wink:

mtscout6:
I also ran `npm run docs-prod` which is more in line with how it's run in production. Icon is rendering, but I agree with @AlexKVal in that it needs more contrast. However, I don't think that's your concern right now.

I talked with our design guys and they said they could rethink the logo and provide us something that will work with a favicon. They asked for a week to do that, and I'm confident that they'll come up with something that will work.

With that said, this looks good to me and we can merge as is for now, knowing the logo will get an upgrade in the next week or two.

mtscout6:
@dozoisch I didn't realize it when I brought this in, but the code examples where dependent on React being defined....

mtscout6:
@dozoisch No worries though it was an easy fix.

dozoisch:
Ohhhhh, that's why. I see now why it needed a global React. My bad, on my end everything seemed to work without it.

AlexKVal:
You did it OK. :+1: 
It is a very subtle bug caused by using `eval()` with `babel@5` version.
You can read more here https://github.com/react-bootstrap/react-bootstrap/pull/539

It seems sooner or later it needs to be refactored without `eval()`,
because by default `babel` inserts `"use strict;"` with ES6 code.

Issue #617, opened 2015-05-05
Input addonBefore does not allow setting hidden-xs on input-group-addon


cymen:
Say I have a modal that has been customized to look like this:

![screen shot 2015-05-05 at 11 50 47 am](https://cloud.githubusercontent.com/assets/121034/7480366/05a00e82-f31d-11e4-98f0-8039d1b19d97.png)

When the screen is shrunk down, I want the addonBefore to disappear. So in regular bootstrap, the input-group-addon would have the `hidden-xs` class applied. But react-bootstrap does not expose a way to set it on the input-group-addon div that it wraps around the addonBefore property.

cymen:
I'd be happy to do a PR to fix this but I'm not sure how to fix it cleanly. If I was to go the most direct route, I would add to Input these properties:

addonAfterClassName
addonBeforeClassName
buttonBeforeClassName
buttonAfterClassName

Or it could be more like:

addonAfterProperties
addonBeforeProperties
buttonBeforeProperties
buttonAfterProperties

Or maybe the option to customize the input-group-addon with say:

addonAfterInputGroup
...

Any preferences?

cymen:
I'm temporarily working around this issue with an ugly hack:

```javascript
  componentDidMount: function() {
    // TODO work around for https://github.com/react-bootstrap/react-bootstrap/issues/617
    this.getDOMNode().getElementsByClassName('input-group-addon')[0].className += ' hidden-xs';
  },
```

aabenoja:
Related to #342.

aabenoja:
@cymen What are you passing in to `addonBefore` exactly?

dozoisch:
piece of code in question

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Input.js#L53-L57

cymen:
My usage looks like this:

```javascript

  renderPrefix: function() {
    return (
      <span className="hidden-xs">
        <i className="fa fa-search"></i> Keywords
      </span>
    );
  },

  renderButton: function() {
    return <Button bsStyle="primary" onClick={this.onFilterApply}>Search</Button>;
  },

  render: function() {
    return (
      <Modal
        className="keyword-filter-modal"
        onRequestHide={this.props.onFilterHideRequest}>

        <div className="keyword-filter">
          <Input
            addonBefore={this.renderPrefix()}
            buttonAfter={this.renderButton()}
            bsSize="large"
            type="text"
            placeholder="Internet of things"
[...]

PR #618, opened 2015-05-05
Add a customBefore and customAfter to Input

cymen:
How about adding a `customBefore` and `customAfter` to `Input` to allow complete customization? When either is used, the passed in JSX gets rendered as is -- no wrapper at all.  That way, I can control the wrapper completely. The downside is I have to make sure my wrapper markup stays in sync but I see that as a small price to pay for that amount of flexibility.

After mulling it over, I prefer to interact with react-bootstrap in this way. This would also be a solution for an issue I opened around not being able to add `hidden-xs` to the `addonBefore` wrapping markup (https://github.com/react-bootstrap/react-bootstrap/issues/617#issuecomment-99214646)

### DIFF
--- a/docs/examples/InputAddons.js
+++ b/docs/examples/InputAddons.js
@@ -6,6 +6,18 @@ const innerDropdown = (
   </DropdownButton>
 );
 
+const customBefore = (
+  <span className="input-group-addon hidden-xs">
+    Hide this on narrow view
+  </span>
+);
+
+const customAfter = (
+  <span className="input-group-addon hidden-xs">
+    <Button>Search</Button>
+  </span>
+);
+
 const inputAddonsInstance = (
   <form>
     <Input type='text' addonBefore='@' />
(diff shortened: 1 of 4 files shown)


mtscout6:
@cymen I don't believe this is the right solution for this. How would something like this sound:

```
render() {
  return (
    <Input type='text'>
        <AddonBefore className='custom-class' />
        <ButtonBefore className='custom-class' onClick={(event) => { ... } />
        {/* OR for dropdown variations */}
        <ButtonBefore className='custom-class'>
          <MenuItem>Dropdown menu item</MenuItem>
        </ButtonBefore>
        <ButtonAfter />
        <AddonAfter />
    </Input>
  );
}
```

I know it's not quite intuitive at first, but it really opens the doors for extensibility. Should we do this with ES6 classes than you could extend any of those components and override its render call. I don't know if this is possible with Bootstrap but it would allow you to, for example, have two buttons before the input.

We could do some validation logic to warn if you use an after component prior to any befores.

@aabenoja What are your thoughts on this? If it looks good I'd propose amending #342 to reflect this idea and work to do this would fall under that issue.

cymen:
@mtscout6 Yes, I think that is a better approach and one I have been using on my internal components. Basically, having my component take children and rendering my component around the children which makes composition easier. That is what I see in your example and that is what we're talking about, right?

mtscout6:
Yes, that's what I'm thinking.

mtscout6:
@aabenoja Is in the middle of working on #342 so I'll be sure to bring this up with him tomorrow. After we talk one of us will update #342 with the proposed solution.

cymen:
Okay -- feel free to close this and keep going with what you two are working on. Thanks!

PR #621, opened 2015-05-06
Use .babelrc for babel configuration

mtscout6:
This will allow us to configure babel in one central location from now
on. Prior to this any time we ran a babel compile, babel-loader, or
babel-node we had to configure them individually which can be error
prone in the long run.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,5 @@
+{
+  "optional": [
+    "es7.objectRestSpread"
+  ]
+}
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,5 +1,5 @@
 /* eslint no-var: 0 */
-require('./register-babel');
+require('babel/register');
 
 var webpackConfig = require('./webpack/test.config.js');
 var isCI = process.env.CONTINUOUS_INTEGRATION === 'true';
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -46,7 +46,7 @@ module.exports = function (config) {
     browsers: [ isCI ? 'PhantomJS' : 'Chrome' ],
 
     captureTimeout: 60000,
-    browserNoActivityTimeout: 30000,
+    browserNoActivityTimeout: 45000,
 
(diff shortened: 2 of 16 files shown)


AlexKVal:
Nice :+1: 

We are already excluding `node_module` [here](https://github.com/react-bootstrap/react-bootstrap/pull/621/files#diff-e56633f72ecc521128b3db6586074d2cR6)
and it seems excluding them [here](https://github.com/react-bootstrap/react-bootstrap/pull/621/files#diff-28b51d16260380d9fae7857a4ba6dbeaR48
) is superfluous

I've tested it all without `exclude: /node_modules/`
all is green
npm run docs OK
./tools/reslease OK

AlexKVal:
As it turned out... 
this is superfluous also [["node_modules/\*\*/*"]](https://github.com/react-bootstrap/react-bootstrap/pull/621/files#diff-e56633f72ecc521128b3db6586074d2cR6)
`babel-node` ignores `node_modules` by default.
here https://babeljs.io/docs/usage/cli/
I've checked it.

dozoisch:
lgtm

mtscout6:
Ok, I can remove that.

mtscout6:
@AlexKVal I removed that ignore, do you see anything else?

AlexKVal:
Today while I had been trying to untangle `npm run docs` chain,
I've noticed another one https://github.com/react-bootstrap/react-bootstrap/blob/master/webpack/strategies/docs.js#L30, but I'm not sure about it,
because it needs more research for me with your "client-loader".
Maybe you'll learn it faster.

Anyway I would merge this PR already :+1:

mtscout6:
That line is fine, it's still good to tell webpack not to apply loaders to the node_modules folder. That exclude applies to all loaders in the chain not just babel.

AlexKVal:
Eventually I see 3x LGTM and I push `The Big Green Button` :smiling_imp:

PR #624, opened 2015-05-07
Input bsSize now propgates correctly for css form-group-\*

aforty:
`FormGroup` now honors the `bsSize` property by applying form-group-lg and form-group-sm, for bsSize="large" and bsSize="small", respectively. 

This allows us to do something as simple as `<Input bsSize="large" />` which was not possible before.

### DIFF
--- a/src/FormGroup.js
+++ b/src/FormGroup.js
@@ -28,7 +28,13 @@ FormGroup.defaultProps = {
 FormGroup.propTypes = {
   standalone: React.PropTypes.bool,
   hasFeedback: React.PropTypes.bool,
-  bsSize: React.PropTypes.oneOf(['small', 'medium', 'large']),
+  bsSize (props) {
+    if (props.standalone) {
+      return new Error('bsSize will not be used when `standalone` is set.');
+    }
+
+    return React.PropTypes.oneOf(['small', 'medium', 'large']).apply(null, arguments);
+  },
   bsStyle: React.PropTypes.oneOf(['success', 'warning', 'error']),
   groupClassName: React.PropTypes.string
 };
--- a/src/FormGroup.js
+++ b/src/FormGroup.js
@@ -5,6 +5,8 @@ class FormGroup extends React.Component {
   render() {
     let classes = {
       'form-group': !this.props.standalone,
+      'form-group-lg': !this.props.standalone && this.props.bsSize === 'large',
+      'form-group-sm': !this.props.standalone && this.props.bsSize === 'small',
       'has-feedback': this.props.hasFeedback,
       'has-success': this.props.bsStyle === 'success',
       'has-warning': this.props.bsStyle === 'warning',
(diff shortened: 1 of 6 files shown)


mtscout6:
There should already be some mechanics in the BootstrapMixin for this. @aabenoja Was that lost when we switched that to be an ES6 class? If so we need to figure out how to keep that logic consolidated.

aabenoja:
@mtscout6 `bsSize` was only used for the input group in the previous implementation. While we did send it up to `FormGroup`, `FormGroup` itself never did anything with it.

aabenoja:
And `FormGroup` has not used `BootstrapMixin`, from what I see in the file history.

mtscout6:
Fair enough. I also didn't realize the boolean check against `this.props.standalone` there. The classnames module will take multiple objects. It may be cleaner to have two objects that you pass to classnames, to avoid multiple checks for standalone.

This could also use some more tests to ensure the standalone check is being respected with size options.

src/FormGroup.js
```diff
@@ -26,6 +28,7 @@ FormGroup.defaultProps = {
 FormGroup.propTypes = {
   standalone: React.PropTypes.bool,
   hasFeedback: React.PropTypes.bool,
+  bsSize: React.PropTypes.oneOf(['small', 'medium', 'large']),
```
mtscout6: To avoid confusion when those classes don't show up because the standalone prop is used. Can you do a custom prop type validation here to check if the standalone prop is true when using this and warn if so?

mtscout6:
Can you amend your commit message as per the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) so our automated changelog generator will document this in the release notes.

Also, please add some documentation for these new options.

aforty:
Sure. `console.warn(...)` or something else?

mtscout6:
Use a custom prop type validator. See https://facebook.github.io/react/docs/reusable-components.html for complete details.

Example custom prop:

```
    // You can also specify a custom validator. It should return an Error
    // object if the validation fails. Don't `console.warn` or throw, as this
    // won't work inside `oneOfType`.
    customProp: function(props, propName, componentName) {
      if (!/matchme/.test(props[propName])) {
        return new Error('Validation failed!');
      }
    }
```

aforty:
Yeah I figured that, wasn't sure how to handle actually do a warning but it looks like throwing an Error from the custom validation function is then displayed by React as a warning. Got it.

mtscout6:
That's the idea, the custom prop validator concept is great as it allows you to verify multiple props in one validation call.

aforty:
Take a look and let me know if I missed anything.

mtscout6:
LGTM

Need one more approval from a member of @react-bootstrap/collaborators

aabenoja:
LGTM :ship:

AlexKVal:
:+1:

PR #488, opened 2015-04-07
[added] Thumbnail component including docs and tests

JordanTheriault:
For Issue #56 Added the thumbnail component, docs, and tests.

Modeled the docs after the [bootstrap doc](http://getbootstrap.com/components/#thumbnails).

I'm new to this repo, so please let me know if you see anything I should change even if it's just best practices. :)

PR #631, opened 2015-05-08
Enable React-Hot-Reloader with Webpack

mtscout6:
Now that we have the webpack-dev-server running in a separate process it's time to enable hot reload for an even more amazing development experience.

Issues still to resolve.

- [x] Had to remove usage of client loader for performance reasons, which breaks the CodeMirror support
- [x] Remove in browser babel transformer webpack error
- [x] Fix carousel image request urls for examples. _Thumbnails will need this too when #488 comes in_
- [x] DRY up `docs/dev-run` script
- [x] Update docs README


mtscout6:
I put up a [youtube video](https://youtu.be/vViVUbyAWeY) demonstrating these changes.

mtscout6:
@react-bootstrap/collaborators That should be everything for this to work.

mtscout6:
Actually I found one more thing, just a minute.

mtscout6:
Ok, now it should be good.

docs/README.md
```diff
@@ -11,10 +11,26 @@ Pages](http://pages.github.com/).
 From the repository root run `npm run docs` and navigate your browser to
 `http://localhost:4000`. This will start an express base node server with
 webpack-dev middleware that will watch your file changes and recompile all the
-static assets needed to generate the site. 
+static assets needed to generate the site. In the console output you'll see that
+that we bind to two ports. The first port is the one you'll use to load the docs
```
AlexKVal: `you'll see that that` two `that` ?

mtscout6: If that's the worst thing in this you can find, then I guess I can fix it :wink:

docs/server.js
```diff
@@ -21,8 +21,11 @@ if (development) {
   });
 
   app.use(function renderApp(req, res) {
+    res.header('Access-Control-Allow-Origin', target);
```
AlexKVal: https://github.com/webpack/webpack-dev-server/pull/70

in webpack.config.js
```js
devServer: {
  headers: { 'Access-Control-Allow-Origin: '*' },
  headers: { 'Access-Control-Allow-Headers': 'X-Requested-With' }
}
```

mtscout6: No the server.js script is not run in the same process as the webpack-dev-server. We need the process doing the server side rendering to allow requests to a different service. If you took this out then you'd run into errors in the browser due to cross origin request violations.

webpack/strategies/docs.js
```diff
@@ -27,7 +27,6 @@ export default (config, options) => {
               jsLoader = value.loader;
 
               return _.extend({}, value, {
-                loader: jsLoader + '!client',
```
AlexKVal: As I understand we can remove `"client-loader": "0.0.1",` from `package.json` ?
I didn't downloaded this PR yet.
Only looking through diffs on GitHub.
(and you've already made three fixes :smile: what caused me to refresh this page three times)

mtscout6: I'll make one more!

mtscout6: Removed :smile: thanks for finding that.

AlexKVal:
Whaaahh! Coool. 
I press Save and page is self updating. :boom: 
:sparkles: :sparkles: :sparkles: :sparkles: 

And it is so fast now  :airplane:

I like it a lot. Thank you! :+1:

mtscout6:
I've been wanting this for a while myself, just been dragging my feet on the dirty work...

AlexKVal:
Actually today you already marked as done second item in my ToDo list. :smile: :+1:

> // Remove extract text plugin from dev workflow so hot reload works on css.

If we could remove it at all it would be nice :smiley:

mtscout6:
It's still worth keeping that plugin for the production build, since it allows us to parallelize the download of JavaScript and CSS.

AlexKVal:
> ...just been dragging my feet on the dirty work...

If you mean burden of maintenance, I understand you now.
It's a lot of work and it needs to be done everyday.

AlexKVal:
> It's still worth keeping that plugin for the production build, since it allows us to parallelize the download of JavaScript and CSS.

I am not finished yet my learning of the whole process, so I didn't know it.
But that bug in that project is on my ToDo also :wink:

docs/server.js
AlexKVal: I had commented out both lines `Access-*`, 
then  `npm run docs`,
then open in browser all pages with opened dev-console.
All is working and clear. Navigation, components rendering.

Then I had changed code on two pages: Components and Intro 
and those changes showed up in browser as it should be.

No errors at all.

What should I do to reproduce `cross origin request violations errors` ?

dozoisch:
nice job

PR #642, opened 2015-05-10
Don't warn about bsSize when it's not specified

taion:
Fixes a bug with #624

### DIFF
--- a/test/FormGroupSpec.js
+++ b/test/FormGroupSpec.js
@@ -46,6 +46,15 @@ describe('FormGroup', function() {
     assert.ok(ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'form-group-lg'));
   });
 
+  // This test case must come first, since the error only gets logged once.
+  it('throws no warning without bsSize when standalone', function () {
+    ReactTestUtils.renderIntoDocument(
+      <FormGroup standalone />
+    );
+
+    // Warning thrown above would lead to failure from index.
+  });
+
   it('throws warning about bsSize when standalone', function () {
     ReactTestUtils.renderIntoDocument(
       <FormGroup standalone bsSize="large" />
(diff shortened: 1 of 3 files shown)


taion:
Didn't realize `shouldNotWarn` was redundant. Got rid of it in revised commit.

dozoisch:
looks good to me!

test/helpers.js
```diff
@@ -1,9 +1,5 @@
-function shouldWarn(about) {
+export function shouldWarn(about) {
   console.warn.called.should.be.true;
   console.warn.calledWithMatch(about).should.be.true;
   console.warn.reset();
 }
-
-export default {
```
AlexKVal: That's my fault.
At some middle point of review process there was proposed version with additional exported function.
I've moved that function out, but forgot to simplify this.
:+1:

test/FormGroupSpec.js
```diff
@@ -46,6 +46,15 @@ describe('FormGroup', function() {
     assert.ok(ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'form-group-lg'));
   });
 
+  // This test case must come first, since the error only gets logged once.
```
AlexKVal: It is just my curiosity.
I had placed this test after next one and then placed it into the end of file
and in both cases all tests are green.

What did you mean by this comment ?

taion: Per https://github.com/facebook/react/blob/a56d0d6dc58a973d103bfae7e7ef7c04226209c7/src/classic/element/ReactElementValidator.js#L273 each prop type warning only gets logged once. Therefore, if I put this test case after the next one, no warning would be thrown even if my code didn't work.

AlexKVal: Wow. I should be cautious now in my tests.
Thank you! :cherries:

mtscout6: I really wish React had a way to turn that off, @aabenoja got pinned down on that for a while when working on the `ButtonInput` PR.

taion: I agree. It was annoying enough that I was thinking of deleting this test case for a bit. "Doesn't throw a warning" isn't really an interesting or useful test in most cases.

Issue #465, opened 2015-03-31
OverlayTrigger and parent-based contexts


taion:
OverlayTrigger renders the overlay element into a new `<div />`, by default under `document.body`. When used with react-router, the overlay element is no longer a child of the overall router... things. This means that the overlay element does not receive the parent-based router context.

This is currently only causing a warning in React 0.13, so it's not urgent, but it could be a problem down the road for people who want to use components that want to access e.g. router context within an overlay.

mtscout6:
Thanks for opening an issue on this, one of the devs in my shop brought this to my attention last week. I can't say that I have a solution figured out yet.

taion:
The best I can think of is something like this

```jsx
<OverlayTrigger
  overlayHolder={() => this._overlayHolder}
>
  <Button {...} />
</OverlayTrigger>

<OverlayHolder
  ref={component => this._overlayHolder = component}
>
  <Popover {...} />
</OverlayHolder>
```

Where `OverlayHolder` is a size-0 div or something, and `OverlayTrigger` calls `setState` on the `OverlayHolder` to get it to display the overlay element.

taion:
Rather:

```jsx
<OverlayHolder
  ref={component => this._overlayHolder = component}
  overlay={<Popover {...} />}
/>
```

like the current syntax, so it doesn't look like the overlay is getting rendered when it's not active.

This is morally the same as what's happening right now, except (when needed) the user explicitly sets the parent element of the overlay, and has it all be within the React element hierarchy. This is cumbersome enough that I'd only want to use this when I care about preserving context, but I don't see another way to do this, given that you can't actually render anything but the wrapped element in `OverlayTrigger` without breaking formatting.

taion:
On thinking of this again, what I have above is probably just dumb.

What do you think about having `OverlayTrigger` explicitly take a context object to forward down to the rendered overlay?

mtscout6:
That would be simple and worth it. I would not want to document this feature though until React actually "officially" releases context support and documentation.

taion:
Do you think it might be worth exposing a special variant of OverlayTrigger in react-router-bootstrap that forwards the router context? @ryanflorence mentioned that they were working on removing the dependency on context, but it'd be useful to have an at least somewhat-documented way to access router context from popovers.

mtscout6:
For react-router-bootstrap I'd be fine with a wrapper that could do that, or at least a mention in its readme.

taion:
[...]

PR #644, opened 2015-05-11
Add context-forwarding trigger factory methods

taion:
For react-bootstrap/react-bootstrap#465

No need to document this. If this is accepted, I'm going to add an `OverlayTriggerWithRouterContext` (better name welcome) to react-router-bootstrap to make that use case easier.

I don't think there's a better way to do this than to make a new component class every time because of `childContextTypes` and `contextTypes` both being static.

### DIFF
--- a/src/utils/createContextWrapper.js
+++ b/src/utils/createContextWrapper.js
@@ -1,0 +1,48 @@
+import React from 'react';
+
+/**
+ * Creates new trigger class that injects context into overlay.
+ */
+export default function createContextWrapper(Trigger, propName) {
+  return function (contextTypes) {
+    class ContextWrapper extends React.Component {
+      getChildContext() {
+        return this.props.context;
+      }
+
+      render() {
+        // Strip injected props from below.
+        const {wrapped, ...props} = this.props;
+        delete props.context;
+
+        return React.cloneElement(wrapped, props);
+      }
+    }
+    ContextWrapper.childContextTypes = contextTypes;
+
+    class TriggerWithContext {
+      render() {
+        const props = {...this.props};
+        props[propName] = this.getWrappedOverlay();
+
+        return (
+          <Trigger {...props}>
+            {this.props.children}
+          </Trigger>
+        );
+      }
+
+      getWrappedOverlay() {
+        return (
+          <ContextWrapper
+            context={this.context}
+            wrapped={this.props[propName]}
+          />
+        );
+      }
+    }
+    TriggerWithContext.contextTypes = contextTypes;
+
+    return TriggerWithContext;
+  };
+}
--- a/test/ModalTriggerSpec.js
+++ b/test/ModalTriggerSpec.js
@@ -4,72 +4,101 @@ import ModalTrigger from '../src/ModalTrigger';
 
 describe('ModalTrigger', function() {
   it('Should create ModalTrigger element', function() {
-    let instance = ReactTestUtils.renderIntoDocument(
+    const instance = ReactTestUtils.renderIntoDocument(
       <ModalTrigger modal={<div>test</div>}>
         <button>button</button>
       </ModalTrigger>
     );
-    let modalTrigger = instance.getDOMNode();
+    const modalTrigger = React.findDOMNode(instance);
     assert.equal(modalTrigger.nodeName, 'BUTTON');
   });
 
   it('Should pass ModalTrigger onMouseOver prop to child', function() {
-    let called = false;
-    let callback = function() {
-      called = true;
-    };
-    let instance = ReactTestUtils.renderIntoDocument(
+    const callback = sinon.spy();
+    const instance = ReactTestUtils.renderIntoDocument(
       <ModalTrigger modal={<div>test</div>} onMouseOver={callback}>
         <button>button</button>
       </ModalTrigger>
     );
-    let modalTrigger = instance.getDOMNode();
+    const modalTrigger = React.findDOMNode(instance);
     ReactTestUtils.Simulate.mouseOver(modalTrigger);
-    assert.equal(called, true);
+    callback.called.should.be.true;
   });
 
   it('Should pass ModalTrigger onMouseOut prop to child', function() {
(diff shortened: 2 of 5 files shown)


src/utils/createContextWrapper.js
```diff
@@ -0,0 +1,48 @@
...
+  return function (contextTypes) {
+    class ContextWrapper extends React.Component {
+      getChildContext() {
+        return this.props.context;
+      }
+
+      render() {
+        // Strip injected props from below.
+        const {wrapped, ...props} = this.props;
+        delete props.context;
```
AlexKVal: @mtscout6 will disagree with this line :smile: 

I had been also trying to use such trick.
`props` eventually become immutable.
We have to avoid it mutation.

src/utils/createContextWrapper.js
```diff
@@ -0,0 +1,48 @@
...
+            {this.props.children}
+          </Trigger>
+        );
+      }
+
+      getWrappedOverlay() {
+        return (
+          <ContextWrapper
+            context={this.context}
+            wrapped={this.props[propName]}
```
AlexKVal: why don't you use just `this.props.propName` ?

test/ModalTriggerSpec.js
```diff
@@ -4,72 +4,101 @@ import ModalTrigger from '../src/ModalTrigger';
...
     assert.equal(modalTrigger.nodeName, 'BUTTON');
   });
 
   it('Should pass ModalTrigger onMouseOver prop to child', function() {
-    let called = false;
-    let callback = function() {
-      called = true;
-    };
-    let instance = ReactTestUtils.renderIntoDocument(
+    const callback = sinon.spy();
```
AlexKVal: :+1:

src/utils/createContextWrapper.js
mtscout6: Agreed, you should not be mutating props.

src/utils/createContextWrapper.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+/**
+ * Creates new trigger class that injects context into overlay.
+ */
+export default function createContextWrapper(Trigger, propName) {
+  return function (contextTypes) {
+    class ContextWrapper extends React.Component {
```
mtscout6: A class declaration within a function is not very efficient, since this class would be redefined with every invocation. It would be better to declare this class in a separate module and instantiate it here....

Or I think I'm understanding that you are doing this because the `childContextTypes` is being appended as a static property. I'm wondering if there's at least some means to memoize this?

src/utils/createContextWrapper.js
taion: I'm not mutating props here. I'm pulling the props from this component, removing the keys that should not be forwarded to the child element, and forwarding only the subset of the props that should be forwarded.

I initially had this written as
```js
const {wrapped, context, ...props} = this.props;
```
but this produced a lint warning on the unused `context` variable.

src/utils/createContextWrapper.js
taion: `propName` is only available via the closure scope, and I want the actual prop with the key of `propName`, rather than e.g. `this.props['propName']`.

src/utils/createContextWrapper.js
taion: In actual use, I would set up a module that does e.g.

```js
export default OverlayTrigger.withContext{
  router: React.PropTypes.func
});
```

and just accomplish the memoization that way.

I definitely don't want people to be calling `OverlayTrigger.withContext` every time - just do it once at module level. Not too worried about this since we're not documenting this anyway.

src/utils/createContextWrapper.js
mtscout6: I see, you're right that `this.props` is not being mutated, my bad. @AlexKVal I think this should be ok here.

mtscout6: I prefer `const {wrapped, context, ...props} = this.props;` but at the same time I don't want to keep adding eslint exceptions inline with our code.

src/utils/createContextWrapper.js
mtscout6: Just because we aren't documenting it right now does not mean that we won't later. Once React starts documenting it, then we should too.

I know I said no documentation, but maybe we should add some comments to the code explaining how to use this. So, only those that really dig in will know how it's intended to be used.

src/utils/createContextWrapper.js
taion: I prefer it too, which is why I wrote it that way at first (:

But I didn't think it was worth adding ESLint disables for something that could be written in a not-much-less-nice way.

src/utils/createContextWrapper.js
taion: Fair enough. I'll add some comments on a new commit, and we can squash before merging.

mtscout6: :+1:

taion: Added in https://github.com/taion/react-bootstrap/commit/6f2e03646207c1e31a24ecf68866d69e33d0add4

mtscout6: Thanks for the code comments that helps a lot.

mtscout6:
LGTM

Feel free to squash and force push.

Any other thoughts from @react-bootstrap/collaborators?

src/utils/createContextWrapper.js
AlexKVal: Guys. It's all my fault. Sorry about that. :cake: 
I really mis-interpreted the code.
And then I misdirected Matt :blush:

src/utils/createContextWrapper.js
AlexKVal: Again. I got it wrong. Sorry.
I'd seen it as `this.props['propName']`.
It seems I was sleeping yet at morning. :blush: 

Thank you for your patience with my wrong questions :cherries:

idolize:
LGTM

Issue #217, opened 2014-09-05
Can not disable a TabbedArea tab


tomgoldsmith:
One of the props on a Nav is "disabled". This prop should be usable on the TabbedArea tabs to disable tabs.

AlexKVal:
Can you please clarify ?
Is that a question or a feature proposal / request ?
Does it relevant yet ?

AlexKVal:
In case of need, we'll reopen it.
I'm closing it.

traviscooper:
Having same issue.

mtscout6:
I'm not seeing [documentation](http://getbootstrap.com/javascript/#tabs) for this with Bootstrap. Though I can see the benefit to supporting something like this. Since the underying tab is just an anchor, I don't think this would be hard to implement nor would it be compromising to Bootstrap parity.

I'm assuming that you are looking for something like this:

![](http://cl.ly/image/3x2q3o1m2N3C/Screen%20Recording%202015-05-13%20at%2008.46%20AM.gif)

I monkey patched things to make that screen recording work, but it should be as simple as adding the `disabled` class to the `li` and adding some JavaScript to not activate the tab when it's disabled.

Of course the change will need tests and added documentation.

@traviscooper and @tomgoldsmith I think a PR for this would be awesome, and incredibly useful.

traviscooper:
I have a small patch I was preparing, shall I send it? :)

traviscooper:
If you pass the 'disabled' prop, straight through from TabArea -> TabPane -> NavItem, then it properly adds the disabled class to the tab and picks up the disabled functionality to not show the content as well.

traviscooper:
https://github.com/react-bootstrap/react-bootstrap/pull/655 to fix this issue.

mtscout6:
Well that was fast! Give me some time to review it :smile:

PR #655, opened 2015-05-13
#217 TabbedArea updated to allow 'disabled' Tab

traviscooper:


mtscout6:
Can you add a test to verify that when the tab is clicked that it does not become active?

mtscout6:
Also, after you add that last test can you squash the commits and follow the commit message pattern as prescribed in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes)? Doing so will allow for our automated changelog tool to document this on the next release.

traviscooper:
Will address!  Thanks @mtscout6

src/TabbedArea.js
```diff
@@ -105,12 +105,13 @@ const TabbedArea = React.createClass({
   },
 
   renderTab(child) {
-    let {eventKey, className, tab } = child.props;
+    let {eventKey, className, tab, disabled } = child.props;
```
mtscout6: This child should be a `TabPane` if I'm not mistaken, can you add `disabled` to the list of `propTypes` on `TabPane`? This will ensure folks get warned if they use the wrong data type.

traviscooper: Will do

traviscooper:
Please see #657

PR #658, opened 2015-05-13
Add source maps when optimizing

taion:


mtscout6:
LGTM

I ran `npm run build` locally and saw that the source map file was in the `dist` and `amd` folders as I would expect.

Issue #173, opened 2014-07-30
Popover does not stay within viewport


bjohn465:
When an `OverlayTrigger` is used with a `Popover` (I haven't tried other components for the `overlay` property), and the trigger is near the edge of the browser's viewport (e.g., the top of the screen), a portion of the popover can't be seen, since it goes outside the viewport.

[Test case](http://jsbin.com/qefor/1/edit?html,output)

When Bootstrap places the popover, [it uses a `getViewportAdjustedDelta` function](https://github.com/twbs/bootstrap/blob/v3.2.0/js/tooltip.js#L252) to adjust the popover's position so it stays within the viewport.

taion:
I've been seeing this too. For a bottom popover, styling it with a fixed `width` makes it go off the side of the viewport when it's too close to the boundary. Styling it with `max-width` leaves the arrow in an odd spot, with behavior inconsistent with what stock Bootstrap does. Compare -

React Bootstrap:
![image](https://cloud.githubusercontent.com/assets/3112159/6876918/a20dd2f2-d4a2-11e4-9466-549237e4b475.png)

Bootstrap JS:
![image](https://cloud.githubusercontent.com/assets/3112159/6876923/ac3e210a-d4a2-11e4-8d8f-34a7e6116021.png)

Notably, on the former, the arrow is offset from the center of the button; on the latter, the arrow is still centered to the button, but offset from the center of the popover body.

taion:
Forgot to fix `propTypes` for the `arrowOffset` props.

PR #662, opened 2015-05-14
[fixed] Fix propTypes for overlays

taion:
Fixes react-bootstrap/react-bootstrap#173

Adds test cases to catch this as well.


mtscout6:
LGTM

I tested locally against babel-eslint v3.1.1 and all is good.

PR #669, opened 2015-05-15
Cleanup Lint command

mtscout6:
There's no need to explicitly white list AND black list what we are linting. So, let's just use the black list from the .eslintignore file.


taion:
LGTM

AlexKVal:
Nice :+1:

PR #678, opened 2015-05-15
[added] DropdownButton now applies onClick prop to Button

aforty:
I needed this and noticed that the property was in DropdownButton but was not wired up. Hope this was simply an oversight and not intentional.

Did I miss anything?


dozoisch:
mostly seems like something that was forgotten!

mtscout6:
This is fine, however be aware that the functionality and signature of this may change as I finally wrap up work on #526.

Issue #653, opened 2015-05-13
Animation for Popovers


jardakotesovec:
Original bootstrap [supports](http://getbootstrap.com/javascript/#popovers-examples) animation property for popovers. Are here any plans to support it with react-bootstrap?

It would be great!

mtscout6:
We take pull requests :smile:

taion:
Fixed via #685.

PR #684, opened 2015-05-16
[added] property animation on Popover (true by default)

jportela:
Adds the FadeInMixin by default, as [original bootstrap(http://getbootstrap.com/javascript/#popovers-examples)] does. Solves #653 

As the original bootstrap, the animation can be disabled by doing animation={false} on the Popover component. I was going to add this to the documentation, but not found any table with the properties on Popover (maybe we should open a new issue for improving the documentation?)

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import classNames from 'classnames';
 import BootstrapMixin from './BootstrapMixin';
+import FadeMixin from './FadeMixin';
 
 const Popover = React.createClass({
-  mixins: [BootstrapMixin],
+  mixins: [BootstrapMixin, FadeMixin],
 
   propTypes: {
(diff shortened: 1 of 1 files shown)


taion:
I checked out your branch and checked the docs with `animation` set to `true` and `false`, and unspecified. This does what I expect. Can I ask, though, that you also add this to `Tooltip`, so we're fully consistent with upstream in this regard?

thetimbanks:
I didn't see this PR until I just submitted mine. The PR I submitted does include the property for Tooltip. I also included a basic test for Popover and Tooltip since there wasn't a spec file for either. If @jportela wants to update his PR I am happy to close mine out in favor of his.

jportela:
Tim, if yours has tests and covers Tooltip go ahead and merge it :) thanks :)

taion:
Closing this per https://github.com/react-bootstrap/react-bootstrap/pull/684#issuecomment-102699612 to focus on #685.

PR #685, opened 2015-05-16
[added] Property for animation on Popover and Tooltip

thetimbanks:
This adds the `animation` property to the `Tooltip` and `Popover` components by adding the `FadeMixin` and an `animation` property to those components.

This will use the have the `animation` property set to `true` by default since that is consistent with Bootstrap.

This also adds a test spec for Popover and Tooltip since one didn't exist before.

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -1,9 +1,10 @@
 import React from 'react';
 import classNames from 'classnames';
 import BootstrapMixin from './BootstrapMixin';
+import FadeMixin from './FadeMixin';
 
 const Popover = React.createClass({
-  mixins: [BootstrapMixin],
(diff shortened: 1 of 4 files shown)


taion:
#653 

Would supersede #684.

taion:
Would you mind adding a comment indicating that `.in` is applied by `FadeMixin` in the case that `animation` is set, and squash to your existing commit?

thetimbanks:
Just got it updated. Is that what you were looking for?

taion:
That's great. LGTM. Need one more reviewer from @react-bootstrap/collaborators to merge.

aabenoja:
:+1:

Issue #352, opened 2015-01-28
add active prop where suitable


flying-sheep:
e.g. bootstrap has special variables for active `MenuItem`s .

we should add an `active={this.state.active}` prop instead of requiring `className={classSet({active: this.state.active})}`

mtscout6:
:+1: We take pull requests.

taion:
Was `MenuItem` the only component that was missing an `active` prop?

mtscout6:
Before closing we should take a survey of all the components that would need this.

AlexKVal:
I'm closing it.
If a user has a problem with the `active` in any component, 
then them will create a specific issue.
(as this one. that is particularly about `MenuItems`)
:cherries:

mtscout6:
I think this is a valid point to ensure that it is correct for Bootstrap parity. Which is why I was saying that a more thorough survey is needed before closing. Unless you are proposing a better way to track progress towards Bootstrap parity.

AlexKVal:
My personal point of view is:
there has to be some compromise
between `Bootstrap parity` and `practical value` of implementing this `parity`.

`Bootstrap`-way and `React plus Boostrap`-way - IMHO is not the same.

`Bootstrap` deals with web-develomplent tasks - its own way.
That way is defined by `jQuery` constraints.

`React-Bootstrap` has the powers of `React`. 
This project can and does deal with the same web-develomplent tasks another way.

`Bootstrap` API is dictated by `Bootstrap`+`jQuery` constraints 
and this is not automatically mean that its API is good and best.
I'm sure there are good parts and bad parts. As everywhere.

Therefore - IMHO - this project should not strive to full parity.

Demand and Supply define what is needed and what is not.
`R-B` already 2+ years old. 
If we take a look through issues and PRs history, we will see,
that those not implemented components and features - they are just not in demand.
Hence it is not necessary to implement them.
Implementing parity for only the sake of parity - it is a bad business :smile: 

Let's market - i.e. our users - define what has to be implemented. And only when it is needed :wink: 

Something like that.
Sorry for the missive
:cherries:

mathieumg:
> Therefore - IMHO - this project should not strive to full parity.

Just to be clear, we mean feature parity, not API-parity.


> If we take a look through issues and PRs history, we will see,
> that those not implemented components and features - they are just not in demand.

[...]

PR #686, opened 2015-05-16
[added] Active property to MenuItem component

thetimbanks:
This adds an `active` property to the `MenuItem` component in order to set the `active` class. The result is a highlighted item in the dropdown menu.

The default value for the property is `false` in order to adhere to the Bootstrap default.

This should take care of #352

### DIFF
--- a/docs/examples/DropdownButtonBasic.js
+++ b/docs/examples/DropdownButtonBasic.js
@@ -5,7 +5,7 @@ function renderDropdownButton (title, i) {
     <DropdownButton bsStyle={title.toLowerCase()} title={title} key={i}>
       <MenuItem eventKey='1'>Action</MenuItem>
       <MenuItem eventKey='2'>Another action</MenuItem>
-      <MenuItem eventKey='3'>Something else here</MenuItem>
+      <MenuItem eventKey='3' active={true}>Active Item</MenuItem>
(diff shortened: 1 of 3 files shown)


taion:
LGTM.

At your discretion, it might make sense to duplicate that test case and verify that the `active` class is not set if the `active` prop isn't, but that's not a big deal to me.

We should also update `MenuItemLink` on react-router-bootstrap to use this new prop.

Need one more sign-off before we can merge.

dozoisch:
LGTM

chilts:
I was wondering if there is something going on with this, since on the example site for the Button Dropdowns, the MenuItem says active for item three, but item one is highlighted.

* http://react-bootstrap.github.io/components.html#btn-dropdowns # see first example

Does that look wrong to anyone else? Perhaps a bug which isn't related to this issue.

chilts:
I've sent a PR here : https://github.com/react-bootstrap/react-bootstrap/pull/1266

I think this should be okay, but you guys are better placed to say if this is the correct solution or not. :)

thetimbanks:
@chilts Looks like when the code was refactored to use es6 classes, some code may have gotten lost. Your PR looks like it should fix it.

chilts:
Cool, thanks.

PR #687, opened 2015-05-17
[fixed] Put AMD modules under correct path

### LINKED ISSUE #679: react-bootstrap-bower file/dir layout should be consistent with npm
Currently all files are located in the root folder, which caused bug https://github.com/react-bootstrap/react-router-bootstrap/issues/53 and leads to inconsistency of paths to components (NPM vs AMD for example).

taion:
Fixes #679


mtscout6:
LGTM

Issue #677, opened 2015-05-15
Minified exception in production docs


AlexKVal:
`npm run docs-prod` exposes this error as well.
With `npm run docs` there is no exception in console.

I've managed to nail the culprit #662 down but not further.

Update:
> Matt: This is because react-router is placing an active class name on the brand in the Navbar.

AlexKVal:
It doesn't exist anymore on the site already.
It is weird.

But.. Anyway it is a good news :smile:

AlexKVal:
Hmm.. github cloud caching weirdness aside
the error is still exists in
`npm run docs-prod` :smiling_imp:

AlexKVal:
Update:
@taion has prompted the faster way to reproduce it
With
```sh
git checkout 96baa15
webpack --config webpack.docs.js --progress && NODE_ENV=production babel-node docs/server.js
```
the error exists.

But two commits back (because one commit is just merging)
```sh
git checkout 96baa15^^ // it is 83d6220
webpack --config webpack.docs.js --progress && NODE_ENV=production babel-node docs/server.js
```
there is no that error.

taion:
I just spot-checked against `96baa15^^` and I still see it there. I don't think it's related to these commits.

taion:
This looks like it's coming from some mismatch between stuff before the `<body>` between the server and the browser, somehow. It doesn't seem to be related to any recent component changes.

mtscout6:
My changes with #688 allow us to see this error:

```
Uncaught Error: Invariant Violation: You're trying to render a component to the document using server rendering but the checksum was invalid. This usually means you rendered a different component type or props on the client from the one on the server, or your render() methods are impure. React cannot handle this case due to cross-browser quirks by rendering at the document root. You should look for environment dependent code in your components and ensure the props are the same client and server side:
 (client)    <meta http-equiv='X-UA-Compatible' co
 (server)    <meta http-equiv="X-UA-Compatible" co
```

Looks like s single quote vs double quote problem.

dozoisch:
Can we close this? Seems to be resolved by #689

AlexKVal:
http://react-bootstrap.github.io/

![screen shot 2015-06-11 at 8 56 06 pm](https://cloud.githubusercontent.com/assets/847572/8114319/65e32456-107c-11e5-8a1e-554595f771f8.png)

AlexKVal:
Actually this is of low no none priority and it is only it this place.
[...]

PR #689, opened 2015-05-17
Fix single quote bugs for docs in production

mtscout6:
Fixes most of #677 

The only part it does not fix is on the home page if the url does not include `index.html`. This is because react-router is placing an `active` class name on the brand in the `Navbar`. I don't see a quick easy way to deal with that at the moment.


AlexKVal:
LGTM

I've updated https://github.com/react-bootstrap/react-bootstrap/issues/677#issue-76821104

Issue #612, opened 2015-05-04
Focus on modal dialogs


ewal:
It's not currently possible to set the focus on the modal dialog (in an easy way) after it has been opened making it keyboard unfriendly. In Bootstrap, the outer modal container automatically gets focus when opened.

mtscout6:
@ewal We'd love a PR to make that work. It is along the lines of what's needed for Accessibility concerns as per our [1.0.0 Roadmap](https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap).

AlexKVal:
Solved by #690

PR #690, opened 2015-05-17
Modals are focused on opening, for improved acessibility

jportela:
Modals are now focused when opening them, as requested by #612 . I've also stored the element that was previously focused, so its focused can be restored when the Modal is closed (it is the same behaviour Bootstrap does).

I was unable to do tests for this, as it has a dependency on document.activeElement (and the tests I've looked into only used TestUtilities.renderIntoDocument, which renders to a detached DOM node, so I'm guessing we can't test this without attaching it to the DOM).

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -126,6 +126,8 @@ const Modal = React.createClass({
           domUtils.ownerDocument(this).body;
     container.className += container.className.length ? ' modal-open' : 'modal-open';
 
+    this.focusModalContent();
+
     if (this.props.backdrop) {
       this.iosClickHack();
     }
@@ -142,6 +144,8 @@ const Modal = React.createClass({
     let container = (this.props.container && React.findDOMNode(this.props.container)) ||
           domUtils.ownerDocument(this).body;
     container.className = container.className.replace(/ ?modal-open/, '');
+
+    this.restoreLastFocus();
   },
 
   handleBackdropClick(e) {
@@ -155,6 +159,19 @@ const Modal = React.createClass({
   handleDocumentKeyUp(e) {
     if (this.props.keyboard && e.keyCode === 27) {
       this.props.onRequestHide();
+    }
+  },
+
+  focusModalContent () {
+    this.lastFocus = document.activeElement;
+    let modalContent = React.findDOMNode(this.refs.modal);
+    modalContent.focus();
+  },
+
+  restoreLastFocus () {
+    if (this.lastFocus) {
+      this.lastFocus.focus();
+      this.lastFocus = null;
     }
   }
 });
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -97,4 +97,65 @@ describe('Modal', function () {
     assert.match(dialog.props.className, /\btestCss\b/);
   });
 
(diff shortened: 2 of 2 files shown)


mtscout6:
You can test this, checkout how I'm doing so with the Dropdown work I'm doing on the [dropdown-revisted branch tests](https://github.com/react-bootstrap/react-bootstrap/blob/dropdown-revisited/test/revisited/DropdownButtonSpec.js#L99-L121).

mtscout6:
This is a great start for accessibility of the modal! Have you considered ensuring that keyboard navigation via tabbing does not allow you to tab to the components under the modal? This is another concern to deal with for Modal accessibility. I understand if that's more than you'd like to tackle in this PR, but if you're up to it we can wait till that's also available. (Note: http://www.w3.org/WAI/PF/aria-practices/#modal_dialog)

jportela:
Thanks @mtscout6 for pointing out the tests, I'll start by adding the test case. And I'm happy to add the extra for not allowing to focus on the components under de modal, which makes sense. I'll try to tackle these today.

mtscout6:
You rock, thanks!

jportela:
added tests. Will still develop the feature to only focus on the components under the modal, today I hope

mtscout6:
The test looks great!

Can you squash the commits and use the commit message standard outlined in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes)? That way this change will be documented in the automated changelog when we push a release.

jportela:
sure, would this be considered a fix ([fixed])?

mtscout6:
Perfect!

Everything here LGTM, you can do the additional work to keep focus within the modal on another PR. I opened issue #723 which defines that separate problem.

Need another review by one of the @react-bootstrap/collaborators

AlexKVal:
:+1: 
Perfect. Simple and clean. I like it.
Thank you :cherries:

src/Modal.js
```diff
@@ -156,6 +160,19 @@ const Modal = React.createClass({
     if (this.props.keyboard && e.keyCode === 27) {
       this.props.onRequestHide();
     }
+  },
+
+  focusModalContent () {
+    this.lastFocus = document.activeElement;
```
taion: Should this be `ownerDocument`?

mtscout6: @taion Good catch, though since this PR has already been merged lets create an issue (#729) for this and follow up on another PR.

AlexKVal: Is it connected with `iframes` ?
I'm just not aware of use cases for this.
(have no such expirience)

mtscout6: As far as I can tell yes it is in relation to iFrames. I wasn't even really aware of that property and why to use it till I read up on it this weekend. So thanks @taion for steering us in a good direction.

AlexKVal: Thank you. 
It is important for me to know about it.
Also I'll be able to do more comprehensive code reviews with regard to `iFrames`.

taion: iframes aren't that high on my list of concerns, I was just following the other code.

PR #695, opened 2015-05-18
Update ESLint to 0.21.2

mtscout6:


dozoisch:
if the test passes LGTM, though I prefer using `~0.21.2` instead of `^0.21.2` because `^` is a bit greedy.

taion:
It's the same for anything v0. And I added test cases to eslint with my PR, so this use case hopefully won't break again in the same way.

mtscout6:
For pre 1.0 packages `~` and `^` make no difference. See: http://semver.npmjs.com/

dozoisch:
@mtscout6 oh wow didn't know that!

taion:
LGTM

PR #696, opened 2015-05-18
[added] dialogClassName prop to modal to be able to pass custom css c…

corysnave:
…lass to modal-dialog div

*Added a custom css class to demonstrate custom dialog sizing
*Added ModalCustomSizing.js to live demo the custom dialog sizing
*Added ModalDefaultSizing.js to live demo the default bootstrap modal sizing.(Wasn't previously in docs)
*Added ModalCustomSizing.js and ModalDefaultSizing.js to Samples.js
*Added the new prop in Modal.js to get passed to the modal-dialog div
*Added test in ModalSpec.js to confirm the class is being passed down to the modal-dialog div

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -20,7 +20,8 @@ const Modal = React.createClass({
     keyboard: React.PropTypes.bool,
     closeButton: React.PropTypes.bool,
     animation: React.PropTypes.bool,
-    onRequestHide: React.PropTypes.func.isRequired
+    onRequestHide: React.PropTypes.func.isRequired,
+    dialogClassName: React.PropTypes.string
   },
 
   getDefaultProps() {
@@ -55,7 +56,7 @@ const Modal = React.createClass({
         className={classNames(this.props.className, classes)}
         onClick={this.props.backdrop === true ? this.handleBackdropClick : null}
         ref="modal">
-        <div className={classNames(dialogClasses)}>
+        <div className={classNames(this.props.dialogClassName, dialogClasses)}>
           <div className="modal-content">
             {this.props.title ? this.renderHeader() : null}
             {this.props.children}
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -85,4 +85,16 @@ describe('Modal', function () {
     assert.ok(dialog.className.match(/\bmodal-sm\b/));
   });
 
+  it('Should pass dialogClassName to the dialog', function () {
+    let noOp = function () {};
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Modal dialogClassName="testCss" onRequestHide={noOp}>
+        <strong>Message</strong>
+      </Modal>
+    );
+
+    let dialog = ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'modal-dialog');
+    assert.match(dialog.props.className, /\btestCss\b/);
+  });
+
 });
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -1077,6 +1077,9 @@ h1[id] {
   margin-right: auto;
   margin-left: auto;
 }
+.custom-modal {
+  width: 90%;
+}
 
 /* Example dropdowns */
 .bs-example > .dropdown > .dropdown-toggle {
--- a/docs/examples/ModalCustomSizing.js
+++ b/docs/examples/ModalCustomSizing.js
@@ -1,0 +1,37 @@
+const MyModal = React.createClass({
+  render() {
+    return (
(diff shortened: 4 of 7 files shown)


test/ModalSpec.js
```diff
@@ -85,4 +85,16 @@ describe('Modal', function () {
...
+  it('Should pass dialogClassName to the dialog', function () {
+    let noOp = function () {};
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Modal dialogClassName="testCss" onRequestHide={noOp}>
+        <strong>Message</strong>
+      </Modal>
+    );
+
+    let dialog = instance.getDOMNode().getElementsByClassName('modal-dialog')[0];
+    assert.ok(dialog.className.match(/\btestCss\b/));
```
aabenoja: `React.findDOMNode(instance)` should be used in place of `component.getDOMNode()`. You could also use array destructuring to grab the first element rather than reference directly by index:
```javascript
let [dialog] = node.getElementsByClassName('modal-dialog');
```

You can use `should` or `expect` in place of this `assert.ok` to chain together a more readable assertion:
```javascript
dialog.className.should.match(/\btestCss\b/); //with should
expect(dialog.className).to.match(/\btestCss\b/); //with expect
assert.match(dialog.className, /\btestCss\b/); //with assert
```

src/Modal.js
```diff
@@ -55,7 +55,7 @@ const Modal = React.createClass({
         className={classNames(this.props.className, classes)}
         onClick={this.props.backdrop === true ? this.handleBackdropClick : null}
         ref="modal">
-        <div className={classNames(dialogClasses)}>
+        <div className={classNames(this.props.dialogClassName, dialogClasses)}>
```
aabenoja: Should we not have validation on `dialogClassName`, at least to ensure that this is a string?

test/ModalSpec.js
```diff
@@ -85,4 +85,16 @@ describe('Modal', function () {
...
 
+  it('Should pass dialogClassName to the dialog', function () {
+    let noOp = function () {};
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Modal dialogClassName="testCss" onRequestHide={noOp}>
+        <strong>Message</strong>
+      </Modal>
+    );
+
+    let dialog = instance.getDOMNode().getElementsByClassName('modal-dialog')[0];
```
mtscout6: Use of ReactTestUtils is preferred here: https://facebook.github.io/react/docs/test-utils.html#findrendereddomcomponentwithclass

src/Modal.js
aabenoja: Also, tests for prop validation to prevent regression.

corysnave:
Added prop type validation and used the test utils func Matt linked and assert.match in my test.

src/Modal.js
mtscout6: I think adding tests around a React.PropType validator is overkill. Our linter should have picked up that a new prop type was getting used without being declare though... @AlexKVal Have you seen this issue with the react eslint plugin not picking that up?

aabenoja:
Squash these commits, please.

aabenoja:
LGTM

@react-bootstrap/collaborators

mtscout6:
LGTM

src/Modal.js
AlexKVal: I have not seen such failures with `eslint-plugin-react`.

I don't see a problem with this code:
```jsx
propTypes: {
  ...
  dialogClassName: React.PropTypes.string // type check
}
... // and then it is used
<div className={classNames(this.props.dialogClassName, dialogClasses)}>
```

Or type checking wasn't there at the first place ?
Then I'll do more thorough investigation about the situation.

AlexKVal: Anyway I've checked it and I understand what the issue is now.

```js
let modal = (
  <div
...
return this.props.backdrop ? 
  this.renderBackdrop(modal) : modal; // here is the cause. 
```
When `return` returns something other than outer element (e.g. expression, variable or constant),
then `elint-plugin-react` linter has no clue how to check props usage and type checks.
It doesn't perform the full code evaluation.

AlexKVal:
Anyway I've checked it and I understand what the issue is now.

```js
let modal = (
  <div
...
return this.props.backdrop ? 
  this.renderBackdrop(modal) : modal; // here is the cause. 
```
When `return` returns something other than outer element (e.g. expression, variable or constant),
then `elint-plugin-react` linter has no clue how to check props usage and type checks.
It doesn't perform the full code evaluation.

aabenoja:
@AlexKVal Good find!

mtscout6:
Good to know, thanks for figuring that out.

PR #697, opened 2015-05-18
Improve code appearance and copyediting on docs

### LINKED ISSUE #94: code in Getting Started docs should be syntax highlighted

taion:
Fixes #94


dozoisch:
:+1:

mtscout6:
@joemcbride Did you see this one?

mtscout6:
Doesn't look like the mixed html content really helped that much:
screen shot 2015-05-18 at 8 31 14 pm

taion:
It's worth noting that we might want to revise this to use `codemirror/addon/runmode/runmode.node.js` so that the syntax highlighting doesn't have that noticeable delay on the browser side.

The `htmlmixed` mode is actually just for the parts of the "Introduction" and "Getting Started" pages that are actually displaying HTML. That odd syntax highlighting is just the normal CodeMirror JavaScript mode behavior.

mtscout6:
Fair enough, and I was thinking that delay in rendering was partly due to it not getting rendered server side when we generate the docs. @joemcbride Has been doing some work in this area to build this component in another library since we need it internally at work for our application's style guide. Ideally we would be able to move to use that and improve changes there. See: https://github.com/joemcbride/react-playground

Joe and I have talked about transferring it here to the org once it had matured enough, but I don't know where he stands with that at the moment. I can check with him on that in the morning.

taion:
The code path for applying syntax highlighting to a `code` block (the `runMode` bit that I'm using here) is unfortunately not the same as the path for applying syntax highlighting to a `textarea`. I think only the former can be used server-side.

taion:
I played around with the Node version of `runMode` but I couldn't figure out how to actually get any modes loaded. I'm okay with this for now, but I agree that the equivalent of `CodeExample` can find a happy home in something like react-playground.

joemcbride:
This is a nice little component.  I wasn't aware of the node runMode.  I'll look into enabling that with the `textarea`.  I think adding a component like `CodeExample` would fit perfectly into the react-playground project.

taion:
FWIW I already do use `runMode` in the `CodeMirrorEditor` component as the mobile fallback. I don't think it will work well with `textarea`s per se... CodeMirror does a lot of other stuff for its editor blocks than just the markup injection.

The nice thing with `runMode` is the 3rd arg can be a callback, which will let you use React to render an array of `span` elements with the markup... should be great for server-side rendering of examples, but you cannot import the provided mode files on the server, because they end up importing the main CodeMirror lib, which does some browser-specific things.

PR #701, opened 2015-05-19
Don't show random alert on components page

taion:


mtscout6:
LGTM, this has been annoying to me when trying to test keyboard accessibility. Thanks!

PR #703, opened 2015-05-19
Add link for new logo survey

mtscout6:
Our design team has come up with a few logos that are favicon friendly. I'd like to put it up to an anonymous vote for a week and see what comes of it. Thoughts?

https://www.surveymonkey.com/s/X7Q9VX7

### DIFF
--- a/docs/src/NavMain.js
+++ b/docs/src/NavMain.js
@@ -28,6 +28,9 @@ const NavMain = React.createClass({
     let links = Object.keys(NAV_LINKS).map(this.renderNavItem).concat([
       <li key='github-link'>
         <a href='https://github.com/react-bootstrap/react-bootstrap' target='_blank'>GitHub</a>
+      </li>,
+      <li key='new-logo-survey'>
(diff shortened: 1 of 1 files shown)


mathieumg:
I like the last one (React colors, yay!) but think it needs improving. In the spirit of the original logo, perhaps the B could use the Bootstrap purple it used to have? Also, it seems the B is not the same as Bootstrap's, its not as bold as it should be.

AlexKVal:
3 or 4 would do.
1 and 2 contains lines that will not bee seen in small icon IMHO.

`The great things are the simplest one` (Internet)

And it would be nice to assess small icon versions for toolbars.

mtscout6:
The survey results are in:

screen shot 2015-05-20 at 5 59 56 pm

I couldn't export them with the free plan of survey monkey, so a screen shot will have to do.

The winner:

screen shot 2015-05-19 at 2 29 44 pm

I will be updating this on the site shortly.

dozoisch:
:+1:

mathieumg:
So this is set in stone and not open to discussion? :grin: (See my message above)

mtscout6:
I was going to talk to my design guys about it today.

PR #710, opened 2015-05-20
Webpack Babel cacheDirectory option and Mocha eslint plugin addition

mtscout6:
These are a few changes I've been stewing on in the Dropdown rewrite branch which can come into the main line. Changes include adding eslint-plugin-mocha, which fails if someone accidentally commits a mocha test with the `.only` flag. Also, this adds a `.babel-cache/` directory to the webpack babel loader to speed up the webpack initialization time a little bit.


mtscout6:
@AlexKVal This is one such change I was talking about that would have required a change to every test you have in #706

AlexKVal:
Wow.. second run of `npm run docs` is almost instant :+1: 

As for tests for webpack-configs: I think I understand now why I want it to be tested.
I am new to `webpack` and I need more control over changes in configs.
I want to see something like colored diffs to know what are the consequences of such (as this) changes.
For me - there is no problem to mark changed lines, `cmd+C` and `cmd+V` them into test files to make tests green again :smiley: . (I know it is a kind of `monkey business`.)

I think at some point of time I will have enough experience with it and I will see the full picture without
"supportive" tools as those tests. As clear as you see it now :wink:

AlexKVal:
@mtscout6 It seems travisCI fails sometimes (io.js branch)
https://travis-ci.org/react-bootstrap/react-bootstrap/jobs/63383053
screen shot 2015-05-20 at 11 10 36 pm

AlexKVal:
I've created new branch and pushed it for TravisCI check.
https://travis-ci.org/AlexKVal/react-bootstrap/builds/63387191

AlexKVal:
I've restarted it a couple of times. No error.

mtscout6:
Ah, you know what I can see what that would happen now.

The build process runs the `webpack` cli multiple times and it's possible that two of those processes have passed that conditional exists check and both attempt to create the directory. Of course the first would win and the second would fail. I'll add a try catch around that for this potential race condition.

AlexKVal:
:+1:

PR #711, opened 2015-05-20
[added] Row: support a 'fluid' boolean, passing the row-fluid Bootstrap class

ronjouch:
Hi, any reason not to support this bootstrap feature that is already supported elsewhere like in `Grid`?

I took inspiration from the `Grid` component; `Grid` and `Row` are now mostly identical. ESLint and tests pass, and I added a simple example.


ronjouch:
Eh, turns out `row-fluid` is bootstrap2 only, forget about that.

PR #713, opened 2015-05-21
Add capability to run Phantom locally

mtscout6:
There is a bug with Chrome right now that makes testing rather painfully
slow if no part of the Chrome windows is visible. This has made it
really difficult to do much very fast since I'm constantly waiting for
Chrome to complete running tests. The bug is defined in
https://github.com/karma-runner/karma-chrome-launcher/issues/44.

The difference in test execution time is currently 30 seconds if not
visible, or 1.5 seconds when it is. Since I get the same results via
Phantom I'd like have the option to quickly do so when needed.


aabenoja:
:shipit:

PR #714, opened 2015-05-21
Handle race condition with creation of babel-cache directory

mtscout6:
This should handle the race condition @AlexKVal encountered with the build process as addressed in https://github.com/react-bootstrap/react-bootstrap/pull/710#issuecomment-104021320.


AlexKVal:
LGTM

Issue #717, opened 2015-05-21
Modals can't be used with server side rendering


maxmalov:
Modal#render method uses `document` reference at [L46](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js#L46) and [L78](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js#L78), so `React.renderToString` stucks and throws an error.

I see that this `in` class is used by `FadeInMixin` as well as `document.querySelectorAll`? Is it possible to move `document` reference somewhere else out of `render` method?

mtscout6:
Ick, that is less than ideal. We'd love a PR to fix that up.

aabenoja:
Fixed by #719

PR #719, opened 2015-05-22
[fixed] server side rendering for Modal component

maxmalov:
Using document reference in the render method will throw an error
on React.renderToString call on the server side.
See react-bootstrap/react-bootstrap#717.

The proper solution is to remove document.querySelector check. Obviously
we cant detect this feature on the server side, besides it leads to
differencies btw server and client side rendering output, so React will warn
about it, ex:

1. `render` on the server side will apply classes: `modal fade`
2. `render` on the client side will apply clasess: `modal fade in`

Also karma test environment is not suitable for testing server side rendering,
so mocha test run againt nodejs was added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "build": "babel-node tools/build-cli.js",
     "test-watch": "karma start",
-    "test": "npm run lint && npm run build && karma start --single-run",
+    "test": "npm run lint && npm run build && karma start --single-run && _mocha --compilers js:babel-core/register './test/server/*Spec.js'",
     "lint": "eslint ./",
     "docs-build": "babel-node tools/build-cli.js --docs-only",
     "docs": "docs/dev-run",
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -43,7 +43,7 @@ const Modal = React.createClass({
     let classes = {
       modal: true,
       fade: this.props.animation,
-      'in': !this.props.animation || !document.querySelectorAll
+      'in': !this.props.animation
     };
 
     let modal = (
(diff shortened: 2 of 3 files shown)


mtscout6:
I'm banging my head against the wall trying to figure out why `document.querySelectorAll` was introduced in the first place and if it's still needed. It seems to have been introduced in 96d13602b7589958d8441bbf5c9aa59d362ce4ef by @stevoland. Everything still seems to work without it now though. I'm wondering if there was something about the code base or even React at the time that necessitated this, whereas now it's not even needed. 

I like the idea of testing the components for Server side rendering, though I wonder if there is a way to survey which of the current tests should work in both and then do so with both. Though this can be done later as proposed in #720.

LGTM

Need an additional review by on of the @react-bootstrap/collaborators

aabenoja:
:+1:

taion:
@mtscout6 

It looks like it was due to how `FadeMixin` was implemented: https://github.com/react-bootstrap/react-bootstrap/commit/d3d3ddc6653c4854306fcecf0365592b5e60f356#diff-485e50263d15ff215f56744914b84ba9

Looks like the idea was to use `FadeMixin#componentDidMount` if `document.querySelectorAll` was available, but not otherwise.

mtscout6:
All seemed to work without it, but I only checked in Chrome. @taion Are you concerned with this being removed? It's not clear from your last comment.

taion:
I am not. FadeMixin no longer works like that. Just commenting on why that errant check was there in the first place.

mtscout6:
Good to know thanks!

PR #725, opened 2015-05-24
[fixed] Fix scroll top calculation for overlays

### LINKED ISSUE #724: Tooltips and Popovers not appearing when left/right aligned
In the documentation, at http://react-bootstrap.github.io/components.html#tooltips you can hover over the first and last Tooltip examples and they do not appear. Same thing clicking on the first and last examples for the Popover component.

taion:
Fixes #724 

Was calculating bod scrollTop in a way that didn't work in Firefox. This bug is not present on IE11.

Checked by hand with Chrome 44, Firefox 40, Safari 8 on OS X 10.10 because I'm still not sure how I'd write proper tests for this.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -281,19 +281,21 @@ const OverlayTrigger = React.createClass({
 
   _getContainerDimensions() {
     const containerNode = this.getContainerDOMNode();
-    let width, height;
+    let width, height, scroll;
+
     if (containerNode.tagName === 'BODY') {
       width = window.innerWidth;
       height = window.innerHeight;
+      scroll =
+        domUtils.ownerDocument(containerNode).documentElement.scrollTop ||
+        containerNode.scrollTop;
     } else {
       width = containerNode.offsetWidth;
       height = containerNode.offsetHeight;
+      scroll = containerNode.scrollTop;
     }
 
-    return {
-      width, height,
-      scroll: containerNode.scrollTop
-    };
+    return {width, height, scroll};
   },
 
   getPosition() {


src/OverlayTrigger.js
```diff
@@ -281,19 +281,21 @@ const OverlayTrigger = React.createClass({
...
   _getContainerDimensions() {
     const containerNode = this.getContainerDOMNode();
-    let width, height;
+    let width, height, scroll;
+
     if (containerNode.tagName === 'BODY') {
       width = window.innerWidth;
       height = window.innerHeight;
+      scroll =
+        domUtils.ownerDocument(containerNode).documentElement.scrollTop ||
```
AlexKVal: eventually this line sometimes is:
```js
domUtils.ownerDocument(domUtils.ownerDocument(this).body).documentElement.scrollTop ||
```

because of:
```js
// domUtils.js
function ownerDocument(componentOrElement) {
  let elem = React.findDOMNode(componentOrElement);
  return (elem && elem.ownerDocument) || document;
}
```
```js
// OverlayMixin.js
getContainerDOMNode() {
  return React.findDOMNode(this.props.container) || domUtils.ownerDocument(this).body;
}
```
```js
// this file
const containerNode = this.getContainerDOMNode();
...
domUtils.ownerDocument(containerNode).documentElement.scrollTop ||
```

It seems there are some technical debt exists in those files already and you just trap into it.

Have you seen this ? http://stackoverflow.com/a/20478983/3874570
```js
function getScrollTop() {
  return window.pageYOffset || document.documentElement.scrollTop || document.body.scrollTop || 0;
};
```
And this ? https://miketaylr.com/posts/2014/11/document-body-scrollTop.html
```js
(document.documentElement ||
 document.body.parentNode ||
 document.body).scrollTop;
```
I hope it will do some help. :cherries:

taion: I need specifically `ownerDocument` there for consistency with other positioning calculations. While there is some duplicated effort there, I think it's cleaner to do everything off the `containerNode` element than to e.g. duplicate that branch in `getContainerDOMNode`.

AlexKVal: Understandable. In that case it all LGTM :+1: 

--
What do you think about this cosmetic stroke:
```js
if (containerNode.tagName === 'BODY') {
  width = window.innerWidth;
  height = window.innerHeight;
  scroll = domUtils.ownerDocument(containerNode).documentElement.scrollTop
} else {
  width = containerNode.offsetWidth;
  height = containerNode.offsetHeight;
}

scroll = scroll || containerNode.scrollTop;
```
or even
```js
return {width, height, scroll || containerNode.scrollTop};
```
I'm totally OK with your code. :cherries: 
It's because I'm learning and trying different coding patterns to get any feedback.

taion: I don't think the latter example works w/r/t the object shorthand.

As for the former example, it works but it feels a little wasteful - on the forward branch I explicitly know that I need to use `containerNode.scrollTop`, so that extra comparison seems unnecessary.

AlexKVal: Ah. I see. Indeed I forgot that object shorthand is fairly new.
Thank you for the answers ! :rainbow:

PR #726, opened 2015-05-24
Fix deprecationWarning for node

epeli:
Checking for window object gives reference error in node.js (isomorphic apps) so do typeof check instead.


AlexKVal:
Nice catch. :sparkle:  Thank you :+1:

PR #738, opened 2015-05-27
[added] Now accepting a `block` property on the ButtonGroup component.

mathieumg:
It's pretty simple and straightforward! See the jsfiddle in the original issue for an example.

### DIFF
--- a/docs/examples/ButtonGroupBlock.js
+++ b/docs/examples/ButtonGroupBlock.js
@@ -1,0 +1,8 @@
+const buttonGroupInstance = (
+  <ButtonGroup vertical block>
+    <Button>Full width button</Button>
+    <Button>Full width button</Button>
+  </ButtonGroup>
+);
+
+React.render(buttonGroupInstance, mountNode);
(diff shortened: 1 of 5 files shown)


mtscout6:
Can you add some documentation for this?

mathieumg:
Done! I don't think there is any use-case for this when not using `vertical`, unless I'm mistaken. Hence, I put the example under the vertical section. I found there was a lack of spacing in-between the snippet for vertical and the paragraph detailing the block property so I added a `<br /> in-between. Would you rather I add the spacing a different way?

Also, since `block` needs `vertical` to do anything, should this enforced within the component or keep this as simple as possible and let the user do something nonsensical that has no effect if he so desires?

mtscout6:
I think you could give it a heading of "Block variation" then you wouldn't need the `<br />`

mathieumg:
That's what I did at first, but I wanted it to be clear that it *needs* to be `vertical` as well.

mtscout6:
Fair enough.

LGTM

mtscout6:
Though I wonder if propType validation should ensure that vertical is true when using block?

mathieumg:
Yeah, that's what I was hinting at in my earlier comment. When `vertical` is missing but `block` is set, should we:

* Apply `block` alone regardless and do nothing else. (As we do now)
* Apply `block` alone, as we do now, but warn that `vertical` is missing and required.
* Not apply `block` and warn that `vertical` is missing and required.
* Not apply `block`and fail silently.

I thin the first option is fine, but the second one could be interesting, while keeping things simple. Is there a non-programmatic way to validate the presence of sibling `propTypes`? I thought there wasn't.

mtscout6:
I like option 3 with help from #739

AlexKVal:
LGTM :sparkle:

PR #739, opened 2015-05-27
[added] CustomPropType.all to allow multiple validations

mtscout6:
This will allow us to have multiple validations applied on a single prop.

/cc @mathieumg This is what I'm using in the dropdown rewrite. An example of its usage can be seen [here](https://github.com/react-bootstrap/react-bootstrap/blob/dropdown-revisited/src/revisited/MenuItem.js#L47-L54)


mathieumg:
Welcome addition, thanks! LGTM.

Issue #231, opened 2014-09-15
Accessibility: MenuItem anchor has tabIndex of -1


rygine:
I'm not sure why the tabIndex is being forced to -1, but menu items are not accessible via keyboard navigation. Simply removing the tabIndex fixes the problem.

dozoisch:
Accessibility => putting in 1.0.0

mtscout6:
Related to #526

taion:
This was actually resolved by #526 - menus are keyboard accessible, just not via tab.

PR #740, opened 2015-05-27
[fixed] #231. Accessibility of MenuItem

whawker:
tabIndex is not provided by default, meaning MenuItems can be tabbed through when they are visible, and are ignored when hidden. A specific tabIndex can be provided via props. Also added aria `role="menuitem"`

### DIFF
--- a/src/MenuItem.js
+++ b/src/MenuItem.js
@@ -10,13 +10,15 @@ const MenuItem = React.createClass({
     target:    React.PropTypes.string,
     onSelect:  React.PropTypes.func,
     eventKey:  React.PropTypes.any,
-    active:    React.PropTypes.bool
+    active:    React.PropTypes.bool,
+    tabIndex:  React.PropTypes.string
   },
 
   getDefaultProps() {
     return {
       href: '#',
-      active: false
+      active: false,
+      tabIndex: null
     };
   },
 
@@ -29,7 +31,7 @@ const MenuItem = React.createClass({
 
   renderAnchor() {
     return (
-      <a onClick={this.handleClick} href={this.props.href} target={this.props.target} title={this.props.title} tabIndex="-1">
+      <a onClick={this.handleClick} href={this.props.href} target={this.props.target} title={this.props.title} tabIndex={this.props.tabIndex}>
         {this.props.children}
       </a>
     );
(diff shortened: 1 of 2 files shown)


mtscout6:
Thank you for your efforts here, but I'm in the middle of re-writing that component int he [dropdown-revisited branch (MenuItem)](https://github.com/react-bootstrap/react-bootstrap/blob/dropdown-revisited/src/revisited/MenuItem.js). That link will take you to the new component I'm working on, which is taking accessibility into account as it will support keyboard navigation and have all the appropriate aria roles. Feel free to take a look at it and let me know if you see anything alarming there.

whawker:
Ah fair enough. Is the revisited branch likely to land in v0.24.0?

If not, would this be possible to merge this PR as a short term fix?

src/MenuItem.js
```diff
@@ -10,13 +10,15 @@ const MenuItem = React.createClass({
...
+    active:    React.PropTypes.bool,
+    tabIndex:  React.PropTypes.string
   },
 
   getDefaultProps() {
     return {
       href: '#',
-      active: false
+      active: false,
+      tabIndex: null
```
mtscout6: This would be a breaking change since it is currently `-1` and not null. I've been planning to keep this as a `-1` in the revisited menu item to comply with keyboard accessibility as outlined in the [W3C WAI Aria guidelines](http://www.w3.org/WAI/PF/aria-practices/#menubutton) Is this something that you would still like configurable once I'm done with my re-write? #526

mtscout6:
If you can fix that default tab index to be the current default then this would not be a breaking change and I'd be ok with bringing it knowing it would be changed later.

whawker:
Hmm this is much more complicated than I first realised, I was just wanting to enable the focusing of MenuItems using the tab key.
I did not require tabIndex to be configurable, I added the tabIndex prop as I thought it might be useful for others.

I will take a more in-depth look into that Aria guideline link you have provided.

mtscout6:
As far as I understand the Aria spec the menu items should be reachable via the arrow keys, and not tab. Which is what I have working on the re-write branch. I'm still working through event handling there and then I should nearly be complete with the first of the re-write.

mtscout6:
I'm going to close this for now, if you find anything new or are able to change this to not have breaking changes then we can reconsider.

Issue #743, opened 2015-05-27
expose utils


tomatau:
Seems many of the utils files contain useful functionality, e.g.

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/domUtils.js
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/createChainedFunction.js
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/EventListener.js

Currently I'm requiring them directly:
```
const domUtils = require('react-bootstrap/src/utils/domUtils')
```

but this feels vulnerable to change.

It would be nice if these were exposed through your `index.js` to avoid duplicating functionality in projects or link to to third party implementations.

mtscout6:
You appear to be using ES6, and we do just transpile our src dir to the lib directory. You should be able to use:

`import domUtils from 'react-bootstrap/lib/utils/domUtils';`

I'd be open to adding a namespaced `Utils` object to the public api which exposed some of these things. The exception being anything copied from React like the `EventListener`. I've said this before but I'd rather we removed those from our library and just pull them from React. Though we can't remove them util we do #693.

If you're up to it go ahead and put together a PR that sets up the namespace. This would be done in a similar fashion as the `FormControls` namespace.

mathieumg:
I know it's a lot more work and effort to maintain, but I was thinking that perhaps any original utility (that wasn't copied from somewhere else or that isn't available in another way) could be released independently on npm? A bit like Lo-Dash have decided to do with the 3.x release and like Express/Connect do it with their middlewares starting with the 4.x release.

For instance, I'd love to be able to use the newly added https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/CustomPropTypes.js in a different React project that is independent of react-bootstrap.

If there is interest, I'd be ready to help with this.

mtscout6:
[__External Conversation__](https://gitter.im/react-bootstrap/react-bootstrap?at=5565db48dacb05b3394a1b3a)

taion:
I brought that up on Gitter specifically because it wasn't directly related to this discussion :p

tomatau:
Yes as said in my OP, either:
- expose through the index (via a Utils namespace or whatever)
- third party dependencies.

Both have merits, I would prefer the third party dependencies (library) -- but there's more potential for slight duplication when multiple versions conflict for a larger project, etc...

[...]

PR #745, opened 2015-05-27
exposed utils namespace

tomatau:
Exposed all utils into utils namespace.

fixes https://github.com/react-bootstrap/react-bootstrap/issues/743

Kept the reference to `EventListener` until a third party option is used instead.

Mapped Object.assign.js to just `assign`

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -2,10 +2,8 @@ import childrenValueInputValidation from './childrenValueInputValidation';
 import createChainedFunction from './createChainedFunction';
 import createContextWrapper from './createContextWrapper';
 import CustomPropTypes from './CustomPropTypes';
-import deprecationWarning from './deprecationWarning';
 import domUtils from './domUtils';
 import EventListener from './EventListener';
-import assign from './Object.assign';
 import ValidComponentChildren from './ValidComponentChildren';
 
 export default {
@@ -13,9 +11,7 @@ export default {
   createChainedFunction,
   createContextWrapper,
   CustomPropTypes,
-  deprecationWarning,
   domUtils,
   EventListener,
-  assign,
   ValidComponentChildren
 };
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -3,7 +3,6 @@ import createChainedFunction from './createChainedFunction';
 import createContextWrapper from './createContextWrapper';
 import CustomPropTypes from './CustomPropTypes';
 import domUtils from './domUtils';
-import EventListener from './EventListener';
 import ValidComponentChildren from './ValidComponentChildren';
 
 export default {
@@ -12,6 +11,5 @@ export default {
   createContextWrapper,
   CustomPropTypes,
   domUtils,
-  EventListener,
   ValidComponentChildren
 };
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,6 +1,5 @@
 import childrenValueInputValidation from './childrenValueInputValidation';
 import createChainedFunction from './createChainedFunction';
-import createContextWrapper from './createContextWrapper';
 import CustomPropTypes from './CustomPropTypes';
 import domUtils from './domUtils';
 import ValidComponentChildren from './ValidComponentChildren';
@@ -8,7 +7,6 @@ import ValidComponentChildren from './ValidComponentChildren';
 export default {
   childrenValueInputValidation,
   createChainedFunction,
-  createContextWrapper,
   CustomPropTypes,
   domUtils,
   ValidComponentChildren
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,0 +1,21 @@
+import childrenValueInputValidation from './childrenValueInputValidation';
+import createChainedFunction from './createChainedFunction';
+import createContextWrapper from './createContextWrapper';
+import CustomPropTypes from './CustomPropTypes';
+import deprecationWarning from './deprecationWarning';
+import domUtils from './domUtils';
+import EventListener from './EventListener';
+import assign from './Object.assign';
+import ValidComponentChildren from './ValidComponentChildren';
+
+export default {
+  childrenValueInputValidation,
+  createChainedFunction,
+  createContextWrapper,
+  CustomPropTypes,
+  deprecationWarning,
+  domUtils,
+  EventListener,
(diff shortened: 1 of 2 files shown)


src/utils/index.js
```diff
@@ -0,0 +1,21 @@
...
+import ValidComponentChildren from './ValidComponentChildren';
+
+export default {
+  childrenValueInputValidation,
+  createChainedFunction,
+  createContextWrapper,
+  CustomPropTypes,
+  deprecationWarning,
+  domUtils,
+  EventListener,
```
mtscout6: I'd rather not have this one here, since there is one available in React. So, long as you are using npm and CommonJS or ES6 Modules.

src/utils/index.js
```diff
@@ -0,0 +1,21 @@
...
+  assign,
```
mtscout6: Same applies to this one.

src/utils/index.js
tomatau: If EventListener is being removed from this namespace I think it should be removed from the libraries internals entirely and specified as a dependency.

Object.assign should be included via the ES6 -> ES5 build step.

mtscout6: That is the goal in the long run, but removing an internal usage is easier than removing something with explicit public exposure.

taion:
I want to limit the line notes, but it seems odd to me to expose `deprecationWarning`, and I'd strongly prefer that `createContextWrapper` not be exposed because it's not meant to be public.

Maybe instead of taking this approach, we should only expose `domUtils`, and add others as it becomes relevant?

tomatau:
I agree to limit exposure and deprecationWarning doesn't feel suitable to expose.

I've already found use for `domUtils`, `createChainedFunction` and `EventListener` when building a custom overlay trigger.

Also, it felt far more suitable to use this libs EventListener than duplicate the functionality of a cross browser event listener in my codebase. I feel this should be moved elsewhere.

Also `CustomPropTypes` seems like it would be popular.

mtscout6:
The `EventListener` is available today from the React npm module. Just use:

```
import EventListener from 'react/lib/EventListener';
```

taion:
This is in the context of using the bundles, no? Is `EventListener` exposed in ReactWithAddons?

Although I think I'm still inclined to say that if you are using the bundle, chances are you're not going to want/need to deal with this sort of thing.

mtscout6:
No `EventListener` is not available on ReactWithAddons, would be nice if it was. With that said though it is still there, and if they remove it then we can revisit the idea of including it in this library or another one.

tomatau:
I wasn't aware it was available through the lib :+1:

mtscout6:
Can you squash those commits and use the commit message format outlined in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes). That will allow your change to be reflected in our generated changelog on the next release.

src/utils/index.js
```diff
@@ -0,0 +1,15 @@
...
+import createChainedFunction from './createChainedFunction';
+import createContextWrapper from './createContextWrapper';
+import CustomPropTypes from './CustomPropTypes';
+import domUtils from './domUtils';
+import ValidComponentChildren from './ValidComponentChildren';
+
+export default {
+  childrenValueInputValidation,
+  createChainedFunction,
+  createContextWrapper,
```
taion: I'd really rather this not be exposed. There's intentionally no user-facing documentation for any internal users of this function.

taion:
One more thought - both this and `src/index.js` would be a lot cleaner if we enabled `es7.exportExtensions` under Babel Stage 1. You'd just write this whole thing as

```js
export childrenValueInputValidation from './childrenValueInputValidation';
export createChainedFunction from './createChainedFunction';
export CustomPropTypes from './CustomPropTypes';
export domUtils from './domUtils';
export ValidComponentChildren from './ValidComponentChildren';
```

Probably not necessary here, but it'd make things cleaner as a separate change.

tomatau:
squashed and added a commit prefix

NB. It's my first time squashing - is it supposed to persist the previous commit messages? The squash claims to be successful but I can still see them in the log and this PR history...

mathieumg:
> is it supposed to persist the previous commit messages

Not if you squashed them together with a new message. Did you force-push, or do you still see them in your local log?

tomatau:
@mathieumg 

I reset -i HEAD~4, picked one and squashed 3, changed the commit message of the one I picked to be more suitable, saved and wrote.. then was prompted for another commit message so repeated the previous one.

At this point my local log was only showing 1 commit.

After this - I realised that I'd made these changes onto my local master - so branched out before pulling origin master (on my fork) then merged my squashed changes back in and pushed.

At this point my local log was showing all commits :palm_tree:

dozoisch:
Basically

```
git rebase -i HEAD~X
```

Then you can do pick, s, s, s, s, etc. (f to squahs ignoring the commit message)

Then

```
git push --force origin branch_name
```

You have to push force with a squash because you are rewriting history.

mtscout6:
If you haven't seen this before it's a great video highlighting many of the cool things you can do with git. Squashing is one of the topics covered.

https://vimeo.com/43659036

AlexKVal:
> Also CustomPropTypes seems like it would be popular.

It is already the second voice for it (the first was in Gitter).

npm search `react-custom-prop-types` is free :wink:

mtscout6:
Other available names:

- `react-prop-types` (I'm partially in favor of this one)
- `react-enhanced-prop-types`
- `react-prop-type-utils`

mathieumg:
If `react-prop-types` is free, why not go with it!

tomatau:
could the chat about custom-prop-types be moved to another issue?

AlexKVal:
Yeah. I am sorry about starting it here :cherries: 
We should continue it in gitter for maintainers.

mtscout6:
@tomatau Sorry about that, will do

Did that video help you understand how to squash commits a little easier?

tomatau:
@mtscout6 thanks for the video - didn't add anything I didn't know about squash (something more complicated broke my push) but it did have a lot of really nice tricks and advice that I'm very grateful for!

I'm going to close this PR and submit a new one with a cleaner commit history

PR #747, opened 2015-05-27
Introduce Release Candidate process for breaking changes

mtscout6:
Documents proposed changes to the process by which we introduce breaking
changes and or remove functionality.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -87,7 +87,18 @@ violation of semver, than a patch release reverting the offending change should
 be pushed as soon as possible to correct the error. The offending change can
 then be re-applied and released with the proper version bump.
 
-## Live releasing the documentation
+### Release Candidates
+
+In an effort to reduce the frequency with which we introduce breaking changes we
+should do our best to first push deprecation warnings in a Minor or Patch
+release. Also, Pull Requests with breaking changes should be submitted against
+the `vX-rc` branch, where X is the next Major version. Which we will in turn
+release as an `alpha` release of the next Major version. When we are ready to
+release the next Major version bump we will merge the `vX-rc` branch into the
+`master` branch and cut a `beta` release.  Once bugs have been addressed with
+the `beta` release then we will release the Major version bump.
+
+### Live releasing the documentation
 
 The documentation release script does a similar job to the release script except
 that it doesn't publish to npm. It will auto tag the current branch with
@@ -122,7 +133,7 @@ Example usage of release-docs script:
 $ ./tools/release-docs
 ```
 
-## Check everything is OK before releasing
+### Check everything is OK before releasing
 
 Release tools have a very useful option `--dry-run`.
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 [![devDependency Status][dev-deps-badge]][dev-deps]
 [![peerDependency Status][peer-deps-badge]][peer-deps]
 
-__Under active development - APIs will change.__ Check out the [1.0.0 Roadmap](https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap) and [Contributing Guidelines][contributing] to see where you can help out.  
+__Under active development - APIs will change.__ Check out the [1.0.0 Roadmap](https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap) and [Contributing Guidelines][contributing] to see where you can help out. Prior to the 1.0.0 release, breaking changes should result in a Minor version bump.
 
 ## Docs
 
--- a/docs/src/HomePage.js
+++ b/docs/src/HomePage.js
@@ -26,7 +26,8 @@ export default class HomePage extends React.Component{
             href='https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap'>1.0.0
             Roadmap</a> and <a
             href='https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md'>Contributing
-            Guidelines</a> to see where you can help out.
+            Guidelines</a> to see where you can help out. Prior to the 1.0.0 release, breaking changes should
+            result in a Minor version bump.
           </Alert>
         </Grid>
 


docs/src/HomePage.js
```diff
@@ -26,7 +26,8 @@ export default class HomePage extends React.Component{
             href='https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap'>1.0.0
             Roadmap</a> and <a
             href='https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md'>Contributing
-            Guidelines</a> to see where you can help out.
+            Guidelines</a> to see where you can help out. Prior to the 1.0.0 release breaking changes should
```
mathieumg: comma after release?

README.md
```diff
@@ -12,7 +12,7 @@
 [![devDependency Status][dev-deps-badge]][dev-deps]
 [![peerDependency Status][peer-deps-badge]][peer-deps]
 
-__Under active development - APIs will change.__ Check out the [1.0.0 Roadmap](https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap) and [Contributing Guidelines][contributing] to see where you can help out.  
+__Under active development - APIs will change.__ Check out the [1.0.0 Roadmap](https://github.com/react-bootstrap/react-bootstrap/wiki#100-roadmap) and [Contributing Guidelines][contributing] to see where you can help out. Prior to the 1.0.0 release breaking changes should result in a Minor version bump.
```
mathieumg: Same here, comma after release?

docs/src/HomePage.js
mtscout6: Fixed.

mathieumg:
The text sounds good! I'm eager to see how this flow pans out.

README.md
mtscout6: Fixed

aabenoja:
:+1: I especially like the idea of commiting against the alpha and beta branches.

@react-bootstrap/collaborators Let's get some more feedback before pulling this in

dozoisch:
:+1: makes sense, specially since we're going toward 1.0.0. I would put that in the 1.0.0 milestone just so that it's easy to find for future ref!

mtscout6:
@dozoisch I was going off @mathieumg's response on [gitter](https://gitter.im/react-bootstrap/react-bootstrap?at=5565e97dd21e5ed02ff0607b). I don't think we'd need it in the 1.0.0 roadmap on the wiki if we have it the both the Contributing and Mainaining Guides.

dozoisch:
Alright, yeah makes sense

AlexKVal:
I like it too. :+1: 
It seems everything is in place.

taion:
I think the issue with having persistent `alpha` and `beta` branches to develop against is that they add the extra burden of continually rebasing onto a moving target of `master`, which is going to keep proceeding with non-breaking changes. I think git-flow works better when there are relatively few changes being made to the master branch, which is not likely to be the case here, since we should be making relatively few breaking changes. In general I think trunk-based development is preferred these days anyway.

I would say that the right way to handle this is to have persistent per-feature branches for breaking changes, and maybe a single `alpha` (or your choice of pre-release) branch from which releases are snapshotted when there are major breaking changes that people might want to run their code against. I don't think having a single parallel "development" branch makes sense when there's meaningful development happening against `master`.

taion:
I should say, rather than a single `alpha` branch, probably best to branch off separate e.g. `v0.24-rc` branches as needed, when we feel that RCs are needed. That way all of these branches can be minimally persistent.

mtscout6:
@taion I can see where your coming from when there's still active development that's viable on the master branch. I also like the idea of a less persistent, and as needed branching model. I'll update the documentation to reflect that idea and see what everyone thinks.

taion:
BTW, https://github.com/facebook/react is a good example of the trunk-based branching model, where `master` is the latest set of changes, and there are e.g. `0.13-stable` branches for maintenance releases.

Again, not relevant to us unless we start having maintenance releases, but you can think of an `v0.x-rc` branch as something like a maintenance branch that you're just doing releases off of, maybe. I don't know.

taion:
New revision looks great to me. So basically

1) Most development happens on master.
2) Breaking changes happen on specific per-release rc branches as needed.
3) As we release pre-releases, we merge from master to the rc branch.
4) I guess really big changes can live on a feature branch that then gets merged into the rc branch?

I think this accomplishes the goals we have, and is pretty lightweight. I like it.

MAINTAINING.md
```diff
@@ -100,17 +111,17 @@ that you start from that tag.
...
   - *Note: Checkout the tag and not master directly because some live
    documentation changes on master that could related to new components
    or updates for the upcoming release*
-0. Create a new branch from there (for example `git checkout -b docs/v0.22.1`)
-0. Cherry-pick the commits you want to include in the live update
+2. Create a new branch from there (for example `git checkout -b docs/v0.22.1`)
+3. Cherry-pick the commits you want to include in the live update
 `git cherry-pick <commit-ish>...`
-0. Use the release-docs script to push and tag to the documentation repository.
+4. Use the release-docs script to push and tag to the documentation repository.
```
dozoisch: why not letting 0.? They auto number

mtscout6: My vim plugin that fires up a rendered copy while I a make changes was not showing the auto number so I presumed it was not working, looking at github rendering it does work. So, it's a mistake on my part.

mtscout6:
@taion yes that's the vision I have for this.

taion:
:+1:

MAINTAINING.md
dozoisch: Alright! Yeah it's by definition in markdown normally

mtscout6: Fixed

AlexKVal:
:shipit:

MAINTAINING.md
mathieumg: I personally prefer when they're explicit, even though there is auto numbering. Markdown was meant to be easy to read even in its text form.

PR #755, opened 2015-05-29
[changed] Fix button dropdown not closing when clicking button on IE8

whawker:
Fixes a bug where attempting to close a button dropdown menu, by clicking the *button* itself did not close the menu.

`e.target` is undefined in IE8 so `isNodeInRoot` returned false, allowing the click event to bubble to the document and re-open the menu.

### DIFF
--- a/src/DropdownStateMixin.js
+++ b/src/DropdownStateMixin.js
@@ -49,7 +49,9 @@ const DropdownStateMixin = {
   handleDocumentClick(e) {
     // If the click originated from within this component
     // don't do anything.
-    if (isNodeInRoot(e.target, React.findDOMNode(this))) {
+    // e.srcElement is required for IE8 as e.target is undefined
+    let target = e.target || e.srcElement;
+    if (isNodeInRoot(target, React.findDOMNode(this))) {
       return;
     }
 
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -42,7 +42,9 @@ export default class RootCloseWrapper extends React.Component {
 
   handleDocumentClick(e) {
     // If the click originated from within this component, don't do anything.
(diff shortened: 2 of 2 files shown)


mtscout6:
Can you amend your commit message to follow the commit message format outlined in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes)?

src/DropdownStateMixin.js
```diff
@@ -49,7 +49,8 @@ const DropdownStateMixin = {
   handleDocumentClick(e) {
     // If the click originated from within this component
     // don't do anything.
-    if (isNodeInRoot(e.target, React.findDOMNode(this))) {
+    let target = e.target || e.srcElement;
```
mtscout6: Can you add a comment above this line stating that `e.srcElement` is specifically for IE8? This will allow us to deprecate this easier when Bootstrap v4 comes out.

src/RootCloseWrapper.js
```diff
@@ -42,7 +42,8 @@ export default class RootCloseWrapper extends React.Component {
 
   handleDocumentClick(e) {
     // If the click originated from within this component, don't do anything.
-    if (isNodeInRoot(e.target, React.findDOMNode(this))) {
+    let target = e.target || e.srcElement;
```
mtscout6: A similar comment would be good here.

mtscout6:
LGTM

Need one more set of eyes from one of the @react-bootstrap/collaborators

dozoisch:
LGTM 

do we need to ref this, when BS4 gets out and we drop ie8?

mtscout6:
That's why I had him mention the IE8 specific bits in code comments. We'll worry about a bootstrap v4 milestone when they finally push an alpha release.

PR #757, opened 2015-05-29
Keyboard accessibility for anchors serving as buttons

mtscout6:
Bootstrap uses a lot of styling that is specifically targeting anchor tags that may also serve in the capacity as a button. Unfortunately since Bootstrap does not style said buttons we have to use an anchor tag instead. But in order to maintain keyboard functionality for accessibility concerns even those anchor tags must provide an href.

The solution is to add an internal `SafeAnchor` component which ensures that something exists for the href attribute, and calls `event.preventDefault()` when the anchor is clicked. It will then continue to invoke any additional `onClick` handler provided.

_Note: I've been debating in my own mind as to whether this is a breaking change or not. Essentially anywhere that someone can choose to provide an href and they opt not to; then the `event.preventDefault()` function will get invoked. This enables keyboard support, which is technically a bug fix. Although it does change the behavior of events that would use this. Thoughts?_

### DIFF
--- a/src/SafeAnchor.js
+++ b/src/SafeAnchor.js
@@ -1,0 +1,38 @@
+import React from 'react';
+
+/**
+ * Note: This is intended as a stop-gap for accessibility concerns that the
+ * Bootstrap CSS does not address as they have styled anchors and not buttons
+ * in many cases.
+ */
+export default class SafeAnchor extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick(event) {
+    if (this.props.href === undefined) {
+      event.preventDefault();
+    }
+
+    if (this.props.onClick) {
+      this.props.onClick(event);
+    }
+  }
+
+  render() {
+    return (
+      <a role={this.props.href ? undefined : 'button'}
+        {...this.props}
+        onClick={this.handleClick}
+        href={this.props.href || ''}/>
+    );
+  }
+}
+
+SafeAnchor.propTypes = {
+  href: React.PropTypes.string,
+  onClick: React.PropTypes.func
+};
--- a/src/ListGroupItem.js
+++ b/src/ListGroupItem.js
@@ -1,7 +1,7 @@
 import React, { cloneElement } from 'react';
 import BootstrapMixin from './BootstrapMixin';
 import classNames from 'classnames';
-
+import SafeAnchor from './SafeAnchor';
 
 const ListGroupItem = React.createClass({
   mixins: [BootstrapMixin],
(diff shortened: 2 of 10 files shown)


taion:
This is really cool. It also helps with styling - no more having to do `cursor: pointer;` for such links with `onClick` but no `href`.

aabenoja:
:+1:

dportzline83:
Just curious, has this be tested in multiple browsers? I'm not sure if any browsers would have issues with an empty ```href``` or not.

mtscout6:
Thanks for the feedback @dportzline83 I'll give it run through the different browsers.

mtscout6:
Here's what I've found

- Win7 IE9 - Works!
- Win8.1 IE11 - Works!
- Safari - Works!
- FireFox - I cannot tab to any links on the page, odd...
- Chrome - Works!

I didn't expect that outcome. But seeing as I can't tab to any link in FF then I think this should be fine.

mtscout6:
Thinking this through a little more I think I'm confident to call this a bug fix and not a breaking change. Unless someone disagrees.

mtscout6:
@dportzline83 Brought up a concern on an internal chat that this should also need the `role='button'` when no href is provided. I will plan to update that after lunch.

mtscout6:
@dportzline83 Tested Firefox on windows and it worked... Firefox on Mac doesn't like tabbing to anchors for some reason.

AlexKVal:
>Firefox on Mac doesn't like tabbing to anchors for some reason.

https://bugzilla.mozilla.org/show_bug.cgi?id=1034338
question:
> In your Mac's System Preferences > Keyboard > Shortcuts, do you have "Full Keyboard Access" toggled to "Text boxes and lists only" instead of "All controls" ?

answer:
> Setting Full Keyboard Access to All controls does fix the problem. That setting did not affect other browsers such as Chrome/Safari. Thanks!

mtscout6:
@AlexKVal Thanks for the tip about Firefox, I have changed that setting on my Mac and verified that all works as expected in Firefox on Mac. I have also pushed up corrections for the issues you found.

src/SafeAnchor.js
```diff
@@ -0,0 +1,38 @@
...
+      this.props.onClick(event);
+    }
+  }
+
+  render() {
+    return (
+      <a role={this.props.href ? undefined : 'button'}
+        {...this.props}
+        onClick={this.handleClick}
+        href={this.props.href || ''}/>
```
AlexKVal: Also I was going to propose to change it to `javascript:void(0)`
but then learned that problems with `href=''` exists only in IE6 & IE7.

AlexKVal:
:shipit: 

We need a feedback from users to find out hidden pitfalls with it :smiling_imp:

mtscout6:
Would you rather I put this in the `v0.24-rc` branch and push with an alpha release?

AlexKVal:
hey.. we have alpha branches now to get feedback :+1: 
I would be for it :smiley: , but my voice doesn't count on such kind of questions.

mtscout6:
That's partly what I was considering in my initial description of this PR :tongue:

mtscout6:
Re-submitted against `v0.24-rc` in #760

PR #758, opened 2015-05-29
Add means to load local dev docs on a different machine

mtscout6:
This is useful for cross-browser testing if you have VMs with dedicated versions of browsers.

Since the webpack-dev-server runs on a different port and we need to add http headers to allow cross-origin requests we need this in order to run the docs on one machine and test on another.

The drawback is that in order to reach the webpack-dev-server directly you will have to use the ip address and not `localhost`. Normal usage of the docs can still be reached with `http://localhost:4000` though.

### DIFF
--- a/docs/dev-run
+++ b/docs/dev-run
@@ -3,6 +3,7 @@
 import 'colors';
 import portfinder from 'portfinder';
 import { exec } from 'child-process-promise';
+import ip from 'ip';
 
 portfinder.basePort = 4000;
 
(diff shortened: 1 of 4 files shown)


taion:
This actually came up for me the other day while trying to test with IE. Cool!

LGTM

AlexKVal:
:+1: 
Thank you :cherries: 
It allows me to test on `iPad` with `npm run docs` also.
I was using `npm run docs-prod` only till now because of that.

PR #760, opened 2015-05-29
Keyboard accessibility for anchors serving as buttons

mtscout6:
Bootstrap uses a lot of styling that is specifically targeting anchor
tags that may also serve in the capacity as a button. Unfortunately
since Bootstrap does not style said buttons we have to use an anchor
tag instead. But in order to maintain keyboard functionality for
accessibility concerns even those anchor tags must provide an href.

The solution is to add an internal `SafeAnchor` component which ensures
that something exists for the href attribute, and calls
`event.preventDefault()` when the anchor is clicked. It will then
continue to invoke any additional `onClick` handler provided.

### DIFF
--- a/src/ListGroupItem.js
+++ b/src/ListGroupItem.js
@@ -1,7 +1,7 @@
 import React, { cloneElement } from 'react';
 import BootstrapMixin from './BootstrapMixin';
 import classNames from 'classnames';
-
+import SafeAnchor from './SafeAnchor';
 
 const ListGroupItem = React.createClass({
   mixins: [BootstrapMixin],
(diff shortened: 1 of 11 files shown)


mtscout6:
@react-bootstrap/collaborators How does this look? Any more concerns to getting this merged? It will be going to the v0.24-rc branch.

dozoisch:
I like the new SafeAnchor!

LGTM

mathieumg:
I like this and I also think exposing it is a great idea! :+1:

mtscout6:
I guess it wouldn't hurt since we pretty much expose everything anyways. We tried this with the `FormGroup` component and ended up getting requests to expose it anyways.

I didn't want to expose it since the goal would be to remove it when we migrate bootstrap v4.

mtscout6:
@mathieumg It's added!

mathieumg:
Awesome! Thanks. If it's to be removed, it can always be deprecated first and then be moved to a separate module if there is an outcry or just die silently if not.

PR #761, opened 2015-05-29
Added pre-release changelog to release script

### LINKED ISSUE #746: Pre-releases should not generate a changelog
When releasing with the `--pre` option we should not generate a changelog. This may require a deviation from the current changelog tool to pick up changes since the last non-pre-release release.

mtscout6:
Closes #746

Creates a different changelog when releasing a pre-release, then when an actual release occurs it is removed.

The idea is that the actual release would contain the same release information but in a single entry.


tools/release-scripts/release.js
```diff
@@ -42,6 +42,10 @@ const yargsConf = yargs
 const argv = yargsConf.argv;
 setExecOptions(argv);
 
+if (argv.dryRun) {
+  console.log('DRY RUN'.magenta);
```
mathieumg: I'm just curious, where is `console` overridden to support these colors?

mathieumg:
LGTM.

tools/release-scripts/release.js
dozoisch: `import 'colors'`

mathieumg: I was looking for it in this file, thanks!

mtscout6:
It's not quite ready to merge I found a kink in how it works. I'm working on it now and should have it wrapped up later today.

tools/release-scripts/release.js
mtscout6: @mathieumg You're right I needed to add that to this file as well, good catch.

mtscout6:
Should be good to go now.

mtscout6:
@react-bootstrap/collaborators Any other concerns with this?

AlexKVal:
:shipit:

PR #764, opened 2015-05-31
Added Style prop to MenuItem

RichardLitt:
I'm having some issues understanding the build process, so I've edited just the one file for now. What more do I need to make this PR a reality?

### DIFF
--- a/src/MenuItem.js
+++ b/src/MenuItem.js
@@ -10,7 +10,8 @@ const MenuItem = React.createClass({
     target:    React.PropTypes.string,
     onSelect:  React.PropTypes.func,
     eventKey:  React.PropTypes.any,
-    active:    React.PropTypes.bool
+    active:    React.PropTypes.bool,
+    style:     React.PropTypes.object
   },
 
   getDefaultProps() {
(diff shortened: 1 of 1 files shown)


dozoisch:
Can't you style it using css?

RichardLitt:
Not using an inline style. I'd have to pull in a separate CSS sheet to override.

AlexKVal:
This issue exists because
- it is possible to apply any additional `props` (including `style`) to the `<li>` [MenuItem.js#L53](https://github.com/react-bootstrap/react-bootstrap/blob/0904adc2aedf2c0c0c0f8ba00eefce426c3d2e2a/src/MenuItem.js#L53)
- but not to the inner `<a>` [MenuItem.js#L32](https://github.com/react-bootstrap/react-bootstrap/blob/0904adc2aedf2c0c0c0f8ba00eefce426c3d2e2a/src/MenuItem.js#L32)

It seems a solution has to be something like this:
```js
<a {...this.props} onClick={...
```

But it is not so straightforward and simple as for now.

RichardLitt:
Exactly. What's stopping that from being the solution?

mtscout6:
Stuff like this is difficult. The concern is consistency through the rest of the library. Generally we've been trying to put stuff like this on the parent element of the component. So, in this case it would need to be applied to the `<li>`, but that doesn't help you out. In the case where it's been getting applied to a child element within the same component then we've been prefixing the prop with that elements name. Granted this has been the approach we've been taking for `className`. We haven't really considered anything similar for `style`. The other issue with this implementation is that the `style` prop would be equally applied to the `<li>` and the `<a>`, which would be undesirable in many cases. I agree with @dozoisch in this case that you're better off applying custom CSS.

RichardLitt:
Alright, I'll do that. Thanks for the explanation! Seems a bit odd, though.

Issue #754, opened 2015-05-29
Error during npm run build


lyf78062919:
[31mError: Command failed: ��ʱ��Ӧ�� <<��
 `babel  --out-file D:\www\javascript\temp\react-bootstrap\lib\factories\index.js <<EOF
 
import Accordion from './Accordion';

import Affix from './Affix';

import Alert from './Alert';

import Badge from './Badge';

import Button from './Button';

import ButtonGroup from './ButtonGroup';

import ButtonInput from './ButtonInput';

import ButtonToolbar from './ButtonToolbar';

import CollapsibleNav from './CollapsibleNav';

import Carousel from './Carousel';

import CarouselItem from './CarouselItem';

import Col from './Col';

import DropdownButton from './DropdownButton';

import DropdownMenu from './DropdownMenu';

import Glyphicon from './Glyphicon';

import Grid from './Grid';

import Input from './Input';

import Interpolate from './Interpolate';

import Jumbotron from './Jumbotron';

import Label from './Label';

import ListGroup from './ListGroup';

import ListGroupItem from './ListGroupItem';

import MenuItem from './MenuItem';

import Modal from './Modal';

import Nav from './Nav';

import Navbar from './Navbar';

import NavItem from './NavItem';

import ModalTrigger from './ModalTrigger';

import OverlayTrigger from './OverlayTrigger';

import PageHeader from './PageHeader';

import Panel from './Panel';

import PanelGroup from './PanelGroup';

import PageItem from './PageItem';

import Pager from './Pager';

import Popover from './Popover';

import ProgressBar from './ProgressBar';

import Row from './Row';

import SplitButton from './SplitButton';

import SubNav from './SubNav';

import TabbedArea from './TabbedArea';

import Table from './Table';

import TabPane from './TabPane';

import Thumbnail from './Thumbnail';

import Tooltip from './Tooltip';

import Well from './Well';

export default {

  Accordion,

  Affix,

  Alert,

  Badge,

  Button,

  ButtonGroup,

  ButtonInput,

  ButtonToolbar,

  CollapsibleNav,

  Carousel,

  CarouselItem,

  Col,

  DropdownButton,

  DropdownMenu,

  Glyphicon,

  Grid,

  Input,

  Interpolate,

  Jumbotron,

  Label,

  ListGroup,

  ListGroupItem,

  MenuItem,

  Modal,

  Nav,

  Navbar,

  NavItem,

  ModalTrigger,

  OverlayTrigger,

[...]

PR #762, opened 2015-05-30
Rely on babel api instead of executing the babel cli

Cellule:
fixes #754
Implements some utility function to build with babel using the api.
`buildGlob` is not use at this moment and can be removed if you wish. It was mostly an attempt to mimic babel-cli.
`'./test/server/*Spec.js'` was not recognised by mocha. I'm not sure if it's particularly related to Windows, but it works this way and I'm confident it will work for Linux.

PR #769, opened 2015-06-01
[fixed] #754. Error during npm run build on Windows

lyf78062919:
```javascript
return exec(`babel ${babelOptions} --out-file ${outpath} <<EOF\n ${content}`);
```
It doesn't work on Windows platform and this is the solution to write a temporary file then run babel and delete the temp file.

### DIFF
--- a/tools/generateFactories.js
+++ b/tools/generateFactories.js
@@ -13,7 +13,10 @@ export default function generateFactories(destination, babelOptions='') {
 
   let generateCompiledFile = function (file, content) {
     let outpath = path.join(destination, `${file}.js`);
-    return exec(`babel ${babelOptions} --out-file ${outpath} <<EOF\n ${content}`);
+    return fsp.writeFile(outpath + '_temp', content)
+        .then(() => exec(`babel ${babelOptions} --out-file ${outpath} ${outpath}_temp`)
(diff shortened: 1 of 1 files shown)


tools/generateFactories.js
```diff
@@ -13,7 +13,10 @@ export default function generateFactories(destination, babelOptions='') {
 
   let generateCompiledFile = function (file, content) {
     let outpath = path.join(destination, `${file}.js`);
-    return exec(`babel ${babelOptions} --out-file ${outpath} <<EOF\n ${content}`);
+    return fsp.writeFile(outpath + '_temp', content)
+        .then(() => exec(`babel ${babelOptions} --out-file ${outpath} ${outpath}_temp`)
+          .then(() => fsp.remove(outpath + '_temp'))
```
aabenoja: I'm not a huge fan of this `then` being inside another `then`, but that is mainly my personal preference. Is there a more technical reason behind this?

mtscout6:
I'd like to put this on hold pending the resulting discussion from https://github.com/react-bootstrap/react-bootstrap/pull/762#issuecomment-107562909

mtscout6:
@lyf78062919 Thanks for your effort on this. We are going to close this in favor of #762.

Issue #756, opened 2015-05-29
Enable AppVeyor for CI testing on Windows


mtscout6:
I tried to get AppVeyor working a while ago, but after fighting with it for more than a day I had to put it to the side (#476). Having it would be really helpful to identify when we break the build on Windows machines. This would help us avoid issues like #754, #491, and #471.

For anyone interested in trying to tackle this there is an `appveyor.yml.tmp` file at the root of this repo with my first stab at that configuration. Renaming that file to `appveyor.yml` should renable the existing AppVeyor build at https://ci.appveyor.com/project/mtscout6/react-bootstrap  (I don't know how to have AppVeyor register under the organization name, hence my username in the url)

PR #775, opened 2015-06-02
Re-enable AppVeyor

mtscout6:
I finally figured out how to get this project building on AppVeyor. This change will enable all PRs to be built on windows machines.

One thing that kinda stinks with this is that the free plan for AppVeyor only allows for one build at a time, that coupled with slower build times on windows means it may take a long time to know that a PR is good to merge.

It may stink but at least we can get feedback for Windows builds long before we get issues saying they don't work anymore.

Some options to help with this:

1. Sponsors
  - If we have any sponsors out there that would be interested in fronting some cost for some added concurrent jobs that may be an option. The benefit of this is that the CI testing agents would essentially be available to the collaborator team to maintain.
  - The question about this approach is where do we house these donations. (This get's scary for me, unless someone has a good idea on what to do here)
2. Alternative CIs
  - The [FubuMVC](https://github.com/DarthFubuMVC) organization has graciously given permission to use their CI at http://build.fubu-project.org/. That build agent that they've got would be more responsive and have quicker builds since we would not need to install Node on each build. Although these are private machines that only a handful of collaborators can maintain. Namely myself (@mtscout6), @aabenoja, and @joemcbride. There may be more in the future but that VM can be reached by most of the guys here in my development shop.
3. Don't do anything yet and see how painful it is in practice.
  - AppVeyor _may_ be adding more concurrent build options for OSS projects within the year: http://help.appveyor.com/discussions/questions/774-can-we-get-more-concurrent-builds

@react-bootstrap/collaborators Thoughts? Which option would you like to pursue?


AlexKVal:
> AppVeyor only allows for one build at a time

I don't think that will be a problem. We have not so much PRs to check.

LGTM

mathieumg:
I vote in order: 3, 2 and then 1. I don't have an issue with 2, and I imagine it could even be enabled in parallel with this. I imagine this closes #756?

mtscout6:
Yes it does Close #756

AlexKVal:
3, then 2. 

I don't believe in sponsors for windows CI at this stage.

If there was a lot of IE6 and IE7 usage out there, then there will be some interest for sponsors.
Todays windows browsers are more and more adhere to web standards.
And `shims` and `polifills` are doing their job well.

Maybe... when v-1.0.0 and a lot more popularity.

mtscout6:
Sounds good to me we can give this a try for a while and see how it goes.

PR #776, opened 2015-06-02
Fixed pre-release to tag as such with npm publish

mtscout6:
The default is `latest` which would install the pre-release for users
installing via `npm install react-boostrap`, which we don't want.

### DIFF
--- a/tools/release-scripts/tag-and-publish.js
+++ b/tools/release-scripts/tag-and-publish.js
@@ -1,10 +1,18 @@
+import 'colors';
 import { safeExec } from '../exec';
+import semver from 'semver';
 import tag from './tag';
 
 export default (version) => {
   console.log('Releasing: '.cyan + 'npm module'.green);
+  let parsed = semver.parse(version);
+  let additionalPublishArgs = '';
+
+  if (parsed.prerelease.length > 0) {
+    additionalPublishArgs = `--tag ${parsed.prerelease[0]}`;
+  }
 
   return tag(version)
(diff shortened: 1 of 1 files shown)


tools/release-scripts/tag-and-publish.js
```diff
@@ -1,10 +1,17 @@
...
+  let additionalPublishArgs = '';
+
+  if (parsed.prerelease.length > 0) {
+    additionalPublishArgs = `--tag ${parsed.prerelease[0]}`;
+  }
 
   return tag(version)
-    .then(() => safeExec('npm publish'))
+    .then(() => safeExec(`npm publish ${additionalPublishArgs}`))
     .then(() => console.log('Released: '.cyan + 'npm module'.green));
```
mathieumg: You won't like me, but following your remark from yesterday, shouldn't you import `colors` here as well? :mag:

mtscout6: Technically yes, and thanks for finding the deficiency.

mtscout6: Fixed

AlexKVal:
LGTM

mathieumg:
:shipit:

AlexKVal:
Though some DRY technical debt is already accrue in `tools`.
Maybe some day they will be refactored also :smiley:

mtscout6:
Yes, we can DRY up a few things in there.

PR #777, opened 2015-06-02
In docs footer don't show docs version.

mtscout6:
Without this you would see v0.23.1-docs.0 in the footer of the docs.

screen shot 2015-06-02 at 12 20 32 pm


AlexKVal:
:+1:

PR #778, opened 2015-06-02
Do not release docs when publishing pre-release

mtscout6:
When I published v0.24.0-alpha.0 today the docs were published with it which is less than desirable. This should prevent that from happening again.


AlexKVal:
LGTM

dozoisch:
LGTM

Issue #502, opened 2015-04-09
Allow specifying alternative icon classes in Input and Carousel


LordSputnik:
Currently, the icon classes for rendering validation icons in Input components, and for rendering previous and next icons in Carousel components are both hardcoded to use specific glyphicons. It should be possible to select which icon classes to use here, since someone might not like the default icons, or might want to switch over to font-awesome.

So, I think it would be good if it were possible to pass icon settings as properties of the Input and Carousel components, either as class strings (eg. &lt;Input iconSet="glyphicon" successIcon="ok" /&gt;), or as icon components (eg. &lt;Input successIcon=MyOKIconComponent /&gt;) (if that's possible in React).

mtscout6:
I agree that would be nice, PRs Welcome!

thetimbanks:
I was taking a look at making this change and found that Bootstrap has the `glyphicon-chevron-left` and `glyphicon-chevron-right` hard coded in the carousel.less file.

https://github.com/twbs/bootstrap/blob/master/less/carousel.less#L129

I tried adding the `icon-next` and `icon-prev` classes, but the content is overwritten in that file as well to a different icon. 

I scanned the project and couldn't find any examples where the default Bootsrap CSS has been modified/overwritten (which I assume shouldn't be done). Are there any suggestions on how to tackle this?

mtscout6:
I'd imagine that we'd need to expose a means to override these: https://github.com/react-bootstrap/react-bootstrap/blob/v0.22.5/src/Carousel.js#L158-L172

I've been playing around with child components for things like this so we don't end up with a wall of prop configurations. Something similar to https://github.com/react-bootstrap/react-bootstrap/issues/342#issuecomment-99283154 or with what I'm doing in the new dropdown work here: https://github.com/react-bootstrap/react-bootstrap/blob/dropdown-revisited/src/revisited/DropdownButton.js#L99-L116

thetimbanks:
For a component that uses `this.props.children` in many locations, how would you suggest handling a situation where one or more of the children should be handled differently?

For example, in the Carousel component there are CarouselItem children that are used to create the indicators and the slides. If other children are added for custom controls, how should those be ignored? Is it as simple as just removing those children from `this.props.children` inside of a method similar to the example you gave?

mtscout6:
[...]

Issue #431, opened 2015-03-04
Investigate use of react-docs for document generation from jsdocs


mtscout6:
@morlay Has presented an Idea for generating the docs based on JsDocs in our components. I really like this idea, and would like to track it's discussion here. First presented in #428

mtscout6:
How will this work with multiple examples per component? For example the current [react-bootstrap button docs](http://react-bootstrap.github.io/components.html#buttons) show multiple examples and documentation with various things you can do with buttons. We do this same thing with other components as well.

Is there a way to have things like Mixins land on one page and components on another? @joemcbride and I have talked about adding another docs page specifically for the various Mixins.

Lastly, I know this is primarily based on jsdoc, but how hard would it be to pull the props directly from the [`propTypes`](https://github.com/morlay/react-doc/blob/master/example/components/button/Button.jsx#L31-L40) declaration?

morlay:
For this version it is support the multiple examples.
`@example` and `@exampleFile` can use after `@component`, `@property` and `@method`, then they will be convert to  an item of `[Array] examples`

    @component Button
    @example code here directly
    @exampleFile ./relative/path/to/exampleFile

    @property {String} bsStyle - button style - Use any of the available button style types to quickly create a styled button. Just modify thebsStyle prop.
    @example
    @example 

then the UI will list all the example. there is an example in my another project:

![demo](https://cloud.githubusercontent.com/assets/1667873/6499627/b85050c8-c339-11e4-90a0-9d3bcb19e666.png)

and for a example demo, we will get 

     {
        "requireList": [
              {
                "path": "react",
                "src": "require('react')",
                "dest": "require(\"react\")"
              },
              {
                "path": "/Users/morlay/gitRepo/react-doc/example/components/button/Button.jsx",
                "src": "require('../Button.jsx')",
                "dest": "require(\"/Users/morlay/gitRepo/react-doc/example/components/button/Button.jsx\")"
              }
            ],
      "contents": ""
   }

this will be use for generating a global requireList  https://github.com/morlay/react-doc/blob/master/example/docs/result.jsx to use injecting to ReactPlayground (https://github.com/morlay/react-doc/blob/master/ui/ReactPlayground.jsx#L147). replacing configure all in this file, I try to use the requireList to generate a `globalRequire`, then we can use it by this way.

     React.render(<ReactDoc globalRequire={require('./result.jsx')} info={require('./result.json')} />, document.body);

morlay:
[...]

PR #793, opened 2015-06-05
Added prevIconClass and nextIconClass props to Carousel with defaults…

mcraiganthony:
Allow specifying alternative icon classes in Carousel (refer to issue #502). I added two new props to the Carousel component that allow alternative icon classes to be passed to the previous and next buttons. If alternative icon classes are not passed, the Carousel defaults to using glyphicons. This gives developers the ability to use their own set of icons instead of being forced to use glyphicons.

Added two new props to the Carousel component:
prevIconClass: React.PropTypes.string
nextIconClass: React.PropTypes.string

Added default props for the two new props above:
prevIconClass: 'glyphicon glyphicon-chevron-left',
nextIconClass: 'glyphicon glyphicon-chevron-right'

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -2,6 +2,7 @@ import React, { cloneElement } from 'react';
 import classNames from 'classnames';
 import BootstrapMixin from './BootstrapMixin';
 import ValidComponentChildren from './utils/ValidComponentChildren';
+import Glyphicon from './Glyphicon';
 
 const Carousel = React.createClass({
   mixins: [BootstrapMixin],
@@ -17,7 +18,9 @@ const Carousel = React.createClass({
     onSlideEnd: React.PropTypes.func,
     activeIndex: React.PropTypes.number,
     defaultActiveIndex: React.PropTypes.number,
-    direction: React.PropTypes.oneOf(['prev', 'next'])
+    direction: React.PropTypes.oneOf(['prev', 'next']),
+    prevIcon: React.PropTypes.node.isRequired,
+    nextIcon: React.PropTypes.node.isRequired
   },
 
   getDefaultProps() {
@@ -27,7 +30,9 @@ const Carousel = React.createClass({
       pauseOnHover: true,
       wrap: true,
       indicators: true,
-      controls: true
+      controls: true,
+      prevIcon: <Glyphicon glyph="chevron-left" />,
+      nextIcon: <Glyphicon glyph="chevron-right" />
     };
   },
 
@@ -158,7 +163,7 @@ const Carousel = React.createClass({
   renderPrev() {
     return (
       <a className="left carousel-control" href="#prev" key={0} onClick={this.prev}>
-        <span className="glyphicon glyphicon-chevron-left" />
+        {this.props.prevIcon}
       </a>
     );
   },
@@ -166,7 +171,7 @@ const Carousel = React.createClass({
   renderNext() {
     return (
       <a className="right carousel-control" href="#next" key={1} onClick={this.next}>
-        <span className="glyphicon glyphicon-chevron-right"/>
+        {this.props.nextIcon}
       </a>
     );
   },
--- a/test/CarouselSpec.js
+++ b/test/CarouselSpec.js
@@ -112,4 +112,22 @@ describe('Carousel', function () {
     assert.equal(backButtons.length, 0);
     assert.equal(nextButtons.length, 1);
   });
+
+  it('Should allow user to specify a previous and next icon', function () {
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Carousel activeIndex={1} controls={true} wrap={false}
+        prevIcon={<span className='ficon ficon-left'/>}
+        nextIcon={<span className='ficon ficon-right'/>}>
+        <CarouselItem ref="item1">Item 1 content</CarouselItem>
+        <CarouselItem ref="item2">Item 2 content</CarouselItem>
+        <CarouselItem ref="item3">Item 3 content</CarouselItem>
+      </Carousel>
+    );
+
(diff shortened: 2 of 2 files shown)


taion:
Huh. Why don't:

https://github.com/react-bootstrap/react-bootstrap/blob/d9b08d31d38cd5a9c5cbb608008b7457192b92dc/src/Carousel.js#L161
https://github.com/react-bootstrap/react-bootstrap/blob/d9b08d31d38cd5a9c5cbb608008b7457192b92dc/src/Carousel.js#L169

use the `Glyphicon` component?

I'd actually implement this as a replacement for the element inside the anchor, and default to the respective `Glyphicon`s.

mfunkie:
I would say as long as the glyphicons don't use a component, that class should be swappable.  Maybe otherwise there could be a `renderPrev` and `renderNext` prop that defaults to

```js
renderPrev: () => <span className="glyphicon glyphicon-chevron-left" />
renderNext: () => <span className="glyphicon glyphicon-chevron-right" />
```

However, many people using Bootstrap don't use glyphicons and this pull request seems to follow an existing pattern in code.

mcraiganthony:
@mfunkie presents the ideal solution. Being able to dump whatever you like inside of the anchor tag would give us the most flexibility.

Since the glyphicon component wasn't being used in the Carousel, I was simply trying to stick with the established pattern.

taion:
I mean set the default value to: `<Glyphicon icon="chevron-right" />`, &c, and make the `propType` be `React.PropTypes.node.isRequired`.

The former is just a cleanup, and the latter seems to be a more correct way to model it than accepting a function or class name or whatever.

mtscout6:
I agree that allowing for the Glyphicon component and possibly swapping for something else would be ideal. For example allowing use of: https://github.com/andreypopp/react-fa.

I don't know if the Carousel pre-dates the Glyphicon component or not, but definitely precedes my time with the library.

A few more issues:

- This change should also include tests that assert that the changed props are applied.
- Added documentation for this feature, I wouldn't bring in FontAwesome or anything. Simply showing use of either different icons from Glyphicon or a custom component would suffice.

mfunkie:
@taion Would it be a required prop if we have a default value?

@mtscout6 The existing documentation for Carousel is a bit sparse, and thus overriding prev/next might seem out of place as far as I can see on the GitHub Pages site.  Are there any plans to bring in usage of something like [react-docgen](https://github.com/reactjs/react-docgen) and start documenting the propTypes in React Bootstrap?

taion:
I'm not 100% sure on the `isRequired` part, but it seems like it makes sense, independent of the presence of a default value.

mtscout6:
@mfunkie See #431 (I know you got that message from the slack room, just sharing for posterity)

mtscout6:
@mfunkie It would still be required if that default prop was removed so yes I think it should still be required.

mfunkie:
Maybe my understanding of isRequired is flawed.  As far as I know, isRequired's use case is only if there are no default props, otherwise it's not technically a required prop.  Otherwise, all props would be 'isRequired', even if the component fails gracefully.

mtscout6:
I could be flawed as well, either should be fine I guess. Since there is a default than there will always be one, we can easily revisit the isRequired option if the default is ever removed. Which is partly why I don't think it hurts for it to be there.

mcraiganthony:
Per @taion 's suggestion ...

Added two new props to the Carousel component:
prevIcon: React.PropTypes.node.isRequired
nextIcon: React.PropTypes.node.isRequired

Added default props for the two new props above:
prevIcon: <Glyphicon glyph="chevron-left" />,
nextIcon: <Glyphicon glyph="chevron-right" />

taion:
Looks good now. Do we need tests for this? I think https://github.com/react-bootstrap/react-bootstrap/blob/12d6ddb5e69cb87436123b6a61a11edaf347a65c/test/CarouselSpec.js#L109-L110 covers that the anchors get rendered appropriately.

Will need to squash before merge.

mtscout6:
Tests would be appropriate to ensure those props are respected and not ignored.

mcraiganthony:
@taion @mtscout6 Added unit test and squashed commits.

mtscout6:
One last minor issue, can you amend your commit message to follow our guidelines as outlined in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) This will ensure that this change in documented in our changelog via our automated changelog tool.

mathieumg:
> As far as I know, isRequired's use case is only if there are no default props, otherwise it's not technically a required prop. Otherwise, all props would be 'isRequired', even if the component fails gracefully.

I was thinking the same, it allows you to glance at `propTypes` only (or documentation generated from it) and know what you *have to* provide when rendering that component. Otherwise, in what case would a prop not be required?

taion:
LGTM

mtscout6:
@mathieumg I wasn't thinking of auto generated docs, in which case you're right we shouldn't use `isRequired`. We aren't there yet with auto docs, but definitely good to consider.

@mcraiganthony Sorry to go back and forth on that. It's an issue that I myself have not investigated thoroughly enough, so I'll take the blame for how that went down. Can you go ahead and remove the `isRequired` flag? Otherwise this looks good to me.

mtscout6:
I'll go ahead and remove those, sorry for all the confusion.

mtscout6:
`isRequired` prop removed in 5e4ff2fe193db1e48f4abfab30f3fb3e97c7e2e9

AlexKVal:
:+1:

mcraiganthony:
Thanks @mtscout6 . Much appreciated.

Issue #802, opened 2015-06-09
OverlayTrigger trigger='click' is being closed when rootClose==true


luisrudge:
Can anyone help me with this issue?

![popover closing](https://cloud.githubusercontent.com/assets/941075/8048136/52ac8742-0e24-11e5-8949-60b5b5cace8c.gif)

taion:
Do you have a code repro?

Main thing this could likely be is that rootClose looks for clicks outside the popover element. If you replace the popover element entirely, then the existing check does the wrong thing.

luisrudge:
@taion yeah. I was thinking something like that. Because it only happens when I click in a date of the next month, which replaces almost the entire calendar component.

taion:
Okay, looking more closely at typical context menu style behavior, it really should be `onMouseDown` in the handler anyway, which would probably fix the behavior here.

taion:
Hmm, except that's not what Bootstrap dropdowns do.

Any opinions what the correct behavior here is?

luisrudge:
well, I think that if you're clicking anywhere inside the popover, it shouldn't close at all.

taion:
The problem is that because you're replacing the popover entirely, your click is no longer landing inside the popover, since the old iteration of the popover is gone by the time that click is processed. Switching to mousedown would make life easier.

luisrudge:
how do I try with the mousedown?

luisrudge:
@taion any thoughts?

jquense:
In this particular use case I think the problem is that "click outside" is the wrong way to handle it (just an opinion from making a bunch of inputs like this). I would turn off rootClose, and handle closing the popup when your date "widget" loses focus (ie. focus moves from the input or the popover to something else). If you really need finer toothed control over it I don't know that the popover is the right tool for the job here, as it is mostly designed to handle simple content.

luisrudge:
@jquense if I remove rootClose, how do I manually close the popover?

jquense:
The comonent child of `OverlayTrigger`is given a a hide prop it can call, or (uglier) actually call the `.hide()` method on the overlay trigger instance. `this.refs.overlayTrigger.hide()`

taion:
It's actually always injected as `props.onRequestHide` on the overlay... same regardless of whether `rootClose` is used.

AlexKVal:
Superceded by #855

taion:
This was actually fixed by #839.

taion:
[...]

PR #807, opened 2015-06-09
Add contains "polyfill" to domUtils

jquense:
just a small internal change, replace `isNodeInRoot()` in RootCloseWrapper, with a elem.contains polyfill in `domUtils`. It offers better performance than recursively walking up parentNodes, not that it matters much in the instances we use it, but its a few lines of code, so why not.

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -4,24 +4,6 @@ import EventListener from './utils/EventListener';
 
 // TODO: Merge this logic with dropdown logic once #526 is done.
 
-/**
- * Checks whether a node is within
- * a root nodes tree
- *
- * @param {DOMElement} node
- * @param {DOMElement} root
- * @returns {boolean}
- */
-function isNodeInRoot(node, root) {
-  while (node) {
-    if (node === root) {
-      return true;
-    }
-    node = node.parentNode;
-  }
-
-  return false;
-}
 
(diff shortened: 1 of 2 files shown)


taion:
LGTM

taion:
On second thought - I think the right way to handle `RootCloseWrapper` is to `stopPropagation` in an `onClick` handler of the popover element itself. This ought to resolve #802 as well. If you don't mind, I might remove this if I make that change.

jquense:
feel free to remove it from RootCloseWrapper if need be, though if you could leave the polyfill in general I am using it in #810.

taion:
Ah - that makes sense. Got it.

mtscout6:
Since our `domUtils` are technically part of our public API this should have had the `[added]` commit message for our changelog. I resolved it my including https://github.com/react-bootstrap/react-bootstrap/commit/9ca26e944d77c8522091a16b11a219ff46dd5c3c. I just wanted to mention it here so we are mindful of the utils in our changelog.

jquense:
Good to know, is there a list of what we consider public anywhere and I am just missing it?

mtscout6:
Any index.js file will reference "public" items. Even though we do encourage direct imports via `react-bootstrap/lib/Alert` for example. For the most part if it's in the src folder than consider it public. Hopefully when we can enhance the docs to be generated this will be easier to identify.

Issue #354, opened 2015-01-30
When you open a modalwhen the vertical scrollbar is present, the page's content shifts


dumbmatter:
Compare [the "Custom Trigger" demo from React Bootstrap](http://react-bootstrap.github.io/components.html#modals) and [the "Large Modal" or "Small Modal" demo from Bootstrap](http://getbootstrap.com/javascript/#modals-sizes). Notice how in both cases the vertical scrollbar disappears, but only in the React Bootstrap version there is a jarring shift in all the content to adjust for the extra horizontal space available.

Probably related to [this code from Bootstrap's version](https://github.com/twbs/bootstrap/blob/c683e7dea763b19e9bb172c34f74f99b926607e0/js/modal.js#L255-L258).

Ofer-Sadgat:
Also, in chrome, if you have recently downloaded something and have the download bar at the bottom, it doesnt successfully hide the vertical scroll bar.

taion:
Consequence of #321. Need this logic from upstream: https://github.com/twbs/bootstrap/blob/efe2023014ec90f35c0f722af6f3ec7a8eab720e/js/modal.js#L267-L271

Issue #723, opened 2015-05-23
When a Modal is open keep focus on the Modal


mtscout6:
You can currently "tab" off the `Modal` into elements under the `Modal`. This is both bad practice and required for Accessibility.

/cc @jportela

citelao:
The code [here](https://github.com/gdkraus/accessible-modal-dialog/blob/master/modal-window.js#L97) is fairly robust, aside from relying on a constant selector string of tabbable elements (`a[href], area[href], input:not([disabled]), ...`).

Note, the selector string should be `*[tabindex]:not([tabindex=-1])` at the end, instead of `*[tabindex]`.

See [here](https://github.com/hull-ships/hull-login/blob/master/src/components/overlay/index.jsx#L131) for an implementation in React

jquense:
fixed in #810

PR #810, opened 2015-06-10
Modal improvements

jquense:
First few bits of getting the Modal logic up to par with upstream tbs. No breaking changes, no major api rewrites...yet

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -207,6 +207,12 @@ const Modal = React.createClass({
 
     this._containerIsOverflowing = container.scrollHeight > containerClientHeight(container, this);
 
+    this._originalPadding = container.style.paddingRight;
+
+    if (this._containerIsOverflowing) {
+      container.style.paddingRight = parseInt(this._originalPadding || 0, 10) + scrollbarSize + 'px';
+    }
+
     if (this.props.backdrop) {
       this.iosClickHack();
     }
@@ -236,6 +242,8 @@ const Modal = React.createClass({
     }
 
     let container = getContainer(this);
+
+    container.style.paddingRight = this._originalPadding;
 
     container.className = container.className.replace(/ ?modal-open/, '');
 
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -11,7 +11,46 @@ import EventListener from './utils/EventListener';
 // - Add `modal-body` div if only one child passed in that doesn't already have it
 // - Tests
 
+/**
+ * Gets the correct clientHeight of the modal container
+ * when the body/window/document you need to use the docElement clientHeight
+ * @param  {HTMLElement} container
+ * @param  {ReactElement|HTMLElement} context
+ * @return {Number}
+ */
+function containerClientHeight(container, context) {
+  let doc = domUtils.ownerDocument(context);
+
+  return (container === doc.body || container === doc.documentElement)
+      ? doc.documentElement.clientHeight
+      : container.clientHeight;
+}
+
+function getContainer(context){
+  return (context.props.container && React.findDOMNode(context.props.container)) ||
+    domUtils.ownerDocument(context).body;
+}
+
+
+if ( domUtils.canUseDom) {
+  let scrollDiv = document.createElement('div');
+
+  scrollDiv.style.position = 'absolute';
+  scrollDiv.style.top = '-9999px';
+  scrollDiv.style.width = '50px';
+  scrollDiv.style.height = '50px';
+  scrollDiv.style.overflow = 'scroll';
+
+  document.body.appendChild(scrollDiv);
+
+  scrollbarSize = scrollDiv.offsetWidth - scrollDiv.clientWidth;
+
+  document.body.removeChild(scrollDiv);
+  scrollDiv = null;
+}
+
 const Modal = React.createClass({
+
   mixins: [BootstrapMixin, FadeMixin],
 
   propTypes: {
@@ -35,8 +74,10 @@ const Modal = React.createClass({
   },
 
   render() {
-    let modalStyle = {display: 'block'};
+    let state = this.state;
+    let modalStyle = { ...state.dialogStyles, display: 'block'};
     let dialogClasses = this.getBsClassSet();
+
     delete dialogClasses.modal;
     dialogClasses['modal-dialog'] = true;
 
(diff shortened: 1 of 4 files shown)


AlexKVal:
the header of the third commit: `no no longer`
`[fixed] container content no no longer shifts when overflowing`

taion:
This will resolve #354, BTW.

src/utils/domUtils.js
```diff
@@ -1,5 +1,13 @@
 import React from 'react';
 
+
+let canUseDom = !!(
```
dozoisch: should that be a const?

taion:
And #723, I believe. Very nice.

src/Modal.js
```diff
@@ -11,7 +11,68 @@ import EventListener from './utils/EventListener';
...
+/**
+ * Firefox doesn't have a focusin event so using capture is easiest way to get bubbling
+ * IE8 can't do addEventListener, but does have onfocus in, so we use that in ie8
+ * @param  {ReactElement|HTMLElement} context
+ * @param  {Function} handler
+ */
+function onFocus(context, handler) {
+  let doc = domUtils.ownerDocument(context);
+  let useFocusin = !doc.addEventListener
+    , remove;
```
mtscout6: In generall we haven't been following a single var pattern in this library. For consistency can this just be another let definition. Webpack Uglify plugin should reduce it to single var in the optimized bundle.

mtscout6:
This all looks really good, minor style tweaks from single var pattern for consistency and it should be ready to go.

jquense:
I will make that change :) there is one thing that bothers me here that I don't know how to fix given the current of how overlays work. 

`.modal-open` is removed on componentWillUnmount, which should be the same thing as "Modal is closed" but given how the FadeMixin works it doesn't actually mean that, since there is now a cloned element in the DOM mascarading as the modal Component, this leads to the body class being removed too quickly, and the screen "jumps" during that 300ms transition.

Generally I have solved things like this by creating something like the TransitionGroup addon that will keep a component child mounted (by caching it in state) until the transition completes, and then unmounts, but I am not sure that would work here, since the Overlay's don't ask the component if they should be rendered or not.

mtscout6:
Latest CI failures should resolve if you rebase on the latest master

mtscout6:
@jquense Is that concern something you are considering in #811? Or is it blocking this PR from getting resolved?

jquense:
@mtscout6 I don't think it blocks this PR, just noting a known limitation to the fix here that doesn't bring it completely inline with upstream bootstrap. I 'd be happy to fix it before we merge if anyone can help figure out a simple way to address it, otherwise we can revisit when #810 is done.

One thought was just to have FadeMixin call an optional callback when it has faded out, but I wasn't quite sure how to safely implement that without breaking compat

mtscout6:
I'd say this is a smaller scoped problem that can be solved in another issue.

src/Modal.js
```diff
@@ -30,13 +92,20 @@ const Modal = React.createClass({
...
       keyboard: true,
       animation: true,
-      closeButton: true
+      closeButton: true,
+      enforceFocus: true
     };
   },
 
+  getInitialState(){
+    return { };
```
AlexKVal: empty ?
itsn't it the same as there is no `getInitialState()` at all ?

src/Modal.js
```diff
@@ -30,13 +92,20 @@ const Modal = React.createClass({
...
+  },
+
   render() {
-    let modalStyle = {display: 'block'};
+    let state = this.state;
+    let modalStyle = { ...state.dialogStyles, display: 'block'};
```
AlexKVal: `{ ...state.dialogStyles, display: 'block'}`
just for symmetry
`{ ...state.dialogStyles, display: 'block' }`
or
`{...state.dialogStyles, display: 'block'}`

src/Modal.js
```diff
@@ -119,30 +188,63 @@ const Modal = React.createClass({
...
   componentDidMount() {
+    const doc = domUtils.ownerDocument(this);
+    const win = domUtils.ownerWindow(this);
+
     this._onDocumentKeyupListener =
-      EventListener.listen(domUtils.ownerDocument(this), 'keyup', this.handleDocumentKeyUp);
+      EventListener.listen(doc, 'keyup', this.handleDocumentKeyUp);
+
+    this._onWindowResizeListener =
+        EventListener.listen(win, 'resize', this.handleWindowResize);
```
AlexKVal: why don't put this and L195 on the one line ?
anyway L208 and L213 are longer than when these are one-liners

src/Modal.js
```diff
@@ -119,30 +188,63 @@ const Modal = React.createClass({
...
+    let container = getContainer(this);
 
-    let container = (this.props.container && React.findDOMNode(this.props.container)) ||
-          domUtils.ownerDocument(this).body;
     container.className += container.className.length ? ' modal-open' : 'modal-open';
 
-    this.focusModalContent();
+    this._containerIsOverflowing = container.scrollHeight > containerClientHeight(container, this);
+
+    this._originalPadding = container.style.paddingRight;
```
AlexKVal: the `State` is used anyway here
L227 `this.setState(this._getStyles() //eslint-disable-line react/no-did-mount-set-state`
so why mix private fields with it ?

I'm just kinda an apprentice so I just don't know `cons` and `pros` with this approach.
The link to any article where I can learn about why here is used such approach 
(of course if you have any) would be awesome for me. 
Or you can just send me to google with the right search-phrase-query (I just don't know what is the right one)
:cherries:

src/Modal.js
jquense: It is pretty much the same yeah, except that providing an empty object lets you access properties on it for a falsey value instead of an error (Null Exception)
```js
// without the empty object in getInitialState this would throw the first render
var style = this.state.dialogStyles // state is null
```

I could be slightly more correct and set state to `{ dialogStyles: undefined }` but it seems unneeded

src/Modal.js
jquense: no real reason other then it struck me as easier to read  and was in line with the formatting pattern of the handler above it :)

src/Modal.js
jquense: you are right `state` would work just fine for these properties. In general though they don't ever change after mount, so they don't strike me as "view state" but just static properties. In general though, the reason you would put something on `this` instead of `state` is because you don't want to trigger a render when changing it. This particular instance doesn't require that, so it could be on state if folks find that easier to understand

AlexKVal: Thank you for the explanation :cake:

src/Modal.js
AlexKVal: Ah.. I see.. I should try it myself. Remove and see what's happening. :blush:

mtscout6:
LGTM

Is there anything else preventing this from getting merged today?

jquense:
not that I know of :)

dozoisch:
:shipit:

PR #812, opened 2015-06-11
Change Logo that is Favicon Friendly

mtscout6:
Changes the icon to the one proposed in #703 

It's finally here guys! My design guys were a little strapped for time so there was a little delay to a few minor tweaks to make it look good on the home page.


taion:
:+1:

dozoisch:
:shipit:

mathieumg:
Nobody would be interested by something like the following?

image

AlexKVal:
I like @mathieumg idea

AlexKVal:
Two colors from two projects that are addressed by this library. Nice idea :+1:

jquense:
I personally like the single color version, but I'm fairly minimal in taste

mtscout6:
I agree after seeing the multi-color that I like the single color better. The reason I say that is that if we did this we would also need to update the top left brand as well. Then what do we do with the hyphen? I'd like to stick with the single color.

screen shot 2015-05-29 at 11 04 26 am

taion:
I thought we'd explicitly agreed on the single color one anyway.

mtscout6:
We did, @mathieumg had asked if I'd have my design guys look into the multi-color, which I asked, but they got swamped with internal issues so were unable to. I'm sorry @mathieumg but I think I'm going to say that we take this as it is for now.

mathieumg:
@taion I mentioned it in https://github.com/react-bootstrap/react-bootstrap/pull/703#issuecomment-103659688 but got no feedback whatsoever, I figured people might have wanted to see it to have a better idea. Looks like most people are against it anyway, that's fine!

AlexKVal:
> Looks like most people are against it anyway, that's fine!

No. Looks like there are no marketing guys among us. That's it :smile:

Jokes aside: I forgot about the logo 

In that case the icon has to be of one color too.

Anyway if somewhere in the future some company would decide to do marketing for this product,
they could figure out what is the best. 
As for now - it is more than sufficient to have any simple, but recognizable logo :cherries:

PR #815, opened 2015-06-11
Trim bower release

### LINKED ISSUE #693: Trim down the Bower release
We have been running into problems recently with the bower release not being as usable as the npm release primarily due to the different distribution mechanism that bower proposes.

Seeing as React, as well as many Bower packages, only provide the concatenated and minified build output I'm inclined to do the same here.

Pros:

- Easier to maintain since dependency differences between the two will be less complicated.
- React addons/utils can be cherry-picked as needed, and not committed to this repo.
- Allow us to use [`babel-runtime`](http://babeljs.io/docs/usage/runtime/) without declaring it as a bower dependency.
- Use `require('babel-core/polyfill')` in `Interpolate.js` and `OverlayTrigger.js` and remove `babel-plugin-object-assign` dependency.

Cons:

- Bower users will be forced to load the entire library all the time. They will not be able to optimize their release.
- This will complicate the react-router-bootstrap bower release and source. (Will be resolved by react-bootstrap/react-router-bootstrap#74)

My questions for bower users, since I don't personally use it:

- What is bower gaining you that you can't get with npm?
- What would you expect from a bower package?
- What am I missing here?

mtscout6:
Resolves #693

Dependent on PR #814

### DIFF
--- a/src/utils/deprecationWarning.js
+++ b/src/utils/deprecationWarning.js
@@ -8,7 +8,7 @@ export default function deprecationWarning(oldname, newname, link) {
     console.warn(message);
 
     if (link) {
-      console.warn(`You can read more about it here ${link}`);
+      console.warn(`You can read more about it at ${link}`);
     }
   }
 }
--- a/tools/amd/bower.json
+++ b/tools/amd/bower.json
@@ -16,7 +16,6 @@
     "**/.*"
   ],
   "dependencies": {
-    "react": ">= 0.13.0",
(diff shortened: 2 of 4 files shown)


AlexKVal:
LGTM

a bit less work for me with `tools` refactoring.

dozoisch:
LGTM

AlexKVal:
So be it. Minified 120Kb library :smiling_imp:

Issue #303, opened 2014-12-09
Can't access DropdownButton instance by ref


myagoo:
I would like to be able to set open state of my dropdownbutton instance by callling this.refs.dropdown.setDropDownState(false) by example.
Howerver, it seems that the ref property is lost(?) during the render method of the dropdownbutton component...

```javascript
return this[renderMethod]([
      <Button
        {...this.props}
        ref="dropdownButton"
        className={joinClasses(this.props.className, className)}
        onClick={this.handleDropdownClick}
        key={0}
        navDropdown={this.props.navItem}
        navItem={null}
        title={null}
        pullRight={null}
        dropup={null}>
        {this.props.title}{' '}
        <span className="caret" />
      </Button>,
      <DropdownMenu
        ref="menu"
        aria-labelledby={this.props.id}
        pullRight={this.props.pullRight}
        key={1}>
        {ValidComponentChildren.map(this.props.children, this.renderMenuItem)}
      </DropdownMenu>
    ]);
  }
```

tjwebb:
+1. says ref is undefined.

mtscout6:
I think the proper way to handle something like this would be through props. So the dropdown would have to be extended to allow external props to control it's state.

AlexKVal:
Superceded by #526

PR #817, opened 2015-06-11
[added] open and onOpenStateChange props for controlling Dropdowns.

rgrwkmn:
Allows an owner component to set the drop down state and be informed of drop down state changes from the dropdown.
Add tests for controlled dropdown.
Add documentation at the end of the DropdownButton section.

fixes react-bootstrap/react-bootstrap#303
also see react-bootstrap/react-bootstrap#526


mathieumg:
You might want to discuss with @mtscout6, I know he has something going in https://github.com/react-bootstrap/react-bootstrap/tree/dropdown-revisited

jquense:
PR, looks great though I wish, in general, if the components could implement this in a more predictable and maintainable way...has anyone thought about using something like [react-controllables](https://github.com/matthewwithanm/react-controllables) or, and I'm biased b/c i worte it, [uncontrollable](https://github.com/jquense/uncontrollable) to handle controlling props optionally?

I do this in react-widgets and it has been a huge benefit in maintainability, and removes a ton of complexity from the core components themselves

mathieumg:
> it has been a huge benefit in maintainability, and removes a ton of complexity from the core components themselves

@jquense I personally have no opposition against something that brings such improvements, no matter the implementation / the source.

mtscout6:
This is a lot more complex than what I was thinking to solve this. I was envisioning the idea of adding an `expanded` prop to `DropdownButton`. Then checking open state with something like:

```
let open = this.props.expanded !== undefined ? this.props.expanded : this.state.open;
```

Doing something like this not too complicated and easy to apply throughout the library without a ton of wrappers. Also this is easy to choose when to control or not without changing underlying components.

Also I'd prefer any major changes to the dropdown get applied on the dropdown-revisited branch, or discussed in #526.

I may not be understanding everything about your uncrontrollable library @jquense but it feels overly complicated to me.

jquense:
@mtscout6 The readme is a bit verbose but it essentially in encapsulates this pattern:

> let open = this.props.expanded !== undefined ? this.props.expanded : this.state.open;

rather then having the component do these checks everywhere, the Dropdown just always assumes `expanded` is a prop, and never state, meanwhile an intermediate component handles that should i use props or state logic. This works b/c the pattern is almost _always_ the same, there is a prop and a handler that sets that prop.

```js
ControllableDropdown = uncontrollable(Dropdown, { expanded: 'onToggle' })

// now you can do any of the following
<ControllableDropdown /> // expanded will be managed in state
<ControllableDropdown defaultExpanded={true} /> // still in state but initial value set

//and finally the user can control everything
let expanded = false;
<ControllableDropdown expanded={exapanded} onToggle={()=> expanded = !expanded}/>
```

So uncontrollable essentially pulls the "prop or state" logic out of the component and puts it in a wrapping component which you give to the user, that way you can just code the underlying component doesn't need to worry about it, but you get the convenience of controllable/uncontrolled component like react form components

jquense:
case in point: https://github.com/jquense/react-widgets/blob/master/src/DropdownList.jsx#L365

the dropdown doesn't have any logic in it to check state or props for `value`, or `open` but the user can optionally controll them, or leave them off and the component will handle it themselves

mtscout6:
Ok, I think I'm wrapping my head around this better this morning. I can see value in the simplicity in writing components that don't concern themselves with so much state.

To make sure I'm understanding this properly, with your uncontrollable library when you declare:

```
uncontrollable(Dropdown, { expanded: 'onToggle' });
```

I read in the readme that the result of `onToggle` would be supplied to the `expanded` prop/state. So, what provides the default function for `onToggle`? I didn't see an `onToggle` function in your dropdown list example there.

jquense:
if the user doesn't  provide the `expanded` prop the wrapping component will automatically provide an onToggle prop to the dropdown that is essentially just `value => this.setState({ expanded:  value})` if you do provide `expanded` the wrapping component will just get out of the way and pass through the props to the dropdown and make no attempt manage that value. 

so there is an assumed contract that is the same as form inputs (value/onChange) a prop that is optionally managed by an outside owner and a handler the component can call that indicates the value it wishes that prop to be.

lm I bit scattered today,  and off site so if it's unclear let me know, I can provide some more examples later, also the tests may be helpful to look at in either project for clearer demonstrations

taion:
Couldn't you get most of the benefit of this with just a simple helper function, something like `this::propOrState('expanded')` that pulls from either the prop or state as appropriate?

Then you can just set e.g. `state.expanded` as normal. I like the terseness of the HOC approach, but I think it loses a little in clarity in that it's less obvious what's going on, whereas using a simple helper function to pick between `props` and `state` will enable avoiding errors while still not compromising on readability.

taion:
Specifically I mean just to have

```js
function propOrState(key) {
  if (this.props[key] !== undefined) {
    return this.props[key];
  } else {
    return this.state[key];
  }
}
```

And as needed call e.g. `this::propOrState('expanded')` (if using the ES7 function bind syntax, else `propOrState.call(this, 'expanded')`) and just manage `state.expanded` ourselves.

rgrwkmn:
@mtscout6 I'll try out implementing your idea and send pull requests to the dropdown-revisited branch. When I ran tests on that branch last night it said it ran zero tests so I wasn't sure what the state of it was.

@jquense I'll check out react-controllables and uncontrollable to see if their patterns can be used effectively here.

Thanks for the feedback, I was surprised everyone was so talkative last night (I sent the PR right before leaving work) but I guess you guys could be in different timezones!

mtscout6:
@rgrwkmn That branch has a linting error at the moment causing some of the pain. Try running `npm run build && npm run test-watch` as that does not attempt to use linting. I have been rebasing and force pushing that branch on occasion so be mindful of that.

mtscout6:
I added the external state management to that branch as I proposed last night, so you may not need to do anything. But feel free to review the changes, any feedback would be good.

jquense:
@talion that's a viable minimal approach, there are few benefits to the HOC that a helper doesn't as easily allow.  first you get the default prop behavior for free, so `defaultExpanded` in our example, for initial values, piggybacking on that point you also get more explicit propTypes validations, included. such as warnings for including a prop without its handler. 

perhaps most beneficial is just the refactorability, any prop/handler pair can be made uncontrollable without altering the underlying component, since components can be completely dumb to that logic


there are cons as well. components that expose public apis via instance methods instead of props need to be explicitly forwarded since the component is wrapped, though I tend to think method apis are a bad idea in general. there is also theoretically a performance cost for the extra element creation, but I have not seen a case personally or otherwise that suffered from that (I'm sure they exist tho). it is also as you say, another level of indirection which may be confusing, or hard to follow

mtscout6:
@taion Did you see the concern I addressed about event chaining on [gitter](https://gitter.im/react-bootstrap/react-bootstrap?at=557b026f6fee6add0cc21339)?

mtscout6:
@rgrwkmn Since there are potential breaking changes with the dropdown-revisited work when I do wrap that up the PR would be against the v0.24.0-rc branch and not master.

rgrwkmn:
@mtscout6 I checked out your changes and updated my example in the docs just to see how it works. Using onClick, onRequestClose, and onSelect I can keep a my component and the DropdownButton in sync. 

Some suggestions on semantics:

1. I had to look at the source to know what onClick meant whereas if it was called onToggle I'd know right away.

2. Seems to me it should be either onRequestToggle and onRequestClose or onToggle and onClose since when you're controlling the component they do the same thing. Maybe internally the naming makes more sense as is.

3. It might make sense to have a single callback like onOpenChange that passes a boolean as the only argument to indicate whether the internals of DropdownButton are trying to open or close the dropdown.

I merged your changes into my branch and committed my example so you can check it out.

mtscout6:
@rgrwkmn Thanks for the feedback, I don't have time to look at it today but I updated #526 with the comment so I don't forget about it.

PR #824, opened 2015-06-13
[changed] Use named exports in index files

### LINKED ISSUE #350: ES6 imports and exports are not completly valid for ES6
I'm using jspm to import modules the ES6 way. At first I tried this:

```
import {ListGroup, ListGroupItem} from 'react-bootstrap';
```

this resulted in ``undefined`` for both though. Instead I had to use this:

```
import ListGroup from 'react-bootstrap/ListGroup';
import ListGroupItem from 'react-bootstrap/ListGroupItem';
```

I may be getting this entirely wrong as I'm still learning all this, but I thought I'd point it out in case it's an issue on your side.

taion:
Fixes #350


mtscout6:
LGTM

taion:
Squashed in an update explicitly showing the ES6 import syntax on the "getting started" page.

mtscout6:
Perfect I didn't think of those docs for this.

taion:
Going to leave this up a bit longer in case anybody has thoughts as to whether requiring e.g.

```js
import * as ReactBootstrap from 'react-bootstrap';
import * as FormControls from 'react-bootstrap/lib/FormControls';
```

v. the current

```js
import ReactBootstrap from 'react-bootstrap';
import FormControls from 'react-bootstrap/lib/FormControls';
```

when using ES6 module syntax is objectionable.

mtscout6:
Only other concern I just thought about https://github.com/react-bootstrap/react-bootstrap/issues/350#issuecomment-111728159

jquense:
I don't think that is an big issue...I think the transpilations/umd build will handle that correctly, just like the es6 -> commonJs works correctly, there wouldn't be any other way to handle that. But i may just being hopeful.

taion:
I expect this won't break users of the concatenated build. Checking real quick.

taion:
The global `ReactBootstrap` variable is still a keyed object with the public components as keys.

`utils` and `FormControls` would have worked, except I messed up the re-exports in the merge. They'll work after I squash a fix.

zerotacg:
I kind of feel like I set of an avalanche. I didn't mean to say the current way is wrong I merely was giving a possible workaround.
As for the fix. I'd vote for keeping the current default export along side the extra exports for compatibility.
In the long run I'd say it depends on how developers in general want to use the module syntax.
Is there an advantage of using

    import {Alert} from ...
    <Alert>...</Alert>

over

    import ReactBootstrap from ...
    <ReactBootstrap.Alert></ReactBootstrap.Alert>

well except from the length obviously? The shorter version sure looks sexy though

taion:
The former is more semantically correct, due to some stuff w/potentially having static module structure. Note that the latter pattern remains possible, you'd just have to do

```js
import * as ReactBootstrap from 'react-bootstrap';
```

taion:
AppVeyor failure is resolved by #826, I believe.

mtscout6:
@zerotacg It's no avalanche, just healthy discussion as we all learn and progress.

@taion I'll merge master into v0.24.0-rc so that change is on this branch.

LGTM

taion:
Okay, I'll merge for now, and we can revert if anybody comes back and objects.

AlexKVal:
```js
import {Alert} from ...
<Alert>...</Alert>
```
> ...is more semantically correct

:+1:

PR #826, opened 2015-06-13
Unlink babel internal api for build

Cellule:
Remove links to babel internal api and only use what we really need.
Addresses discussion at #762 
@sebmck Sorry for using your private api, I didn't know how to do what I needed at the time and I wanted to remain as close as possible to the original behavior. I should have asked to expose the api instead of using it like this. 

After some inspection, I realized that I didn't really need the internal api since our need is very simple.

### DIFF
--- a/tools/buildBabel.js
+++ b/tools/buildBabel.js
@@ -1,13 +1,12 @@
 import { transform } from 'babel-core';
-import resolveRc from 'babel-core/lib/babel/tools/resolve-rc';
-import * as babelUtil from 'babel-core/lib/babel/util';
 import glob from 'glob';
 import fs from 'fs';
 import path from 'path';
 import outputFileSync from 'output-file-sync';
 
 export function buildContent(content, filename, destination, babelOptions={}) {
-  const result = transform(content, resolveRc(filename, babelOptions));
+  babelOptions.filename = filename;
+  const result = transform(content, babelOptions);
   outputFileSync(destination, result.code, {encoding: 'utf8'});
 }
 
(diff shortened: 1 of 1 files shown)


sebmck:
Oh this is embarrassing, I didn't realise the error message was so harsh and unprofessional. My extreme apologies.

Cellule:
I'm the one who feels embarrassed. I'm still a young programmer and I need to be told when I do things I shouldn't.

jquense:
I might be missing something but doesn't `babel.transform` already resolve the babelrc it seems unecessary to do it ourselves?

sebmck:
If you pass an absolute `filename` option to `babel.transform` then it does the `.babelrc` lookup for you.

jquense:
great! @Cellule can you just pass the filename to buildContent instead of reading it first? Then we can avoid this whole thing completely :)

Cellule:
Thanks, this is exactly what I was looking for at first, but I couldn't find it in the doc. 
It's much cleaner that way :)

jquense:
tested and it does pickup the .babelrc and LGTM :+1:

mtscout6:
@cellule it's not entirely your fault for that either. We had a discussion at length on whether we should allow that dependency to occur. @sebmck I apologize for not recognizing the need to open that discussion with you as well. We as a team felt at the time that it didn't seem to be too bad of a dependency to introduce. I have pushed both my colleagues and other OSS contributors to do so on many occasions. The hard part here was this was preventing us from building on Windows which was excluding a subset of our contributors. Seeing as all OSS projects are different and approached with varying levels of acceptance for issues we did not weigh this decision to be as detrimental as it seems to have been. I'm glad we are working together to accomplish some success. It's not something you see between different projects at times. @sebmck Again thank you for the great tool you provide, you are moving the community in great directions.

taion:
Much cleaner. LGTM as well.

mtscout6:
LGTM

PR #829, opened 2015-06-13
Fix docs command for windows users

jquense:
:( Windows doesn't do shebangs, so it has no idea how to run that without
the command.

### DIFF
--- a/docs/dev-run
+++ b/docs/dev-run
@@ -1,4 +1,3 @@
-#!/usr/bin/env babel-node
 /* eslint no-process-exit: 0 */
 import 'colors';
 import portfinder from 'portfinder';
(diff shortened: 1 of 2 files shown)


taion:
Let's just get rid of that line on `docs/dev-run` altogether then. LGTM otherwise.

jquense:
the shebang you mean? remove it from the file? it doesn't do any harm on windows, but removing it might break other users calling the file directly?

mtscout6:
Works fine for me on Mac

LGTM

taion:
I say remove the shebang from the file, because none of the rest of our scripts (e.g. `build-cli.js`) have the shebang.

mtscout6:
That's fair, we run all our scripts with either babel-node or node anyways to play well with windows. I'll try and remember not to do that again :blush:

jquense:
removed! good to merge when everything passes?

AlexKVal:
LGTM

PR #830, opened 2015-06-14
Enable loose option for Babel

jquense:
enable loose, mode for Babel. This transpiles with a slightly lower
level of spec "correctness" but in the vast majority of cases this
doesn't matter, and leads to better performance. Admittedly I am not
backing that up with numbers, but there is almost no downside to
enabling it. Specifically though it doesn't use DefineProperty in
classes which is notoriously slow

http://babeljs.io/docs/advanced/loose/

Everyone might want to take a look through the caveats in the above list. If folks think that we are unlikely to run into said caveats then its probably worth it to enable it.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "optional": [
     "es7.objectRestSpread"
-  ]
+  ],
+  "loose": ["all"]
 }


AlexKVal:
Some data for analysis:

`npm run docs` and `npm run docs-prod` 
no warnings in console and all seems still works fine.

Sizes: before / after
`react-bootstrap.js`
243,381 bytes
232,895 bytes

`react-bootstrap.min.js`
118,709 bytes 
113,326 bytes

For all files:
```js
Object.defineProperty(exports, '__esModule', {
  value: true
});
```
into
```js
exports.__esModule = true;
```

For ES6 `class` components: added `var _createClass = (function () { `
https://gist.github.com/AlexKVal/121b0a79062ff76eaed6#file-button-diff-L13

For Input.js one more change `var _get = function get(_x, _x2, _x3) {`
https://gist.github.com/AlexKVal/832d18cf8c4c7ca6c349#file-input-diff-L15

etc.

Here is the full diff between two realizations for the `lib` folder: 
https://gist.github.com/AlexKVal/1091ac6919f8138c71a6

taion:
Loose mode had issues with decorators a while back, right? Granted, we're not using them anyway, but were they fixed?

jquense:
thanks for the data @AlexKVal :) 

as for decorators my current understanding is that now their use essentially just disables loose mode for classes, so it goes back to using its class helper to define properties. that may just be for method decorators which seem to be less common than class ones. property initializers (stage 0) also seem to so this.

taion:
I see - that makes sense, thanks.

mtscout6:
I think this sounds reasonable, the question is whether this constitutes a breaking change or not. While our docs and tests may still be working I'm slightly concerned that there may be some trivial thing that we don't see that breaks someone downstream. In that light I'd prefer that this change was brought into the v0.24-rc branch. Also, since this does effect the src output the commit message should follow our changelog format so downstream users are aware of this non-visible change.

jquense:
sounds good to me I'll make those adjustments

PR #835, opened 2015-06-14
[fixed] rootClose behavior on replaced elements

### LINKED ISSUE #802: OverlayTrigger trigger='click' is being closed when rootClose==true

taion:
Fixes #802

### DIFF
--- a/src/RootCloseWrapper.js
+++ b/src/RootCloseWrapper.js
@@ -4,6 +4,15 @@ import EventListener from './utils/EventListener';
 
 // TODO: Merge this logic with dropdown logic once #526 is done.
 
+// TODO: Consider using an ES6 symbol here, once we use babel-runtime.
+const CLICK_WAS_INSIDE = '__click_was_inside';
+
+function suppressRootClose(event) {
+  // Tag the native event to prevent the root close logic on document click.
+  // This seems safer than using event.nativeEvent.stopImmediatePropagation(),
+  // which is only supported in IE >= 9.
+  event.nativeEvent[CLICK_WAS_INSIDE] = true;
+}
 
 export default class RootCloseWrapper extends React.Component {
   constructor(props) {
@@ -23,10 +32,8 @@ export default class RootCloseWrapper extends React.Component {
   }
 
   handleDocumentClick(e) {
-    // If the click originated from within this component, don't do anything.
-    // e.srcElement is required for IE8 as e.target is undefined
-    let target = e.target || e.srcElement;
-    if (domUtils.contains(React.findDOMNode(this), target)) {
+    // This is now the native event.
+    if (e[CLICK_WAS_INSIDE]) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


src/RootCloseWrapper.js
```diff
@@ -54,7 +60,22 @@ export default class RootCloseWrapper extends React.Component {
...
+    // Wrap the child in a new element, so the child won't have to handle
+    // potentially combining multiple onClick listeners.
+    const child = React.cloneElement(
+      React.Children.only(this.props.children),
+      {ref: component => this._wrappedElement = component}
+    );
+
+    return (
+      <div onClick={suppressRootClose}>
+        {child}
```
jquense: Does adding an extra element potentially mess with layout? What would be the downside with just chaining the onClick? Or if that fails maybe just adding a native handler with addEventListener?

The other concern is that is that will replace any refs the child already has, which is not unlikely if someone needs to call any methods, or reference there overlay component. One alternative is to use the chained function helper to there, that will work if there is an existing ref of the new style, but not if the ref is a string.

src/RootCloseWrapper.js
```diff
@@ -4,6 +4,14 @@ import EventListener from './utils/EventListener';
 
 // TODO: Merge this logic with dropdown logic once #526 is done.
 
+const clickWasInside = Symbol('clickWasInside');
+
```
jquense: This is a really clever approach and seems like a better option than stopImmediatePropagation, since the clicks will still be passed through. One thing though is that using a symbol, while probably better, introduces a lot of extra code to the project to polyfill them. We should look at how much code is introduced by babel/corejs for symbols just so we are all aware of the additional bytes.

taion: Oh crap. I thought we were already using `babel-runtime`. We're not. This can't be merged as-is, then.

src/RootCloseWrapper.js
taion: I didn't want to just pass through `onClick` because I was concerned about the case of custom overlay components, or ones that already have `onClick` specified.

I didn't want to add a native handler because I was concerned about the overlay DOM node getting replaced by various mechanisms.

The extra element does not mess with the layout (there already are wrapping elements for ovelrays in new layers anyway), though I needed the `getWrappedElement` workaround below to get size calculations to work correctly.

I agree about your point with refs - I misread this part:
>Note: React.cloneElement(child, { ref: 'newRef' }) DOES override the ref so it is still not possible for two parents to have a ref to the same child, unless you use callback-refs.
to mean that `cloneElement` would do something smart with callback-refs, but I can see that it doesn't, so I need to re-write this bit.

taion:
Addressed issues above -

1. Use a pseudo-symbol key instead of an actual `Symbol`
2. Don't steal `ref` on overlay

taion:
Tests are passing now - can someone take a look?

jquense:
I think the added wrapper div constitutes a breaking change, It doesn't matter in the context of its use in OverlayTrigger, but if someone was using it outside of that, with say an inline span, the div is going to break that.

mtscout6:
Everything looks good to me, but I agree that the extra div can be a breaking change. Can you re-open against the v0.24-rc branch?

PR #841, opened 2015-06-15
Use babel-runtime transformer for helpers

### LINKED ISSUE #579: Use Babel runtime
A lot of components have duplicated code in their generated code.

Ex:
badge

We could fix that, by adding [babel-runtime][babel] in the build. This way, it requires babel-runtime modules instead of duplicating the output. This is one more dep for us but could help reduce the lib output size.

[babel]:https://babeljs.io/docs/usage/runtime/

### LINKED ISSUE #840: Do not use Babel cache for release build
Changes to e.g. `.babelrc` do not invalidate `babel-loader`'s cache. I don't think it's too onerous to require developers messing with Babel settings to clear the cache themselves, but we should make sure we either `rimraf` the cache in the publish script or bypass the cache when building for release, just to guarantee correctness.

taion:
Fixes #579

Reduces size of `dist/react-bootstrap.min.js` from 121,221B to 116,282B at this commit.

Will need to resolve #840 or manually clear `.babel-cache` before building with this change.

As this is against v0.24 RC, no need to update e.g. Bower dependencies, but this is now a runtime dependency when using the standard build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "stage": 1
+  "stage": 1,
+  "optional": ["runtime"]
 }
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "yargs": "^3.5.4"
   },
   "dependencies": {
+    "babel-runtime": "^5.1.10",
     "classnames": "^2.0.0"
   }
 }


jquense:
did we address the amd concern? I am not sure what the overall plan for that build is to begin with.

mtscout6:
This will work with amd since we are not declaring it as an external dependency: https://github.com/react-bootstrap/react-bootstrap/blob/master/webpack/webpack.config.js#L16-L24 Which means it will get inlined to the concatenated build.

taion:
There is no more AMD build for 0.24. It's just the bundles in `amd/`.

mtscout6:
LGTM

@taion Are you following this up with a PR to clear the cache in the release scripts?

AlexKVal:
LGTM

PR #843, opened 2015-06-15
[fixed] Do not use Babel cache for release build

### LINKED ISSUE #840: Do not use Babel cache for release build

taion:
~~Also use default Babel cache directory~~

Fixes #840

I don't know if this is the best/clearest way to do this. What this does is

1. Use cache by default in `build-cli.js`
2. Use cache by default when running `webpack`
3. Do not use cache by default in exported function from `tools/build.js`, and implicitly call that without setting `useCache` from `release.js`

### DIFF
--- a/docs/build.js
+++ b/docs/build.js
@@ -32,8 +32,11 @@ function generateHTML(fileName) {
   });
 }
 
-export default function BuildDocs({ dev }) {
+export default function BuildDocs({useCache, dev}) {
   console.log('Building: '.cyan + 'docs'.green + (dev ? ' [DEV]'.grey : ''));
+
+  const useCacheOption = `--use-cache=${Boolean(useCache)}`;
+  const devOption = dev ? '' : '-p';
 
   return exec(`rimraf ${docsBuilt}`)
     .then(() => fsp.mkdir(docsBuilt))
@@ -41,7 +44,7 @@ export default function BuildDocs({ dev }) {
       let pagesGenerators = Root.getPages().map(generateHTML);
 
       return Promise.all(pagesGenerators.concat([
-        exec(`webpack --config webpack.docs.js ${dev ? '' : '-p '}--bail`),
+        exec(`webpack --config webpack.docs.js ${useCacheOption} --bail ${devOption}`),
         copy(license, docsBuilt),
         copy(readmeSrc, readmeDest)
       ]));
--- a/tools/build-cli.js
+++ b/tools/build-cli.js
@@ -8,9 +8,16 @@ import { setExecOptions } from './exec';
 import yargs from 'yargs';
 
 const argv = yargs
+  .help('h')
   .option('docs-only', {
     demand: false,
     default: false
+  })
+  .option('use-cache', {
+    type: 'boolean',
+    demand: false,
+    default: true,
+    describe: 'Use Babel cache when running webpack'
   })
   .option('verbose', {
     demand: false,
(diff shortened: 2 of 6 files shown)


jquense:
LGTM :)

taion:
Rebased and fixed imports. Is this setup of the `use-cache` option confusing? I'm totally willing to change around the defaults if necessary.

.gitignore
```diff
@@ -11,4 +11,3 @@ dist/
 docs-built/
 tmp-bower-repo/
 tmp-docs-repo/
-.babel-cache/
```
mtscout6: I'm wondering if we should leave this in for a while until this change makes it into the master branch and has been there for a week or two. Mainly because I can see people accidentally adding this in a commit and submitting it in a PR. It won't hurt anything to be there for a time, and we can create an issue that we target with the v0.25 release to remember to remove it.

AlexKVal:
Now if I need to clean up the cache I do `rm -rf .babel-cache`
How could I clean it up after this PR ?

I've checked it:
when I change options in `.babelrc` the babel-cache with this option `?cacheDirectory`
doesn't update and needs to be cleaned up somehow to reflect the new changes in `.babelrc`.

mtscout6:
That is a good question, how do we invalidate the cache if we use the system OS directory? @sebmck Is there a way to invalidate the cache when the .babelrc file changes or should we open an issue against babel for this?

sebmck:
@mtscout6 There's not currently a way to and I don't think there's an automatic way to do this that wont hurt performance.

mtscout6:
I don't know the mechanics entirely for your cache, but the filenames I see in there I imagine are some sort of hash of the file contents. Perhaps a hash of the babelrc concatenated with the hash of the file?

sebmck:
@mtscout6 That still requires doing a lot of file i/o for every transform. `.babelrc` files don't change enough for it to be worth it.

mtscout6:
This is only in babel-loader right? Or is it in the core of babel?

sebmck:
`.babelrc` resolution is done in Babel core.

mtscout6:
It looks like you are already building the hash with a JSON representation of the source, options, and identifier. Would those options be from the .babelrc? 

https://github.com/babel/babel-loader/blob/master/lib/fs-cache.js#L75-L79

mtscout6:
I wonder why we would need cache invalidation if that's the case.

sebmck:
I don't write or maintain `babel-loader`.

mtscout6:
Rgr, thanks for taking a look @sebmck it's really appreciated

mtscout6:
@couto Do you have any feedback on this?

taion:
Hmm. Okay. Sounds like I'll switch back to `.babel-cache` for the Babel cache, then, to allow manual invalidation.

mtscout6:
@taion That sounds fair.

taion:
Back to using `.babel-cache`.

mtscout6:
LGTM

Couto:
I'm not entirely sure I understood the problem, but you can pass any string you want into `cacheIdentifier` so if you want to bust babel-loader's cache when the `.babelrc` changes... You can just `JSON.stringify` `.babelrc`'s contents into the loader's option.

taion:
The issue is that it's unintuitive that `babel-loader` doesn't do this by default - I think in general you'd expect a change to `.babelrc` to force a cache invalidation, because it may change the output of the transpilation.

PR #848, opened 2015-06-16
Transitions Components

jquense:
__EDIT__: Should be good to go now!

An experiment in mixin alternatives for transitions. I was already working on a generic version for a project at work and thought it might be valuable upstream here.

`Transition` component is sort of like a streamlined CSSTransitionGroup from react.addons, but expects a single child and doesn't depend on components mounting/unmounting.

If it looks like the sort of thing we might like in core here I'll clean it up and add tests. Feedback appreciated, I added examples to the Component page for easy viewing :P.

Benefits in my eyes:
- Fade doesn't need to clone its element anymore, and we have proper callbacks
- Collapse feels a bit more manageable in component form, rather than having to mixin the behavior
- Transition is nice and extensible for its specific purpose.
- no mixins means non createClass consumers can use em
- looks like there is no need for breaking changes for components that use the new components, and mixins could be depreciated reasonably

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -1,5 +1,3 @@
-/*eslint-disable react/prop-types */
-'use strict';
 import React from 'react';
 import Transition from './Transition';
 import domUtils from './utils/domUtils';
@@ -122,7 +120,7 @@ Collapse.propTypes = {
   in:       React.PropTypes.bool,
 
   /**
-   * Provide the durration of the animation in milliseconds, used to ensure that finishing callbacks are fired even if the
+   * Provide the duration of the animation in milliseconds, used to ensure that finishing callbacks are fired even if the
    * original browser transition end events are canceled.
    */
   duration:          React.PropTypes.number,
@@ -140,7 +138,7 @@ Collapse.propTypes = {
 
   /**
    * A function that returns the height or width of the animating DOM node. Allows for providing some custom logic how much
-   * Collapse component should animation in its specified dimension.
+   * Collapse component should animate in its specified dimension.
    *
    * `getDimensionValue` is called with the current dimension prop value and the DOM node.
    */
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -1,0 +1,200 @@
+/*eslint-disable react/prop-types */
+'use strict';
+import React from 'react';
+import Transition from './Transition';
+import domUtils from './utils/domUtils';
+import createChainedFunction from './utils/createChainedFunction';
+
+let capitalize = str => str[0].toUpperCase() + str.substr(1);
+
+// reading a dimension prop will cause the browser to recalculate,
+// which will let our animations work
+let triggerBrowserReflow = node => node.offsetHeight; //eslint-disable-line no-unused-expressions
+
+const MARGINS = {
+  height: ['marginTop', 'marginBottom'],
+  width:  ['marginLeft', 'marginRight']
+};
+
+function getDimensionValue(dimension, elem){
+  let value = elem[`offset${capitalize(dimension)}`];
+  let computedStyles = domUtils.getComputedStyles(elem);
+  let margins = MARGINS[dimension];
+
+  return (value +
+    parseInt(computedStyles[margins[0]], 10) +
+    parseInt(computedStyles[margins[1]], 10)
+  );
+}
+
+class Collapse extends React.Component {
+
+  constructor(props, context){
+    super(props, context);
+
+    this.onEnterListener = this.handleEnter.bind(this);
+    this.onEnteringListener = this.handleEntering.bind(this);
+    this.onEnteredListener = this.handleEntered.bind(this);
+    this.onExitListener = this.handleExit.bind(this);
+    this.onExitingListener = this.handleExiting.bind(this);
+  }
+
+  render() {
+    let enter = createChainedFunction(this.onEnterListener, this.props.onEnter);
+    let entering = createChainedFunction(this.onEnteringListener, this.props.onEntering);
+    let entered = createChainedFunction(this.onEnteredListener, this.props.onEntered);
+    let exit = createChainedFunction(this.onExitListener, this.props.onExit);
+    let exiting = createChainedFunction(this.onExitingListener, this.props.onExiting);
+
+    return (
+      <Transition
+        ref='transition'
+        {...this.props}
+        aria-expanded={this.props.in}
+        className={this._dimension() === 'width' ? 'width' : ''}
+        exitedClassName='collapse'
+        exitingClassName='collapsing'
+        enteredClassName='collapse in'
+        enteringClassName='collapsing'
+        onEnter={enter}
+        onEntering={entering}
+        onEntered={entered}
+        onExit={exit}
+        onExiting={exiting}
+        onExited={this.props.onExited}
+      >
+        { this.props.children }
+      </Transition>
+    );
+  }
+
+  /* -- Expanding -- */
+  handleEnter(elem){
+    let dimension = this._dimension();
+    elem.style[dimension] = '0';
+  }
+
+  handleEntering(elem){
+    let dimension = this._dimension();
+
+    elem.style[dimension] = this._getScrollDimensionValue(elem, dimension);
+  }
+
+  handleEntered(elem){
+    let dimension = this._dimension();
+    elem.style[dimension] = null;
+  }
+
+  /* -- Collapsing -- */
+  handleExit(elem){
+    let dimension = this._dimension();
+
+    elem.style[dimension] = this.props.getDimensionValue(dimension, elem) + 'px';
+  }
+
+  handleExiting(elem){
+    let dimension = this._dimension();
+
+    triggerBrowserReflow(elem);
+    elem.style[dimension] = '0';
+  }
+
+  _dimension(){
+    return typeof this.props.dimension === 'function'
+      ? this.props.dimension()
+      : this.props.dimension;
+  }
+
+  //for testing
+  _getTransitionInstance(){
+    return this.refs.transition;
+  }
+
+  _getScrollDimensionValue(elem, dimension){
+    return elem[`scroll${capitalize(dimension)}`] + 'px';
+  }
+}
+
+Collapse.propTypes = {
+  /**
+   * Collapse the Component in or out.
+   */
+  in:       React.PropTypes.bool,
+
+  /**
+   * Provide the durration of the animation in milliseconds, used to ensure that finishing callbacks are fired even if the
+   * original browser transition end events are canceled.
+   */
+  duration:          React.PropTypes.number,
+
+  /**
+   * Specifies the dimension used when collapsing.
+   *
+   * _Note: Bootstrap only partially supports this!
+   * You will need to supply your own css animation for the `.width` css class._
+   */
+  dimension: React.PropTypes.oneOfType([
+    React.PropTypes.oneOf(['height', 'width']),
+    React.PropTypes.func
+  ]),
+
+  /**
+   * A function that returns the height or width of the animating DOM node. Allows for providing some custom logic how much
+   * Collapse component should animation in its specified dimension.
+   *
+   * `getDimensionValue` is called with the current dimension prop value and the DOM node.
+   */
+  getDimensionValue: React.PropTypes.func,
+
+  /**
+   * A Callback fired before the component starts to expand.
+   */
+  onEnter: React.PropTypes.func,
+
+  /**
+   * A Callback fired immediately after the component starts to expand.
+   */
+  onEntering: React.PropTypes.func,
+
+  /**
+   * A Callback fired after the component has expanded.
+   */
+  onEntered: React.PropTypes.func,
+
+  /**
+   * A Callback fired before the component starts to collapse.
+   */
+  onExit: React.PropTypes.func,
+
+  /**
+   * A Callback fired immediately after the component starts to collapse.
+   */
+  onExiting: React.PropTypes.func,
+
+  /**
+   * A Callback fired after the component has collapsed.
+   */
+  onExited: React.PropTypes.func,
+
+  /**
+   * Specify whether the transitioning component should be unmounted (removed from the DOM) once the exit animation finishes.
+   */
+  unmountOnExit:     React.PropTypes.bool,
+
+  /**
+   * Specify whether the component should collapse or expand when it mounts.
+   */
+  transitionAppear: React.PropTypes.bool
+};
+
+Collapse.defaultProps = {
+  in:       false,
+  duration: 300,
+  dimension: 'height',
+  transitionAppear: false,
+  unmountOnExit: false,
+  getDimensionValue
+};
+
+export default Collapse;
+
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -130,7 +130,7 @@ Collapse.propTypes = {
   /**
    * Specifies the dimension used when collapsing.
    *
-   * _Note: Bootstrap only partially supports this!
+   * _Note: Bootstrap only partially supports 'width'!
    * You will need to supply your own css animation for the `.width` css class._
    */
   dimension: React.PropTypes.oneOfType([
--- a/src/CollapsibleNav.js
+++ b/src/CollapsibleNav.js
@@ -1,14 +1,13 @@
 import React, { cloneElement } from 'react';
 import BootstrapMixin from './BootstrapMixin';
-import CollapsibleMixin from './CollapsibleMixin';
+import Collapse from './Collapse';
 import classNames from 'classnames';
-import domUtils from './utils/domUtils';
 
 import ValidComponentChildren from './utils/ValidComponentChildren';
 import createChainedFunction from './utils/createChainedFunction';
 
 const CollapsibleNav = React.createClass({
-  mixins: [BootstrapMixin, CollapsibleMixin],
+  mixins: [BootstrapMixin],
 
   propTypes: {
     onSelect: React.PropTypes.func,
@@ -19,41 +18,48 @@ const CollapsibleNav = React.createClass({
     eventKey: React.PropTypes.any
   },
 
-  getCollapsibleDOMNode() {
-    return React.findDOMNode(this);
-  },
 
-  getCollapsibleDimensionValue() {
-    let height = 0;
-    let nodes = this.refs;
-    for (let key in nodes) {
-      if (nodes.hasOwnProperty(key)) {
+  // getCollapsibleDimensionValue() {
+  //   let height = 0;
+  //   let nodes = this.refs;
+  //   for (let key in nodes) {
+  //     if (nodes.hasOwnProperty(key)) {
 
-        let n = React.findDOMNode(nodes[key]);
-        let h = n.offsetHeight;
-        let computedStyles = domUtils.getComputedStyles(n);
+  //       let n = React.findDOMNode(nodes[key]);
+  //       let h = n.offsetHeight;
+  //       let computedStyles = domUtils.getComputedStyles(n);
 
-        height += (h +
-          parseInt(computedStyles.marginTop, 10) +
-          parseInt(computedStyles.marginBottom, 10)
-        );
-      }
-    }
-    return height;
-  },
+  //       height += (h +
+  //         parseInt(computedStyles.marginTop, 10) +
+  //         parseInt(computedStyles.marginBottom, 10)
+  //       );
+  //     }
+  //   }
+  //   return height;
+  // },
 
   render() {
     /*
      * this.props.collapsible is set in NavBar when an eventKey is supplied.
      */
-    const classes = this.props.collapsible ? this.getCollapsibleClassSet('navbar-collapse') : null;
+    const classes = this.props.collapsible ? 'navbar-collapse' : null;
     const renderChildren = this.props.collapsible ? this.renderCollapsibleNavChildren : this.renderChildren;
 
-    return (
+    let nav = (
       <div eventKey={this.props.eventKey} className={classNames(this.props.className, classes)} >
         {ValidComponentChildren.map(this.props.children, renderChildren)}
       </div>
     );
+
+    if ( this.props.collapsible ){
+      return (
+        <Collapse in={this.props.expanded}>
+          { nav }
+        </Collapse>
+      );
+    } else {
+      return nav;
+    }
   },
 
   getChildActiveProp(child) {
--- a/src/CollapsibleNav.js
+++ b/src/CollapsibleNav.js
@@ -19,25 +19,6 @@ const CollapsibleNav = React.createClass({
   },
 
 
-  // getCollapsibleDimensionValue() {
-  //   let height = 0;
-  //   let nodes = this.refs;
-  //   for (let key in nodes) {
-  //     if (nodes.hasOwnProperty(key)) {
-
-  //       let n = React.findDOMNode(nodes[key]);
-  //       let h = n.offsetHeight;
-  //       let computedStyles = domUtils.getComputedStyles(n);
-
-  //       height += (h +
-  //         parseInt(computedStyles.marginTop, 10) +
-  //         parseInt(computedStyles.marginBottom, 10)
-  //       );
-  //     }
-  //   }
-  //   return height;
-  // },
-
   render() {
     /*
      * this.props.collapsible is set in NavBar when an eventKey is supplied.
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -1,0 +1,90 @@
+'use strict';
+import React from 'react';
+import Transition from './Transition';
+
+class Fade extends React.Component {
+
+  constructor(props, context){
+    super(props, context);
+  }
+
+  render() {
+    return (
+      <Transition
+        {...this.props}
+        in={this.props.in}
+        className='fade'
+        enteredClassName='in'
+        enteringClassName='in'
+      >
+        { this.props.children }
+      </Transition>
+    );
+  }
+}
+
+Fade.propTypes = {
+  /**
+   * Fade the Component in or out.
+   */
+  in:       React.PropTypes.bool,
+
+  /**
+   * Provide the durration of the animation in milliseconds, used to ensure that finishing callbacks are fired even if the
+   * original browser transition end events are canceled.
+   */
+  duration:          React.PropTypes.number,
+
+  /**
+   * A Callback fired before the component starts to fade in.
+   */
+  onEnter: React.PropTypes.func,
+
+  /**
+   * A Callback fired immediately after the component has started to faded in.
+   */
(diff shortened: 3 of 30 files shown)


aabenoja:
I like that this is pushing us towards a more composed approach than relying on mixins to provide shared shared functionality. I've only skimmed it a bit, but I like what you've done with `Collapse`.

src/Transitions/Collapse.js
```diff
@@ -0,0 +1,94 @@
...
+class Collapse extends React.Component {
+
+  constructor(props, context){
+    super(props, context);
+  }
+
+  render() {
+    let entering = createChainedFunction(this.handleEntering.bind(this), this.props.onEntering);
+    let entered = createChainedFunction(this.handleEntered.bind(this), this.props.onEntered);
+    let exit = createChainedFunction(this.handleExit.bind(this), this.props.onExit);
```
aabenoja: Just to pick your brain a little, what led to your decision to do the `bind(this)` calls here instead of using the constructor? This is a very clean approach, I just want your opinion.

jquense: Mostly I just prefer it, it feels cleaner and I don't need to save a bunch of handler references along with the unbound method on the class instance. It is not particularly "correct" to do it this way, it is less performant (almost always negligible tho in this case), and it can break the purerender functionality of the child component if it has it. In cases where it doesn't matter tho I don't optimize for it a head of time :)

mtscout6: If this could break pure render then this should be in the constructor right? Since this is a HoC for other components. I'd rather nip future bugs with pure render in the but now if we can.

mtscout6:
I assume that a lot of this can come from React itself once facebook/react#3150 is available?

src/Transitions/Collapse.js
jquense: yeah definitely, I will clean it up. its mostly just left over style from when this was a work thing

jquense:
> I assume that a lot of this can come from React itself once facebook/react#3150 is available?

I hadn't seen that PR but even with that improvement I don't think it provides enough hooks into the transition life-cycle to implement something like Collapse; specifically the _entering_ and _exiting_ hooks. There may be other concerns as well, I have tended to roll my own focused versions of the TransitionGroup component for a while now, as the provided one tends to be too general an attempt at an animation component to be good a specific things, but that is more of a personal opinion :)

src/Transitions/Collapse.js
```diff
@@ -0,0 +1,94 @@
...
+import React from 'react';
+import Transition from './Transition';
+import domUtils from '../utils/domUtils';
+import createChainedFunction from '../utils/createChainedFunction';
+
+let capitalize = str => str[0].toUpperCase() + str.substr(1);
+
+// reading a dimension prop will cause the browser to recalculate,
+// which will let our animations work
+let triggerBrowserReflow = node => node.offsetHeight; //eslint-disable-line no-unused-expressions
```
AlexKVal: if you use
```js
function triggerBrowserReflow(node) { return node.offsetHeight; }
```
then there is no need in `//eslint-disable-line no-unused-expressions`

src/Transitions/Collapse.js
```diff
@@ -0,0 +1,94 @@
...
+
+  render() {
+    let entering = createChainedFunction(this.handleEntering.bind(this), this.props.onEntering);
+    let entered = createChainedFunction(this.handleEntered.bind(this), this.props.onEntered);
+    let exit = createChainedFunction(this.handleExit.bind(this), this.props.onExit);
+
+    return (
+      <Transition
+        {...this.props}
+        in={this.props.in}
```
AlexKVal: it is passed in with previous line `{...this.props}` anyway.
do you want to make it explicit ?

src/Transitions/Collapse.js
```diff
@@ -0,0 +1,94 @@
...
+    triggerBrowserReflow(elem);
+
+    elem.style[dimension] = '0';
+  }
+}
+
+Collapse.propTypes = {
+  ...Transition.propTypes,
+  dimension: React.PropTypes.oneOf(['height', 'width']), //not entirely supported by tbs
+  getDimensionValue: React.PropTypes.funcs
```
AlexKVal: funcs => func

src/Transitions/Collapse.js
```diff
@@ -0,0 +1,94 @@
...
+};
+
+Collapse.defaultProps = {
+  in:       false,
+  dimension: 'height',
+  getDimensionValue
```
AlexKVal: What for do you make it changeable through property ?
I cannot find anywhere else its usage.

For user to provide their own `getDimensionValue` function realization them should dig into all this code to understand how to implement it. Isn't it ?

src/Transitions/Transition.js
```diff
@@ -0,0 +1,173 @@
...
+import TransitionEvents from '../utils/TransitionEvents';
+import classnames from 'classnames';
+
+function omit(obj, keys) {
+  let included = Object.keys(obj).filter( k => keys.indexOf(k) === -1)
+  let newObj = {};
+
+  included.forEach( key => newObj[key] = obj[key] )
+  return newObj;
+}
```
AlexKVal: this is good candidate to use `lodash` :wink: 
`lodash` performance is faster
screen shot 2015-06-26 at 5 21 05 pm

src/Transitions/Transition.js
```diff
@@ -0,0 +1,173 @@
...
+    if (this.state.in && this.state.transitioning) {
+      classes += this.props.enteringClassName;
+    }
+
+    if (!this.state.in && !this.state.transitioning) {
+      classes += this.props.exitClassName;
+    }
+
+    if (!this.state.in && this.state.transitioning) {
+      classes += this.props.exitingClassName;
```
AlexKVal: with all this `classes += `
isn't there a possible problem with missing space ?

E.g. `classes = 'class1 class2'`
then
`classes += this.props.nextClassName;` where `nextClassName = 'class3'`
in result:
`classes = 'class1 class2class3'` ?

src/Transitions/Transition.js
```diff
@@ -0,0 +1,173 @@
...
+  in:       false,
+  duration: 300,
+  transitioningNode: ()=>{},
+  onEnter:    ()=>{},
+  onEntering: ()=>{},
+  onEntered:  ()=>{},
+
+  onExit:     ()=>{},
+  onExiting:  ()=>{},
+  onExited:   ()=>{}
```
AlexKVal: I thought that here would be possible perf concerns because fat arrow binds all scope,
but as it turns out, `babel` is smart enough here. 

It transpiles this into:
`onExited: function onExited() {},`

AlexKVal:
`Nav` components needs to be expanded by default
```js
// src/Nav.js
getDefaultProps() {
  return {
    bsClass: 'nav',
    expanded: true // <- need to add this
  };
},
```
otherwise all examples on the `components` page with `<Nav ..>` don't show up.
And `aside` navigation on the right also doesn't show up without it.
screen shot 2015-06-26 at 4 14 00 pm

AlexKVal:
As others already mentioned - this new API is awesome :rainbow: :+1: 

(I could try to polish it in the case you are complitely out of time.
through cherry-pick I presume)

src/Transitions/Transition.js
AlexKVal: I took more deep look at this code and I've noticed,
that all these four branches are mutually exclusive,
so there is no need at all in `+=`. 
I've checked it.
`=` works just fine.

src/Panel.js
```diff
@@ -23,6 +25,19 @@ const Panel = React.createClass({
...
+  getInitialState(){
+    let defaultExpanded = this.props.defaultExpanded != null ?
+      this.props.defaultExpanded :
+        this.props.expanded != null ?
+        this.props.expanded :
+        false;
+
+    return {
+      expanded: defaultExpanded,
+      collapsing: false
```
AlexKVal: This is superfluous.

src/Transitions/Transition.js
jquense: Sure but the performance isn't remotely a bottleneck here so it wouldn't have any practical impact on speed here, especially since it's looping through a just few keys, we are talking about differences in nano seconds

src/Panel.js
jquense: collapsing? true!

src/Transitions/Collapse.js
jquense: yes though to me that hides the purpose of the function which is not use the value, but just trigger a recalc

src/Panel.js
AlexKVal: No. Just remove it.

It is used by `CollapsibleMixin` and you are removing this mixin,
then this property becomes obsolete :wink:

src/Transitions/Transition.js
AlexKVal: Agree. :cherries: 
`lodash`-refactoring could be done in the far future if an issue arise.

src/Transitions/Collapse.js
AlexKVal: I do not propose to remove explaining comment
```js
// reading a dimension prop will cause the browser to recalculate,
// which will let our animations work
```
With this explaining comment
```js
// reading a dimension prop will cause the browser to recalculate,
// which will let our animations work
function triggerBrowserReflow(node) { return node.offsetHeight; }
```
it looks very nice :wink:

AlexKVal: We can remove `return`
```js
// reading a dimension prop will cause the browser to recalculate,
// which will let our animations work
function triggerBrowserReflow(node) { node.offsetHeight; }
```
then it looks good :wink:

AlexKVal: Pff.. no :) eslint throws warning in that case :smile: 
then `return` + explaining comment
```js
// reading a dimension prop will cause the browser to recalculate,
// which will let our animations work
function triggerBrowserReflow(node) { return node.offsetHeight; }
```

AlexKVal: Anyway..
It's just only proposal. :wink: 
I'm just trying to find a better solution. That's all.

src/Transitions/Collapse.js
jquense: hmm i don't think so...I will remove it

src/Transitions/Collapse.js
jquense: good catch!

src/Transitions/Collapse.js
jquense: It isn't currently being used in this PR but the CollabsibleNav did use it (I was seeing if it was not necessary). More importantly though it was part of the mixin API so I included it here in case there were ppl who needed it.

src/Transitions/Transition.js
jquense: good point, and I should be using `else if`s to make that more explicit as well

src/Transitions/Transition.js
jquense: yup :)

src/Transitions/Collapse.js
AlexKVal: I believe we should remove it.
It will not be in need anywhere after this patch.

No ppl would need it because otherwise they should undestand old mechanic and new mechanic and dig into the code very deeply to make it right.

And the most importantly: `getDimensionValue` is part of private API, so no ppl should be using it anyway.
So let's not make it public interface. (otherwise we should document it well and add tests for it)

jquense: I don't think it was a private API, it was required to use the collapsibleMixin, so any user using that mixin had to implement that method. the only difference in my component is that I provide a default implementation which is not possible in mixin.

AlexKVal: I was wrong about the 'private API' term. Sorry.
As you said: it was required for the `collapsibleMixin`, hence if this patch will deprecate `collapsibleMixin`, do we still need it as a part of API with `Collapse` ?

jquense: I think it's a good escape hatch for dealing with weird height calculations. it's a bonus we can provide over the basic bootstrap. 

I will of course test and document it

AlexKVal: It has a meaning in that case :+1:

src/CollapsibleNav.js
```diff
@@ -19,41 +18,48 @@ const CollapsibleNav = React.createClass({
...
-  getCollapsibleDOMNode() {
-    return React.findDOMNode(this);
-  },
 
-  getCollapsibleDimensionValue() {
-    let height = 0;
-    let nodes = this.refs;
-    for (let key in nodes) {
-      if (nodes.hasOwnProperty(key)) {
+  // getCollapsibleDimensionValue() {
```
mathieumg: Any reason why this is not straight-out removed?

src/Collapse.js
```diff
@@ -0,0 +1,200 @@
...
+   * You will need to supply your own css animation for the `.width` css class._
+   */
+  dimension: React.PropTypes.oneOfType([
+    React.PropTypes.oneOf(['height', 'width']),
+    React.PropTypes.func
+  ]),
+
+  /**
+   * A function that returns the height or width of the animating DOM node. Allows for providing some custom logic how much
+   * Collapse component should animation in its specified dimension.
```
AlexKVal: maybe `should animate` ?

src/CollapsibleNav.js
```diff
@@ -19,41 +18,48 @@ const CollapsibleNav = React.createClass({
...
-    return height;
-  },
+  //       height += (h +
+  //         parseInt(computedStyles.marginTop, 10) +
+  //         parseInt(computedStyles.marginBottom, 10)
+  //       );
+  //     }
+  //   }
+  //   return height;
+  // },
```
AlexKVal: yeah :smile: 
do not forget to remove those comments :balloon:

src/Position.js
```diff
@@ -61,13 +61,18 @@ class Position extends React.Component {
       return;
     }
 
+    let overlay = React.findDOMNode(this);
     let target = React.findDOMNode(this.props.target(this.props));
     let container = React.findDOMNode(this.props.container) || domUtils.ownerDocument(this).body;
 
+    // if ( !overlay || !target || !container ){
+    //   return;
+    // }
```
AlexKVal: and here

src/Transition.js
```diff
@@ -0,0 +1,286 @@
...
+    let out = !this.state.in && !this.state.transitioning;
+
+    if ( !child || (this.props.unmountOnExit && out && !starting) ){
+      return null;
+    }
+
+    let classes = '';
+
+    // for whatever reason classnames() doesn't actually work here,
+    // maybe because they aren't always single classes?
```
AlexKVal: and this I believe also is extraneous

src/Transition.js
```diff
@@ -0,0 +1,286 @@
...
+
+    else if (!this.state.in && this.state.transitioning) {
+      classes = this.props.exitingClassName;
+    }
+
+    return React.cloneElement(child, {
+      ...childProps,
+      className: classnames(
+          child.props.className
+        , this.props.className
```
AlexKVal: you like that `comma first` yeah :smile:

src/Transition.js
```diff
@@ -0,0 +1,286 @@
+'use strict';
```
AlexKVal: remove :wink:

src/Fade.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
```
AlexKVal: remove :wink:

src/Collapse.js
```diff
@@ -0,0 +1,200 @@
+/*eslint-disable react/prop-types */
+'use strict';
```
AlexKVal: remove :wink:

src/Transition.js
jquense: this one is just so future devs don't accidentally refactoring the code using classnames(), which appears like it would work but causes subtle bugs

src/Collapse.js
```diff
@@ -0,0 +1,200 @@
+/*eslint-disable react/prop-types */
```
AlexKVal: You can remove it. I've checked: eslint OK now.

src/CollapsibleNav.js
jquense: I was hoping someone might know why this was used instead of the normal default method of getting height, it tested it without and it seems to work fine

src/Overlay.js
```diff
@@ -1,13 +1,26 @@
 /*eslint-disable object-shorthand, react/prop-types */
```
AlexKVal: Why don't you just add propTypes for
`container`
`containerPadding`
`target`
`placement`
?

--
It is not a problem. I can take care of it after merging. Just added it to my todo-list.

src/Modal.js
```diff
@@ -466,4 +485,7 @@ Modal.Header = Header;
 Modal.Title = Title;
 Modal.Footer = Footer;
 
+Modal.TRANSITION_DURATION = 300;
+Modal.BACKDROP_TRANSITION_DURATION = 150;
```
AlexKVal: I bet users would ask it to be configurable. :smile: 

--
I'm fine with it anyway. :sparkle:

src/Overlay.js
```diff
@@ -17,30 +30,60 @@ class Overlay extends React.Component {
...
-    );
+    let child = null;
 
-    if (rootClose) {
-      positionedChild = (
-        <RootCloseWrapper onRootClose={this.props.onHide}>
-          { positionedChild }
-        </RootCloseWrapper>
+    if ( Transition === true ){
+      Transition = Fade;
```
AlexKVal: Why `Transition` var is caps-ed ?

src/Panel.js
```diff
@@ -23,6 +25,18 @@ const Panel = React.createClass({
     };
   },
 
+  getInitialState(){
+    let defaultExpanded = this.props.defaultExpanded != null ?
+      this.props.defaultExpanded :
```
AlexKVal: `defaultExpanded` property here is OK.

src/Panel.js
```diff
@@ -23,6 +25,18 @@ const Panel = React.createClass({
...
+        this.props.expanded != null ?
+        this.props.expanded :
```
AlexKVal: but this...
Isn't it an anti-pattern here http://facebook.github.io/react/tips/props-in-getInitialState-as-anti-pattern.html ?

src/Panel.js
```diff
@@ -69,13 +75,16 @@ const Panel = React.createClass({
...
-        ref='panel'
-        aria-expanded={this.isExpanded() ? 'true' : 'false'}>
-        {this.renderBody()}
-      </div>
+      <Collapse in={this.isExpanded()}>
+        <div
+          className={collapseClass}
+          id={this.props.id}
+          ref='panel'
+          aria-expanded={this.isExpanded() ? 'true' : 'false'}>
```
AlexKVal: cool.. didn't know that it is possible to write it without commas.
but it seems it is possible only in JSX.

src/Transition.js
```diff
@@ -0,0 +1,286 @@
...
+   */
+  unmountOnExit:     React.PropTypes.bool,
+
+  /**
+   * Specify whether transitions should run when the Transition component mounts.
+   */
+  transitionAppear: React.PropTypes.bool,
+
+  /**
+   * Provide the durration of the animation in milliseconds, used to ensure that finishing callbacks are fired even if the
```
AlexKVal: duration

src/Transition.js
```diff
@@ -0,0 +1,286 @@
...
+   * original browser transition end events are canceled.
```
AlexKVal: cancelled

src/Transition.js
```diff
@@ -0,0 +1,286 @@
...
+   */
+  onExiting:   React.PropTypes.func,
+  /**
+   * A callback fired after "exit" classes are applied
+   */
+  onExited:    React.PropTypes.func
+};
+
+// name the function so it is clearer in the documentation
+const noop = ()=>{};
```
AlexKVal: I like it :+1:

test/CollapseSpec.js
```diff
@@ -0,0 +1,216 @@
...
+    });
+  });
+
+  it('Should default to collapsed', function () {
+    instance = ReactTestUtils.renderIntoDocument(
+      <Component>Panel content</Component>
+    );
+
+    assert.ok(
+      instance.collapse.props.in === false);
```
AlexKVal: broken one short line ? :smile:

test/FadeSpec.js
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+import ReactTestUtils from 'react/lib/ReactTestUtils';
+import Fade from '../src/Fade';
+//import classNames from 'classnames';
```
AlexKVal: too

test/CollapseSpec.js
AlexKVal: oh.. nevemind. there are many such lines next :smile:

src/Modal.js
mathieumg: From what I understand, this *is* configurable. It's a static property on `Modal`, so you only have to change it directly, i.e.

```js
require("react-bootstrap/lib/Modal").TRANSITION_DURATION = 1234;
```

Anywhere in your code, to modify the default value. Did you mean that it should be configurable per-instance? That doesn't sound very consistent to me.

test/PopoverSpec.js
```diff
@@ -11,16 +11,16 @@ describe('Popover', function () {
...
-    assert.equal(React.findDOMNode(instance).className.match(/\bfade\b/), null, 'The fade class should not be present');
-  });
+  // it('Should not have the fade class if animation is false', function () {
+  //   let instance = ReactTestUtils.renderIntoDocument(
+  //     <Popover title="Popover title" animation={false}>
+  //       <strong>Popover Content</strong>
+  //     </Popover>
+  //   );
+  //   assert.equal(React.findDOMNode(instance).className.match(/\bfade\b/), null, 'The fade class should not be present');
+  // });
```
AlexKVal: too

src/Panel.js
mathieumg: It would be an anti-pattern if your component doesn't need to change the value, which is not the case here, no?

test/TooltipSpec.js
```diff
@@ -10,15 +10,15 @@ describe('Tooltip', function () {
...
   });
+
+  // it('Should not have the fade class if animation is false', function () {
+  //   let instance = ReactTestUtils.renderIntoDocument(
+  //     <Tooltip animation={false}>
+  //       <strong>Tooltip Content</strong>
+  //     </Tooltip>
+  //   );
+  //   assert.equal(React.findDOMNode(instance).className.match(/\bfade\b/), null, 'The fade class should not be present');
+  // });
```
AlexKVal: Also

src/Transition.js
mathieumg: Both spellings are fine! Depends on your locale, I imagine.

test/helpers.js
```diff
@@ -1,5 +1,30 @@
...
   console.warn.calledWithMatch(about).should.be.true;
   console.warn.reset();
 }
+
+/**
+ * Helper for rendering and updating props for plain class Components
+ * since `setProps` is deprecated.
+ * @param  {ReactElement} element     Root element to render
+ * @param  {HTMLElement?} mountPoint  Optional mount node, when empty it uses an unattached div like `renderIntoDocument()`
+ * @return {ComponentInstance}        The instance, with a new method `renderWithProps` which will return a new instance wiht updated props
```
AlexKVal: with

AlexKVal:
Actually I think this is good to go. :cherries: 

--
All those comments / eslints / typos etc. are not significant enough.
I can hangle them all by myself later, it you wish :+1:

src/Transition.js
AlexKVal: then it need to be reworded, because in that form which it is now,
it literally provokes to try and refactor :smile: 

--
and again - this is not essential for merging.

src/Modal.js
AlexKVal: No.. I just thought about public api and documentation.

--
Anyway as I said it already - this is not essential.

It is more because of my curiosity and my wish to learn all the things.
And now - because you clarified it - I know that it is configurable in fact.
I just hadn't given it more thought in the first place :blush:

src/Overlay.js
mathieumg: Because most of the time it will be a component (except in this case where it seems to be a boolean, only to be replaced by the `Fade` component however) and since it's used in JSX below, if it wasn't capitalized React would think it is a regular DOM element and not a React component.

src/Panel.js
AlexKVal: This is still too high for me :smile: 
Just trying to ask the right questions :wink:

src/Transition.js
AlexKVal: Oh. Really sorry about that!
It is indeed.

mathieumg: No problem! English is not my main tongue either, and this kind of stuff can be really confusing at first...

src/Overlay.js
jquense: they are already added with:

```
Overlay.propTypes = {
  ...Portal.propTypes,
  ...Position.propTypes,
```

re-adding them isn't very DRY and would require that we document them again. I think in this case the linting rule is just doing a bad job, I don't want to write worse code to satisfy it.

src/Modal.js
jquense: I did it this way because that's how [bootstrap does it](https://github.com/twbs/bootstrap/blob/master/js/modal.js#L38). They are public, but only begrudgingly, and don't really control the duration of the animation (the css class does). it is just an unfortunate hack to get around the fact that transition end events in browsers just don't fire sometimes.

jquense:
I think I just addressed most of them, I do think that #942 should probably be merged before this tho, since this one sort of depends on it.

src/Overlay.js
AlexKVal: Ah.. Then I would re-write it similarly to: [src/ProgressBar.js#L1](https://github.com/react-bootstrap/react-bootstrap/blob/0683df74ce3510ef7d525b7e4dd01274178400c5/src/ProgressBar.js#L1)
as
```js
/* eslint react/prop-types: [2, {ignore: ["container", "containerPadding", "target", "placement", "children"] }] */
/* These properties are validated in 'Portal' and 'Position' components */
```

--
again: I can do it myself later, if you wish

taion:
I haven't been following this PR specifically, but do we want this for the v0.24 milestone, or should we defer to v0.25?

jquense:
@talion since its done, can't we just put it in 0.24?

jquense:
We should be good to go here as well, I just rebased against that last PR and its all green

taion:
This is the last open PR against the `v0.24-rc` branch. Looks like @AlexKVal already signed off on it. Can you merge?

jquense:
:rainbow:

PR #853, opened 2015-06-17
Code coverage

mtscout6:
I was wanting some analysis to know if the dropdown re-write is sufficiently tested. So, I set up karma-coverage to report on src code coverage. This is not directly tied to that re-write so I'd like to bring it in sooner and decoupled from that re-write work. The question is if we as a team would also like to have [Coveralls](https://coveralls.io) report code coverage stats in PRs? If so it would only be a minor change which I can also include in this PR.


mathieumg:
> The question is if we as a team would also like to have Coveralls report code coverage stats in PRs?

I have nothing against it conceptually, but I don't like the noise it brings into the PRs. Would it be redundant if the coverage is already checked against in the CI or does it bring something more?

mtscout6:
screen shot 2015-06-17 at 9 17 42 am

These are the setting we can configure per repo. I'd like to only use the status api which is how travis and appveyor are currently reporting. I'm pretty sure that if we leave the thresolds empty they won't fail the build, just report status meaning to see the increase/decrease values we'd need to go to the coveralls site to see them.

mathieumg:
> I'd like to only use the status api

:+1: If there's no comment noise and an easy way to get reports, I'm all for it!

mtscout6:
I'll get it wired up on this PR and see how it looks.

dozoisch:
I'm all for it!

jquense:
i will admit I am not really a code coverage sort of person, but i'm willing to become one :)

mtscout6:
@jquense No worries, I'm not initially proposing that this fail anything. It's primarily useful for analysis of tests to know if they are effective. It's also good at pointing you at holes that you may have missed. Too many devs see it as an end all and worry too much about code coverage which is not the only indicator that code is tested. It is a useful one though.

mtscout6:
There we go coveralls is working and shows up next to travis and appveyor. I like it!

mathieumg:
:+1: :tada:

webpack/test.config.js
```diff
@@ -1,10 +1,17 @@
...
 
 export default _.extend({}, baseConfig, {
   output: {
     pathinfo: true
   },
 
+  module: {
+    loaders: [
+      { test: /\.js/, loader: `${jsLoader}!isparta`, exclude: /node_modules|test[\/\\]/ },
+      { test: /\.js/, loader: `${jsLoader}`, include: /test[\/\\]/ }
```
AlexKVal: \`${jsLoader}\` => is the same as just `jsLoader` :wink:

webpack/test-coverage.config.js
```diff
@@ -0,0 +1,12 @@
+import _ from 'lodash';
+import { jsLoader } from './base.config';
+import testConfig from './test.config';
+
+export default _.extend({}, testConfig, {
+  module: {
+    loaders: [
+      { test: /\.js/, loader: `${jsLoader}!isparta`, exclude: /node_modules|test[\/\\]/ },
+      { test: /\.js/, loader: `${jsLoader}`, include: /test[\/\\]/ }
+    ]
```
AlexKVal: duplicate ?

it is already defined in test.config.js.

I presume this file can be safely removed at all ?

AlexKVal:
I like the idea :100: 

> So in order to see the increase/decrease values we'd need to go to the coveralls site to see them.

There is no problem for me at all to click to see the difference with every PR review :sparkle:

mtscout6:
screen shot 2015-06-17 at 2 56 49 pm

You can see if if you expand the build details :smile:

webpack/test-coverage.config.js
mtscout6: Note the `isparta` loader to the non-test files. Also this is using the testConfig as it's base.

webpack/test.config.js
mtscout6: good call.

AlexKVal:
Oh.. it is simpler than I thought. 
It is just under 'Hide all checks' :smile:  Thank you !

AlexKVal:
I compare those two files:
`test-coverage.config.js` 
```js 
module: {
  loaders: [
    { test: /\.js/, loader: `${jsLoader}!isparta`, exclude: /node_modules|test[\/\\]/ },
    { test: /\.js/, loader: jsLoader, include: /test[\/\\]/ }
  ]
}
```
and `test.config.js`
```js
module: {
  loaders: [
    { test: /\.js/, loader: `${jsLoader}!isparta`, exclude: /node_modules|test[\/\\]/ },
    { test: /\.js/, loader: `${jsLoader}`, include: /test[\/\\]/ }
  ]
}
```

I composed them into line by line: (just changed \`${jsLoader}\` to jsLoader)
```js
module: {
  loaders: [
    { test: /\.js/, loader: `${jsLoader}!isparta`, exclude: /node_modules|test[\/\\]/ },
    { test: /\.js/, loader: `${jsLoader}!isparta`, exclude: /node_modules|test[\/\\]/ },
    { test: /\.js/, loader: jsLoader, include: /test[\/\\]/ }
    { test: /\.js/, loader: jsLoader, include: /test[\/\\]/ }
  ]
}
```
They are absolutely similar. 

If I remove those from `webpack/test-coverage.config.js` then 

Am I missing something ?

AlexKVal:
I was trying to make PR for this PR :smile: 
the result: https://github.com/react-bootstrap/react-bootstrap/pull/858

mtscout6:
Sorry about that @AlexKVal I didn't realize that I hadn't removed the module configuration from the test.config, which was my reasoning for the separate file.

mtscout6:
The reason I wanted it separate is because it does take a little longer to get through webpack and execute the tests when you are gathering code coverage. I didn't want to impose that cost as part of the TDD lifecycle to keep the feedback loop fast.

AlexKVal:
It makes sense then. :+1:

PR #854, opened 2015-06-17
[added] Created class prop for custom styling on ul in Dropdown Button

corysnave:
I have a particular use case for this functionality. The dropdown is used for a set of checkboxes, and in certain scenarios, the dropdown would go off the screen and some content was inaccessible. This addition will allow users to apply styling rules to the <ul> tag, such as an overflow.

### DIFF
--- a/docs/assets/docs.css
+++ b/docs/assets/docs.css
@@ -1092,6 +1092,11 @@ h1[id] {
   margin-bottom: 5px;
   clear: left;
 }
+.custom-dropdown {
+  max-height: 150px;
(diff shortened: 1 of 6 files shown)


mtscout6:
I have the wonderful privilege of working for the same company as @corysnave. I spoke with him in person about this, and he was unaware of my work on #526. The dropdown re-write will allow him to get this done without the added prop. This is not a huge internal priority so he was ok with closing this PR. @corysnave Correct me if I'm wrong about any of this.

Issue #850, opened 2015-06-17
Allow hovering over popover when trigger is hover


datpham23:
Hi,

Is there any way to allow popovers with the trigger 'hover' to stay open as long as you're hovering the trigger or the popover?

taion:
Not at the moment. The `hover` trigger is really meant more for tooltips.

jquense:
this seems out of scope for the OverlayTrigger...but should be possible under the new API (#811) in some form or another.

AlexKVal:
Superceded by #855

mmahalwy:
any update on this?

jquense:
the update is that you can make whatever behavior you'd like using the Overlay component, we aren't directly supporting this use case

lou:
I had the same issue so I did this little component to handle it.
It work just fine for my use case.
https://gist.github.com/lou/571b7c0e7797860d6c555a9fdc0496f9

guruwanabe:
> the update is that you can make whatever behavior you'd like using the Overlay component, we aren't directly supporting this use case

Seems rather odd to me that a simple use case like this is not supported. Yes we can achieve it with custom overlays, but it should be easier to just integrate as a prop to OverlayTrigger something like 'persistShow' instead of building a custom component.

Issue #816, opened 2015-06-11
OverlayTrigger, Popover and positioning


marcojahn:
Hi,
using an OverlayTrigger with a Popover within a Navbar works fine except repositioning the element.

The code I am using
```
<OverlayTrigger trigger='hover' placement='top' overlay={<Popover title='Messages'>more text</Popover>}>
    <NavItem eventKey={1} href='#'>Message box <Label bsStyle="primary">42</Label></NavItem>
</OverlayTrigger>
```

But it seems that providing a positionTop={50} or positionTop={-50} is not handed over to the Popover.
I've tried setting it on both OverlayTrigger and Popover. But none works.

The value of this.props.positionTop in Popover.js render() is always "null"

Any tips on how this must be used to work ?

Kind regards

taion:
Ref #811, which will change how most of this works anyway.

FWIW you're not really supposed to be setting those props manually - `OverlayTrigger` will override them anyway.

AlexKVal:
Superceded by #855

Issue #811, opened 2015-06-10
Proposal: Alternative Overlay API


jquense:
The current overlay API has always felt a bit wonky to me, I think in part because the React ecosystem seems to have now "standardized" around specific patterns with overlays, and the RTBS overlays are showing their age a bit. Specifically, they make the declarative opening and closing of overlays more difficult than it should be, because the logic is specified indirectly via a `OverlayTrigger` child. It would be nice if the below could be possible without much rigmarole.

```js
render(){
 return (<div>
  <Button onClick={e => this.setState({ show: true })}>
   Show Modal
  </Button>

  <Modal show={this.state.show} onHide={e => this.setState({ show: false })}>
     ...
  </Modal>
 </div>)
}
```
This doesn't require much more code than the current implementation, but is a good bit more declarative and flexible, and inline with the React ecosystem. It also doesn't require needing to create a whole new Modal class every time you want to use a Modal.

In the case of tooltips and popovers the problem is a bit harder, because the trigger is _needed_ to position the overlay. I don't have a great solution, and maybe the current API is best for this sort of thing, but this is a possible alternative.

```js
render(){
 return (
  <div>
    <Button onClick={e => this.setState({ show: true, target: e.target })}>
      Show Modal
    </Button>

    <Popover 
      show={this.state.show} 
      onHide={e => this.setState({ show: false })
      target={()=> this.state.target }}>
       ...
     </Popover >
 </div>)
}
```

These api's don't even need to be mutually exclusive, `OverlayTrigger` could encapsulate the above logic fairly easily, while still allowing fine grained control over the overlays if a consumer wants it.

I'll admit that a fair bit of this is a matter of opinion, but to me this strike me as more "React like" in its explicitness. Would love feedback, and any additional thoughts

dozoisch:
:+1:

mtscout6:
Huge :+1: I've been wanting to propose such a change as well. There's a lot less mental overhead to understand this.

taion:
This is cool. Two notes -

1. This pattern would let us avoid the context-forwarding bits of code in overlay triggers, which will be great for a lot of Flux applications (and react-router) that use context.

[...]

PR #855, opened 2015-06-18
Overlay Overhaul

jquense:
Moving the conversation from #811 into a PR b/c I think it may help provide some visibility into the code changes I am thinking about here. be warned this is a really long note

We have two major API's that are Changing: Modals, Overlays (tooltips/popovers)

### Consolidations
  - __deprecations__
    + OverlayMixin replaced by `Portal` component
  - __additions__
    + `Portal` component, _just_ renders its children into a `container` _fixes #404_

Modals and Overlays share some common functions underneath, and Users have come to use our OverlayMixin, the new `Portal` Component offers the same functionality in a way that works for both createClass or plain class consumers

```js
  <Portal container={document.body}>
    <div>rendered into the body!</div>
  </Portal>
```

### Modals
- __deprecations__
  - `ModalTrigger` - _fixes #397, fixes #616_
  - `title` and `closeButton` props on Modal (moved to ModalHeader)
  - `onRequestHide` prop renamed to `onHide` (more idiomatic with our callbacks)
- __additions__ - partly addresses #748
  - `ModalHeader`
  - `ModalTitle`
  - `ModalBody`
  - `ModalFooter`

#### Api

new components are exported on the Modal component Automatically for ease of use, or can be imported from lib if the user wants. Body, Footer, and Title are just convenient wrappers of `<div className='modal-part'/>` and Modal Header will auto wire the closeButton click to the parent `onHide`.

```js
<Modal container={mountNode} onHide={function(){}} aria-labelledby='title'>
    <Modal.Header closeButton>
      <Modal.Title id='title'>Modal title</Modal.Title>
    </Modal.Header>

    <Modal.Body>
        One fine body...
    </Modal.Body>
    <Modal.Footer>
        <Button bsStyle='primary'>Save changes</Button>
     </Modal.Footer>
</Modal>
```

I think its safe to remove (deprecated for now) ModalTrigger, it is a slight convenience over handling the state yourself ,but since you don't generally have 10s of modals in a component at once (as with tooltips potentially) I say we prefer the React philosophy of explicitness over abstraction here.

By explicitly specifying the Header, users get more customization options and better potential accessibility (`aria-labelledby` above). We can gracefully deprecate the current api, by checking if a user provides a `title` prop in the Modal and auto adding the new Header Component to the children, so no breaking changes.

### Overlays

The tooltips and popovers (from here on out "Overlays") present some interesting issues Modals do not. 

- You may have many Overlays on a page, and managing that manually in state is very verbose
- Overlays are targeted to a specific element, as well as potentially being rendered into a new subtree

#### Goals for the API

- maintain a similar level of ease-of-use that the OverlayTrigger offers, for wiring overlays to triggers _and_ rendering custom overlays (non tooltips or Popovers)
- Allow better control over the provided Overlays so they can be tweaked for custom behavior, such as: #850, #816, #802, #611, #436

My Proposed path here is to ultmately end up with a declarative Overlay Component that can be completely controlled in terms of positioning, and display (show/hide). I think that looks something like the Modals above

```js
<Button
  bsStyle='default'
  ref='btn'
  onClick={e => this.setState({ show: !this.state.show })}
>
  Holy guacamole!
</Button>
<Popover
  show={this.state.show}
  target={()=> this.refs.btn}
  placement='bottom'
  container={mountNode}
  containerPadding={20}
  title='Popover bottom'
>
  <strong>Holy guacamole!</strong> Check this info.
</Popover>
```

To acheive this though, both the render-to-container (show/hide), and positioning logic need to be removed from the `OverlayTrigger` Component and moved downstream. me and @taion have had a few long conversations on gitter about how far down to move it.

Before that though I have abstracted out that logic into seperate Components, we've already seen `Portal` as used by Modal, and the to-be-better-named: `Position` Component, for calculating an offset based on the target and container.

that leaves us with the following as the implementation of an Overlay:

```js
class Popover extends React.Component {
  render(){
    let {
        container
      , containerPadding
      , target
      , placement
      , ...props } = this.props;

    return (
      <Portal show={props.show} container={container}>
        <Position {...{ container, containerPadding, target, placement }}>
          <PopoverMarkup>
            { this.props.children }
          </PopoverMarkup>
        </Position>
      </Portal>
    );
  }
}
```

Which feels like a lot of boilerplate for making custom Overlays. To help with that I am bundling both of these components up into an Overlay Component which makes it a bit better...

```js
class Popover extends React.Component {
  render() {
    return (
      <Overlay {...this.props}>
        <PopoverMarkup {...this.props}>
          { this.props.children }
        </PopoverMarkup>
      </Overlay>
    );
  }
}
```

This feels a bit verbose to me but probably reasonable for Custom Overlays. It also means that the Popover and Tooltip Components are no longer "dumb" static markup components, which enables the declarative api way up page. I think that this is worth it, the Alternative is that we needto abandon the declartive api and put the logic back in OverlayTrigger or write this when we want the declarative syntax:

```js
<Button bsStyle='default' ref='btn' onClick={e => this.setState({ show: !this.state.show })}>
  Holy guacamole!
</Button>
<Overlay
  show={this.state.show}
  target={()=> this.refs.btn}
  placement='bottom'
  container={mountNode}
  containerPadding={20}
>
  <Popover title='Popover bottom'>
    <strong>Holy guacamole!</strong> Check this info.
  </Popover>
</Overlay>
```

I find that less optimal than just having the PopOver component but it does have an advantage that @taion thinks is worth it. If we do it this way OverlayTrigger can jsut wrap its `overlay` prop in an `Overlay` Component, and then you can specify static markup to it, without it needing to be a special kind of Component.

##### Simplier Custom Overlays
```js
  class MyOverlay {
    render(){
      let { positionLeft: left, positionTop: top } = this.props
      return <div style={{left, top}}>my overlay</div>
    }
  }

 <OverlayTrigger overlay={<MyOverlay/>}/>
```

which is less verbose than:

##### Simplier Declarative Overlays
```js
  class MyOverlayMarkup {
    render(){
      let { positionLeft: left, positionTop: top } = this.props
      return <div style={{left, top}}>my overlay</div>
    }
  }

  // we could alternatively make this a function instead of a wrapper component like:
  // createOverlay(MyOverlayMarkup) or a decorator like @Overlay
  class MyOverlay {
    render(){
      return (
        <Overlay {...this.props}>
          <MyOverlayMarkup {...this.props}/>
        </Overlay>
      );
    }
  }

 <OverlayTrigger overlay={<MyOverlay/>}/>
```

### GET TO THE POINT ALREADY THIS PR IS SOOOOOOOO LONG

Which Overlay Api do we go with?
- Simplier Custom Overlays
- Simplier declarative Popover/Tooltips
- Something completely different
- Just revert it all


shermanjoshua:
Tested this out on my app.  Everything worked without changes and worked after adding in new components.

LGTM

taion:
As mentioned before, I still think `Position` or whatever should just render its child into a `div` that's appropriately styled with `top` and `left`, and just pass in the arrow position arguments to its child. I don't really like the idea of having to write e.g. `<div style={{left, top}}>` for every custom overlay.

I haven't fully thought through the `ModalTrigger` change. I can see the point you're making, but it just seems nicer to not have explicit state in my component if it doesn't get me anything. To the extent that I have a button that triggers a modal, I'm going to want something like `ModalTrigger` anyway, so I feel like we may as well provide it.

src/Modal.js
```diff
@@ -315,4 +349,37 @@ const Modal = React.createClass({
...
+          { modal }
+        </Portal>
+      );
+    } else {
+      return modal;
+    }
+  }
+});
+
+ModalOverlay.Body = Body;
```
taion: These are most likely not going to work with factories when declared with this syntax. Do we need them? Not sure of the benefit of typing `Modal.Body` rather than `ModalBody`.

src/RootCloseWrapper.js
```diff
@@ -26,7 +26,9 @@ export default class RootCloseWrapper extends React.Component {
     // If the click originated from within this component, don't do anything.
     // e.srcElement is required for IE8 as e.target is undefined
     let target = e.target || e.srcElement;
-    if (domUtils.contains(React.findDOMNode(this), target)) {
```
taion: This is not the v0.24-rc version of `RootCloseWrapper`.

src/Position.js
```diff
@@ -0,0 +1,77 @@
...
+    this.state = {
+      positionLeft: null,
+      positionTop: null,
+      arrowOffsetLeft: null,
+      arrowOffsetTop: null
+    };
+  }
+
+  componentWillMount(){
+    this._needsflush = true;
```
taion: Minor quibble but this should be `_needsFlush`.

src/Position.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  componentWillRecieveProps(){
+    this._needsflush = true;
+  }
+
+  componentDidMount(){
+    if ( !this._needsflush ) { return; }
```
taion: Maybe just DRY this and move the check into `_updatePosition`? Or call it `_maybeUpdatePosition` or whatever.

src/Modal.js
jquense: both are fine, the factories will get the ModalBody, ModalTitle, etc components from lib. I find this way is just more pleasant and saves 5 additional imports for components that you will always use together. There is not prescription to do it this way its merely a convenience

taion:
This should probably be against an RC branch. Do we want this for v0.24, or is this more appropriate for the next release after that?

Also, we should consider adding a convenience `TooltipTrigger` for tooltips.

src/RootCloseWrapper.js
jquense: yeah, the dream is to actual get this into master, for round one. hence all the deprecation warnings about the code. still gotta see if thats feasible

src/CreateOverlay.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import OverlayMixin from './OverlayMixin';
+import RootCloseWrapper from './RootCloseWrapper';
+
+
+export default function createOverlayedComponent (render, { renderChildren = false, rootClose = true } = {}) {
```
taion: This isn't used anywhere and duplicates a lot of code from `Portal` &c. Do we need this?

taion:
New components need to be added to `src/index.js`.

src/RootCloseWrapper.js
taion: I really don't think it's a good idea to introduce such a large change into master. It may not strictly be breaking, but it's a pretty big overhaul.

src/Modal.js
taion: It's weird thing. It seems a little weird to set things up this way - if you were using just ES6 modules, you'd be doing named exports, and they'd still have to be imported separately.

@mtscout6 You mentioned you were considering something similar for the dropdown overhaul. Which approach are you taking for e.g. `DropdownButtonTitle`?

mtscout6: I was taking the same approach. Partly because other components may also have some variant of `Title` that's specific to them. It helps to avoid naming collisions in the long run. I also like that it's clear through it's usage that it is meant to be a child of said component.

mtscout6:
As far as needing this to go to the v0.24 branch, my opinion is that if it's backwards compatible then I'm fine with it in master. That's partly why I asked @shermanjoshua to test it out in his app as I know he's used modals a lot more than me.

src/RootCloseWrapper.js
taion: So e.g. `DropdownButton.Title` instead of `DropdownButtonTitle`? I just think it's a bit confusing to have both.

src/CreateOverlay.js
jquense: no its not needed, I have it deleted downstream here I just haven't add a chance to push it

jquense:
> As mentioned before, I still think Position or whatever should just render its child into a div that's appropriately styled with top and left, and just pass in the arrow position arguments to its child. I don't really like the idea of having to write e.g. <div style={{left, top}}> for every custom overlay

I am trying this out, tho there isn't any way I can make that change in a backwards compatible way which is a limiting factor for this PR (we can break it after if we want but 'd prefer an initial PR that doesn't). A wrapping div is not going to be helpful because I'd have to make a choice about position: fixed vs absolute on that div and there isn't a good of letting the user specify (admittedly absolute is probably fine but positioning stuff is so complicated across browsers) I am trying out the other way you suggested tho and just having it pass position in as the `style` prop, then plain markup gets it automatically and in cases where you need to do something funky you can just not apply it (or adjust the left and top values).

I think that ModalTrigger is not worth the added convenience, for its cost. In most places it is more lines of code (b/c you need to create a class) and your functionality is more limited and state is implicit.

```js
let MyModal = React.createClass({
  render(){
   return (
    <Modal>
      ...
      <ModalFooter>
         button onClick={this.props.onHide}>close<button/>
      </ModalFooter>
    </Modal>)
  }
})

<ModalTrigger modal={modal}>
  <button>open<button/>
</ModalTrigger>
```
#### vs 

```js
let close = ()=> this.setState({ open: false})

<button onClick={this.setState({ open: true})}>open<button/>

<Modal show={this.state.open} onHide={close}>
      ...
   <ModalFooter>
     <button onClick={close }>close<button/>
   </ModalFooter>
</Modal>
```

taion:
I don't think it's ideal to add deprecation warnings without a major release.

Even separate from that, even if this isn't meant to be opaque to users, I'd still be inclined to save larger internal re-writes for larger releases, just out of caution.

src/RootCloseWrapper.js
mtscout6: The main idea is that you wouldn't import both, it's feasible that you could but you don't have to.

taion:
For the modal example? I tend to write it as:

```jsx
<ModalTrigger modal={this.renderModal()}>
  <button>hi</button>
</ModalTrigger>
```

vs

```jsx
<button onClick={this.handleShowModal}>hi</button>
{this.renderModal()}
```

Plus to not create a bunch of extra functions or break pure render, you need to explicitly bind the show/hide handlers at the component level, which also ends up being boilerplate.

jquense:
Point to ToolTip Triggers:

I removed it ultimately b/c from what I was thinking it didn't seem to gain us much

```js
<OverlayTrigger overlay={<Tooltip><strong>Wow</strong>, hi there </Tooltip>}></OverlayTrigger>

<TooltipTrigger tooltip={<span><strong>Wow</strong>, hi there </span>}></TooltipTrigger>
```
Plus the Popover equivalent, as we saw before, is ugly and unhelpful feeling. To me it felt better to try and reduce the api surface area rather then create an asymetric one, but I don't feel that strongly about it if there is some additional benefit I'm not thinking of.

mtscout6:
> I don't think it's ideal to add deprecation warnings without a major release.

What do you mean by that, are there warnings in here that have not been pushed to the v0.24-rc branch yet? If so then yes having a corresponding PR that performs the deprecation to the v0.24-rc branch would be ideal.

I still feel that so long as backwards compatibility is maintained by the public API than we can change whatever we need to under the covers in non Major releases. With that said though, we are still pre 1.0.0 and we are very clear that there will be breaking changes along the way. Though I don't feel that that would be the case here.

src/RootCloseWrapper.js
taion: I don't know - I feel like if you have `DropdownButton.Title`, then having `DropdownButtonTitle` as a separate import-able module rather than an implementation detail is a little bit redundant.

In an ES6 module world, though, I think it'd be odd not to write this as

```js
// Modal.js
export {default as Title} from './impl/ModalTitle';
```

which would have you write

```js
// ModalUser.js
import {Modal, Title as ModalTitle} from 'react-bootstrap/lib/Modal';
```

Or maybe having them be static properties of the `Modal` class is fine.

taion:
I don't think we should mark things as deprecated or add deprecation warnings at all between major releases.

We don't strictly need to follow e.g. the React approach of using "major" releases to mark features as deprecated and then only removing them on the next "major" release, but I believe that it's good practice to do so whenever it isn't excessively burdensome.

taion:
@jquense 

For a simple tooltip:

```js
<TooltipTrigger tooltip="some text">hi</TooltipTrigger>

<OverlayTrigger overlay={<Tooltip>"some text"</Tooltip>}>hi</OverlayTrigger>
```

i.e. in the simplest and probably most common case where you don't apply any styling to the `Tooltip` contents, this is significantly less verbose and more expressive.

jquense:
thats fair, tho in that case we could just add taht behavior to OverlayTrigger, since it reeeally needs to be wrapped in a component. if `this.props.children` is a text node we can just wrap it in a Tooltip vs letting react wrap it in a span

taion:
I don't think it really makes sense for `OverlayTrigger` to do that, though. It's not at all intuitive to me that `OverlayTrigger` should privilege tooltips, and I don't necessarily want to pull in R-B's `Tooltip` from using `OverlayTrigger`.

The difference is that popovers usually have more complicated markups (and most of mine are in custom classes), but I think tooltips are a really clear special case. It saves a lot of extra typing.

taion:
Scratch the `TooltipTrigger` stuff though - it's orthogonal to this and can be a separate PR that will work either way.

amarant:
Currently react-bootstrap creates an empty div for every tooltip or popover, in an application that has many tooltips it slows down the rendering a lot and mess with the react dev tools; it seems this pull request doesn't change that for now; it would be great if this ended up solving it. In the pull request #867 based on the original MixinOverlay it creates an empty div, the portal endpoint, when needed like the original jquery bootstrap. It also refers to the issue #353 that was closed without a solution.

src/Popover.js
```diff
@@ -48,7 +50,7 @@ const Popover = React.createClass({
     };
 
     return (
-      <div {...this.props} className={classNames(this.props.className, classes)} style={style} title={null}>
+      <div role='tooltip' {...this.props} className={classNames(this.props.className, classes)} style={style} title={null}>
```
mmerickel: should be `role='popover'` ?

jquense: Not in this case. There are a preset number of valid roles that that are allowed her according to the HTML/ARIA spec, out of those "tooltip" best describes the role of a popover.

jquense:
OK! passing, polished, and ready for a final check.

mtscout6:
Looks like there are merge conflicts can you rebase on the latest master?

docs/src/ComponentsPage.js
```diff
@@ -301,54 +307,96 @@ const ComponentsPage = React.createClass({
...
+
+                  {/* Popover */}
+                  <div className='bs-docs-section'>
+                    <h2 id='popovers'>Popovers</h2>
 
-                  <p>Positioned popover component.</p>
-                  <ReactPlayground codeText={Samples.PopoverPositioned} />
+                    <p>
+                      The Popover component, like the Tooltip can be used with an <code>Overlay</code> Component, or an <code>OverlayTrigger</code>.
+                      Unlike the Tooltip popovers are designed to display more reobust information about their targets.
```
dozoisch: reobust !

dozoisch:
Very nice job! LGTM! Love the new api

docs/examples/ModalContained.js
```diff
@@ -9,28 +9,40 @@
...
-  }
-});
-
 const Trigger = React.createClass({
+  getInitialState(){
+    return { show: false };
+  },
+
   render() {
+    let close = e => this.setState({ show: false});
```
taion: We should probably avoid binding handlers in `#render()` just to implicitly encourage good practice for our users.

taion:
Should we try to be consistent with code formatting in the examples? e.g. `render(){` v `render() {` for spacing, and in other bits.

mmerickel:
This is fantastic. Below probably belongs in another issue but I'm putting it here for now to avoid 

I would like to be able to control the `Position` more carefully however. Currently `getTopDelta` always adjusts for the scroll position of the element and thus if something is offscreen and I show a popover on it, the positioning is shifted to the top of the viewport (away from the element which is offscreen).

I manually removed the scroll offsets from `getTopDelta` and now things work great for me but I think this should be the default. Currently when the user scrolls the popover does not move back to where it should be (the arrow is clamped to the target but the popover box itself is split off away in space).

```diff
diff --git a/src/utils/overlayPositionUtils.js b/src/utils/overlayPositionUtils.js
index eb4a045..c106059 100644
--- a/src/utils/overlayPositionUtils.js
+++ b/src/utils/overlayPositionUtils.js
@@ -80,11 +80,10 @@ const utils = {

 function getTopDelta(top, overlayHeight, container, padding) {
   const containerDimensions = utils.getContainerDimensions(container);
-  const containerScroll = containerDimensions.scroll;
   const containerHeight = containerDimensions.height;

-  const topEdgeOffset = top - padding - containerScroll;
-  const bottomEdgeOffset = top + padding - containerScroll + overlayHeight;
+  const topEdgeOffset = top - padding;
+  const bottomEdgeOffset = top + padding + overlayHeight;

   if (topEdgeOffset < 0) {
     return -topEdgeOffset;
```

dozoisch:
@taion we should be able to enforce that using eslint no?

taion:
@mmerickel 

That behavior is intentional and matches upstream. The reason is that if you have a tall popover, you don't want to clip the top half of the popover. See http://getbootstrap.com/javascript/#live-demo-1

@dozoisch 

That would be a good way to enforce this.

@jquense 

Why does the "Overlay Contained" example no longer use `OverlayTrigger`? Does `container` no longer work there?

More broadly, I think it might be more clear if the documentation were structured as:

1. Tooltips (w/`OverlayTrigger` and maybe eventually `TooltipTrigger`)
2. Popovers (w/`OverlayTrigger`)
3. Custom overlays - also with the `Overlay` component as needed

Per the way Bootstrap is set up, I'd expect the popover/tooltip cases to cover most user needs, so we'd probably be better served by illustrating the more powerful, general use case last rather than first.

jquense:
@taion 

It works fine with container, and with OverlayTrigger, but I tried to demonstrate how Tooltips and Popovers fit into the new Overlay component. The concern is that just classifying Overlay component use as for "custom" overlays suggests that Tooltips and Popovers don't work with it, or are special. Some of the use cases this newer structure solves is not just for custom Overlays but probably more so for custom Popover and Tooltip behavior, such as having finely grained control over the showing and hiding of Popovers.

I am fine reworking the doc structure, but I did want to also demonstrate the increased flexibility with the existing overlays as well as with custom ones.

taion:
Agree with what you're saying with using tooltips and popovers with the overlay component. However, I think it'd be more clear if that were signposted separately - i.e. show examples with the existing tooltip and popovers in the "custom overlay" example, e.g. with a "display all overlays" button or something.

jquense:
far enough I can work on that. Do we want to hold up the PR for that, or do it in a follow up?

taion:
Your call. It's a third-order concern, but I think that sort of an example (e.g. clicking a button to show multiple overlays) both illustrates more clearly what's happening, and gives a good motivating example for why someone would want to not use `OverlayTrigger`.

I care a lot more about the code formatting examples since I think people are going to notice those. The callback binding is whatever... just a bit nicer to not e.g. show patterns that might break PureRender or whatever downstream.

jquense:
ok reworked the ordering and examples. OverlayTrigger is still a bit oddly placed but maybe for the better.

jquense:
Any additional thoughts or are we good to go? It would be nice to get a release cut and this merged (also I selfishly want this at work)

mmerickel:
In the new docs it looks like the sidebar menu is out of sync with the position on screen. Tooltips and Popovers were moved above overlays for example, and the correct section is not highlighted.

mmerickel:
Are there any tricks to getting a container that can be used for positioning? I'm struggling when showing an overlay where the target element is off-screen and all the examples right now show a container that is either `{this}` or `mountNode`. It'd be nice (although maybe incorrect?) to be able to send a callback down similar to target so that I can use a `ref` that is not `this` as the container.

jquense:
@mmerickel you can use any component or DOM node as a container, I agree it would be nice to also allow passing a function so that can be done lazily (similar to the new `target` props)  but I think that can be tackled in a separate PR

mmerickel:
@jquense I get that I can use any component or DOM node as the container, but it's not clear how if it's not `this` or some other `this` node passed from above. For example how would I set the `wrappedContainer` as the container in this example instead of `this`? Is it only possible with a future callback?

```
render() {
  return (
    <div class="wrapper">
      <div ref="wrappedContainer" class="wrapped">
        <Overlay container={this} placement="left" show={true}>
          <Popover>Hello world</Popover>
        </Overlay>
      </div>
    </div>
  );
}
```

jquense:
No I agree you can't currently do that, and you probably should be able to. I just think that it doesn't need to be done in this PR.  If we can get it merged, then you can open up a PR taking a function as well as a component/Node no problem.

mmerickel:
@jquense That's fine, I'm using this branch right now and struggling with that issue so I appreciate you taking a look. So far things have been working well aside from some container positioning issues that I haven't been able to solve yet. Far and away smoother than when I was using master though.

taion:
A few things -

* As @mmerickel mentioned the ordering on the affix of Overlays, Tooltips, Popovers is wrong
* The initially-shown overlaid tooltip and overlay components end up getting scrolled off the page by the container scroll tracking - they should probably not start off as shown
* I don't know why but the top/bottom tooltips here are off-center: 
image
* Maybe we can merge the "Use the Tooltip in combination with OverlayTrigger to add some interactivity to your tips." and "Positioned tooltip component." examples? Just call the latter e.g. "Positioned tooltips with `OverlayTrigger`" or something?

Looking great otherwise though.

taion:
:shipit: :shipit: :ship: :boat:

(after fixup-ing the commits from #927 and rebasing onto master)

And IMO we should cut 0.23.6 as soon we this and https://github.com/react-bootstrap/react-bootstrap/pull/923 are resolved.

PR #856, opened 2015-06-18
[fixed] Modal Null Exception when react-bootstrap is loaded before the Body tag

### LINKED ISSUE #849: Uncaught TypeError: Cannot read property 'appendChild' of null
Happens with your latest version 0.23.4 but didn't earlier.

Thrown at line `document.body.appendChild(scrollDiv);` in

```js
	if (_utilsDomUtils2['default'].canUseDom) {
	  var scrollDiv = document.createElement('div');

	  scrollDiv.style.position = 'absolute';
	  scrollDiv.style.top = '-9999px';
	  scrollDiv.style.width = '50px';
	  scrollDiv.style.height = '50px';
	  scrollDiv.style.overflow = 'scroll';

	  document.body.appendChild(scrollDiv);

	  scrollbarSize = scrollDiv.offsetWidth - scrollDiv.clientWidth;

	  document.body.removeChild(scrollDiv);
	  scrollDiv = null;
	}
```

jquense:
Make scrollbar size check lazy, Only called after components have
mounted. fixes #849

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -54,7 +54,11 @@ function onFocus(context, handler) {
 
 let scrollbarSize;
 
-if (domUtils.canUseDom) {
+function getScrollbarSize(){
+  if ( scrollbarSize !== undefined ){
+    return scrollbarSize;
+  }
+
   let scrollDiv = document.createElement('div');
 
   scrollDiv.style.position = 'absolute';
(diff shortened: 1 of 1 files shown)


taion:
Should the scrollbar size calculation be per-document? It's a bizarre edge case that I don't expect to show up, but I think it might be possible for scrollbar size to differ on a sub-document.

taion:
Never mind, that has nothing to do with this change. LGTM.

PR #857, opened 2015-06-18
Accessibility

jquense:
Some of this is just hard to do without a good answer to: https://github.com/facebook/react/issues/4000 

I've posted a comment asking if there was anything that was feasible for libraries to do
https://github.com/facebook/react/issues/4000#issuecomment-113044968

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -6,6 +6,7 @@ const NavItem = React.createClass({
   mixins: [BootstrapMixin],
 
   propTypes: {
+    linkId: React.PropTypes.string,
     onSelect: React.PropTypes.func,
     active: React.PropTypes.bool,
     disabled: React.PropTypes.bool,
@@ -26,6 +27,7 @@ const NavItem = React.createClass({
   render() {
     let {
         role,
+        linkId,
         disabled,
         active,
         href,
@@ -43,6 +45,7 @@ const NavItem = React.createClass({
           href,
           title,
           target,
+          id: linkId,
           onClick: this.handleClick,
           ref: 'anchor'
         };
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -10,9 +10,11 @@ const NavItem = React.createClass({
     active: React.PropTypes.bool,
     disabled: React.PropTypes.bool,
     href: React.PropTypes.string,
+    role: React.PropTypes.string,
     title: React.PropTypes.node,
     eventKey: React.PropTypes.any,
-    target: React.PropTypes.string
+    target: React.PropTypes.string,
+    'aria-controls': React.PropTypes.string
   },
 
   getDefaultProps() {
@@ -23,18 +25,21 @@ const NavItem = React.createClass({
 
   render() {
     let {
+        role,
         disabled,
         active,
         href,
         title,
         target,
         children,
+        'aria-controls': ariaControls,  // eslint-disable-line react/prop-types
         ...props } = this.props; // eslint-disable-line object-shorthand
     let classes = {
           active,
           disabled
         };
     let linkProps = {
+          role,
           href,
           title,
           target,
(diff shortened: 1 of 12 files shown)


src/NavItem.js
```diff
@@ -23,32 +26,37 @@ const NavItem = React.createClass({
...
-    if (href === '#') {
+    if (!role && href === '#') {
       linkProps.role = 'button';
     }
 
     return (
-      <li {...props} className={classNames(props.className, classes)}>
-        <a {...linkProps}>
+      <li {...props} role='presentation' className={classNames(props.className, classes)}>
+        <a {...linkProps} aria-selected={active} aria-controls={ariaControls}>
```
AlexKVal: Isn't it supposed to be as `ariaSelected` and `ariaControls` ?
https://facebook.github.io/react/docs/tags-and-attributes.html#supported-attributes
> Note:
All attributes are camel-cased

jquense: Yup with the exception of data and aria attributes :) not sure why that is

> React supports all data-* and aria-* attributes as well as every attribute in the following lists.

AlexKVal: Yep. I've checked it now... 
It has to be like `aria-*`. So pity.

AlexKVal:
Cool :sparkle:

src/NavItem.js
```diff
@@ -23,32 +26,37 @@ const NavItem = React.createClass({
...
     let {
+        role,
+        linkId,
         disabled,
         active,
         href,
         title,
         target,
         children,
+        'aria-controls': ariaControls,  // eslint-disable-line react/prop-types
```
AlexKVal: I created issue
https://github.com/yannickcr/eslint-plugin-react/issues/118

jquense: :+1: its interesting that it isn't supported

jquense:
anyone know why Appveyor is failing?

mathieumg:
npm connectivity issues it would seem?

mtscout6:
I re-triggered AppVeyor, npm installs have struggled to be reliable on AppVeyor for some reason...

jquense:
AppVeyor hates my PRs :P

dozoisch:
a11y for the win! Nice job

jquense:
Does anyone have any suggestion about how we proceed with the ID issue? @mtscout6 did you make any progress in your Dropdown work? I saw that was a task in the issue.

It seems like we practically have about 3 options
- Generate ID's naively and tell users to supply their own id's when server rendering (so they match)
- only add the a11y stuff when a user supplies an id, (like [this](https://github.com/react-bootstrap/react-bootstrap/pull/857/files#diff-a8a38bd1bdd1a56c3f27491f45d5b8e3R8))
- require the user to wrap there App component in a utility we provide like: https://github.com/facebook/react/issues/4000#issuecomment-113097875

Ideally I'd like to add a11y stuff by default and have the user turn it off if necessary vs an opt in.

mtscout6:
I was using [uuid](https://www.npmjs.com/package/uuid) to generate id's but once I saw facebook's issue I ripped it out primarily out of fear of delaying the dropdown any further and just made `id` a required prop.

jquense:
haaaa, I like just making it a required prop. nice and simple. I added it to the list of options above

mtscout6:
What I like about it being a required prop is that once facebook figures out a good solution we can remove the required prop. It would not break backwards compatibility since I see it as a good thing to allow users to customize if they wish.

jquense:
ha finally green. Do folks want to merge this as is, and work on stuff that requires ID's in another PR against 0.24 or just add it all in here?

mtscout6:
Id's should be fine here since required props are just warnings right?

jquense:
true...its fine as long as the components don't break assuming its there b/c its required, but I can make sure that's true.

mtscout6:
I guess if internally you break because it doesn't exist than it would make sense to go in v0.24. Maybe it's just safest to put in v0.24...

src/NavItem.js
AlexKVal: @Cellule has fixed it https://github.com/yannickcr/eslint-plugin-react/pull/124 :sparkle:

mtscout6:
@jquense Were you planning to put the id stuff here or in another PR? If another PR we can get this one merged right?

jquense:
yeah I think let's just merge this and I will work on another PR. these components are done as is anyway.  might as well get some improvements in now

PR #860, opened 2015-06-18
Load custom eslint rules, including a spread aware shorthand rule

jquense:
Fixes that dang false eslint warning about using spread properties in an object.

Prior art on this: https://github.com/facebook/react/pull/4034

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,8 @@
   },
   "parser": "babel-eslint",
   "plugins": [
-    "react"
+    "react",
+    "react-bootstrap-internal"
   ],
   "rules": {
     "comma-spacing": 2,
@@ -18,7 +19,7 @@
     "no-underscore-dangle": 0,
     "no-unused-vars": [2, { "vars": "all", "args": "none" }],
     "no-var": 2,
-    "object-shorthand": 2,
+    "react-bootstrap-internal/object-shorthand": 2,
     "quotes": [2, "single", "avoid-escape"],
     "react/display-name": 0,
     "react/jsx-no-undef": 2,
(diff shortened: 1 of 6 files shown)


mtscout6:
This seems like a very generic problem that is not isolated to react-bootstrap, and it also seems tricky to maintain over time. Perhaps pushing this to a new npm module that adds this rule would be good? That would have the added benefit of being usable in react-router-bootstrap if we end up needing it there. The goal of course being to deprecate that module once eslint truly supports es7.

jquense:
the particular rule is generic and would work great in its own npm repo, and that makes the most sense probably. I don't particularly want to own said repo independently, but if we wanted to add it to the org that seems like a good idea, and I'd be happy to maintain it. Not having rights to make new repos this was the next easiest thing :).

Aside from that particular rule we might want to have custom eslint rules that are bootstrap specific, which was another reason i went this approach. That being said I have no idea what those rules might be, and we can always just cross that bridge then if the need ever arises and feels worth it.

jquense:
We will solve this particular thing in either babel-eslint or a seperate repo in the babel org

taion:
It'd be great if this could go in `babel-eslint` or similar - would like to minimize the number of packages I need in my scaffolding.

jquense:
so as a of a few minutes ago [eslint-plugin-babel](https://github.com/babel/eslint-plugin-babel) is now a thing. but hopefully if this goes through, https://github.com/eslint/eslint/pull/2793 we can just move it into babel-eslint itself

mtscout6:
@jquense do you still need this branch?

jquense:
nope! thanks for the reminder :)

PR #867, opened 2015-06-20
[fixed] Create overlay target only when needed.

amarant:
Currently react-bootstrap creates an empty div for every tooltip or popover, in an application that has many tooltips it slows down the rendering a lot and mess with the react dev tools; this aims to only create an empty div, the portal endpoint, when needed like the original jquery bootstrap. It also refers to the issue #353 that was closed without a solution.
Thanks for looking at it.

### DIFF
--- a/src/OverlayMixin.js
+++ b/src/OverlayMixin.js
@@ -7,8 +7,28 @@ export default {
     container: CustomPropTypes.mountable
   },
 
+  componentDidMount() {
+    this._renderOverlay();
+  },
+
   componentWillUnmount() {
     this._unrenderOverlay();
+    this._unmountOverlayTarget();
+  },
+
+  componentDidUpdate() {
+    this._renderOverlay();
+  },
+
+  _mountOverlayTarget() {
+    if (!this._overlayTarget) {
+      this._overlayTarget = document.createElement('div');
+      this.getContainerDOMNode()
+        .appendChild(this._overlayTarget);
+    }
+  },
+
+  _unmountOverlayTarget() {
     if (this._overlayTarget) {
       this.getContainerDOMNode()
         .removeChild(this._overlayTarget);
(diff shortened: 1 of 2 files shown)


mathieumg:
Take a look at https://github.com/react-bootstrap/react-bootstrap/pull/855

amarant:
It seems it also creates empty divs.

jquense:
creating the divs up front reduces DOM manipulations when opening and closing overlays. I'd be happy to alter how it works but some actual numbers demonstrating a performance increase would be helpful for deciding if it's a worthwhile change

amarant:
Creating the divs at the beginning in my application end up doing 
```javascript
this._overlayTarget = document.createElement('div')
this.getContainerDOMNode().appendChild(this._overlayTarget)
````
times the total number of OverlayMixed component, at the initial rendering. So maybe 200 single div creation and appending.
There is generally at most 1 tooltip viewed at a time, and not every tooltip ends up being viewed in the lifetime of the application, I really don't understand what number are needed. Also note that viewing a tooltip ends up creating other span or div so I don't see the problem.

amarant:
So I benchmarked it :
with the divs upfront : 180 ms,
with divs
lazy loaded: 60ms
lazy divs
With 200 holy guacamoles from the examples.

```javascript
import React from 'react';

import OverlayTrigger from '../../src/OverlayTrigger';
import Tooltip from '../../src/Tooltip';
import Button from '../../src/Button';

const TestPage = React.createClass({
  render() {
    let tooltips = [];
    for (let i = 0; i < 200; i++) {
      tooltips.push(
        <OverlayTrigger
          key={i}
          placement='left'
          overlay={
            <Tooltip>
              <strong>Holy guacamole!</strong>
                Check this info.
              </Tooltip>}>
          <Button bsStyle='default'>Holy guacamole!</Button>
        </OverlayTrigger>
      );
    }
    return (
      <div>
        {tooltips}
      </div>
    );
  }
});

export default TestPage;

``

jquense:
Hi there, 

Sorry I'm not trying to be difficult, and I agree with you that it is probably good for performance to make this change. however, since performance is notoriously difficult to intuit with accuracy and often defies common sense, it is helpful to have some data. Changes cost something in terms of complexity and break current expectations and can have unexpected consequences even if minimal. I think the use case of 200+ overlays is exceptional but this change affects everyone. at the very least it'd be helpful knowing what the difference in initial load this change makes with varying amounts of tooltips. if it's say 5ms vs 250ms that's very useful info.

mathieumg:
@jquense Can't seem to find your post, but you said something along the lines of:

> I think the use case of 200+ overlays is exceptional but this change affects everyone.

I think it's not that exceptional, one can easily imagine a list of 200 items where each item would have a tooltip explaining something, for example.

taion:
@amarant 

Would you mind showing a worse-case situation with only 1 tooltip, just to see how it looks?

I think this is a good change just w/r/t not cluttering up the DOM.

amarant:
@taion 
With 1 tooltip :
lazy loaded: 12ms
lazy portal
with the divs upfront : 15 ms,
upfront portal

taion:
That's extremely surprising to me. Seems like it's doing more work and should be slower. Wow.

amarant:
I don't understand what is surprising, in those the tooltip are not loaded, I only talk about initial rendering, a tooltip load is negligible because it is triggered by a user event once at a time.

taion:
Oh, I see. I misunderstood what numbers you were quoting.

I'm okay in principle with doing this. Understand that it will probably take a bit longer to render the tooltips, but I don't think by enough to meaningfully adversely affect user experience.

@jquense?

jquense:
ya seems like a good idea...I need to make sure the concept plays nicely with the new changes and the ones I'm thinking of with the transitions. It seems like it should be fine, but I need to try it ok. Practically maybe we can close this out @amarant and I can add it in with the Overlay Overhaul if that's ok with you? Thanks so much for the idea, and taking the time to put together a PR and benchmarks

amarant:
Ok, looks fine to me.

PR #872, opened 2015-06-21
eslint plugin babel

jquense:
supersedes #860 

Not the most ideal outcome to this, but its the best we got at this point :/

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,8 @@
   },
   "parser": "babel-eslint",
   "plugins": [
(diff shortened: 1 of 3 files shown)


AlexKVal:
Nice :+1: :sparkle: 

You've done it:
https://github.com/eslint/eslint/issues/2486#issuecomment-100653157
> But we still don't support es7 features, so I think your best bet would be to copy object-shorthand rule, modify it, and package it as a plugin.


LGTM

Issue #878, opened 2015-06-22
Add componentClass property to Jumbotron


lo1tuma:
Hi,

I would like to see the `componentClass` also in the `Jumbotron` component, to be able to use something different than a `div`.

mtscout6:
Can you provide a little more context as to why you want a different componentClass there? I ask because the existing Jumbotron is following the same standard as documented by Bootstrap at http://getbootstrap.com/components/#jumbotron

lo1tuma:
I would like to use a `section` instead of a `div`.

The bootstrap documentation says nothing about in which tag-combination the `.jumbotron` class can be used. I’m a bit confused: do you only support tag names which are also shown in code examples in the bootstrap documentation?

mtscout6:
Not necessarily I'm mainly trying to understand what exactly is needed. In this case I can see how a section is a reasonable alternative. If you want to draft a PR for this and default it to div that'd be ok.

lo1tuma:
Sure, I’ll make a PR soon.

Thanks.

mtscout6:
Closed by #879

PR #879, opened 2015-06-22
Jumbotron componentclass

lo1tuma:
fixes: #878

### DIFF
--- a/src/Jumbotron.js
+++ b/src/Jumbotron.js
@@ -2,11 +2,21 @@ import React from 'react';
 import classNames from 'classnames';
 
 const Jumbotron = React.createClass({
+  propTypes: {
+    componentClass: React.PropTypes.any.isRequired
+  },
+
+  getDefaultProps() {
+    return { componentClass: 'div' };
+  },
+
   render() {
+    const ComponentClass = this.props.componentClass;
+
     return (
-      <div {...this.props} className={classNames(this.props.className, 'jumbotron')}>
+      <ComponentClass {...this.props} className={classNames(this.props.className, 'jumbotron')}>
         {this.props.children}
-      </div>
+      </ComponentClass>
     );
   }
 });
--- a/test/NavbarSpec.js
+++ b/test/NavbarSpec.js
@@ -61,7 +61,7 @@ describe('Nav', function () {
     let instance = ReactTestUtils.renderIntoDocument(
       <Navbar componentClass={'header'}/>
     );
-    assert.ok(React.findDOMNode(instance).nodeName, 'HEADER');
+    assert.equal(React.findDOMNode(instance).nodeName, 'HEADER');
   });
 
   it('Should add header with brand', function () {
--- a/test/JumbotronSpec.js
+++ b/test/JumbotronSpec.js
@@ -9,6 +9,8 @@ describe('Jumbotron', function () {
         <strong>Content</strong>
       </Jumbotron>
     );
+
+    assert.equal(React.findDOMNode(instance).nodeName, 'DIV');
     assert.ok(ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'strong'));
   });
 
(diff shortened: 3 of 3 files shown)


src/Jumbotron.js
```diff
@@ -2,11 +2,21 @@ import React from 'react';
 import classNames from 'classnames';
 
 const Jumbotron = React.createClass({
+  propTypes: {
+    componentClass: React.PropTypes.node.isRequired
```
jquense: I think `node` might be the wrong type here...we are really looking for a string, or a function. i.e. tagname or Component Class.

test/NavbarSpec.js
```diff
@@ -61,7 +61,7 @@ describe('Nav', function () {
     let instance = ReactTestUtils.renderIntoDocument(
       <Navbar componentClass={'header'}/>
     );
-    assert.ok(React.findDOMNode(instance).nodeName, 'HEADER');
+    assert.equal(React.findDOMNode(instance).nodeName, 'HEADER');
```
jquense: is this is just an unrelated fix? thanks :)

src/Jumbotron.js
lo1tuma: Good point. I copied this from the [`Row` component](https://github.com/react-bootstrap/react-bootstrap/blob/1bc32530e1c8cfa861ad7bbfa4bb943a3b0ecd8d/src/Row.js#L5-L7). [`Navbar`](https://github.com/react-bootstrap/react-bootstrap/blob/1bc32530e1c8cfa861ad7bbfa4bb943a3b0ecd8d/src/Navbar.js#L18) uses `node` as well.

The react documentation says the following about the `node` PropType:
> Anything that can be rendered: numbers, strings, elements or an array containing these types.

That said, I think `node` is the correct type, consider the following example:

```js
<Jumbotron componentClass={MyCustomSectionComponent}>
   <strong>foo</strong>
</Jumbotron>
```

test/NavbarSpec.js
lo1tuma: Yes, I've came across this issue as I looked how other components are testing the `componentClass` property.

I’ve made it a separate commit. If desired I can make a separate PR for this change.

src/Jumbotron.js
jquense: A class isn't an Element in this case: https://jsfiddle.net/69z2wepo/10729/ so your example actually would fail 

also this example would not fail 
```
<Jumbotron componentClass={2}>
   <strong>foo</strong>
</Jumbotron>
```

test/NavbarSpec.js
jquense: The separate commit is perfect, no need for a seperate PR, thanks for taking the time

src/Jumbotron.js
lo1tuma: You are correct.

What `PropType` would you suggest to use? React itself [uses `any`](https://github.com/facebook/react/blob/ba81b60ad8e93b747be42a03b797065932c49c96/src/addons/transitions/ReactTransitionGroup.js#L25).

What about the [other occurrences](https://github.com/react-bootstrap/react-bootstrap/search?utf8=%E2%9C%93&q=componentClass) of `componentClass?

jquense: In the past I've personally used:
```
PropTypes.oneOf([PropTypes.string, PropTypes.func])
```
It doesn't catch everything but it gets closer at least :/ the more robust solution might be to add something to the CustomProptypes util like this: https://github.com/jquense/react-widgets/blob/master/src/util/propTypes.js#L10

for now though its probably fine to just go with `any` unless you want to tackle a more complete solution.

mtscout6:
Since the fix to the tests is not relevant to the public api it may just be noise in the changelog so no real need for the [fixed] prefix on that commit. Also on the `[added]` commit you are repeating yourself by saying `[added] Add` you could just drop the second `Add` ;)

src/Jumbotron.js
lo1tuma: Ok, I’ve changed it to `any`. A more complete solution would still make sense, but I think it is out of the scope of this PR, because it should be fixed for all other components with a `componetClass` prop too.

lo1tuma:
@mtscout6 I’ve updated the commit messages :wink:

jquense:
LGTM

taion:
Sorry for the minor comments but I thought we weren't going to do isRequired for props with defaults... not that we're being particularly consistent here.

Should we make an issue to fix other instances of componentClass in this repo? They all use `node`.

mtscout6:
Yes we should make an issue to clean those up. Another issue to clean up the unnecessary `isRequired` specifiers would be good too.

mtscout6:
LGTM

PR #880, opened 2015-06-23
Ensure all src folder js files are loaded for full code coverage analysis

mtscout6:


test/index.js
```diff
@@ -19,5 +19,9 @@ describe('Process environment for tests', function () {
   });
 });
 
+// Ensure all files in src folder are loader for proper code coverage analysis
```
AlexKVal: loader => loaded

AlexKVal:
Ah.. Here is the cause of that -0.01% :smile: 

LGTM

test/index.js
mtscout6: Thanks, fixed!

PR #881, opened 2015-06-23
Do not ignore `offsetParent` in jQuery mode of `domUtils.getPosition`

Morhaus:


taion:
LGTM. Need 1 more reviewer

jquense:
vaguely unrelated why do we even have the jquery branch? also doesn't jquery's position use offsetParent or is the param here not actually an offsetParent?

taion:
jQuery branch for IE8 compat

Explicit offsetParent there is actually only ever the container node for overlays, i.e. not actually the offset parent.

AlexKVal:
It seems the fix for better IE8 support :+1: 
LGTM

PR #887, opened 2015-06-24
Generate Prop Documentation

jquense:
This is a first pass at getting some of the doc info generated. Not having prop lists in our docs was really getting to me. So this is an attempt at a "good enough for now" solution that we can build on. 

you can see that a bunch of info is still missing from the doc info, mostly because there are no doc comments on any Props, but also b/c some stuff isn't statically analyzable. For those cases we can use the jsDoc "doclets" (like @private) to tell the generator how to display the info.

This isn't meant to be a complete solution, but one that is better than what we currently have. I am hoping that others will help start documenting our props as we go.

It should also be said that I am not sure this is working with the doc build command. It does work with the dev server version though

### DIFF
--- a/docs/generate-metadata.js
+++ b/docs/generate-metadata.js
@@ -1,0 +1,97 @@
+import metadata from 'react-component-metadata';
+import glob from 'glob';
+import fsp from 'fs-promise';
+import promisify from '../tools/promisify';
+
+let globp = promisify(glob);
+
+// removes doclet syntax from comments
+let cleanDoclets = desc => {
+  let idx = desc.indexOf('@');
+  return (idx === -1 ? desc : desc.substr(0, idx )).trim();
+};
+
+let cleanDocletValue = str => str.replace(/^\{/, '').replace(/\}$/, '');
+
+let isLiteral = str => str.trim()[0] === '"' || str.trim()[0] === "'";
+
+/**
+ * parse out description doclets to an object and remove the comment
+ *
+ * @param  {ComponentMetadata|PropMetadata} obj
+ */
+function parseDoclets(obj){
+  obj.doclets = metadata.parseDoclets(obj.desc || '');
+  obj.desc = cleanDoclets(obj.desc || '');
+}
+
+/**
+ * Reads the JSDoc "doclets" and applies certain ones to the prop type data
+ * This allows us to "fix" parsing errors, or unparsable data with JSDoc style comments
+ *
+ * @param  {Object} props     Object Hash of the prop metadata
+ * @param  {String} propName
+ */
+function applyPropDoclets(props, propName){
+  let prop = props[propName];
+  let doclets = prop.doclets;
+  let value;
+
+  // the @type doclet to provide a prop type
+  // Also allows enums (oneOf) if string literals are provided
+  // ex: @type {("optionA"|"optionB")}
+  if (doclets.type) {
+    value = cleanDocletValue(doclets.type);
+    prop.type.name = value;
+
+    if ( value[0] === '(' ) {
+      value = value.substring(1, value.length - 1).split('|');
+
+      prop.type.value = value;
+      prop.type.name = value.every(isLiteral) ? 'enum' : 'union';
+    }
+  }
+
+  // Use @required to mark a prop as required
+  // useful for custom propTypes where there isn't a `.isRequired` addon
+  if ( doclets.required) {
+    prop.required = true;
+  }
+}
+
+
+export default function generate(destination, options = { mixins: true }){
+
+  return globp(__dirname + '/../src/**/*.js') //eslint-disable-line no-path-concat
+    .then( files => {
+
+      let results = files.map(
+        filename => fsp.readFile(filename).then(content => metadata(content, options)) );
+
+      return Promise.all(results)
+        .then( data => {
+          let result = {};
+
+          data.forEach(components => {
+            Object.keys(components).forEach(key => {
+              const component = components[key];
+
+              parseDoclets(component);
+
+              Object.keys(component.props).forEach( propName => {
+                const prop = component.props[propName];
+
+                parseDoclets(prop);
+                applyPropDoclets(component.props, propName);
+              });
+            });
+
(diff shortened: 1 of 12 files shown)


taion:
:+1:

Very cool. Biggest missing piece in docs by far.

Should we roughly follow the styling of Bootstrap JS options like http://getbootstrap.com/javascript/#modals-options?

jquense:
grr I cannot for the life of me get solve this confusing travis issue...the pattern is definitely provided. Everything works locally, maybe someone else can understand why its failing?

docs/src/ComponentsPage.js
```diff
@@ -168,11 +177,19 @@ const ComponentsPage = React.createClass({
                   <h3 id='btn-dropdowns-right'>Dropdown right variation</h3>
                   <p>Trigger dropdown menus that align to the right of the button using the <code>pullRight</code> prop.</p>
                   <ReactPlayground codeText={Samples.SplitButtonRight} />
+
+                  <h3 id='buttons-props'>Props</h3>
```
dozoisch: whoops line 105, 149 and this one have the same id!

dozoisch:
Looks good man! Nice job

docs/generate-metadata.js
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  if ( doclets.required) {
+    prop.required = true;
+  }
+}
+
+export default function generate(destination, options = { mixins: true }){
+  //console.log(path.join(__dirname, '../src/**/*.js'))
```
mathieumg: Left this here by mistake?

mathieumg:
> Biggest missing piece in docs by far.

Yes.

Didn't check it out yet to try the look and feel, but welcome addition! Are the `console.log` calls scattered around intentional? (some are commented, some aren't) In any event, good job! :tada:

mathieumg:
@jquense I think you're getting hit by a Babel regression, I just opened an issue to validate that: https://github.com/babel/babel/issues/1815

docs/generate-metadata.js
```diff
@@ -0,0 +1,98 @@
...
+        resolve(result);
+      }
+
+      args = args.concat(finish);
+
+      fn.apply(null, args);
+    });
+
+  };
+}
```
mtscout6: Should this be defined somewhere else? I think it's simple enough that we shouldn't need to find another dependency that provides it, especially since this is for dev tools only. Though it is generic enough to move to say `tools/promisify` maybe?

mtscout6:
Huge thanks for putting this together it is definitely a great place to start from! Can't say I entirely understand everything going on with the metadata generation without taking more time to look at it, but I don't see anything glaring from the few minutes I looked.

I like the idea of styling the table a little more, just so it doesn't look so plain. But that can happen in a follow up PR, if you don't want to in this one. If that's the case though we should create an issue for such so we don't forget about it.

Aside from the minor cleanup already mentioned to remove commented out code this looks good to me.

jquense:
Ok changes made, tho I left the styling mostly the same, anyone is welcome to adjust that as they see fit :)

I did add a note about the generation in the docs section of contributing and added more comments around the code to try and help make it a bit clearer

mtscout6:
> not sure what just happened to the coverage tho....

What's wrong with it?

jquense:
@mtscout6 it was going haywire, but it seems to resolved itself. Originally said that this dropped coverage by 45%

docs/generate-metadata.js
```diff
@@ -0,0 +1,97 @@
...
+
+// removes doclet syntax from comments
+let cleanDoclets = desc => {
+  let idx = desc.indexOf('@');
+  return (idx === -1 ? desc : desc.substr(0, idx )).trim();
+};
+
+let cleanDocletValue = str => str.replace(/^\{/, '').replace(/\}$/, '');
+
+let isLiteral = str => str.trim()[0] === '"' || str.trim()[0] === "'";
```
AlexKVal: just idea
```js
let isLiteral = str => /^("|')/.test(str.trim())
```

docs/generate-metadata.js
```diff
@@ -0,0 +1,97 @@
...
+
+let globp = promisify(glob);
+
+// removes doclet syntax from comments
+let cleanDoclets = desc => {
+  let idx = desc.indexOf('@');
+  return (idx === -1 ? desc : desc.substr(0, idx )).trim();
+};
+
+let cleanDocletValue = str => str.replace(/^\{/, '').replace(/\}$/, '');
```
AlexKVal: just idea
```js
let cleanDocletValue = str => str.replace(/^\{|\}$/g, '')
```

docs/src/PropTable.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
+import Table from '../../src/Table';
+import merge from 'lodash/object/merge';
+
+let cleanDocletValue = str => str.replace(/^\{/, '').replace(/\}$/, '');
```
AlexKVal:
```js
let cleanDocletValue = str => str.replace(/^\{|\}$/g, '')
```

AlexKVal:
LGTM 
(those 'ideas' with some minor DRY-es I could re-write myself after merging :cherries: )

It is definitely good to start with. :+1:

mtscout6:
Any objections from the team about getting this merged and letting follow on PRs do any cleanup?

mathieumg:
Still didn't get around to test/see it, but I trust you all! Move fast, etc. :+1: LGTM

On a side note, perhaps we could try to provide (not mandatorily, but recommend them) screenshot/screencasts when submitting a PR with visual/behavioral implications.

jquense:
Good to go? I'd like to get this in so I can finish the docs on #855 and get that all polished up

AlexKVal:
There is a glance how it looks now @mathieumg

screen shot 2015-06-25 at 7 51 24 pm
screen shot 2015-06-25 at 7 50 43 pm

mathieumg:
@jquense Amazing! Thanks! :grin:

PR #895, opened 2015-06-25
[removed] Don't need to disable Babel cache

taion:
Superseded by babel-loader#5.2.0

### DIFF
--- a/docs/build.js
+++ b/docs/build.js
@@ -34,10 +34,9 @@ function generateHTML(fileName) {
   });
 }
 
-export default function BuildDocs({useCache, dev}) {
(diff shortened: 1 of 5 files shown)


taion:
Note that this takes out the custom cache directory as well, since developers on the correct version of `babel-loader` won't even need to explicitly clear their cache during development.

mtscout6:
Dependent on #896?

taion:
This is on the alpha branch while #896 is on master, so hopefully shouldn't be relevant.

AlexKVal:
I'll fix eslint warnings by next commit.

PR #898, opened 2015-06-25
[changed] Add two-release deprecation policy

### LINKED ISSUE #863: Official two-release deprecation policy before 1.0.0
What do you all think about instituting an official two-release deprecation policy before 1.0.0, akin to upstream React?

For example, we are currently at v0.23. If we want to remove a feature, we won't add a deprecation warning until v0.24, and we won't actually remove the feature until v0.25, barring absolute necessity.

This is more strict than what we currently have, but is in line with what React does.

taion:
Fixes #863 

Already reviewed as #897


AlexKVal:
:hand:

PR #899, opened 2015-06-25
Style props tables like Bootstrap options tables

taion:
Before:
image

After:
image

Compare:
image

ref #887

### DIFF
--- a/docs/assets/style.css
+++ b/docs/assets/style.css
@@ -162,10 +162,6 @@ body {
   position: absolute;
 }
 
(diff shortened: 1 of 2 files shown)


jquense:
sweet, looks great! LGTM

jquense:
also reminds me that if folks are comfortable with using [marked](https://github.com/chjj/marked) we could write markdown in the descriptions which is pretty convenient

dozoisch:
Looks way better nice!

PR #913, opened 2015-06-29
Fix a few Modal bugs

jquense:


AlexKVal:
These are already committed
https://github.com/react-bootstrap/react-bootstrap/commit/97ef41537fef7dc0b6a35921279e0124c6f40152
https://github.com/react-bootstrap/react-bootstrap/commit/a8b177aebfd480eecb0c30d8ea0f226e35558634

and they need https://github.com/react-bootstrap/react-bootstrap/pull/914

AlexKVal:
https://github.com/react-bootstrap/react-bootstrap/commit/97ef41537fef7dc0b6a35921279e0124c6f40152#diff-3ddc036af206541151e86f03cfc0c0eeR297
```js
checkForFocus(){
  if ( domUtils.canUseDom ) { // here
    try {
      this.lastFocus = document.activeElement;
    }
    catch (e) {}
  }
},
```
`if ( domUtils.canUseDom ) {` => `if ( domUtils.canUseDom() ) {`

Tests don't cover this. Though it seems very tricky to do.

AlexKVal:
As for: `Modal won't steal focus from children`
now we have some not intuitive pair of `autoFocus` and `enforceFocus` properties.

1) Tests don't cover all 4x possible `autoFocus & enforceFocus` combinations.
2) And it seems need some explaining documentation about their usage and combinations.

Otherwise we will have a decent number of new issues from users :smile:

AlexKVal:
As for: `Stack overflow with nested Modals`
LGTM
Ideally it would be nice to have test case for it to prevent future regressions.

AlexKVal:
Anyway it seems you've done good lump of job :sparkle: :smiley:

jquense:
Sorry for the eslint issues.  my and local npm test is not catching them, so something is wrong there unrelated to this.

in terms of the props, they don't affect each other so testing each combination is unnecessary. They are each tested individually already. Also I didn't really add anything new I just split the functionality into two props. that being said the only reason you might disable these props is to solve one of the bugs this fixes.

jquense:
so this was accidentally merged directly into master by me already as noted above. oops

AlexKVal:
I was wrong about `if ( domUtils.canUseDom ) {` => `if ( domUtils.canUseDom() ) {`
Sorry. :blush:

Issue #920, opened 2015-06-30
textarea's value change to old value when the modal hide.


snadn:
I have a textarea in a modal. if the default value is `a` and i change it to `b`.when the modal hide,the value of textarea become to `a`.

I found that in 'FadeMixin.js', it use `.cloneNode(true)` to get a mirror element, but textarea can not get the new value by this.

dozoisch:
Is your input controlled or not? Can you share a snippet that shows that issue?

AlexKVal:
Superceded by #921

Vyprichenko:
Seems to me I have quite similar issue: for some reason form in modal is recreated (getInitialState method is called) every time when modal is reopened.
Here is an example: http://jsfiddle.net/v5wzkr7e/6/

jquense:
that makes  actually makes sense because the modal unmounts and mounts its content on show and hide.

Vyprichenko:
Unfortunately it makes almost impossible to place components with complex forms in modals, because user will loss all input on casual close.
I see that I'll have to store my form's and modal's states both in common parent, but initially I intended to keep them separated.

jquense:
I mean that's just sort of how modals work... if you want to avoid casual closes, then set the backdrop to static and check if the form has been saved before allowing it to be closed. similar to avoiding accidental page navigations. that's the better UX pattern anyway :)

Vyprichenko:
I've ended up with combined modal + form component containing state of it's children, which probably is better solution in terms of React: http://jsfiddle.net/v5wzkr7e/8/

Real form is quite big and it doesn't have to be filled at once — user is free to work with another parts of page and then go back to the form. I don't have any other reasons to interfere in user's interaction with such a pretty Modal, which is so exciting to open and close  : )

Also static backdrop can be useful in another conditions, thank you.

taion:
Yes, that's about the right approach for persistence - something Flux-like or whatever.

PR #921, opened 2015-06-30
[fixed] fixed issue #920

snadn:
fixed issue #920

[Review on Reviewable](https://reviewable.io/reviews/react-bootstrap/react-bootstrap/921)

### DIFF
--- a/test/FadeMixinSpec.js
+++ b/test/FadeMixinSpec.js
@@ -56,7 +56,7 @@ describe('FadeMixin', function () {
     }, 25);
   });
 
-  it('Should be b of the textarea value', function (done) {
+  it('should retain the same input value throughout transition', function (done) {
     let instance = ReactTestUtils.renderIntoDocument(<Component />);
     let textarea = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'textarea');
 
@@ -69,7 +69,7 @@ describe('FadeMixin', function () {
       let child = element.children[1];
 
       setTimeout(function(){
-        assert.equal(child.value, 'b')
+        assert.equal(child.value, 'b');
         done();
       }, 25);
     }, 25);
--- a/test/FadeMixinSpec.js
+++ b/test/FadeMixinSpec.js
@@ -13,6 +13,7 @@ describe('FadeMixin', function () {
         return (
           <div {...this.props} className='fade'>
             <span className='fade'/>
+            <textarea></textarea>
           </div>
         );
       }
@@ -50,6 +51,25 @@ describe('FadeMixin', function () {
         assert.ok(element.className.match(/\bfade\b/));
         assert.ok(!child.className.match(/\bin\b/));
         assert.ok(child.className.match(/\bfade\b/));
+        done();
+      }, 25);
+    }, 25);
+  });
+
+  it('Should be b of the textarea value', function (done) {
+    let instance = ReactTestUtils.renderIntoDocument(<Component />);
+    let textarea = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'textarea');
+
+    setTimeout(function(){
+
+      React.findDOMNode(textarea).value = 'b';
+
+      instance.componentWillUnmount();
+      let element = instance._fadeOutEl.children[0];
+      let child = element.children[1];
+
+      setTimeout(function(){
+        assert.equal(child.value, 'b')
         done();
       }, 25);
     }, 25);
--- a/src/FadeMixin.js
+++ b/src/FadeMixin.js
@@ -64,7 +64,16 @@ export default {
     if (els.length) {
       this._fadeOutEl = document.createElement('div');
       container.appendChild(this._fadeOutEl);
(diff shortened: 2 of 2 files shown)


jquense:
could you talk a little bit what this does? It looks like you are just trying to maintain the value of the text inputs as the Modal fades out, but the issue seems to be about values not changing when you type.

This sort of seems like overkill for a value that is visible for 300 milliseconds, plus it doesn't seem to address the other input fields..

test/FadeMixinSpec.js
```diff
@@ -54,4 +55,23 @@ describe('FadeMixin', function () {
       }, 25);
     }, 25);
   });
+
+  it('Should be b of the textarea value', function (done) {
```
mtscout6: Can you expound on what "b" is?

snadn: @mtscout6 I was just a random set of values

mtscout6: Something to the effect of `should retain the same input value throughout transition` would be more clear. The test name should describe your intent, not the implementation details.

snadn:
@jquense yes，do this just for the 300 milliseconds.but It allows users to very clearly see this change. and i think it's not good User experience.

At the same time. I do this, because the user feedback this problem.

test/FadeMixinSpec.js
```diff
@@ -13,6 +13,7 @@ describe('FadeMixin', function () {
         return (
           <div {...this.props} className='fade'>
             <span className='fade'/>
+            <textarea></textarea>
```
AlexKVal: Please do not add this here.
It belongs only to one test.
It would be best to place it exactly with that test case :cherries:

test/FadeMixinSpec.js
```diff
@@ -54,4 +55,23 @@ describe('FadeMixin', function () {
...
+    let instance = ReactTestUtils.renderIntoDocument(<Component />);
+    let textarea = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'textarea');
+
+    setTimeout(function(){
+
+      React.findDOMNode(textarea).value = 'b';
+
+      instance.componentWillUnmount();
+      let element = instance._fadeOutEl.children[0];
+      let child = element.children[1];
```
AlexKVal: This is so cryptic, unreadable and brittle :sweat:

Could you please re-write the test such it is all clear what's going on and what we are testing here ?
:cherries:

AlexKVal:
IMHO the provided solution 
```diff
-      this._fadeOutEl.appendChild(React.findDOMNode(this).cloneNode(true));
+
+      let originNode = React.findDOMNode(this);
+      let cloneNode = originNode.cloneNode(true);
+      let originTextareaList = [].concat.apply([], originNode.querySelectorAll('textarea'));
+      let cloneTextareaList = cloneNode.querySelectorAll('textarea');
+      originTextareaList.forEach((item, index) => {
+        cloneTextareaList[index].value = item.value;
+      });
+
+      this._fadeOutEl.appendChild(cloneNode);
```
is too specific. It solves only `textarea` case.
And what is with other types of `input` fields ?
http://www.w3schools.com/tags/tag_input.asp
```
type:
button
checkbox
color
date 
datetime 
datetime-local 
email 
file
hidden
image
month 
number 
password
radio
range 
reset
search
submit
tel
text
time 
url
week
```

I believe it should be some generalized solution. 

Maybe through some property/callback in which a user of `FadeMixin` will be able to write
a custom logic to preserve those values their need and only if them need it.
:cherries:

jquense:
It should be noted that when/if #848 goes through the problem will go away since we won't be cloning the faded elements anymore.

AlexKVal:
Then I presume we can close this PR in favor of #848 ?
because there is a few value of committing the code that would be thrown away. Isn't it ?

AlexKVal:
/cc @react-bootstrap/collaborators 
what do you think guys about https://github.com/react-bootstrap/react-bootstrap/pull/921#issuecomment-118650476 ?

mathieumg:
> Then I presume we can close this PR in favor of #848 ?
> because there is a few value of committing the code that would be thrown away. Isn't it ?

Yep, this can be closed!

AlexKVal:
@snadn Thank you for your wish to help! It is appreciated :cherries: 
I'm closing this in favor of #848

If you still want to help with anything,
you can start here [help wanted](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22) or [bugs](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aopen+is%3Aissue+label%3Abug) as well :rainbow: :umbrella:

snadn:
@AlexKVal This problem has been fixed in the new version of the other methods used, is not it?

AlexKVal:
The new version which will be `v0.24` is not released yet.

AlexKVal:
#848 will go to `v0.24`
screen shot 2015-07-11 at 7 26 45 pm

PR #930, opened 2015-07-01
[fixed] Accidental breaking change in Modal trigger

### LINKED ISSUE #929: ModalTrigger: no backward in minor version changes
You removed ModalTrigger, now all projects with this component are broken. And all this in minor version changes.

jquense:
fixes #929

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -449,7 +449,7 @@ const Modal = React.createClass({
       <ModalMarkup {...props}>{this.props.children}</ModalMarkup>
     );
     // I can't think of another way to not break back compat while defaulting container
-    if ( show != null ){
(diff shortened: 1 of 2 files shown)


taion:
LGTM

What do you think about separating out the "autoFocus/enforceFocus/backdrop" logic into a separate visible component from the modal dialog component itself? Might help with the kinds of patterns per #929.

taion:
Should we release v0.23.7 for this?

mtscout6:
Yes we should

jquense:
Normally I'd say we should also add a test that catches this, but since ModalTrigger being removed in the next major bump maybe its not needed.

PR #939, opened 2015-07-04
fix negative pagenum when items=1 and active next when items=0

firework1230:
Found two bugs when using Pagination component. Happens when ellipsis=true.
Bug1. items=1 and maxButtons=5, display -3,-2,-1,0,1.
Bug2. items=0, next and last buttons  are active.

### DIFF
--- a/test/utils/CustomPropTypesSpec.js
+++ b/test/utils/CustomPropTypesSpec.js
@@ -167,4 +167,21 @@ describe('CustomPropTypes', function() {
       validators[2].should.not.have.been.called;
     });
   });
+
+  describe('isRequiredForA11y', function () {
+    function validate(prop) {
+      return CustomPropTypes.isRequiredForA11y(React.PropTypes.string)({p: prop}, 'p', 'Component');
+    }
+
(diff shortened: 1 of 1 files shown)


jquense:
hi there!

Thanks for taking the time to help us out with this it is appreciated :). I am not the best judge on the content of the PR so I will leave that for other contributors, but I do have a process related issue.

Could you split the commit into separate commits for each of the two bugs you are fixing here, [following these guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes), thanks.

AlexKVal:
In result there should be two commits something like that:
`[fixed] bug in Pagination ...`
`[fixed] another bug in Pagination ...`
Bug descriptions of course are yours.

These `[fixed]` tags allow our release tools automatically grab and add these descriptions into
`Changelog` like this https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md

--
Tests I've already written and will add them after merging this patch.

PR #942, opened 2015-07-06
[changed] Remove Overlay and Modal deprecations

jquense:
remove overlay overhaul deprecations


src/OverlayTrigger.js
```diff
@@ -178,30 +175,24 @@ const OverlayTrigger = React.createClass({
...
-          'Consider also including the `"focus"` trigger so that touch and keyboard only users can see the overlay as well.');
+    if (isOneOf('click', this.props.trigger)) {
+      props.onClick = createChainedFunction(this.toggle, props.onClick);
+    }
 
-        props.onMouseOver = createChainedFunction(this.handleDelayedShow, this.props.onMouseOver);
-        props.onMouseOut = createChainedFunction(this.handleDelayedHide, this.props.onMouseOut);
-      }
+    if (isOneOf('hover', this.props.trigger)) {
+      warning(!(this.props.trigger === 'hover'),
```
mathieumg: `this.props.trigger !== 'hover'`?

jquense: Ya it's probably just a weird artifact from a refactoring...

jquense:
anyone have any issues with this? should probably go in before #848

AlexKVal:
Sorry. I have nothing to comment. It is too much code complexity for me yet.
As for the code itself (not the algo) I'm fine with it.

jquense:
There is nothing "new" in this PR it just removes the old api and warnings for Overlays and Modals, I would think that passing CI is probably enough of a review? More then happy to have others look over it tho :)

mathieumg:
:+1:

src/OverlayTrigger.js
```diff
@@ -178,30 +175,24 @@ const OverlayTrigger = React.createClass({
...
+        '[react-bootstrap] Specifying only the `"hover"` trigger limits the visibilty of the overlay to just mouse users. ' +
+        'Consider also including the `"focus"` trigger so that touch and keyboard only users can see the overlay as well.');
```
AlexKVal: those who will use it only for `mouse`-users.. will be annoyed and then one of two:
1) they will complain about it on issue trackers, or
2) they will just add focus to suppress this warning

maybe better to move this warning into `props.trigger` doc-definition ?

src/Modal.js
```diff
@@ -226,28 +193,12 @@ const ModalMarkup = React.createClass({
...
-        <Header closeButton={this.props.closeButton} onHide={this._getHide()}>
-          { this.props.title &&
-            <Title>{this.props.title}</Title>
-          }
-        </Header>
-      );
-    }
 
-    return React.Children.map(children, child => {
+    return React.Children.map(this.props.children, child => {
```
AlexKVal: why don't you use `mapValidComponents()`?
[src/utils/ValidComponentChildren.js#L15](https://github.com/react-bootstrap/react-bootstrap/blob/02a8764b196b9268678ded26838a2436a54a2090/src/utils/ValidComponentChildren.js#L15)

src/Portal.js
```diff
@@ -13,20 +13,81 @@ let Portal = React.createClass({
...
+    // Save reference for future access.
+    if (overlay !== null) {
+      this._mountOverlayTarget();
+      this._overlayInstance = React.render(overlay, this._overlayTarget);
+    } else {
+      // Unrender if the component is null for transitions to null
+      this._unrenderOverlay();
+      this._unmountOverlayTarget();
     }
+  },
```
AlexKVal: simplify:
```js
_renderOverlay() {
  // Save reference for future access.
  if (this.props.children) {
    this._mountOverlayTarget();
    const overlay = React.Children.only(this.props.children);
    this._overlayInstance = React.render(overlay, this._overlayTarget);
  } else {
    // Unrender if the component is null for transitions to null
    this._unrenderOverlay();
    this._unmountOverlayTarget();
  }
},
```

src/Portal.js
```diff
@@ -13,20 +13,81 @@ let Portal = React.createClass({
...
+    if (this._overlayInstance) {
+      if (this._overlayInstance.getWrappedDOMNode) {
+        return this._overlayInstance.getWrappedDOMNode();
+      } else {
+        return React.findDOMNode(this._overlayInstance);
+      }
+    }
+
+    return null;
+  },
```
AlexKVal: just proposal:
```js
getOverlayDOMNode() {
  if (!this.isMounted()) {
    throw new Error('getOverlayDOMNode(): A component must be mounted to have a DOM node.');
  }

  if (!this._overlayInstance) {
    return null;
  }

  if (this._overlayInstance.getWrappedDOMNode) {
    return this._overlayInstance.getWrappedDOMNode();
  }

  return React.findDOMNode(this._overlayInstance);
},
```

test/ModalSpec.js
```diff
@@ -1,18 +1,31 @@
...
+    React.unmountComponentAtNode(mountPoint);
+    document.body.removeChild(mountPoint);
+  });
 
   it('Should render the modal content', function() {
     let noOp = function () {};
-    let instance = ReactTestUtils.renderIntoDocument(
-      <Modal onHide={noOp}>
+    let instance = render(
+      <Modal show onHide={noOp} animation={false}>
```
AlexKVal: heh... I missed `<Modal />` addition..
I would propose `disableAnimation` instead of `animation`,
then use could just write it as:
```js
<Modal show onHide={noOp} disableAnimation>
```

though, this isn't really significant

test/ModalSpec.js
```diff
@@ -27,21 +40,30 @@ describe('Modal', function () {
...
-    let instance = ReactTestUtils.renderIntoDocument(
-        <Container />
-    );
+    let instance = render(
+          <Container />
+        , mountPoint);
+
+    let modal = ReactTestUtils.findRenderedComponentWithType(instance, Modal);
+
     assert.ok(React.findDOMNode(instance).className.match(/\modal-open\b/));
```
AlexKVal: the old one, unnoticed semi-bug :smile: 
`/\modal-open\b/` => `/\bmodal-open\b/`

src/OverlayTrigger.js
jquense: this is already in the code base and no one has yet complained. they should just add focus to suppress the warning, that's the point of the warning, to encourage practices that serve all types of users

src/Portal.js
```diff
@@ -13,20 +13,81 @@ let Portal = React.createClass({
...
+      this._overlayInstance = null;
+    }
   },
 
   render() {
     return null;
+  },
+
+  getOverlayDOMNode() {
+    if (!this.isMounted()) {
```
mathieumg: I thought this didn't work in ES6 components?

jquense:
Folks I think the bikeshedding on this PR is getting a bit out of hand. The scope of this PR is simply to remove deprecated files, and warnings that were adding in the Modal Overhaul release.

If folks want to refactor or review the code from that PR more, lets just do it in a separate PR.

taion:
Yes, and this is absolutely a v0.24.0 blocker - let's get this merged.

src/Portal.js
jquense: good catch, i just copied it from the OverlayMixin. fixed now

jquense:
any PR specific comments or are we gtg?

src/Modal.js
taion: @jquense

src/OverlayTrigger.js
taion: I think the warnings are usually written this way for clarity.

src/Portal.js
taion: I like the way @AlexKVal has it but it doesn't really matter.

taion:
I scanned it real quick - just that one comment on mapping only valid children, and we should be good to go.

src/Portal.js
mathieumg: I only looked at the partial diff and saw `getOverlayDOMNode() {`, I always forget about shorthand function properties...

src/Modal.js
jquense: I didn't use it originally because I didn't know about it, and I didn't change it here b/c I wasn't trying to fix bugs, especially since #964 is already handling it.

If we want to roll that into here let me know

taion:
Sorry thought we were addressing that issue in this PR. This is good to merge then. LGTM.

src/Modal.js
jquense: Final though on that actually, mapValidChildren will strip out text nodes which is not the right behavior here.

Issue #99, opened 2014-06-10
Popovers and Tooltips not showin up properly on svg elements


ptzolov:
I'm using this library with [om](https://github.com/swannodette/om), and the interop is quite nice. The only issue I've come across is is that popovers don't show up centered on the svg element.

example:

```
(defn offnet-segment
  [app owner]
  (reify
    om/IRenderState
    (render-state [this {:keys [translate scale width control-chan
                                rx ry cursor title data-content]}]
      (dom/g #js {:stroke "black"
                  :transform (str "translate(" translate ") scale(" scale ")")
                  :width width}
             (ReactBootstrap/OverlayTrigger #js {:trigger "click"
                                                 :placement "top"
                                                 :overlay (ReactBootstrap/Popover #js {:title "ID"}
                                                                                  "DESCRIPTION")}
                                            (dom/g nil
                                                   (dom/rect #js {:className "path-element"
                                                                  :x 600
                                                                  :y 120
                                                                  :rx rx
                                                                  :ry ry
                                                                  :width width
                                                                  :height 70
                                                                  :fill "#0c60b7"})
                                                   (dom/text #js {:className "ne-text submit-button"
                                                                  :x (+ 600 (/ width 2))
                                                                  :y 175
                                                                  :fill "white"} "Test")))))))
```
Example: 

The popover should be centred over the svg rectangle. 
![image](https://cloud.githubusercontent.com/assets/2257349/3231413/687d997c-f0ab-11e3-9371-64ae0170b7f7.png)

taion:
Superseded by #1050.

PR #955, opened 2015-07-07
Fix for Tooltip position in Firefox on SVG's

petesaia:
Possibly related to #99 

The `offsetWidth` and `offsetHeight` were coming in `undefined` thus resulting in a calculated `NaN`. Resolving them to 0 did the trick.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ npm-debug.log
 node_modules
 amd/
 !tools/amd/
-lib/
 !tools/lib/
 dist/
 !tools/dist/
--- a/lib/Accordion.js
+++ b/lib/Accordion.js
@@ -1,0 +1,32 @@
+'use strict';
+
+Object.defineProperty(exports, '__esModule', {
+  value: true
+});
+
+var _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; };
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { 'default': obj }; }
+
+var _react = require('react');
+
+var _react2 = _interopRequireDefault(_react);
+
+var _PanelGroup = require('./PanelGroup');
+
+var _PanelGroup2 = _interopRequireDefault(_PanelGroup);
+
+var Accordion = _react2['default'].createClass({
+  displayName: 'Accordion',
+
+  render: function render() {
+    return _react2['default'].createElement(
+      _PanelGroup2['default'],
+      _extends({}, this.props, { accordion: true }),
+      this.props.children
+    );
+  }
+});
+
+exports['default'] = Accordion;
+module.exports = exports['default'];
--- a/lib/Affix.js
+++ b/lib/Affix.js
@@ -1,0 +1,50 @@
+'use strict';
+
+Object.defineProperty(exports, '__esModule', {
+  value: true
+});
+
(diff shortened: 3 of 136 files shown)


jquense:
looks good.  please squash your commits and follow the commit guidelines here: https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md and we are good to go.  thanks for helping out :)

AlexKVal:
Do we need some test to prevent future regressions ?

taion:
Ideally, but in this case I'd be okay with adding a comment unless you want to set up Travis to also use Firefox.

AlexKVal:
I've tried Firefox tests. 
There are a lot of work to do first. Many tests are failing.
At least it was true 1.5 months ago.
It is on my ToDo list, buy with a low priority.

Then I'm fine with just a comment :smile: 

But a squashing is definitely needed :sweat:

taion:
Okay. @petesaia: can you add a comment explaining why this is necessary (maybe with a TODO to add a test case to cover this), and squash your commits? We'll deal with setting up CI on FF later.

petesaia:
So after looking at this issue again with a clearer head, I'm a little confused. The problem is happening because `offsetWidth` and `offsetHeight` are assumed to always be available on an element. A way of fixing this would be to use `getBoundingClientRect()`.

With that said, it looks like `getOffset()` and `getPosition()` are doing this properly (unless jquery exists). [They are also being called within the same function(s)](https://github.com/petesaia/react-bootstrap/blob/dac93ae652ba5ccc444d0b45e035e26086828e5e/src/utils/overlayPositionUtils.js#L25), yet `offsetWidth` and `offsetHeight` are still being used after the fact. Why not just use the values from `getOffset()` and `getPosition()`? This might be a good time to [remove jquery](https://github.com/petesaia/react-bootstrap/blob/dac93ae652ba5ccc444d0b45e035e26086828e5e/src/utils/domUtils.js#L56).

I squashed my commits, but i'd just recommend using the values from these methods so logic is more centralized. What do you think?

taion:
We use jQuery for IE8 compat - same reason I don't use GBCR there, because it doesn't return width/height on IE8.

taion:
@petesaia Can you add the comment and squash your commits?

petesaia:
@taion I hate to sound so noobish, but could you explain how else you'd like me to squash the commits and where to comment? So far, i've done a `git rebase -i HEAD~4` for my four commits and merged (6e476fe). As detailed as [CONTRIBUTING.md](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) is, I don't see anything about this, specifically. Sorry!

taion:
Don't do the merge yourself - when I click the "merge" button, GH will do it for me.

What you need to do is to do `git rebase -i master` and then choose `s` or `f` for all the commits after the first, fix the commit comment, then do `git push -f`.

Add the comment where you made your changes.

petesaia:
https://github.com/petesaia/react-bootstrap/commit/e9858cea7bd6868081d5f6136164776e5f957658

taion:
No worries. Looking good. Can you just add that comment around there real quick explaining why this is necessary?

jquense:
@petesaia I think he means add a comment to the code itself, not just on the PR diff.

petesaia:
Done.

taion:
Gahhh merge conflict. Mind rebasing and re-submitting? This looks good to go otherwise.

petesaia:
I've rebased, but it's still throwing a:

>  Can’t automatically merge. Don’t worry, you can still create the pull request.

Any ideas?

taion:
You probably need to do: https://help.github.com/articles/syncing-a-fork/

petesaia:
Sent a new one.

mtscout6:
For future reference as well if you don't want to close a PR just to open a new one you can use git to force push your branch after a rebase. That will update the PR with the new branch automatically. Force push can be problematic in some cases which is why some people are weary of it, but it is useful in cases like this.

petesaia:
@mtscout6 That's certainly good to know, thanks.

PR #964, opened 2015-07-09
Fix null children

frodsan:
This fixes the following case:

```javascript
  getModalFooter() {
    if (something) return null;

    return <ModalFooter> ... </ModalFooter>;
  }

  render() {
    // ...
      <Modal>
          <Modal.Header closeButton>
            <Modal.Title id='contained-modal-title'>{ this.props.title }</Modal.Title>
          </Modal.Header>

          <Modal.Body>
            { this.getModalBody() }
          </Modal.Body>

          { this.getModalFooter() }
      <Modal>
  }
```

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -245,7 +245,7 @@ const ModalMarkup = React.createClass({
 
     return React.Children.map(children, child => {
       // TODO: use context in 0.14
-      if (child.type.__isModalHeader) {
+      if (child && child.type.__isModalHeader) {
         return cloneElement(child, {
           onHide: createChainedFunction(this._getHide(), child.props.onHide)
         });


jquense:
Hi there, thanks for putting this together.  Question. Why would you ever have an empty modal?  in that case it seems lile just not rendering the modal is the better solution

frodsan:
I updated the use case.

jquense:
so to probably check I think we also need to check that `.type` exists as well since strings could be children. 

that and adding test case and we are good to go I think

jquense:
O and please follow the guidelines for commit messages here: https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md

AlexKVal:
/cc @react-bootstrap/collaborators 
The `mapValidComponents()` [utils/ValidComponentChildren.js#L15](https://github.com/react-bootstrap/react-bootstrap/blob/02a8764b196b9268678ded26838a2436a54a2090/src/utils/ValidComponentChildren.js#L15) helper exists exactly to prevent such cases and is used instead of `React.Children.map()`

See
[src/CollapsibleNav.js#L54](https://github.com/react-bootstrap/react-bootstrap/blob/55802fdf7dc7c0bb83ded1cacfebd57dab5a2377/src/CollapsibleNav.js#L54)
[src/ListGroup.js#L44](https://github.com/react-bootstrap/react-bootstrap/blob/fef89847c5e9911e26c6ae4ce003dd25612ce587/src/ListGroup.js#L44)
etc. 
https://github.com/react-bootstrap/react-bootstrap/search?utf8=%E2%9C%93&q=ValidComponentChildren.map
:cherries:

jquense:
superceded by #1040

Issue #935, opened 2015-07-03
ListGroup TypeError: Cannot read property 'href' of undefined when children is Immutable iterable


cassus:
From v13 React supports _iterators and immutable-js sequences as children._ (https://facebook.github.io/react/blog/2015/03/10/react-v0.13.html)

ListGroup fails with `TypeError: Cannot read property 'href' of undefined` when the children is an iterable object.

Example:
```jsx
<ListGroup>
            { Immutable.List(['foo']).map((foo) =>
                    <ListGroupItem>
                        {foo}
                    </ListGroupItem>
            )}
        </ListGroup>
```

Workaround: add `.toJS()` after the map

AlexKVal:
Could you please provide reproducible code snippet, then we can work with ?

Otherwise it is not clear what the problem is and how it should be addressed :sweat: 

Thank you :cherries:

apkiernan:
I tried mocking something up to see what the issue is. The issue appears to be that the use of `Immutable.List.map` returns as the children to `ListGroup` an object containing ImmutableJS List objects, meaning they are not React components which are expected. I think the quote about supporting iterables and immutable sequences refers to the ability to map arrays and such from inside the children of a component which would then return React components. 

The workaround I found for this was to use Immutable's `toArray()` method before the `map` call, then use the map method as usual:

```
const Foo = React.createClass({
  render() {
    return (
      <ListGroup>
        {Immutable.List(['foo']).toArray().map((foo) => {
          return (
            <ListGroupItem>
              {foo}
            </ListGroupItem>
          );
        })}
      </ListGroup>
    );
  }
});
```

This is the same even if using generic components such as spans and divs. The result of the Immutable sequence needs to be React component(s), not just the Immutable object.

Please let me know if I'm missing something or am off base, and I will look into it further.

jquense:
@apkiernan you are on track, react supporting iterates is not the same thing as supporting immutable js. if the sequence is not one of ReactElements then react can't render them, even tho it can iterate through it.

cassus:
I don't see how Immutable.List.Map returns not React components when iterated.

```js
const iterator = Immutable.List(['foo', 'bar']).map((foo) => {
    return (
        <ListGroupItem>
            {foo}
        </ListGroupItem>
    )
})

for(const foo of iterator) {
    console.log(foo)
}
```

Logs
```
[...]

PR #970, opened 2015-07-10
[added] ListGroup supports iterator as child

cassus:
Closes https://github.com/react-bootstrap/react-bootstrap/issues/935

### DIFF
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -13,17 +13,11 @@ class ListGroup extends React.Component {
 
     if (!this.props.children) {
       return this.renderDiv(items);
-    } else if (React.Children.count(this.props.children) === 1 && !Array.isArray(this.props.children)) {
-      let child = this.props.children;
-
-      childrenAnchors = this.isAnchor(child.props);
-
     } else {
-
-      childrenAnchors = Array.prototype.some.call(this.props.children, (child) => {
-        return !Array.isArray(child) ? this.isAnchor(child.props) : Array.prototype.some.call(child, (subChild) => {
-            return this.isAnchor(subChild.props);
-        });
+      React.Children.forEach(this.props.children, (child) => {
+        if (this.isAnchor(child.props)) {
+          childrenAnchors = true;
+        }
 
       });
 
--- a/test/ListGroupSpec.js
+++ b/test/ListGroupSpec.js
@@ -147,4 +147,22 @@ describe('ListGroup', function () {
     assert.equal(React.findDOMNode(instance).id, 'testItem');
   });
 
-});
+  it('Should support an iterator as child', function() {
+    let childArray = [<ListGroupItem key={42}>Only Child in array</ListGroupItem>];
+    let childIterator = {};
+
+    childIterator[Symbol.iterator] = childArray[Symbol.iterator].bind(childArray);
+
+    let instance = ReactTestUtils.renderIntoDocument(
+      <ListGroup>
+        {childIterator}
+      </ListGroup>
+    );
+
+    let items = ReactTestUtils.scryRenderedComponentsWithType(instance, ListGroupItem);
+
+    assert.ok(ReactTestUtils.findRenderedDOMComponentWithClass(items[0], 'list-group-item'));
+  });
+
+})
+;


src/ListGroup.js
```diff
@@ -13,17 +13,15 @@ class ListGroup extends React.Component {
...
+      React.Children.forEach(this.props.children, (child) => {
+        if (!Array.isArray(child) && this.isAnchor(child.props)) {
+          childrenAnchors = true;
+        }
 
-      childrenAnchors = Array.prototype.some.call(this.props.children, (child) => {
-        return !Array.isArray(child) ? this.isAnchor(child.props) : Array.prototype.some.call(child, (subChild) => {
-            return this.isAnchor(subChild.props);
-        });
+        if (Array.isArray(child) && child.some.call((subChild) => this.isAnchor(subChild.props))) {
```
taion: Is this needed to support iterators? Seems weird to do the `.some` call here when you're already doing `forEach` above.

cassus: I didn't add it initially but other test cases describe this behavior.

taion: `React.Children.forEach` deals with that for you. You can simplify this to just

```js
      React.Children.forEach(this.props.children, (child) => {
        if (this.isAnchor(child.props)) {
          childrenAnchors = true;
        }
      });
```

and the tests still pass.

jquense: the original code does seem to test two layers down, why just two? @cassus that probably means your `forEach` code needs to be pulled out into a separate function so it can be called recursively

taion: It's for this test case: https://github.com/react-bootstrap/react-bootstrap/blob/v0.23.7/test/ListGroupSpec.js#L77-L89

But `React.Children.forEach` unrolls that for you so you don't need to manually nest the iterations any more.

jquense: Wait `Children.forEach` is recursive? really? I could have sworn it wasn't...

taion: In the general case probably not, but per https://facebook.github.io/react/docs/top-level-api.html#react.children.map:

>If `children` is a nested object or array it will be traversed: `fn` will never be passed the container objects.

And anyway the tests cover this and they still pass with the snippet I have above.

cassus: @taion thx for finding this simpler implementation. I added it to the pull req.

AlexKVal:
I restarted CI-es after fixing `eslint-warning`. 
There is some bug with restarting on `travis-ci` now.
But `appveyor` now is green.

--
LGTM. :sparkle: 

Thank you @cassus :cherries:

cassus:
Well Symbol was defined locally when I ran the tests. I'm not sure how to add the polyfill / something else to the test runner.

jquense:
I _believe_ it's enough to use the string '@@Iterator' which isn't real es6 but how react checks when symbols aren't present. not sure why   why it's undefined for Travis thougj

taion:
Oh that's really weird.

Also, can you squash your two commits?

AlexKVal:
It seems the issue is with the Chrome version used by travis `INFO [Chromium 37.0.2062 (Ubuntu 0.0.0)]`
because our desktop versions of Chrome are near `Version 43.0.2357.132`
I've checked it with exact version of `iojs-v2.3.4`

AlexKVal:
It seems another way to test this has to be found.

(otherwise we should ask Travis to upgrade their Chomium :smile:)

taion:
On v0.24 we have babel-runtime as a dependency, which would polyfill `Symbol` for us.

We'd want to be careful not to use `Symbol` in the release build, but I'd be okay with having that polyfill for tests.

jquense:
maybe we hold off merging until 0.24 is merged in? that really needs to be the next thing we do anyway?

taion:
Agreed on that.

AlexKVal:
Add it to `0.25 Release` milestone. Right ?

taion:
Nah, I'd pull this in for v0.24.0 -  we'll just merge in the v0.24-rc branch first, then rebase this on top of that, and the tests should pass.

AlexKVal:
Ah.. Ok. I changed it to `v0.24` milestone.

taion:
One more thing - we should probably only iterate over valid children, with something like https://github.com/react-bootstrap/react-bootstrap/blob/02a8764b196b9268678ded26838a2436a54a2090/src/utils/ValidComponentChildren.js#L15

cassus:
Commits squashed, waiting for v0.24-rc merge.

taion:
Okay great - it looks like the `null` child case I mentioned above isn't currently handled correctly anyway, so I'm okay with this as-is. We can add handling + a test case later, if everyone else is okay with that.

cassus:
@taion this sounds like a potential bug that existed in the original version as well. I don't want to overload this pull request with adding a fix and test for something else.

taion:
Hah, crossed comments again. Jinx.

AlexKVal:
> One more thing - we should probably only iterate over valid children, with something like https://github.com/react-bootstrap/react-bootstrap/blob/02a8764b196b9268678ded26838a2436a54a2090/src/utils/ValidComponentChildren.js#L15


Pff.. I posted about this here today, but cannot see where it is. It seems I had an internet connection failure :disappointed: 

But it is nice you noted about it @taion :+1: :smile:

taion:
I'm going to close and hijack this PR to fix the test - turns out babel-runtime doesn't polyfill the iterator stuff on ES6 arrays, for whatever reason.

taion:
Hijacked to #998 so I could fix the test and merge.

PR #979, opened 2015-07-12
[added] ModalHeader custom closeButton

### LINKED ISSUE #956: Modal custom close button
It would be great to be able to use a custom close button for Modal instead of being forced to use this markup:
```
<button type="button" className="close" aria-hidden="true" onClick={this.props.onRequestHide}>&times;</button>
```

Possibly closeButton could accept both a boolean and a react component? Or whatever API is elegant.

BurntCaramel:
Resolves #956

### DIFF
--- a/src/ModalHeader.js
+++ b/src/ModalHeader.js
@@ -4,24 +4,35 @@ import classnames from 'classnames';
 
 class ModalHeader extends React.Component {
 
+  renderCloseButton() {
+    if (this.props.closeButton === true) {
+      return (
+        <button
+          className='close'
+          aria-label={this.props['aria-label'] || 'Close'} //eslint-disable-line react/prop-types
+          onClick={this.props.onHide}
+          style={{ marginTop: -2 }}
+        >
+          <span aria-hidden="true">
+            &times;
+          </span>
+        </button>
+      );
+    }
+    else if (React.isValidElement(this.props.closeButton)) {
+      return React.cloneElement(this.props.closeButton, {
+        onClick: this.props.onHide
+      });
+    }
+  }
+
   render() {
     return (
       <div
         {...this.props}
         className={classnames(this.props.className, this.props.modalClassName)}
       >
-        { this.props.closeButton &&
-          <button
-            className='close'
-            aria-label={this.props['aria-label'] || 'Close'} //eslint-disable-line react/prop-types
-            onClick={this.props.onHide}
-            style={{ marginTop: -2 }}
-          >
-            <span aria-hidden="true">
-              &times;
-            </span>
-          </button>
-        }
+        { this.renderCloseButton() }
         { this.props.children }
       </div>
     );
@@ -39,7 +50,10 @@ ModalHeader.propTypes = {
   /**
    * Specify whether the Component should contain a close button
    */
-  closeButton: React.PropTypes.bool,
+  closeButton: React.PropTypes.oneOfType([
+    React.PropTypes.bool,
+    React.PropTypes.element
+  ]),
   /**
    * A Callback fired when the close button is clicked. If used directly inside a Modal component, the onHide will automatically
    * be propagated up to the parent Modal `onHide`.
--- a/src/ModalHeader.js
+++ b/src/ModalHeader.js
@@ -48,7 +48,7 @@ ModalHeader.propTypes = {
    */
   modalClassName: React.PropTypes.string,
   /**
-   * Specify whether the Component should contain a close button
+   * Specify whether the Component should contain a close button, or alternatively specify a custom ReactElement to use
    */
   closeButton: React.PropTypes.oneOfType([
     React.PropTypes.bool,
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -220,6 +220,24 @@ describe('Modal', function () {
     });
   });
 
+  it('Should allow a custom close button', function (done) {
+    let doneOp = function () { done(); };
+
+    const customCloseButton = (
+      <button className='custom-close-button'>Close</button>
+    );
+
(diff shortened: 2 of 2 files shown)


AlexKVal:
LGTM :sparkle: 

You can squash the commits and change the commit message as prescribed in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) 
Doing so will allow for our automated changelog tool to document this on the next release.

Smth like that:
`[added] Allow ModalHeader’s closeButton to take a ReactElement`

Need one more approval from a member of @react-bootstrap/collaborators

:cherries:

src/ModalHeader.js
```diff
@@ -4,24 +4,35 @@ import classnames from 'classnames';
 
 class ModalHeader extends React.Component {
 
+  renderCloseButton() {
+    if (this.props.closeButton === true) {
```
mathieumg: `if(this.props.closeButton)`

src/ModalHeader.js
```diff
@@ -4,24 +4,35 @@ import classnames from 'classnames';
...
+      return (
+        <button
+          className='close'
+          aria-label={this.props['aria-label'] || 'Close'} //eslint-disable-line react/prop-types
+          onClick={this.props.onHide}
+          style={{ marginTop: -2 }}
```
mathieumg: That isn't in the original Bootstrap code, what causes the discrepancy so we need to do this correction here?

src/ModalHeader.js
```diff
@@ -4,24 +4,35 @@ import classnames from 'classnames';
...
+        >
+          <span aria-hidden="true">
+            &times;
+          </span>
+        </button>
+      );
+    }
+    else if (React.isValidElement(this.props.closeButton)) {
```
mathieumg: I imagine `this.props.children` doesn't know about `this.props.onHide`? If it somehow does, I don't think this `else` would be necessary. If it doesn't, I'd pass down `this.props.onHide` as `onHide` and not as `onClick` for all we know that custom "close button" could be a drop-down with options.

Another way would be to sniff whether `this.props.children` is alone or there are many, and pass it (or to the first) the `onHide`, but I'm not a huge fan of that approach.

src/ModalHeader.js
jquense: that will give a false positive for the new case where it's an element won't it?

test/ModalSpec.js
```diff
@@ -220,6 +220,24 @@ describe('Modal', function () {
     });
   });
```
mathieumg: Can you add a test for the case when a `closeButton` is requested but no custom element is provided? (Verify that it displays the default close button)

src/ModalHeader.js
jquense: I think this was accidentally cargo culled in from my modal change. it should be removed

mathieumg:
Good work! I put down some remarks, maybe @jquense can chime in, he's the architect behind the whole `Modal` refactor. :smile:

src/ModalHeader.js
mathieumg: Whoop's you're right.

src/ModalHeader.js
mathieumg: Whoop's, you're right.

src/ModalHeader.js
AlexKVal: The initial author proposition was to make it possible to customize just  `a Button`.
And this PR solves exactly it :wink: 

I think that all other edge cases are a good candidate for another issue / PR
:cherries:

test/ModalSpec.js
AlexKVal: Good point. It will increase [coveralls.io](https://coveralls.io/github/react-bootstrap/react-bootstrap), which has been suffered a little last time :smile: 
[Coverage Status](https://coveralls.io/github/react-bootstrap/react-bootstrap?branch=master)

src/ModalHeader.js
mathieumg: I only asked to change `onClick` to `onHide` since we don't know the intent "below" and want to keep the highest flexibility possible.

jquense: yeah I agree I  the case of a custom element it should just pass onHide as onHide, instead of as onClick.

jquense:
So it occurs to me that this is actually already possible with the current API....

```
<Modal.Header>
  <button className='customClose' onClick={closeModal}>CLOSE</button>
  <Modal.Title>hey</Modal.Title>
</Modal.Header>
```

The only real advantage to rolling it into `ModalHeader` I can see is if we do pass `onHide` as `onClick` so you don't need to wire up a dom element in custom component, which seems much more verbose then just doing the above...

AlexKVal:
> So it occurs to me that this is actually already possible with the current API

Ah-ha-ha.. :smile:  :+1: 

Then it seems it is just a lack of a good example for this case in the documentation :wink: 
:cherries:

BurntCaramel:
@jquense What does closeModal do, does it rely on a ref to get to the Modal?

Yes I think best practices should be in the documentation, happy to help out.

taion:
@jquense Should we merge this or document your solution?

taion:
@jquense ping

jquense:
I think we should just document my solution for now

taion:
Makes sense.

@BurntCaramel `closeModal` would be whatever you'd passed in as `onHide` to the `Modal` itself. Just call that thing directly.

taion:
This is actually already somewhat documented, as it's how we handle the "close" button in the footer of the existing modal example.

Thank you for the PR, though, and sorry for not accepting it at the time.

BurntCaramel:
No worries, thanks to all for your time.

PR #1013, opened 2015-07-17
Ensure tooltips are shown in TooltipBasic

### LINKED ISSUE #1012: Docs 0.24. Tooltips first example doesn't show.
http://localhost:4000/components.html#tooltips

screen shot 2015-07-17 at 6 58 18 pm

taion:
Closes #1012

### DIFF
--- a/docs/examples/TooltipBasic.js
+++ b/docs/examples/TooltipBasic.js
@@ -1,17 +1,18 @@
 const tooltipInstance = (
   <div>
-    <Tooltip placement="right">
+    <Tooltip placement="right" className="in">
       Tooltip right
     </Tooltip>
-    <Tooltip placement="top">
+
+    <Tooltip placement="top" className="in">
       Tooltip top
     </Tooltip>
 
-    <Tooltip placement="left">
+    <Tooltip placement="left" className="in">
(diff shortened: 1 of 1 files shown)


taion:
@jquense Granted this is a contrived example, but it feels a little silly - is there a benefit I don't see to doing https://github.com/react-bootstrap/react-bootstrap/blob/86b0836778d48465a9246387e60473ac81bdeabc/src/Overlay.js#L73 as opposed to using an e.g. `in` or `show` prop on the overlaid item?

AlexKVal:
I think this is the right way to show it in the example:
```jsx
<Tooltip placement="right" className="in">
```
By default it doesn't have to be shown.

LGTM

--
Maybe it is worth to add some clarifying comment about why `className="in"` is added in the example ?
To prevent silly questions about it. I'm not sure. Just a thought.

taion:
It's silly though - there's no good reason to have a tooltip that's shown all the time. This is enough to get the point across anyway.

jquense:
I might be missing what your asking, that branch is there for when animation is `false` you need to ensure the `in` class is added so it shows up. since the twbs css has them hidden by default

PR #1016, opened 2015-07-17
[fixed] Portal doesn't mount extra node

### LINKED ISSUE #990: Tooltip in Modal adds permanent divs to DOM.
It seems that when a `Modal` is opened or closed, one or two `div` elements are added to the `body` element. 

This can be reproduced using the "Live Demo" in the [Modal components example](http://react-bootstrap.github.io/components.html#modals); If you observe the Inspector in Firefox or the Elements view in Chrome, you will see a `<div></div>` is added every time you open and close it.

Note that while the online demo adds a `div` every time I open *and* close the modal, in my local tests 2 'div's are added every time I close the modal.

jquense:
fixes #990

### DIFF
--- a/src/Portal.js
+++ b/src/Portal.js
@@ -24,7 +24,7 @@ let Portal = React.createClass({
 
   componentWillUnmount() {
     this._unrenderOverlay();
-    this._mountOverlayTarget();
+    this._unmountOverlayTarget();
   },
 
   _mountOverlayTarget() {


taion:
That's not quite right - it needs to actually unmount the overlay target.

jquense:
Oops moving too fast. will fix

taion:
LGTM

Issue #825, opened 2015-06-13
Remove Factory support


taion:
The factory generation in the build process only generates factories for top-level public components.

That said, does anybody still use those factories, or can we drop them altogether?

tacomanator:
Will try to look into addressing this week.

AlexKVal:
@tacomanator is there any progress on it ?

It seems we will have to drop `factories support` if there is no additional interest of course.

Because there will be more namespaced components.

tacomanator:
I modified the factory generation to be recursive to handle the nested components, but so far I haven't found a good test for filtering out the non-component properties of the indexes. The current check for a instance of a React constructor has false positives. I'll keep looking for something.

AlexKVal:
@tacomanator is there any progress on it ?

mtscout6:
This is starting to hit me as well with #526, which is impeding my progress on it. I was on the fence about these factories when we first brought them in, and now I'm somewhat regretting it. @tacomanator I understand the convenience this provides you, but it is becoming troublesome to maintain. I'd rather propose you use your wrapper library as you were before and drop this support from this library.

tacomanator:
Yes I understand. I've been a little busy preparing for a several month trip abroad and haven't been able to look into it further. Please do what you need to do, and I will either submit another PR or create a wrapper/fork as appropriate as soon as I can. Sorry about the delay.

AlexKVal:
Closed by #1019 and #1020

PR #1019, opened 2015-07-19
[changed] Add deprecation warning that factories will be removed

mtscout6:
For details see #825 

Since we haven't pushed 0.24 yet and since that will break the Form Static component anyways, I'd like to get this one in as part of v0.24.


AlexKVal:
LGTM.
I'll make them safe.

Issue #977, opened 2015-07-11
Enhance prop docs generation for repeated properties


AlexKVal:
start: https://github.com/react-bootstrap/react-bootstrap/pull/974#issuecomment-120647630

> @jquense the most obvious solution I can think of just to have an object holding default messages somewhere that the prop table can read from if the desc is empty.

taion:
Handled by https://github.com/react-bootstrap/react-bootstrap/blob/v0.28.0/docs/src/defaultPropDescriptions.js now.

PR #1024, opened 2015-07-20
[fixed] Only calculate overlay position on display

### LINKED ISSUE #1018: Transition/Position Nesting on Overlays
Placeholder issue for some final v0.24 fixes.

taion:
Fixes #1018

This restores the old behavior where we only calculate position initially (and when the target changes).

I also ended up modifying the `Transition` implementation a bit to make the state tracking a bit more explicit.

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -118,18 +118,18 @@ class Collapse extends React.Component {
 
 Collapse.propTypes = {
   /**
-   * Whether the component is expanded
+   * Show the component; triggers the expand or collapse animation
    */
   in: React.PropTypes.bool,
 
   /**
-   * Whether the component should be unmounted (removed from DOM) when
-   * collapsed
+   * Unmount the component (remove it from the DOM) when it is collapsed
    */
   unmountOnExit: React.PropTypes.bool,
 
   /**
-   * Whether the component should expand after mounting
+   * Run the expand animation when the component mounts, if it is initially
+   * shown
    */
   transitionAppear: React.PropTypes.bool,
 
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -113,57 +113,61 @@ class Collapse extends React.Component {
   }
 }
 
+// Explicitly copied from Transition for doc generation.
+// TODO: Remove duplication once #977 is resolved.
+
 Collapse.propTypes = {
   /**
-   * Whether the component is entered; triggers the enter or exit animation
+   * Whether the component is expanded
    */
   in: React.PropTypes.bool,
 
   /**
-   * Whether the component should be unmounted (removed from DOM) when exited
+   * Whether the component should be unmounted (removed from DOM) when
+   * collapsed
    */
   unmountOnExit: React.PropTypes.bool,
 
   /**
-   * Whether transition in should run when the Transition component mounts, if
-   * the component is initially entered
+   * Whether the component should expand after mounting
    */
   transitionAppear: React.PropTypes.bool,
 
   /**
-   * Duration of the animation in milliseconds, to ensure that finishing
-   * callbacks are fired even if the original browser transition end events are
-   * canceled
+   * Duration of the collapse animation in milliseconds, to ensure that
+   * finishing callbacks are fired even if the original browser transition end
+   * events are canceled
    */
   duration: React.PropTypes.number,
 
   /**
-   * Callback fired before the "entering" classes are applied
+   * Callback fired before the component expands
    */
   onEnter: React.PropTypes.func,
   /**
-   * Callback fired after the "entering" classes are applied
+   * Callback fired after the component starts to expand
    */
   onEntering: React.PropTypes.func,
   /**
-   * Callback fired after the "enter" classes are applied
+   * Callback fired after the component has expanded
    */
   onEntered: React.PropTypes.func,
   /**
-   * Callback fired before the "exiting" classes are applied
+   * Callback fired before the component collapses
    */
   onExit: React.PropTypes.func,
   /**
-   * Callback fired after the "exiting" classes are applied
+   * Callback fired after the component starts to collapse
    */
   onExiting: React.PropTypes.func,
   /**
-   * Callback fired after the "exited" classes are applied
+   * Callback fired after the component has collapsed
    */
   onExited: React.PropTypes.func,
 
   /**
-   * The dimension used when collapsing
+   * The dimension used when collapsing, or a function that returns the
+   * dimension
    *
    * _Note: Bootstrap only partially supports 'width'!
    * You will need to supply your own CSS animation for the `.width` CSS class._
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -115,21 +115,58 @@ class Collapse extends React.Component {
 
 Collapse.propTypes = {
   /**
-   * Collapse the Component in or out.
+   * Whether the component is entered; triggers the enter or exit animation
    */
-  in:       React.PropTypes.bool,
+  in: React.PropTypes.bool,
 
   /**
-   * Provide the duration of the animation in milliseconds, used to ensure that finishing callbacks are fired even if the
-   * original browser transition end events are canceled.
+   * Whether the component should be unmounted (removed from DOM) when exited
    */
-  duration:          React.PropTypes.number,
+  unmountOnExit: React.PropTypes.bool,
 
   /**
-   * Specifies the dimension used when collapsing.
+   * Whether transition in should run when the Transition component mounts, if
+   * the component is initially entered
+   */
+  transitionAppear: React.PropTypes.bool,
+
+  /**
+   * Duration of the animation in milliseconds, to ensure that finishing
+   * callbacks are fired even if the original browser transition end events are
+   * canceled
+   */
+  duration: React.PropTypes.number,
+
+  /**
+   * Callback fired before the "entering" classes are applied
+   */
+  onEnter: React.PropTypes.func,
+  /**
+   * Callback fired after the "entering" classes are applied
+   */
+  onEntering: React.PropTypes.func,
+  /**
+   * Callback fired after the "enter" classes are applied
+   */
+  onEntered: React.PropTypes.func,
+  /**
+   * Callback fired before the "exiting" classes are applied
+   */
+  onExit: React.PropTypes.func,
+  /**
+   * Callback fired after the "exiting" classes are applied
+   */
+  onExiting: React.PropTypes.func,
+  /**
+   * Callback fired after the "exited" classes are applied
+   */
+  onExited: React.PropTypes.func,
+
+  /**
+   * The dimension used when collapsing
    *
    * _Note: Bootstrap only partially supports 'width'!
-   * You will need to supply your own css animation for the `.width` css class._
+   * You will need to supply your own CSS animation for the `.width` CSS class._
    */
   dimension: React.PropTypes.oneOfType([
     React.PropTypes.oneOf(['height', 'width']),
@@ -137,62 +174,23 @@ Collapse.propTypes = {
   ]),
 
   /**
-   * A function that returns the height or width of the animating DOM node. Allows for providing some custom logic how much
-   * Collapse component should animate in its specified dimension.
+   * Function that returns the height or width of the animating DOM node
    *
-   * `getDimensionValue` is called with the current dimension prop value and the DOM node.
+   * Allows for providing some custom logic for how much the Collapse component
+   * should animate in its specified dimension. Called with the current
+   * dimension prop value and the DOM node.
    */
-  getDimensionValue: React.PropTypes.func,
-
-  /**
-   * A Callback fired before the component starts to expand.
-   */
-  onEnter: React.PropTypes.func,
-
-  /**
-   * A Callback fired immediately after the component starts to expand.
-   */
-  onEntering: React.PropTypes.func,
-
-  /**
-   * A Callback fired after the component has expanded.
-   */
-  onEntered: React.PropTypes.func,
-
-  /**
-   * A Callback fired before the component starts to collapse.
-   */
-  onExit: React.PropTypes.func,
-
-  /**
-   * A Callback fired immediately after the component starts to collapse.
-   */
-  onExiting: React.PropTypes.func,
-
-  /**
-   * A Callback fired after the component has collapsed.
-   */
-  onExited: React.PropTypes.func,
-
-  /**
-   * Specify whether the transitioning component should be unmounted (removed from the DOM) once the exit animation finishes.
-   */
-  unmountOnExit:     React.PropTypes.bool,
-
-  /**
-   * Specify whether the component should collapse or expand when it mounts.
-   */
-  transitionAppear: React.PropTypes.bool
+  getDimensionValue: React.PropTypes.func
 };
 
 Collapse.defaultProps = {
-  in:       false,
+  in: false,
   duration: 300,
-  dimension: 'height',
-  transitionAppear: false,
   unmountOnExit: false,
+  transitionAppear: false,
+
+  dimension: 'height',
   getDimensionValue
 };
 
 export default Collapse;
-
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -21,18 +21,18 @@ class Fade extends React.Component {
 
 Fade.propTypes = {
   /**
-   * Whether the component is faded in
+   * Show the component; triggers the fade in or fade out animation
    */
   in: React.PropTypes.bool,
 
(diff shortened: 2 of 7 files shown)


taion:
There were a few minor race conditions in `Transition` around rapid state changes, which this implementation should resolve.

src/Collapse.js
```diff
@@ -115,84 +115,82 @@ class Collapse extends React.Component {
 
 Collapse.propTypes = {
   /**
-   * Collapse the Component in or out.
+   * Whether the component is entered; triggers the enter or exit animation
```
AlexKVal: some vague description.
sorry but I did not get it. :sweat: 
may be add a couple simple one liner examples ?

--
though maybe it's just advanced English

src/Collapse.js
```diff
@@ -115,84 +115,82 @@ class Collapse extends React.Component {
...
-  duration:          React.PropTypes.number,
+  unmountOnExit: React.PropTypes.bool,
 
   /**
-   * Specifies the dimension used when collapsing.
-   *
-   * _Note: Bootstrap only partially supports 'width'!
-   * You will need to supply your own css animation for the `.width` css class._
+   * Whether transition in should run when the Transition component mounts, if
+   * the component is initially entered
```
AlexKVal: `transition in` => `transition 'in'` ?
`is initially entered` => `is initially in 'entered' state` ?

src/Collapse.js
```diff
@@ -115,84 +115,82 @@ class Collapse extends React.Component {
...
-   * Specify whether the transitioning component should be unmounted (removed from the DOM) once the exit animation finishes.
+   * The dimension used when collapsing
+   *
+   * _Note: Bootstrap only partially supports 'width'!
+   * You will need to supply your own CSS animation for the `.width` CSS class._
    */
-  unmountOnExit:     React.PropTypes.bool,
+  dimension: React.PropTypes.oneOfType([
+    React.PropTypes.oneOf(['height', 'width']),
+    React.PropTypes.func
```
AlexKVal: this `function` variant is not documented.
Is it private ?

src/Fade.js
```diff
@@ -2,87 +2,77 @@ import React from 'react';
 import Transition from './Transition';
 
 class Fade extends React.Component {
-
-  constructor(props, context){
-    super(props, context);
-  }
```
AlexKVal: can I read somewhere about it ?

src/Fade.js
```diff
@@ -2,87 +2,77 @@ import React from 'react';
...
         enteringClassName='in'
       >
-        { this.props.children }
+        {this.props.children}
       </Transition>
     );
   }
 }
 
+// Explicitly copied from Transition for doc generation.
```
AlexKVal: I would mark it for future:
`TODO: remove doc generation duplication with Transition, when some mechanic for it is added`
(phrasing could be better o/c)

src/Position.js
```diff
@@ -15,33 +15,35 @@ class Position extends React.Component {
...
+    }
   }
 
   render() {
-    let { children, ...props } = this.props;
-    let { positionLeft, positionTop, ...arrows } = this.props.target ? this.state : {};
+    const {children, ...props} = this.props;
+    const {positionLeft, positionTop, ...arrowPosition } = this.state;
 
+    const child = React.Children.only(children);
```
AlexKVal: :+1:

src/Overlay.js
```diff
@@ -42,56 +35,63 @@ class Overlay extends React.Component {
...
-    let child = null;
-
-    if ( Transition === true ){
+    if (Transition === true) {
       Transition = Fade;
     }
 
-    if (props.show || (Transition && this.state.exiting && !this.state.exited)) {
+    // Don't un-render the overlay while it's transitioning out.
+    const mountOverlay = props.show || (Transition && !this.state.exited);
```
AlexKVal: :+1:

src/Position.js
```diff
@@ -49,55 +51,62 @@ class Position extends React.Component {
...
-  _maybeUpdatePosition(){
-    if ( this._needsFlush ) {
-      this._needsFlush = false;
-      this._updatePosition();
+  getTargetSafe(props) {
+    if (!props.target) {
+      return null;
     }
+
+    return props.target(props);
```
AlexKVal: why not just ?
```js
getTargetSafe(props) {
  return props.target ? props.target(props) : null;
}
```

src/Position.js
```diff
@@ -49,55 +51,62 @@ class Position extends React.Component {
...
+      container,
+      props.containerPadding
+    ));
   }
 }
 
 Position.propTypes = {
   /**
-   * The target DOM node the Component is positioned next too.
+   * Function mapping props to DOM node the component is positioned next to
```
AlexKVal: :+1:

src/Position.js
```diff
@@ -49,55 +51,62 @@ class Position extends React.Component {
...
-  container:        CustomPropTypes.mountable,
+  container: CustomPropTypes.mountable,
   /**
-   * Distance in pixels the Component should be positioned to the edge of the Container.
+   * Minimum spacing in pixels between container border and component border
    */
   containerPadding: React.PropTypes.number,
   /**
-   * The location that the overlay should be positioned to its target.
+   * How to position the component relative to the target
```
AlexKVal: :+1:

src/Transition.js
```diff
@@ -2,284 +2,274 @@ import React from 'react';
...
+  render() {
+    const status = this.state.status;
+    if (status === UNMOUNTED) {
+      return null;
     }
 
-    else if (!this.state.in && this.state.transitioning) {
-      classes = this.props.exitingClassName;
+    const {children, className, style, ...childProps} = this.props;
+    Object.keys(Transition.propTypes).forEach(key => delete childProps[key]);
```
AlexKVal: :+1:  instead of `omit()`

src/Transition.js
```diff
@@ -2,284 +2,274 @@ import React from 'react';
...
+
+    let transitionClassName;
+    if (status === EXITED) {
+      transitionClassName = this.props.exitedClassName;
+    } else if (status === ENTERING) {
+      transitionClassName = this.props.enteringClassName;
+    } else if (status === ENTERED) {
+      transitionClassName = this.props.enteredClassName;
+    } else if (status === EXITING) {
+      transitionClassName = this.props.exitingClassName;
```
AlexKVal: why not continue and not use `switch/case` then ?

src/Transition.js
```diff
@@ -2,284 +2,274 @@ import React from 'react';
...
-   * A callback fired after "exit" classes are applied
+   * Callback fired after the "exited" classes are applied
    */
-  onExited:    React.PropTypes.func
+  onExited: React.PropTypes.func
 };
 
-// name the function so it is clearer in the documentation
-const noop = ()=>{};
+// Name the function so it is clearer in the documentation
```
AlexKVal: :+1:

AlexKVal:
I think adding disabling eslint comment should be enough.
```js
const {children, className, style, ...childProps} = this.props; // eslint-disable-line react/prop-types
```
to satisfy
screen shot 2015-07-20 at 3 22 50 pm

and it is good to go :sparkle:  :cherries: 

--
I like the idea to use constants here
```js
export const EXITED = 1;
export const ENTERING = 2;
export const ENTERED = 3;
export const EXITING = 4;
```
Code readability and tests readability now is much better.

Even I started to understand at some degree how it works :blush: 
(though it definitely needs some kind of tutorial with diagrams, but it is another issue o/c.
I believe when I am totally grasp mechanic of these components, I could write one for beginners like me)

src/Fade.js
AlexKVal: and maybe add https://github.com/react-bootstrap/react-bootstrap/issues/977 to this TODO

src/Fade.js
jquense: if you only call `super` in the constructor with the same arguments its the same thing as not having it. Classes will essentially create that constructor automatically for you.

src/Overlay.js
```diff
@@ -42,56 +35,63 @@ class Overlay extends React.Component {
...
-      // Position the child before the animation to avoid `null` DOM nodes
+    if (Transition) {
+      // This animates the child by injecting props, so it must be inner-most.
       child = (
-        <Position {...{ container, containerPadding, target, placement }}>
-          { child }
-        </Position>
+        <Transition
+          in={props.show}
+          transitionAppear
```
jquense: why remove the `props.show` here? Originally it was there because you only want the fade in animation to run on mount. if `in=false` initially this will show the overlay then fade it out, instead of just starting closed. Is that the behavior we want?

src/Overlay.js
jquense: What was the rational for removing `this.state.exiting`? I originally had the code like this, but ran into a bug...that I can't remember, that this additional check fixed

src/Transition.js
```diff
@@ -2,284 +2,274 @@ import React from 'react';
...
+          this.props.onEntered(node);
+        });
       });
-    };
-
-    if (maybeNode) {
-      enter(maybeNode);
-    }
-    else if (this.props.unmountOnExit) {
-      this._ensureNode(enter);
```
jquense: sorry its a bit early for me still...in the new code how is this cases handled, where a node has exited and unmounted and now needs to enter, wouldn't the node still be null here, or am i just not tracking the logic through well enough.

jquense:
looks great :) nice and clean. Tangential question why remove all the Fade/Collapse specific documentation comments with the base Transition ones (in which case you can just leave them out completely)?

src/Transition.js
```diff
@@ -2,284 +2,274 @@ import React from 'react';
...
+    return React.cloneElement(
+      child,
+      {
+        ...childProps,
+        className: classnames(
+          child.props.className,
+          className,
+          transitionClassName
+        ),
+        style: {...child.props.style, ...style}
```
jquense: why this? My thought was that if someone wanted to style the child they can just add it there, instead of introducing an additional input vector. I don't have any strong opinion here, just wondering

src/Fade.js
taion: This is not so easy - the current metadata parser actually crashes if I try to do this the "obvious" way with e.g. `_.omit`.

src/Position.js
taion: I think it's a little more clear to write "exceptional" cases as traps at the start of a function rather than branches, e.g. I always use something like

```js
if (thingsAreMessedUp) {
  return;
}

// do actual work
```

src/Transition.js
taion: I had it initially - a lot of people don't like `switch`/`case`, though, so I thought this was a bit more kosher.

src/Transition.js
taion: This actually broke the custom overlay example - I'd changed it to `Position > Transition > {node}`, and the styles from `Position` were clobbering the ones on the node.

That said, looking at the code now, it's actually just fine if I change it back to `Transition > Position > {node}`, so I'm going to get rid of this.

src/Transition.js
taion: For the `unmountAtExit`, the flow on `in` getting set is:

1. In `componentWillReceive` props, we hit the `unmountAtExit` + `UNMOUNTED` case, and set the state to `EXITED`
2. In `render`, we show the component in the exited state
3. In `componentDidUpdate` (after the DOM has already reflowed), we hit the case where we are `in` and `EXITED`, and kick off the enter animation - at this point the node exists again

src/Fade.js
jquense: I have been thinking about this...if we had a standard omit, we could teach the metadata about it. Right now tho it will probably choke on anything that isn't an object literal, so you'd need to use a spread like: `..._.omit(blah, ['bah'])`

src/Overlay.js
taion: In both versions of the code, we only hit this branch if we're showing or exiting the overlay.

I might be misunderstanding the intention of `transitionAppear`, but I interpreted it as meaning "if the component is `in` on mounting, mount it in the exited state, but transition to entering immediately after the initial render", and nothing else.

taion:
RE: custom doc removal - laziness/tiredness on my part, will add back.

src/Transition.js
jquense: ahhh ok so when it is going to transition back in, you change it from `UNMOUNTED` to `EXITED`, nice. :)

src/Overlay.js
jquense: I was more or less trying to copy the meaning from [CSSTransitionGroup](http://facebook.github.io/react/docs/animation.html#animate-initial-mounting) which, now that I think about it, only apply for entering, since state is determined by whether children exist there. Which is to say matches your interpretation.

I was thinking originally though that since the `in` prop determines initial state, it could also apply to the inverse, i.e:  "if the component is not `in` on mounting, mount it in the entered state, but transition to exiting immediately after the initial render". That was the original behavior. That being said its a totally valid to question whether that makes any sense. Would you ever want to animate out on mount? I went for just allowing it for flexibility, but I can't personally think of an actual use case...

src/Transition.js
taion: Yes - let me rewrite this to be more explicit in that it's separately handling the `unmountAtExit` and `transitionAppear` cases.

src/Overlay.js
taion: It shouldn't be difficult to add it back, but I can't think of a good use case for wanting to do so - can add it back if it comes up.

src/Overlay.js
taion: I think there were a few edge cases around the "exited" trigger firing early from `onExited` - since I now explicitly cancel the chained callbacks in `Transition`, those no longer come up.

taion:
Added commit with fixes to above.

taion:
Added a test to handle a bug in my `unmountOnExit` handling.

src/Fade.js
AlexKVal: Gotcha.

AlexKVal:
Thank you for clear doc descriptions for props :+1: 
Now everything is cleaner.

jquense:
This looks great! Thanks for cleaning it up, it's really slick now. I haven't had a chance to actually try it out. but should have a moment to do so in a bit.

AlexKVal:
OT: 
> I switched everything to use expect for consistency. should doesn't work for e.g. null values, which I use to assert non-mounted-ness on the DOM.

What do you thing about I re-write all other tests from `should` to `expect` and somehow prevent (by eslint) all future `should` usage ?

taion:
@AlexKVal 

I replaced `should` with `expect` because I started out with `should`, realized it didn't work, and then did a regex search-and-replace that changed other instances in the file, so I figured "what the hell" and just made everything consistent (previously it was using a mix of `should` and `assert`).

I'd prefer to use `expect` everywhere, but it's pure bikeshedding, and doesn't materially affect the readability of the code. I'd personally be mildly supportive of making such a change to make our test code more consistent and more easily scannable, but I don't think it's a big deal.

AlexKVal:
My motive for `should` => `expect` is exactly this:
> should doesn't work for e.g. null values

thus to prevent such subtle bugs.

taion:
It won't be a subtle bug though, for a test (unless the test is somehow non-deterministic). The test just won't work.

@jquense Did you get a chance to take a look? Waiting to fix the positioning bit before releasing v0.24.

jquense:
went through it on the docs site. Looks great, just one thought. since the updatePosition call is being processed in `componentWillReceiveProps` there might be a chance that that new target doesn't exist yet right, if it was just mounted?

If that's not an issue then :shipit:

taion:
That's a good point. If the target gets switched to point to a ref of a component mounted in the same render cycle, this code will probably break. Let me not try to be as clever in this code.

taion:
@jquense Fixed the Position recalculation.

jquense:
:shipit: :shipit:

taion:
I'm really sorry - one more change incoming. Current implementation is pretty dumpster tier now that I think about it. Sorry to waste your time.

taion:
Okay, last change for real this time - instead of trying to do weird checks on the value of the `target` function, just cache the last target element, and check to see if it changes.

jquense:
LGTM

taion:
LF:KDJSF

AlexKVal:
> LF:KDJSF

Did you mean all those `!fixup` and `fixup!` ? :smile: 
screen shot 2015-07-22 at 3 24 00 pm

AlexKVal:
But hey !!
`v0.24` finally is out :tada: :tada: :tada: :sparkle: :sparkles: :sparkle:

PR #1028, opened 2015-07-20
[fixed] Tooltip accepts a style prop

### LINKED ISSUE #1027: Style passthroughs on Tooltips
We think a user should be able to supply a `style` prop to Tooltips, because this would be consistent with the behavior of most other components. Currently, any `style` prop passed into a tooltip will be ignored.

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Tooltip.js#L72

Style is being defined and supplied with left and top attributes only. We would like to be able to provide other styles as a prop to Tooltips.

```
<Tooltip style={{opacity: 1}} />
```

Should render a tooltip with merged styles (`{opacity: 1, top: Xpx, left: Xpx}`).

Would you be open to a PR?

cc/ @gpleiss @stubbornella @matt-royal @ctaymor

kennyw12:
fixes #1027

### DIFF
--- a/src/Tooltip.js
+++ b/src/Tooltip.js
@@ -43,7 +43,11 @@ const Tooltip = React.createClass({
     /**
      * Title text
      */
-    title: React.PropTypes.node
+    title: React.PropTypes.node,
+    /**
+     * Style hash
+     */
+    style: React.PropTypes.object
   },
 
   getDefaultProps() {
@@ -60,7 +64,8 @@ const Tooltip = React.createClass({
 
     const style = {
       'left': this.props.positionLeft,
-      'top': this.props.positionTop
+      'top': this.props.positionTop,
+      ...this.props.style
     };
 
     const arrowStyle = {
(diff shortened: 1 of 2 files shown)


src/Tooltip.js
```diff
@@ -43,7 +43,11 @@ const Tooltip = React.createClass({
     /**
      * Title text
      */
-    title: React.PropTypes.node
+    title: React.PropTypes.node,
+    /**
+     * Style hash
+     */
+    style: React.PropTypes.object
```
taion: I'd actually rather not explicitly document this since it just adds to the clutter, but that's totally just personal preference on my part.

taion:
LGTM. Need one more from @react-bootstrap/collaborators to sign off.

dozoisch:
Yeah Looks good! 

Would it be worth mentioning that style top/left overrides positionTop and positionLeft ?

taion:
@dozoisch We okay with cleaning up docs after merging this?

dozoisch:
@taion I am! Alright merging, then!

PR #1037, opened 2015-07-21
Lock Babel to old version due to linting issues

taion:
This fixes the build issues. See https://github.com/facebook/react/pull/4449.


jquense:
LGTM

Issue #748, opened 2015-05-27
Make Modal wrapper configurable


mathieumg:
Following discussion in https://github.com/mzabriskie/react-draggable/issues/56 it would be great if we could configure some more internals of the `Modal`, in this case configure what wraps the `Modal` content. We recently made changes to allow greater flexibility in that regard by making configurable some hardcoded `className`s, this would be the next step in that direction.

mtscout6:
:+1: Bootstrap themselves don't want to support this, but if we can with minimal effort I see no reason why we shouldn't.

cosminnicula:
:+1:

cosminnicula:
You might find this useful https://github.com/jquense/react-bootstrap-modal

Notice how Modal is composed of Modal.Header, Modal.Body and Modal.Footer. In this way you have more control over the markup.

taion:
:+1: I would love to be able to access just the overlay + hide on whatever logic and not use any of the standard Bootstrap modal styling.

mtscout6:
@cosminnicula Thanks for sharing that modal by @jquense. I have to say that I REALLY like that approach, and it's in line with a lot of changes that I'd like to make to more of the components within this library. In fact I was doing something similar on the Dropdown re-write for the title so you could do:

```
<DropdownButton>
  <DropdownButtonTitle>Text</DropdownButtonTitle>
  <MenuItem>...</MenuItem>
</DropdownButtom>
```

I think that designing components in this fashion would also yield smaller simpler components to maintain and test.

@joemcbride and I just came up with a use case internally at work that would be a lot easier with @jquense's approach.

This would be a breaking change, but I think it would be for the better.

jquense:
We've had great success with this sort of approach (with the modal and other components), its a bunch more markup but the added configurablility is well worth it. The convention of adding the "children" (`Modal.Header`) types as static properties reduces some of the added typing by still allowing a single import.

Please take anything ya'll can use from my code, its originally based of the modal here, so I'd be happy to contribute back.

mtscout6:
@jquense I'd be open to seeing what a PR for this would look like. The hard part would be a sensible deprecation warning for the existing setup.

jquense:
depends on what you want to break and what features are needed it seems.

[...]

PR #1040, opened 2015-07-22
Modal tweak

jquense:
Not sure that `Modal.Markup` is the best name...taking suggestions

### DIFF
--- a/docs/examples/ModalStatic.js
+++ b/docs/examples/ModalStatic.js
@@ -1,0 +1,22 @@
+
+const modalInstance = (
+  <div className='static-modal'>
+    <Modal.Dialog>
+      <Modal.Header>
+        <Modal.Title>Modal title</Modal.Title>
+      </Modal.Header>
+
+      <Modal.Body>
+        One fine body...
+      </Modal.Body>
+
+      <Modal.Footer>
+        <Button>Close</Button>
+        <Button bsStyle='primary'>Save changes</Button>
+      </Modal.Footer>
+
+    </Modal.Dialog>
+  </div>
+);
+
+React.render(modalInstance, mountNode);
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -260,6 +260,10 @@ const ComponentsPage = React.createClass({
                 <div className='bs-docs-section'>
                   <h1 className='page-header'><Anchor id='modals'>Modals</Anchor> <small>Modal</small></h1>
 
+                  <h3><Anchor id='modals-static'>Static Markup</Anchor></h3>
+                  <p>A modal dialog component</p>
+                  <ReactPlayground codeText={Samples.ModalStatic} />
+
                   <h3><Anchor id='modals-live'>Basic example</Anchor></h3>
                   <p></p>
                   <p>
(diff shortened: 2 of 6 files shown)


taion:
Maybe `ModalDialog`? We should add back the "static modal dialog" example to the components page, though.

jquense:
good call, let me add that back

jquense:
@taion does it make more sense to split the backdrop and onmount container logic out into the Modal component and have the Dialog just be the actual dialog markup? I am not sure its useful to anyone but it makes more sense to me conceptually. then the Dialog component can be dumb

taion:
I think that's the way to do it. Make `ModalDialog` be just a static dialog component. This will also have the added benefit of preventing that example from causing random scrolling bugs every couple of weeks.

jquense:
whelp that ended up being a good bit more work then I thought :P still I think this is better. It also mirrors the upstream twbs modal a bit more structurally which is helpful I think.

taion:
Oof, yeah.

Do you think it makes sense to add a prop to `Modal` such that the user can supply his own `ModalDialog` or equivalent (i.e. make it `cloneElement` a single supplied child rather than use the current render logic)? I think that would address some of the issues @mfunkie brought up on Reactiflux.

I think it'd also otherwise make this `Modal` a lot more useful to people who want to more deeply customize how the dialog itself appears, rather than just using the TWBS modal dialog.

jquense:
ya I think that's a great idea, I wanted to do something similar to address #748 when the original PR went through but couldn't think of a not-ugly way to handle it. A prop is the simplest, maybe we could also do some sniffing with children but I am not sure how that be unambiguous (and not breaking). I also don't want to have to make everyone do: `<Modal><Modal.Dialog>...</Modal.Dialog></Modal>` to use it, but maybe that's ultimately the best way to go?

mfunkie:
I like the idea behind this.  Basically separating the driver of the Modal from the Modal's Markup, correct?  I think it's moving closer to bringing a lot of value to people who might not even use Bootstrap, but want a lot of the functionality that this library provides.

taion:
@jquense 

I feel like a prop is the best way to handle this - child sniffing seems really dangerous and trouble-prone.

I think it'd be reasonable to have a `customModal` prop that defaults to `false` that controls this. I'd be okay with requiring `ModalDialog` going forward as well. Don't feel strongly either way.

jquense:
Appveyor being a doofus again...

I added `dialogComponent` as a prop that should solve most of the need in that regard I think, I also updated the classes to use bsClass directly so that will be changeable by folks who need to separate out the styles. I think we are gtg if anyone can take a peak

taion:
Cool - so the idea is that the API would look like `<Modal dialogComponent={MyModalDialog} ...`?

@mfunkie How does this look to you?

jquense:
Ya exactly. if you don't want to add a new wrapping element changing the default you can also do something like: 

```js
Object.assign(Modal.defaultProps, {
  dialogComponent: MyCoolDialog,
  bsClass: 'my-modal'
})
```

or whatever

mfunkie:
Since you're kinda merging props with the dialogComponent, can you allow className to be passed also?  It doesn't look like you're using that prop and it would get overwritten if I remember right (className doesnt get merged in prop spread right?)

taion:
I guess the alternative would be something like

```js
<Modal customDialog={true} /* ... */>
  <MyModalDialog />
</Modal>
```

I dunno if that's any good.

jquense:
@mfunkie in this case you aren't passing `dialogComponent` an Element but a class, so merging classnames would be the responsibility of MyCoolDialog

jquense:
Backed the dialog API out for now, so we can get a release out. will open a new PR to bike shed the api a bit more

taion:
LGTM

PR #1045, opened 2015-07-23
OverlayTrigger: mount overlay if mount point is defined

salper:
I came accross some weird React component life cycle where didUpdate was called after willUnmount. Thus, mounting overlay when there is no mount point naturally raised an error.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -138,7 +138,9 @@ const OverlayTrigger = React.createClass({
   },
 
   componentDidUpdate(){
-    React.render(this._overlay, this._mountNode);
+    if (this._mountNode) {
+      React.render(this._overlay, this._mountNode);
+    }
   },
 
   getOverlayTarget() {


salper:
Yipikaye Windows...

taion:
Wow, I didn't realize `componentDidUpdate` could fire after `componentWillUnmount`. LGTM. Might have to revise this if we were make `OverlayTrigger` create the mount nodes lazily the way `Overlay` does.

jquense:
:shipit: since the mount node here, doesn't ever actually get added to the DOM i figured it's probably fine to not do the lazy creation part of it. Perhaps it could be a memory issue tho?

taion:
Oops, my bad, I misunderstood what was happening. Wow - I didn't realize that worked. TIL.

jquense:
hehe yeah, its too clever-in-the-bad-way, but it solved a somewhat intractable problem. (stole it from `testUtils.renderIntoDocument`)

PR #1047, opened 2015-07-23
Show anchor icons when links are focused as well as hovered

jquense:


AlexKVal:
before:
screen shot 2015-07-23 at 7 55 46 pm

after:
screen shot 2015-07-23 at 7 56 00 pm

LGTM

PR #1048, opened 2015-07-23
[added] Overlay and OverlayTrigger accept Transition callbacks

### LINKED ISSUE #1036: OverlayTrigger should notify when the overlay is shown or hidden
Perhaps it could accept `onOverlayShown` and `onOverlayHidden` properties.

Currently if you need to know if the overlay is shown or not, you have to ditch `OverlayTrigger` all together and roll your own (as far as I can tell).

jquense:
Fixes #1036

### DIFF
--- a/src/Overlay.js
+++ b/src/Overlay.js
@@ -57,13 +57,20 @@ class Overlay extends React.Component {
     );
 
     if (Transition) {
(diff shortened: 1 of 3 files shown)


taion:
:shipit:

benekastah:
Awesome! Thanks for this.

PR #1052, opened 2015-07-23
[added] WrapperElement for Pagination buttons

Torwori:
usage: `<Pagination ... WrapperElement={<Element/>}/>`
default behavior remains unchanged
use `this.props.children` in Element where you want the numbers to be

Useful for people using react-router because you can wrap the buttons in Link.

note: Bootstrap css targets `.pagination > li > a`

### DIFF
--- a/src/Pagination.js
+++ b/src/Pagination.js
@@ -2,6 +2,8 @@ import React from 'react';
 import classNames from 'classnames';
 import BootstrapMixin from './BootstrapMixin';
 import PaginationButton from './PaginationButton';
+import CustomPropTypes from './utils/CustomPropTypes';
+import SafeAnchor from './SafeAnchor';
 
 const Pagination = React.createClass({
   mixins: [BootstrapMixin],
@@ -15,7 +17,11 @@ const Pagination = React.createClass({
     last: React.PropTypes.bool,
     prev: React.PropTypes.bool,
     next: React.PropTypes.bool,
-    onSelect: React.PropTypes.func
+    onSelect: React.PropTypes.func,
+    /**
+     * You can use a custom element for the buttons
+     */
+    buttonComponentClass: CustomPropTypes.elementType
   },
 
   getDefaultProps() {
@@ -28,6 +34,7 @@ const Pagination = React.createClass({
       prev: false,
       next: false,
       ellipsis: true,
+      buttonComponentClass: SafeAnchor,
       bsClass: 'pagination'
     };
   },
@@ -40,7 +47,8 @@ const Pagination = React.createClass({
       activePage,
       items,
       onSelect,
-      ellipsis
+      ellipsis,
+      buttonComponentClass
     } = this.props;
 
     if(maxButtons){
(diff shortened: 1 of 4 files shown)


AlexKVal:
Maybe it is better to do it this way ?
```js
propTypes: {
    /**
     * You can use a custom element for this component
     */
    componentClass: CustomPropTypes.elementType
  },
...
render() {
    let ComponentClass = this.props.componentClass;
    return (
      <ComponentClass ...
```
[src/Grid.js#L16](https://github.com/react-bootstrap/react-bootstrap/blob/5d9db003810174287fe6fb4567896157af8e8400/src/Grid.js#L16)
[src/Row.js#L10](https://github.com/react-bootstrap/react-bootstrap/blob/31dac7f5602591456f0568746b3f494d49ea0a0c/src/Row.js#L10)
and so on [componentClass](https://github.com/react-bootstrap/react-bootstrap/search?utf8=%E2%9C%93&q=componentClass)

Torwori:
Ah, I didn't check the way you did it in other components. I'll update it as soon as I can.

Torwori:
Looking good now? I'm also willing to write the docs if this gets pulled in. :smiley:

src/Pagination.js
```diff
@@ -15,7 +17,11 @@ const Pagination = React.createClass({
     last: React.PropTypes.bool,
     prev: React.PropTypes.bool,
     next: React.PropTypes.bool,
-    onSelect: React.PropTypes.func
+    onSelect: React.PropTypes.func,
+    /**
+     * You can use a custom element for this component
+     */
+    componentClass: CustomPropTypes.elementType
```
jquense: since this is not being used as a custom element for _this_ component but is being passed down I would name it something a bit more descriptive like `buttonComponentClass`

Torwori: I agree, changed it.

jquense:
looks good can you squash your commits and then we just need another contributor to sign off

dozoisch:
LGTM

stasyaner:
does it still work? I can wrap page number into a Link but it seems like it then loses some events or whatever was making the pagination work.

taion:
You'd presumably need another component that actually renders the `<Link>`.

stasyaner:
Am I missing something?

```
import {Link} from 'react-router';

class LinkWrapper extends Component{
  render() {
    return <Link to='/something'>{this.props.children}</Link>;
  }
}
```

taion:
Seems fine. Might depend on your `onSelect`.

stasyaner:
@taion So I should add to props.children my onSelect?
```
<Pagination
        prev
        next
        first
        last
        ellipsis
        boundaryLinks
        bsSize='small'
        items={props.pageNumber}
        maxButtons={props.pagesToShow}
        activePage={props.activePage}
        onSelect={(eventKey) => {props.setActivePage(eventKey)}}
      />
```

taion:
Both approaches should work.

stasyaner:
but it's not... seems like onSelect doesn't go down to it if I wrap it into that LinkWrapper, it doesn't change activePage prop

davidecarpini:
stasyaner,  Your component should be:

class LinkWrapper extends Component{
  render() {
    return <Link to='/something' onClick={ e => this.props.onClick(e) } >{this.props.children}</Link>;
  }
}

PR #1053, opened 2015-07-23
Overlays don't position themselves properly when target is an SVG

### LINKED ISSUE #1050: Overlays don't position themselves properly when target is an SVG
`getBoundingClientRect` should provide enough information to make it possible.

benekastah:
Fixes #1050

I started trying to get tests going for this, but it ended up taking a while for me for various reasons. I can come back to it later if needed.

### DIFF
--- a/src/utils/domUtils.js
+++ b/src/utils/domUtils.js
@@ -138,6 +138,25 @@ function getPosition(elem, offsetParent) {
 }
 
 /**
+ * Get an element's size
+ *
+ * @param {HTMLElement} elem
+ * @returns {{width: number, height: number}}
+ */
+function getSize(elem) {
+  let rect = {
+    width: elem.offsetWidth,
+    height: elem.offsetHeight
+  };
+  if (typeof elem.getBoundingClientRect !== 'undefined') {
+    let {width, height} = elem.getBoundingClientRect();
+    rect.width = width || rect.width;
+    rect.height = height || rect.height;
+  }
+  return rect;
+}
+
+/**
  * Get parent element
  *
  * @param {HTMLElement?} elem
@@ -187,6 +206,7 @@ export default {
   getComputedStyles,
   getOffset,
   getPosition,
+  getSize,
   activeElement: getActiveElement,
   offsetParent: offsetParentFunc
 };
--- a/src/utils/domUtils.js
+++ b/src/utils/domUtils.js
@@ -145,8 +145,8 @@ function getPosition(elem, offsetParent) {
  */
 function getSize(elem) {
   let rect = {
-    width: elem.offsetWidth,
-    height: elem.offsetHeight
+    width: elem.offsetWidth || 0,
+    height: elem.offsetHeight || 0
   };
   if (typeof elem.getBoundingClientRect !== 'undefined') {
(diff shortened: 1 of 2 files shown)


src/utils/domUtils.js
```diff
@@ -138,6 +138,24 @@ function getPosition(elem, offsetParent) {
 }
 
 /**
+ * Get an element's size
+ *
+ * @param {HTMLElement} elem
+ * @returns {{width: number, height: number}}
+ */
+function getSize(elem) {
+  if (typeof elem.getBoundingClientRect !== 'undefined') {
```
taion: This is not the right check. Non-compliant browsers may return an object that lacks the `width` and `height` entries.

benekastah:
Not sure how to get the checks to pass. I ran tests locally, which all passed.

src/utils/domUtils.js
benekastah: Ah, good point. I changed it. Does it look good to you now?

taion:
Should make sure this plays well w/#1058, and unwind what's there if necessary.

benekastah:
@taion Not sure if we still needed that change, but I kept it in `getSize` just in case.

taion:
I'd drop it if your fix covers the Firefox + SVG use case. The less cruft the better.

benekastah:
Ah, ok. I'll leave it as it was then.

jquense:
do we need a test case for this? Seems like a nice thing to have, not sure what others feel like

AlexKVal:
I would be nice to have tests for the new functionality:
`utils/domUtils.js`
```js
getSize(elem)
```

`utils/overlayPositionUtils.js`
```js
getContainerDimensions(containerNode)
getPosition(target, container)
```

taion:
Well, we don't run CI against Firefox anyway, and this is really just a better implementation of existing behavior. Assuming the overlays still look right on the components docs page, I'm okay with this going in as is.

LGTM

benekastah:
Ok, I merged the "Properly test position calculation" stuff in and got the tests working. Everything should be good to go.

src/utils/domUtils.js
```diff
@@ -138,6 +138,25 @@ function getPosition(elem, offsetParent) {
...
+ * @returns {{width: number, height: number}}
+ */
+function getSize(elem) {
+  let rect = {
+    width: elem.offsetWidth,
+    height: elem.offsetHeight
+  };
+  if (typeof elem.getBoundingClientRect !== 'undefined') {
+    let {width, height} = elem.getBoundingClientRect();
+    rect.width = width || rect.width;
```
taion: What if gBCR actually returns `width` or `height` of 0?

benekastah: If that happens, `offsetWidth` or `offsetHeight` ought to be 0 too.

taion: Right, it's the case where `offsetWidth` is `null`-y rather than 0 that I'm concerned about here.

benekastah: Good call. Just fixed.

taion:
LGTM.

Need one more reviewer from @react-bootstrap/collaborators.

jquense:
LGTM

PR #1060, opened 2015-07-24
[added] Allow custom Modal dialog components

jquense:
pulled out from #1040, needs some additional bike shedding 

cc @mfunkie

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -4,10 +4,11 @@ import classNames from 'classnames';
 import domUtils from './utils/domUtils';
 import EventListener from './utils/EventListener';
 import createChainedFunction from './utils/createChainedFunction';
+import CustomPropTypes from './utils/CustomPropTypes';
 
 import Portal from './Portal';
 import Fade from './Fade';
-import Dialog from './ModalDialog';
+import ModalDialog from './ModalDialog';
 import Body from './ModalBody';
 import Header from './ModalHeader';
 import Title from './ModalTitle';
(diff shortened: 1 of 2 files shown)


taion:
Reprise - the question to me is whether we like the current API better, or if it would be better to do e.g.

```js
<Modal customDialog={true} ... >
  <MyModalDialog />
</Modal>
```

AlexKVal:
I vote for @taion proposal:
```js
<Modal customDialog={true} ... >
  <MyModalDialog />
</Modal>
```

I'm proposing the same API for another issue: https://github.com/react-bootstrap/react-bootstrap/pull/1071

AlexKVal:
Matt also is a proponent of such API :smile:  https://github.com/react-bootstrap/react-bootstrap/issues/748#issuecomment-108510673

taion:
I retract what I said earlier.

For handling the "generic modal" use case, it makes more sense to split out a new package from this with the core utilities around e.g. `Modal`, `Overlay`, `Transition`, `Position`, and friends.

It makes more sense for the `Modal` component in React-Bootstrap itself to map well to the Bootstrap modal, which this accomplishes quite well, while still giving some flexibility for customization.

I'm okay with merging this as-is. Will open a new issue regarding pulling out core components.

taion:
@jquense Can you rebase this? Let's merge this and defer the generic API to #1084, unless there are further objections.

taion:
I rebased - the conflict was spurious (some overlapping lines with another added test).

jquense:
I am fine with this as is if you are.

PR #1067, opened 2015-07-26
Properly test position calculation

taion:
With @jquense's new code it's easy to do the right thing here and actually properly test `calcOverlayPosition` via the exposed API points, rather than stubbing out a bunch of random functions in `overlayPositionUtils`.

This should largely address the concerns around tests in #1053.

The only caveat is that the new test path no longer exercises the scroll handling, because overlay positioning ignores current scroll value when using a non-`BODY` container.


AlexKVal:
Cool :+1:

jquense:
sweet :) looks good

Issue #1078, opened 2015-07-29
Modal (and probably others) should use componentWillUnmount


zerkms:
Currently all listeners and some other routines are handled in `onHide` method:

      componentDidUpdate(prevProps) {
        let { animation } = this.props;

        if ( prevProps.show && !this.props.show && !animation) {
          //otherwise handleHidden will call this.
          this.onHide();
        }
        else if ( !prevProps.show && this.props.show ) {
          this.onShow();
        }
      }

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js#L268

So the only way `this.onHide` is invoked is when it was about to update.

If you have a layout like:

    <MyComponent>
        <Modal ... />
    </MyComponent>

and at some moment I remove the `MyComponent` from the page then the `Modal` does not have chance to unbind its listeners and clean up mess after itself.

This might be solved if `this.onHide()` is invoked in `componentWillUnmount` additionally.

As a workaround in `MyComponent` I used a dirty

    componentWillUnmount() {
        this.refs.modal.onHide();
    }

and assigned a `ref="modal"` to the corresponding `Modal` element.

jquense:
nice catch, would like you like to PR that change?

zerkms:
sure, I'll give it a try this evening

taion:
Isn't this conflating a couple of concepts here? i.e. `onHide` means the user requested a hide and will do stuff like `this.setState({showModal: false})`.

I'd think you'd actually want to handle this with a `componentWillUnmount` on whatever is inside the `Modal`.

zerkms:
@taion not sure I see your point, could you please elaborate?

taion:
Let's say you have:

```js
<Modal show={...}>
  <MyComponent />
</Modal>
```

Why can't you put the cleanup in `MyComponent#componentWillUnmount`?

zerkms:
@taion cleanup of *what*? I don't have anything I own that I need to clean up, it's `Modal` that does not clean up after itself.

taion:
Your `onHide` method is going to be something like `() => this.setState({show: false})`, right? Are you saying that it's necessary to call this as the `Modal` unmounts?

zerkms:
I don't have any `onHide` method, it's an internal `Modal`'s method.

taion:
Ah! Sorry, I misunderstood. That makes sense, thanks.

zerkms:
[...]

PR #1079, opened 2015-07-29
Modal: added finalisation for the component when it was unbound from the tree

zerkms:
Fix for #1078

The component must clean up after itself when it was unmounted

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -277,6 +277,12 @@ const Modal = React.createClass({
     }
   },
 
+  componentWillUnmount() {
+    if (this.props.show) {
+      this.onHide();
+    }
+  },
+
   onShow() {
     const doc = domUtils.ownerDocument(this);
     const win = domUtils.ownerWindow(this);
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -197,6 +197,22 @@ describe('Modal', function () {
       , mountPoint);
   });
 
+  it('Should unbind listeners when unmounted', function() {
+    render(
+        <div>
+          <Modal show onHide={() => null} animation={false}>
+            <strong>Foo bar</strong>
+          </Modal>
+        </div>
+    , mountPoint);
+
+    assert.include(document.body.className, 'modal-open');
+
+    render(<div />, mountPoint);
+
+    assert.notInclude(document.body.className, 'modal-open');
+  });
+
   describe('Focused state', function () {
     let focusableContainer = null;
 


src/Modal.js
```diff
@@ -277,6 +277,11 @@ const Modal = React.createClass({
     }
   },
 
+  componentWillUnmount() {
+    this.onHide();
+    this.props.onHide();
```
wgcrouch: Does it make sense to call the onHide prop. If its unmounted the caller would have chosen to not display it. In our case we use the onHide callback to set some state which causes the modal not to render. So props.onHide would be called twice then

zerkms: @wgcrouch are you sure it will be called twice?

wgcrouch: sorry, that was a probably a bit unclear. 

If a users clicks something to hide the modal, eg the backdrop, props.onHide is called. The parent component handles the callback and sets some state which causes the modal to not be rendered, which results in a call to componentWillUnmount, which then calles props.onHide again.

zerkms: oh right, that's my fault indeed.

I cannot think of something better than creating a dedicated method with a flag that would invoke `this.props.onHide()` once.

But I'm not sure if there are scenarios when the component can be shown twice without recreating from the scratch.

jquense: Ya the modal should not call props.onHide, all that needs to happen is that the listeners get cleaned up which is done in this.onHide

zerkms: @jquense why should not it be called?

jquense: because the modal isn't being hidden it is being unmounted. Which brings up another issue actually. you need to check that `onHide` actually needs to be called. If `props.show` is `false` then it should not be called again.

not calling the onHide prop callback is consistent with every the other components, and with the react ecosystem in general (what I am familiar with of course)

zerkms: @jquense 

> because the modal isn't being hidden it is being unmounted. 

when it's unmounted it's being hidden without doubts (if it's shown, that's what I missed indeed)

> not calling the onHide prop callback is consistent with every the other components, and with the react ecosystem in general (what I am familiar with of course)

That is arguable: I as a developer and consumer of `Modal` component expect `onHide` to be invoked when the component disappears, irregardless on what actually caused that.

jquense: I can appreciate that though I disagree @taion perhaps can also weigh in.

regardless of that point right now the PR needs to add a guard that checks that the modal is even open before calling anything, could you make that change?

zerkms: wip :-)

jquense: thanks a lot! :)

zerkms:
I've added few more commits, after we decide if the implementation is acceptable I'll squash them into a single changeset.

src/Modal.js
```diff
@@ -309,6 +315,8 @@ const Modal = React.createClass({
 
     this.setState(this._getStyles() //eslint-disable-line react/no-did-mount-set-state
       , () => this.focusModalContent());
+
+    this.needsFinalization = true;
```
jquense: this is probably not needed, `onShow` is only ever called with props.show is set to true (either initially or updated) so checking if its true is enough to know if `onHide` needs to be called, which makes it consistent with `componentDidMount` as well

zerkms: @jquense it was my initial though, but then I came into conclusion that explicit is better than implicit.

src/Modal.js
taion: I don't think it makes sense to call `props.onHide` there. We have `props.onHide` because you can't control when the user will interact with the `Modal` in such a way as to want to close it. However, you (presumably) have control over when you are unmounting the `Modal`, so you have the opportunity to take whatever actions are relevant there. If you're unmounting the `Modal` for a reason that is similar to the user initiating a "hide" action, then you can invoke that callback yourself. If you're unmounting the `Modal` for some different reason, then I don't want to force that behavior upon you.

jquense:
Ok looks good, squash and adjust the commit message to follow the [guidlines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) and we are good to go. thanks again for helping out

taion:
LGTM as well once it's squashed.

taion:
That AppVeyor failure looks totally random. I'm re-running it.

TimDaub:
Still, there is another rather big problem with the modal component in my opinion.

When `<Modal>` is unmounting, its not telling its children (especially stuff that is in `<Modal.Body>`) that it's doing that.

So when I have a structure that looks like:

```
<Modal>
  <Modal.Body>
    <MyComponent />
  </Modal.Body>
</Modal>
```

I cannot use `MyComponent#componentWillUnmount`, since its never triggered.
This poses a rather big problem since its breaking the standard Reactjs API and keeps me from removing store event listeners properly before unmounting the component.

1. Is there a way to work around this?
2. How can this be fixed? (I can offer my help)

jquense:
@TimDaub what do you mean? when Modal unmounts so will Body and MyComponent, all the lifecycle hooks will be called, we don't do anything to keep that from happening (in fact there is nothing we can do to do that even if we wanted to)

TimDaub:
@jquense OK, sorry I was wrong.
https://jsbin.com/xuceyefobu/edit?html,js,output

jquense:
Ahh good glad to here there wasn't something really wrong :)

jareware:
Would love to see this land in a release!

PR #1085, opened 2015-07-29
[fixed] Carousel checks if it is mounted before setting state

Torwori:
Using carousel in a modal can sometimes lead to the modal being closed and carousel being unmounted before `setState` is called, causing the `Warning: setState(...): Can only update a mounted or mounting component...` message.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -266,26 +266,28 @@ const Carousel = React.createClass({
   handleSelect(index, direction) {
     clearTimeout(this.timeout);
 
-    let previousActiveIndex = this.getActiveIndex();
-    direction = direction || this.getDirection(previousActiveIndex, index);
+    if (this.isMounted()) {
+      let previousActiveIndex = this.getActiveIndex();
+      direction = direction || this.getDirection(previousActiveIndex, index);
 
-    if (this.props.onSelect) {
-      this.props.onSelect(index, direction);
-    }
-
-    if (this.props.activeIndex == null && index !== previousActiveIndex) {
-      if (this.state.previousActiveIndex != null) {
-        // If currently animating don't activate the new index.
-        // TODO: look into queuing this canceled call and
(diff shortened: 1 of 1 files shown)


taion:
We clear the timeout in `componentWillUnmount` though. Do you have any more details for what triggers this?

Torwori:
I can't reliably reproduce it. It most often seems to occur when there is an animation in progress and I close the modal, though sometimes it happens for seemingly no reason.

Stack Trace

jquense:
I have no idea if this is the same issue, but when things are in different "subtrees" (portals) and report up to a different subtree (modal calling back to thing rendering it) the ordering of prop updates and setState calls gets wonky. AFAICT this is due to React.render() being asynchronous. Since its not very reproducable that _may_ be related to the bug here, or completely unrelated :P.

ref: https://github.com/facebook/react/issues/1698

Torwori:
It could be the same issue, I have no idea. :smile:
The `isMounted()` stops the warning from showing up, but I can see how it looks unnecessary since timeouts seem to be properly cleared.

taion:
Let me take a look at this real quick once I get a chance. If I can find an easy fix that doesn't involve the check, I'd like to use that instead.

taion:
Okay, I can't reproduce this, but I'll take your word for there being some weird race condition here. Can you amend your commit to wrap all of https://github.com/Torwori/react-bootstrap/blob/carousel-mounted-fix/src/Carousel.js#L269-L291 in the `isMounted` check, so we don't fire `onSelected` when this sort of thing happens?

Torwori:
Done. Seems like there is an issue with webpack and travis though.

taion:
Seeing the same thing. Oh well.

LGTM. Need one more reviewer from @react-bootstrap/collaborators.

jquense:
LGTM

PR #1086, opened 2015-07-30
Initially render examples on the server

taion:
Got a little carried away here, maybe.

This does the initial render of the examples on the server side. It's not perfect - you still need to wait a while for all the JS to finish running on the client before you can interact with any of the examples, but at least you can see them on initial page load.

### DIFF
--- a/docs/src/ReactPlayground.js
+++ b/docs/src/ReactPlayground.js
@@ -1,113 +1,65 @@
-import * as modReact from 'react';
-import * as modClassNames from 'classnames';
-import * as modAccordion from '../../src/Accordion';
-import * as modAlert from '../../src/Alert';
-import * as modBadge from '../../src/Badge';
-import * as modButton from '../../src/Button';
-import * as modButtonGroup from '../../src/ButtonGroup';
-import * as modButtonInput from '../../src/ButtonInput';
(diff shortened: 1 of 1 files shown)


jquense:
so excited to try this out

taion:
Oops bug, updating.

taion:
The initial implementation was having issues getting the example to always update. This appears to fix them.

docs/src/ReactPlayground.js
```diff
@@ -1,113 +1,63 @@
...
-const TabPane = modTabPane.default;
-const Thumbnail = modThumbnail.default;
-const Tooltip = modTooltip.default;
-const Well = modWell.default;
-const Overlay = modOverlay.default;
+import babel from 'babel-core/browser';
+import CodeExample from './CodeExample';
 
-/* eslint-enable */
+import SafeAnchor from '../../src/SafeAnchor';
```
AlexKVal: `import` => `require` too.
This is also a component.

AlexKVal: Ah.. I see.. this is not necessary here.

AlexKVal:
Nice :+1:

taion:
Squashed and added a comment on the `SafeAnchor` import.

You guys okay with my merging this?

PR #1087, opened 2015-07-30
[fixed] adds missing 'role' to .panel-collapse

kennyw12:
To be ARIA-compliant, elements with "aria-expanded" must have a "role" set. ["note"](http://www.w3.org/TR/wai-aria/roles#note) seemed like the appropriate role.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -80,7 +80,8 @@ const Panel = React.createClass({
           className={collapseClass}
           id={this.props.id}
           ref='panel'
-          aria-expanded={this.isExpanded()}>
+          aria-expanded={this.isExpanded()}
+          role='note'>
           {this.renderBody()}
 
         </div>
--- a/test/PanelSpec.js
+++ b/test/PanelSpec.js
@@ -219,5 +219,14 @@ describe('Panel', function () {
       assert.equal(anchor.getAttribute('aria-controls'), 'panel-1');
     });
 
+    it('Should add role="note"', function() {
+      let instance = ReactTestUtils.renderIntoDocument(
+        <Panel id='panel-1' collapsible expanded header="Heading">Panel content</Panel>
+      );
+
+      let collapse = React.findDOMNode(instance).querySelector('.panel-collapse');
+
+      assert.equal(collapse.getAttribute('role'), 'note');
+    });
   });
 });


taion:
Are you sure it shouldn't be `tabpanel`? That's what TWBS uses for e.g. accordions.

kennyw12:
We considered [tabpanel](http://www.w3.org/TR/wai-aria/roles#tabpanel), but it has a more specific semantic meaning:

> A container for the resources associated with a tab, where each tab is contained in a tablist.

Panels aren't necessarily used with tabs, and it isn't clear what the "tablist" would be if we tried to apply the analogy to the panel itself.

jquense:
`tabpanel` makes sense in the context of the accordion, not sure what to do when there is no other tab... `tabpanel` also needs a bunch of associated roles like `tab` and `aria-describedBy` wondering if it makes more sense to allow the role as a prop and only add aria-selected when it exists?

kennyw12:
It makes sense to allow that configuration for users who want to control the role more specifically, but for now we think that `note` is a good default. Perhaps making it configurable could be a future enhancement if anyone requests it?

taion:
In the "Collapsible list group" example in the TWBS JS docs (scroll down from http://getbootstrap.com/javascript/#collapse-example-accordion), they use a `.panel-group` element with `role="tablist"` wrapping the whole thing, and assign `role="tab"` to the solitary `.panel-heading`.

kennyw12:
It's a common use-case for us to use panels in isolation, without a panel-group. In this situation, the "tab" and "tablist" roles seem inappropriate. Do you have any ideas for reconciling these two cases?

taion:
In the upstream example, they're using a `.panel-group` with just the single panel in there.

kennyw12:
Unfortunately we have many existing cases where other teams in our organization are using isolated panels. It isn't an option for us to force them all to now add a panel-group wrapper. This would leave us in a situation where panels are "invalid by default" where ARIA is concerned.

jquense:
let me ask around the a11y slack channel. my concern is that we are giving a role to the wrong thing when we should be moving the aria-selected attribute yo something more semantically meaningful markup

taion:
I understand and sympathize with your concerns. My concern is just that `"note"` is too opinionated, not generically suitable, and not in-line with what the upstream examples show.

jquense:
for what it's worth the tbs examples aren't always the best reference for pragmatic a11y implementations (tho they have gotten better recently).

stubbornella:
Hmm, note sounds a bit odd to me too. I don't think we can be sure that the content is parenthetical to the document since the component is context agnostic. Assuming I've understood this correctly. 

http://www.w3.org/WAI/PF/aria-practices/#accordion

I found these docs regarding accordions. I don't think it makes sense to apply this to all panels, but rather only to collapsable/accordion type panels.

atomanyih:
Considering all this discussion, here's the options we are considering:

## Note
### User writes:
```html
<Panel id="myPanel" />
```

### Pros:
* works by default

### Cons: 
* semantic meaning of `note` possibly not applicable


## TabPanel
### User writes:
```html
<div role="tablist>
    <button role="tab" aria-controls="myPanel" aria-expanded="true"/>
</div>
<Panel id="myPanel"/>
```

### Pros:
* semantic meaning of `tabpanel` possibly more applicable

### Cons:
* requires a control with the role `tab` contained in a `tabList`
* not applicable to single panels

## Note:
If we did this we'd probably want to hide the controls behind another component
```html
<PanelControl target="myPanel"/>
<Panel id="myPanel"/>
```
However, any panel will need an associated `PanelControl`

## BYO role
### User writes:
```html
<button arial-controls="myPanel" />
<Panel id="myPanel" role="user-provided" />
```

### Pros:
* we don't have to figure out what the user needs

### Cons:
* onus is on the user to properly use roles (which is hard. see above discussion)

---
It's possible that we actually don't need a role for the panel body at all. The [W3C documents](http://www.w3.org/TR/wai-aria/states_and_properties#aria-expanded) define `aria-expanded` as:

> Indicates whether the element, **or another grouping element it controls**, is currently expanded or collapsed.

This suggests that it's possible to leave `aria-expanded` off of the panel body as long as there is a controlling element:

```html
<button aria-expanded="true" aria-controls="myPanel" />
<div id="myPanel" />
```

The chrome accessibility developer tools see this as valid, but to me it feels ambiguous.
We're going to try a screen reader against it to see if it's an option.

PR #1090, opened 2015-07-31
[fixed] regression when clicking "static" modal backdrops

jquense:
Not sure where this fell to the floor, but clearly we have no tests around it (will get to that)

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -216,10 +216,14 @@ const Modal = React.createClass({
     let { animation, bsClass } = this.props;
     let duration = Modal.BACKDROP_TRANSITION_DURATION;
 
+    // Don't handle clicks for "static" backdrops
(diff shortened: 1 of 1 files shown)


taion:
Can you elaborate a bit? It looks like we don't go down this code path at all if `this.props.backdrop` is false: https://github.com/react-bootstrap/react-bootstrap/blob/v0.24.2/src/Modal.js#L183-L185

jquense:
hehe sorry, yeah its not for when its `false` its for when its `"static"`, right now static backdrops are also closing the modal when its clicked.

taion:
Oh, I see. Might want to drop a comment (or explicitly compare to `"static"`) there just to make it more clear what's happening.

taion:
:shipit:

Issue #1054, opened 2015-07-23
TabPane should pass different classNames to NavItem and tab-pane


kennyw12:
We would like to style the NavItem and tab-pane div within TabPane differently by passing in different classNames. 

This functionality currently isn't supported by passing in className on a TabPane, since a TabPane className prop is passed to both the TabPane and NavItem.
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/TabbedArea.js#L85

```
<TabPane navClassName="my-nav-item" paneClassName="my-tab-pane" {...myotherProps}>My Tab Pane</TabPane>
```

Would you be open to a PR?

/cc @stubbornella @gpleiss @matt-royal @ctaymor

stubbornella:
To give broader context, we are trying to be as consistent as we can about how className, id, and style attributes are passed into an element so that folks who want to modify components can anticipate this behavior. 

What we've settled on so far is that ```className```, ```id```, and ```style``` are passed to the outer-most wrapper element. ```InnerClassName``` is being applied to the inner element of a container. So, for example, on a panel, innerClassName gets passed through to the `.panel-body` element.

Tabs are a little more complicated because there are multiple tab panes, so we had thought that classNames applied to individual tabs would apply to only those tab-panes (not the tab itself), and innerClassName would apply to all tab-panes (we'd merge classNames). 

I'm realizing this is a little hard to explain in words... ;) we're happy to jump on a hangout or just do a PR so you can see what we are suggesting in code!

jquense:
bah yeah TabbedArea is a leaky component right now...I am having a bit of a hard time following what you folks are looking for, but it seems that passing the tabpane className to the navItem is causing the issue? We are definitely open to a PR, not sure if trying to maintain compatibility here makes sense...the current behavior of passing the className to both the navItem and the tabPane strike me as odd, and possibly a bug.

@react-bootstrap/collaborators anyone know _why_ this logic is here in the first place, it seems likes its always been there?

mtscout6:
The Tab component is another component in dire need of some love. If you're open to doing some work on it to clean it so the external API is clean and intuitive that would be great. This is right in line with our goals for our 1.0.0 release.

kennyw12:
We are happy to help with the design with this component. Are there any issues your team has encountered that we can keep in mind when considering changes to the API? We are already planning to review open PRs and issues regarding TabbedArea and TabPane.

mtscout6:
[...]

PR #1020, opened 2015-07-19
Remove factories

mtscout6:
Part of #825 

Superseded by #1019 note this one will go into the v0.25 release.

PR #1091, opened 2015-07-31
[changed] Reworks TabbedArea and TabPane API

kennyw12:
Renames TabbedArea to Tabs
Renames TabPane to Tab
Renames `tab` prop on Tab to `title`
className on Tab only applies to .tab-pane
className, id, and style on Tabs applies to outermost div
Removes BootstrapMixin from Tabs to remove faulty documentation.
  namely: bsSize and unsupported bsStyle options

#1054


taion:
Taking a very, very cursory glance, but it looks like this doesn't deprecate the components under the old names, but instead removes the old ones completely?

jquense:
This is super cool :+1: I haven't had a chance to dig in but I like the changes. As @taion tho we need to talk about a deprecation path for big changes like this where they are possible (the smaller className changes probably aren't deprecatable). Luckily it doesn't look like it will be too difficult in these instances :)

AlexKVal:
Looks nice :+1: :sparkle: 

But it should be pull requested against `v0.26-rc` and added to `v0.26 Release` Milestone.

And an additional PR with deprecations should be pull requested against `v0.25-rc`.

Like this: `factories deprecation`
https://github.com/react-bootstrap/react-bootstrap/pull/1019 deprecation messages => v0.24
https://github.com/react-bootstrap/react-bootstrap/pull/1020 removing of factories => v0.25

:cherries:

timv88:
+1 for adding className to the most outer element of the tabs. I think this should be a standard for every react-bootstrap component.

Currently I have to work around this by having the following CSS selector `> div:not([class]):not([id]) {...}`. Not cool.

kennyw12:
Hi, I don't see a `v0.26-rc` tag or branch on react-bootstrap/react-bootstrap. I also don't quite know what you mean by adding something to a Milestone. How do I do that?

I'm also not sure if I added deprecation warnings correctly in #1099.
Sorry.

jquense:
@kennyw12 what I would do is submit this PR against the next major version: `v0.25-rc` and without that PR also provide aliased components under the old names that warn that they are deprecated. So if a user uses `TabPane` they will get a warning but it will just render the now `Tab` component. That way users can switch to the new components slowly (over the course the `25` version anyway).

I probably would also split the commits up into maybe two commits for the sake of the change log, one for renaming components, and the other documenting the className shifts, since that is also a breaking chang.

kennyw12:
Just to clarify, sounds like we should have 3 commits:
1. Deprecates TabbedArea and TabPane(aliasing them to the new Tabs and Tab), and adds in Tabs and Tab (including both sets of tests and docs, the old docs marked with a deprecation warning). This is against v0.25-rc
2. Removes TabbedArea and TabPane, and all associated docs and tests. This is against v0.26-rc, which still needs to be created as a tag
3. Make the className shifts against 0.26-rc, on top of #2. (This change seems like it would not be able to be deprecated since it changes where classNames go.)

Would that work for you? If so, could you make the 0.26-rc tag?
Thanks!

The Pivotal UI Team
@atomanyih @ctaymor @gpleiss @kennyw12 @matt-royal @nicw @stubbornella

jquense:
There is no need to put the className changes in `0.26` not every breaking change needs a deprecation, since `0.25` is the next "major" version its fine to have breaking changes in it. The deprecation policy is just: "when a deprecation is possible, the features will be deprecated for a full major version before removed",

I will look into a `0.26-rc` branch, sorry our actual process is lagging behind the new deprecation policy.

kennyw12:
When a branch for `0.26-rc` is created, we will create a second PR against it to remove TabbedArea and TabPane.

The Pivotal UI Team
@atomanyih @ctaymor @gpleiss @kennyw12 @matt-royal @nicw @stubbornella

jquense:
@kennyw12 hey there, this PR is still against `master` you'll have to close this probably and re-PR against the `0.25` branch. Sorry for all the rigamarole, we appreciate the help :)

AlexKVal:
I like the proposal https://github.com/react-bootstrap/react-bootstrap/pull/1091#issuecomment-127578923 of @jquense :+1: 
@mtscout6 points to about the same here https://github.com/react-bootstrap/react-bootstrap/pull/1099#issuecomment-127624838

Then you won't need to make any additional PRs against `v0.26-rc`. :sparkle: 
(i.e. this https://github.com/react-bootstrap/react-bootstrap/pull/1099 is not needed then.)

And `react-bootstrap`'s team will remove deprecated code in the future with `v0.26-rc` version.
:cherries:

AlexKVal:
this is now like this:
screen shot 2015-08-05 at 11 02 14 am
and it should be re-based like this:
screen shot 2015-08-05 at 11 03 01 am

--
This is the way as I would doing it:
```sh
> git remote -v
origin ...
upstream	https://github.com/react-bootstrap/react-bootstrap.git (fetch)
upstream	https://github.com/react-bootstrap/react-bootstrap.git (push)
```
```sh
> git checkout master
> git fetch upstream
> git checkout v0.25-rc
> git branch tabs-rework
> git checkout tabs-rework
```
then you make your changes and PR them. 
And on the `GitHub` site then you choose branch against which you want merge it: `react-bootstra/v0.25-rc`

I hope it will help :cherries:

kennyw12:
Alright, we made the same PR against `v0.25-rc` (#1109). Closing this PR.

Thanks!

PR #1093, opened 2015-07-31
[changed] Update dependencies

taion:
Specifically, we need to update both `karma-webpack` and `karma` for our tests to stop failing.


jquense:
LGTM

Issue #646, opened 2015-05-11
BootstrapMixin solution for ES6 Classes


mtscout6:
Seeing as mixins are not supported with ES6 classes, we need to figure out a different way to handle what's going on in the `BootstrapMixin`.

As much as I love that I don't need to declare the three PropTypes `bsClass`, `bsStyle`, and `bsSize` they are not a good one size fits all solution. For example the `Button` component does not have css defined for `tabs` or `pills` which this implementation will allow. However, the current solution allows you to easy impose your own custom rules into the validation check; see #306, #440, and #496. @trevorr has done a great job with this so far and I'd still like to be able to support something like this, though I imagine it will get more complicated.

The current `getBsClassSet` function shouldn't be too diffucult to extract to a purely static function without any dependency on `this` so I'm not too concerned with that.

This is an issue that needs to be figured out sooner rather than later to continue allowing us to use ES6 Classes. (For example lack of support for this is blocking suggestions such as those proposed in https://github.com/react-bootstrap/react-bootstrap/pull/626#issuecomment-100986044)

taion:
My 2 cents, I think the cleanest way to do this is to add a class decorator that attaches the additional props types. Something like:

```js
@BootstrapProps
class Label extends React.Component { ... }
```

The syntax is a little bit unfamiliar because decorators are fairly new, but hopefully with time it will look normal/acceptable.

taion:
That said, there's nothing wrong with ES5 classes + mixins. Facebook still use that approach internally, per the comment on: https://www.codementor.io/ama/questions/4965867003/how-soon-after-a-react-update-does-facebook-update-their-react-dependent-projects

mtscout6:
The decorator still does not address the concern about the `Button` component not having support for `tabs` or `pills`, though still an interesting idea to think about.

AlexKVal:
I also don't see any issues with using existing `React.createClass` function 
with mixins support with ES6 code.

Facebook
> We just don't use the ES6 class syntax for anything that requires a mixin.

It is an official API for near future.

When Facebook (and community as a whole) finds new `the right way` instead of mixins,
then it will become feasible and clear how to approach.

I made a research about current - ideas - instead of mixins.
There are no any prevalent yet.

They all have cons and pros and no one of them completely solves the issue with mixins with ES6 classes.

My vision of it:
`pros`:
1) `React.createClass` function is an official API
2) Facebook uses it also in their main products
[...]

PR #1096, opened 2015-07-31
DropdownButton, SplitButton, DropdownMenu, MenuItem Rewrite

mtscout6:
Major part of #526 

Adds:
- Keyboard Navigation
- Aria properties for Assistive Technology
- Option to use custom menu with any of the dropdown variations
- `NavDropdown` component (Similar to `DropdownButton` but specifically for use within `Nav` components.
- `DropdownToggle` Component which can be used as an alternative to the `DropdownButton` `title` prop. Can either be directly imported or used as a static property of the `DropdownButton` with `DropdownButton.Toggle`. _(Useful should you want to use glyphicons or custom html within the toggle that's not a simple string.)_
- `SplitToggle` Similar to the `DropdownToggle` but targeted at the `SplitButton`'s toggle.

Changed:
- Event handling of all these components to be in line with #419 
- Written with ES6 class syntax

Removed:
- DropdownStateMixin - Everything is using ES6 class syntax so no more mixin usage

Work Remaining: (If anyone would like to help here I'd greatly appreciate it)
- [x] Update all doc examples
- [x] Implement all test stubs
- [ ] Add generic doc section for `onSelect` handlers per #419

Potential Problems that may still exist:
- [x] `Nav` component `onSelect` handler is passed to `NavDropdown` the signatures don't match so we will either need to switch over `Nav`, `NavBar` and `NavItem` to the new signature or add some kind of temporary adaptor till we can address them. I'm not sure which option will involve the least amount of work.

Minor inconveniences to figure out: (Should not prohibit this PR)
- PropType validation with objectRestSpread usage. (Part of #646)

I'm hoping that that's everything needed to put a bow tie on this.


jquense:
woohoo! gonna dig into this a bit tomorrow, I have a few cases at work where I can try it out.

mtscout6:
> I'm not sure which option will involve the least amount of work.
>  -mtscout6 (myself earlier)

Digging deeper this may be complicated pending the outcome of https://github.com/react-bootstrap/react-bootstrap/issues/419#issuecomment-126925902

taion:
:clap:

src/DropdownMenu.js
```diff
@@ -1,43 +1,155 @@
...
+      this._onDocumentClickListener.remove();
+    }
+  }
+
+  componentWillUnmount() {
+    this.unbindRootCloseHandler();
+  }
+
+  handleDocumentClick(event) {
+    let inTree = domUtils.isNodeInTree(event.target, React.findDOMNode(this));
```
jquense: I think this is probably now a duplicate of `domUtils.contains` which was added recently-ish

src/DropdownBase.js
```diff
@@ -0,0 +1,307 @@
...
+    this.refineMenu = this.refineMenu.bind(this);
+    this.refineToggle = this.refineToggle.bind(this);
+
+    this.childExtractors = [{
+      key: 'toggle',
+      matches: child => child.type === this.Toggle,
+      refine: this.refineToggle
+    }, {
+      key: 'menu',
+      matches: child => child.type === DropdownMenu || child.type !== MenuItem,
```
jquense: I am always a bit wary of type checks like this...I am not sure why someone might have multiple instances of RB but this would break if they did, perhaps that's correct...it does limit the ability to create custom menu components though. nit picky concerns i guess.

src/DropdownBase.js
```diff
@@ -0,0 +1,307 @@
...
+    singleMenuValidation(DropdownToggle, MenuItem),
+    menuWithMenuItemSiblings(DropdownToggle)
+  ]),
+
+  noCaret: React.PropTypes.bool,
+
+  pullRight: React.PropTypes.bool,
+  open: React.PropTypes.bool,
+
+  onRequestClose: React.PropTypes.func,
```
jquense: `onRequestClose` seems a bit anachronistic to me..I believe we have moved other handler prop names over to the more traditional event naming convention of just `onClose` `onHide`. It's definately a more accurate name, although i don't usually see this pattern in the React ecosystem lately (maybe just me). Perhaps there is a larger question of what consistency looks like on naming these things

src/DropdownBase.js
```diff
@@ -0,0 +1,307 @@
...
+  if (props.title !== undefined && title !== undefined) {
+    return new Error(`(title|children) ${component} - Must provide either a 'title' prop or a '${component}.Toggle' child, not both.`);
+  }
+
+  if (props.title === undefined && title === undefined) {
+    return new Error(`(title|children) ${component} - Must provide either a 'title' prop or a '${component}.Toggle' child`);
+  }
+}
+
+function childrenAsArray(children) {
```
jquense: children can also be a keyed object, which bit me doing modal changes. Perhaps we should have a proper util for this using `React.Children.forEach`

src/DropdownBase.js
```diff
@@ -0,0 +1,307 @@
...
+      } else {
+        children.push(child);
+      }
+    });
+
+    this.childExtractors.forEach(extractor => {
+      set[extractor.key] = extractor.refine(set[extractor.key], children, open);
+    });
+
+    return {
```
jquense: this whole method is super cool

jquense:
gonna try and help out with some docs tomorrow if I get some time :)

src/DropdownBase.js
mtscout6: Come to think of it this may be a problem with react-router-bootstrap's MenuItemLink component. Will need to rethink this.

src/DropdownBase.js
mtscout6: That's a reasonable name change.

src/DropdownBase.js
mtscout6: I'll lean on your experience for this one.

src/DropdownBase.js
mtscout6: :blush: I built it based on ideas you presented by namespacing components in your modal project before you joined the org.

src/DropdownBase.js
jquense: in Modal we just with with a static prop like `__isModalHeader` works well enough i think

mtscout6: That sounds reasonable, and easy to achieve with the MenuItem and MenuItemLink components.

I won't have time to work on it till the weekend though.

jquense:
Do we have further consensus on whether we are pushing forward with the onSelect event, or going to try a different approach?

src/DropdownMenu.js
```diff
@@ -1,43 +1,155 @@
...
+
+  componentDidUpdate() {
+    if (this.props.open) {
+      this.bindRootCloseHandler();
+    } else {
+      this.unbindRootCloseHandler();
+    }
+  }
+
+  bindRootCloseHandler() {
```
jquense: It would be great to just use RootCloseWrapper here, but I don't think it'll work as it adds a wrapping div, @taion any thoughts?

mtscout6:
@jquense I like your earlier proposal for the events, it dramatically simplifies things.

src/DropdownMenu.js
taion: w/r/t `RootCloseWrapper`, I originally wrote it to use a wrapping `div` because I couldn't be sure that whatever custom popover I wrapped it with wouldn't do something dumb with `onClick`.

For something like this where we control what gets rendered (i.e. with the outer-most `ul`), I'd modify `RootCloseWrapper` to have a separate path that just injects the `onClick` prop instead of using the wrapping `div`.

jquense: Thats sort of the rub here as well, folks can now provide custom menu components, so we don't know for certain that the component will handle it...

taion: `#render()` here always renders a `ul` as the outer-most DOM node, and we control it directly, no? Unless I'm misreading this code. i.e. we can just inject `onClick` into that `ul` from `RootCloseWrapper` painlessly.

jquense: bah sorry yes. I was conflating this question along with another one about whether it makes sense to push the root click to the DropdownBase instead of the menu, so all menus' benefit from it.

taion: Oh sorry, I've barely gotten started reading through the rest of the PR.

How much does the rendering break if we add the wrapping `div`?

src/DropdownBase.js
taion: The usage in `menuWithMenuItemSiblingsValidator` looks like it can just be `React.Children.forEach` (or maybe our valid children only one?).

The usage in `singleMenuValidator` seems like it could benefit from e.g. a `ValidComponentChildren.filter`, though.

src/DropdownBase.js
```diff
@@ -0,0 +1,307 @@
...
+import MenuItem from './MenuItem';
+import CustomPropTypes from './utils/CustomPropTypes';
+import createChainedFunction from './utils/createChainedFunction';
+
+import find from 'lodash/collection/find';
+
+const TOGGLE_REF = 'toggle-btn';
+
+export default class DropdownBase extends React.Component {
+  constructor(props, Toggle=DropdownToggle) {
```
taion: I believe the 2nd param to `React.Component#constructor` is supposed to be `context`.

Could we just use `this.constructor.Toggle` below? Or set `Toggle` on the prototype?

taion: On second thought, `this.constructor.Toggle` will break everyone using IE <= 10 because static inheritance won't work there, so I'd suggest putting `Toggle` on the prototype, e.g. below do `SplitButton.prototype.Toggle = SplitToggle`.

taion:
This is super cool.

src/DropdownBase.js
```diff
@@ -0,0 +1,307 @@
...
+  }
+
+  refineMenu(menu, children, open) {
+    const menuProps = {
+      ref: 'menu',
+      open,
+      labelledBy: this.props.id
+    };
+
+    if (menu === undefined) {
```
taion: cc @jquense 

This is actually the opposite of what we decided to do for `Modal`, i.e. the support for custom modal dialogs is going into react-overlays, while the R-B `Modal` intentionally has fairly limited support (and a different API) for specifying a custom modal dialog class.

It might be nice to be consistent here if possible.

jquense: I was thinking about that myself and i am a bit torn about this one. If there is a more generic dropdown implementation that could be refactored out, its probably to early to know what that'd look like yet. On the other hand i think the level of customizability here is probably on par with what you could do with bootstrap proper. We, for instance gut the built in dropdown (in our non react projects), and essentially just use the styles. this api would let me do that in react, which I really appreciate.

taion: Sure - the right answer might just be to make the R-B `Modal`'s `customDialogComponent` API look a little more like this. I don't think it'd affect anybody deciding between using R-B `Modal` and react-overlays, but it'd make things line up better between R-B modals and dropdowns.

jquense: Ya definitely, as it is, the custom dialog prop is sort of a 'good enough' solution for the time anyway. I'm all for consistency

mtscout6:
@jquense With that PR complete did you see anything else that's needed before we bring this in? If it's complete I can do a squash and rebase

taion:
@mtscout6 I think we still need to fix the way we pass in `Toggle` to the constructors, and figure out if we're okay with being inconsistent between here and `Modal` for how we specify custom menus v custom modal dialogs.

jquense:
So there is a good point here: https://github.com/react-bootstrap/react-bootstrap/issues/526#issuecomment-130473990 the current PR will break that usecase, using a custom menuItem, since the extractors look for not known types and assume they are menu's.

We can always say that these cases need to have a menu wrapped around them, or perhaps create a more public api for telling the dropdown about types.

taion:
@jquense An alternative might be to use something like `dialogComponent` (maybe `menuComponent`?) per `Modal` to specify the custom menu class.

jquense:
@taion i would like to keep the current way if possible, it seems the more ergonomic of the two api (to me anyway). One issue with that approach tho is that you'd have to do something similar for the Button and Toggle components as well and that seems ugly...

Ultimately the issue is that the Dropdown really needs to know what things are menu items, or toggles, or menus. The most straight forward way to me is just to duck type it with a static property or something, not super elegant.

taion:
That makes sense. What do you think of doing the same for `Modal`, then, and moving toward deprecating `dialogComponent` there?

jquense:
seems reasonable to me, tho that case is a little different, since the user is almost always going to specify a custom dialog, whereas here they may just specify a menu to add some classes or something.

The other thing might be that maybe we can jsut get rid of that feature on the modal completely, and point ppl to react-overlays if they need something more custom...I still need to see what the diff is between that and our implementation tho, before I can say its easy to do.

taion:
To make sure I'm not misunderstanding you, is your proposal that the `children` of a dropdown should be:

1. Components that are explicitly labeled with a role (toggle, button, or menu); at most one of each
2. Some set of components, none of which are explicitly labelled with a role

If the user does (1), then we use the explicitly labeled components. If the user does (2), then everything proceeds per the default.

The alternatively would be something like exposing `toggleComponent` (is this just the existing `title` prop?) and `dropdownButton` props and forcing the explicit use of those.

Thinking a little more, I'm not sure which semantically feels better. I like the elegance of specifying everything under `children`, but the difference between the custom and non-custom API looks a little magical; I don't really have a better answer, though.

taion:
To add to that, one difference between this and the API for `Modal` is that the children of the `Modal` have no special semantic meaning (except for the purposes of injection `onHide` into the header), and otherwise just behave as a children array.

This is a bit different in that the different `children` have different semantic meanings; e.g. it looks like `Toggle` and `Menu` would be invariant to things like re-orderings, and they won't get passed around as-is.

jquense:
I think maybe some sort of `type` prop might be a good general API. If we could do something fairly foolproof we could not have to mess with things like linkProps or other awkward child pass throughs. not sure it would work, but it would be cool if it did. I don't like the magic, but elements as props always feels really awkward to me as well

jquense:
examples of what I am thinking of...

```js
<DropdownButton>
  <Button bsRole='toggle' >open</Button>

  <CustomItem bsRole='menuitem'>Item 1</CustomItem>
  <MenuItem>Item 2</MenuItem> //role set via DefaultProps
</DropdownButton>
```

I want to say that whatever the API is it is really an escape hatch. Most folks shouldn't need this sort of behavior, In general it will be other library authors (like react-router-bootstrap).

In this particular case, I think the "magic" is in assuming certain components are Menus. It seems like the perhaps we should be explicitly checking for menus and toggles, and just assume everything else is a MenuItem, which preserves the current behavior.

essentially a component has two "modes" for dealing with children, if a bsRole is defined on any child, then there is a strict whitelist of components allowed, only children with a role, and only whitelisted roles for that component. Otherwise it just does its default, such as wrapping children in a menu.

jquense:
ok I checked a few things off I think, Are more pending items before we merge this? In liee of general onSelect documentation perhaps we should document controlled/uncontrolled props in getting started or something? If so does that block the PR here?

jquense:
anything else here? LGTM otherwise...

mtscout6:
Looks good to me, though I don't have time to do the squashing for this before tomorrow.

webpack/docs.config.js
```diff
@@ -21,6 +21,9 @@ if (options.debug) {
 }
 
 export default _.extend({}, baseConfig, {
+
+  devtool: options.debug ? 'source-map' : null,
```
taion: Probably shouldn't change this here either.

src/RootCloseWrapper.js
```diff
@@ -75,13 +85,23 @@ export default class RootCloseWrapper extends React.Component {
...
 RootCloseWrapper.propTypes = {
-  onRootClose: React.PropTypes.func.isRequired
+  onRootClose: React.PropTypes.func.isRequired,
+
+  /**
+   * Passes the suppress click handler directly to the child component instead of
+   * placing it on a wrapping div. Only use when you can be sure the child will properly handle the
+   * click event.
+   */
+  noWrap: React.PropTypes.bool
```
taion: Is this getting backported into react-overlays? :D

taion:
This is really cool BTW. Thanks so much for all the work you've put into this - the new "custom" API is fantastic.

src/RootCloseWrapper.js
jquense: I think i already did? :P I actually have a better idea about this idea...

jquense:
Ya it really is super sweet and going to save me a ton of work in the long run, thanks @mtscout6 !

src/RootCloseWrapper.js
taion: I mean the `noWrap` bit - I don't think that's in react-overlays yet?

mtscout6:
I'm going to give the rebase and squash a go right now.

mtscout6:
Well, that rebase went quicker than I was expecting.

mtscout6:
:sad: Linting killed it, I'll get that fixed. Give me a few minutes.

mtscout6:
Linting is all fixed and everything is now squashed.

taion:
Well what are we waiting for then? :shipit: :shipit: :shipit: :D

AlexKVal:
Yay :tada: Finally :rocket: 
It was a long way.

adjohnson916:
+1 for sub-components instead of `title` attributes on components like `NavDropdown`.

PR #1099, opened 2015-08-03
[changed] Deprecating TabPane and TabbedArea

kennyw12:
TabPane and TabbedArea are now deprecated. Please use Tab and Tabs
instead.

Signed-off-by: Kenny Wang <kwang@pivotal.io>

#1091

### DIFF
--- a/src/TabPane.js
+++ b/src/TabPane.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import classNames from 'classnames';
 import TransitionEvents from './utils/TransitionEvents';
+import deprecationWarning from './utils/deprecationWarning';
 
 const TabPane = React.createClass({
   propTypes: {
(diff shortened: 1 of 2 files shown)


mtscout6:
Generally when we plan to rename something or move it then the deprecation process is to do the rename with some backwards compatible port. 

For example see: 

- https://github.com/react-bootstrap/react-bootstrap/blob/v0.24.3/src/Input.js#L8-L11
- https://github.com/react-bootstrap/react-bootstrap/commit/51a205f

We moved the `static` input type into a separate component, but still left a temporary bridge that redirects to the new component. The deprecation warning is then there notifying folks about the bridge, since we will remove that bridge in the future.

Also with that `collapsable` => `collapsible` spelling correction we let both be legal but logged a deprecation warning for those using the old spelling.

This process allows downstream users to fix one problem at a time without needing a version bump that breaks their site.

bman654:
Would have been nice to see this mentioned in the [ChangeLog](https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md)...

taion:
It's right under: https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md#v0250---tue-25-aug-2015-185633-gmt

PR #1100, opened 2015-08-04
[fixed] only add aria-expanded to Collapse when an ARIA role is present

kennyw12:
Also remove redundant aria-expanded from Panel collapsible body

We decided to close #1087 because it didn't really make sense to add a specific role for a generic component.
Removing aria-expanded from components that don't have roles doesn't change the functionality of JAWS or VoiceOver, so we decided that it may be better for the consumer to specify the role of the content.

The Pivotal UI Team
@atomanyih @ctaymor @gpleiss @kennyw12 @matt-royal @nicw @stubbornella

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -48,7 +48,7 @@ class Collapse extends React.Component {
       <Transition
         ref='transition'
         {...this.props}
-        aria-expanded={this.props.in}
+        aria-expanded={this.props.role ? this.props.in : null}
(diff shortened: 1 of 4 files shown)


jquense:
LGTM I think this a better way to go for now.

hopefully when we take another look at the panel component (already thinking about it) we can provide a more "default on" solution to the ARIA here

taion:
:+1: :shipit:

PR #1103, opened 2015-08-04
[fixed] prevent click if PaginationButton disabled

### LINKED ISSUE #1102: Prev, Next, Last, First buttons on Pagination handle clicks even if they are disabled
When active page is the last one then button "Next" is disabled but click event is still handled and `activePage` is incremented  

It can be checked on the demo page http://react-bootstrap.github.io/components.html#pagination-more

krzysiek1507:
Fixes #1102


krzysiek1507:
Superseded by #1104

PR #1108, opened 2015-08-05
Remove reference to factories from documentation

mtscout6:
Forgot to remove this when we removed factory support.


AlexKVal:
Yup  and I am too :blush:

PR #1110, opened 2015-08-05
[added] Custom feedback icon for Input

bzf:
Reference: #502

- Added a new prop on Input named feedbackIcon
- Added tests for ensuring Glyphicons as default
- Added tests for rendering the feedbackIcon

This allows for passing custom icons that may be used instead of the default Glyphicons when an Input is having feedback.

### DIFF
--- a/src/InputBase.js
+++ b/src/InputBase.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import classNames from 'classnames';
 import FormGroup from './FormGroup';
+import Glyphicon from './Glyphicon';
 
 class InputBase extends React.Component {
   getInputDOMNode() {
@@ -91,17 +92,20 @@ class InputBase extends React.Component {
   }
 
   renderIcon() {
-    let classes = {
-      'glyphicon': true,
-      'form-control-feedback': true,
-      'glyphicon-ok': this.props.bsStyle === 'success',
-      'glyphicon-warning-sign': this.props.bsStyle === 'warning',
-      'glyphicon-remove': this.props.bsStyle === 'error'
-    };
+    if (this.props.hasFeedback) {
+      if (this.props.feedbackIcon) {
+        return React.cloneElement(this.props.feedbackIcon, { formControlFeedback: true });
+      }
 
-    return this.props.hasFeedback ? (
-      <span className={classNames(classes)} key="icon" />
-    ) : null;
+      switch (this.props.bsStyle) {
+        case 'success': return <Glyphicon formControlFeedback glyph="ok" key="icon" />;
+        case 'warning': return <Glyphicon formControlFeedback glyph="warning-sign" key="icon" />;
+        case 'error': return <Glyphicon formControlFeedback glyph="remove" key="icon" />;
+        default: return <span className="form-control-feedback" key="icon" />;
+      }
+    } else {
+      return null;
+    }
   }
 
   renderHelp() {
@@ -214,6 +218,7 @@ InputBase.propTypes = {
   bsSize: React.PropTypes.oneOf(['small', 'medium', 'large']),
   bsStyle: React.PropTypes.oneOf(['success', 'warning', 'error']),
   hasFeedback: React.PropTypes.bool,
+  feedbackIcon: React.PropTypes.node,
   id: React.PropTypes.string,
   groupClassName: React.PropTypes.string,
   wrapperClassName: React.PropTypes.string,
--- a/src/Glyphicon.js
+++ b/src/Glyphicon.js
@@ -7,12 +7,14 @@ const Glyphicon = React.createClass({
   mixins: [BootstrapMixin],
 
   propTypes: {
-    glyph: React.PropTypes.oneOf(styleMaps.GLYPHS).isRequired
+    glyph: React.PropTypes.oneOf(styleMaps.GLYPHS).isRequired,
+    formControlFeedback: React.PropTypes.bool
   },
 
   getDefaultProps() {
     return {
-      bsClass: 'glyphicon'
+      bsClass: 'glyphicon',
+      formControlFeedback: false
     };
   },
 
@@ -20,6 +22,7 @@ const Glyphicon = React.createClass({
     let classes = this.getBsClassSet();
 
     classes['glyphicon-' + this.props.glyph] = true;
+    classes['form-control-feedback'] = this.props.formControlFeedback;
 
     return (
       <span {...this.props} className={classNames(this.props.className, classes)}>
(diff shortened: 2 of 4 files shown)


AlexKVal:
I did not look through it deeply enough, I just want to propose:
you could use `Glyphicon` component http://react-bootstrap.github.io/components.html#glyphicons then 
instead of all those: [`<span className="form-control-feedback glyphicon glyphicon-remove" key="icon" />`](https://github.com/react-bootstrap/react-bootstrap/pull/1110/files#diff-7aadaedfaad3ec27f373abd724524f81R231)
:wink:

bzf:
@AlexKVal: Gotcha! :smile: :+1:

ewal:
Nice! :+1: for this feature.

taion:
It does make sense to use `Glyphicon` here. I wonder, though, if it wouldn't be better just to have a `feedbackIcon` prop that could give you the default when set to e.g. `true`, but give you the custom one of your choice (unconnected to `bsStyle`) when set to a component.

bzf:
@taion: That sounds great! Maybe we just would like to set icons on an `Input` regardless of it uses `hasFeedback` or not.

Since we're the ones who set the feedback on the `Input`, maybe we could just add a `customIcon` prop. We know which icon we want for each case, so wouldn't it make sense to just let us set it when we `render`? Then you'd be able to use whatever icon you want, even when you don't want the feedback functionality.

```js
let myIcon = (
  <Glyphicon glyph="star" />
);

// Fallback on the regular Glyphicons
<Input hasFeedback bsStyle="success" />

// Use a custom icon
<Input customIcon={myIcon} />

// Use the custom icon over the default Glyphicon
<Input hasFeedback bsStyle="success" customIcon={myIcon} />
```

Thoughts?

dozoisch:
@bzf I like that api

taion:
I'd call it `feedbackIcon` just to indicate the role the icon (typically) plays - distinguishes it from e.g. the input group stuff.

bzf:
I updated the PR to use a `feedbackIcon` prop instead.

However, I couldn't get a custom validation working for the prop that checked if the class `form-control-feedback` is set on the given prop. I think this should be added before merging to avoid confusion and show an error when using it.

src/InputBase.js
```diff
@@ -91,17 +92,20 @@ class InputBase extends React.Component {
...
-    let classes = {
-      'glyphicon': true,
-      'form-control-feedback': true,
-      'glyphicon-ok': this.props.bsStyle === 'success',
-      'glyphicon-warning-sign': this.props.bsStyle === 'warning',
-      'glyphicon-remove': this.props.bsStyle === 'error'
-    };
+    if (this.props.hasFeedback) {
+      if (this.props.feedbackIcon) {
+        return this.props.feedbackIcon;
```
jquense: Maybe this should clone and add the `form-control-feedback` class instead of relying on the user to provide it? I am not sure if that unnecessarily limits use cases though.

taion: I think you pretty much need `.form-control-feedback` for the icon to render correctly, no? Not sure how to enforce this, though.

bzf: @taion: Yeah exactly, so I guess there's two ways to do this. Either
* do as @jquense said: clone the prop, add the class and render it
* add a validation which shows a warning and put the responsibility on the user passing the prop

I'm not sure which way would be preferred though...

jquense: I am actually running into a similar issue with custom menus in the new dropdown stuff. One one hand forcing the user to reset the style if they want different behavior than `form-control-feedback` stinks, but I think that use case is less likely then someone just wanting to swap out the actual icon, in which case adding the class is the better option.

taion: Perhaps we could at least make `formControlFeedback` or whatever a prop on `Glyphicon`? Saves a bit of typing.

jquense: I like that idea ^^ I think that's a great compromise

bzf:
@jquense, @taion: I added a `formControlFeedback` prop onto the `Glyphicon` component. :smiley:

src/InputBase.js
```diff
@@ -91,17 +92,20 @@ class InputBase extends React.Component {
...
+      }
 
-    return this.props.hasFeedback ? (
-      <span className={classNames(classes)} key="icon" />
-    ) : null;
+      switch (this.props.bsStyle) {
+        case 'success': return <Glyphicon formControlFeedback glyph="ok" key="icon" />;
+        case 'warning': return <Glyphicon formControlFeedback glyph="warning-sign" key="icon" />;
+        case 'error': return <Glyphicon formControlFeedback glyph="remove" key="icon" />;
+        default: return <span className="form-control-feedback" key="icon" />;
```
jquense: hmm this seems weird, I think if it doesn't match a supported style we should render nothing (`null`), especially since now they can handle it themselves if they want some custom thing.

I guess removing this would amount to a breaking change? It seems crazy that someone would be relying on this behavior but i suppose they could be...

src/InputBase.js
jquense: I think we still need to clone the element and inject the formFeedback prop so passed in Glyphicons will automatically render correctly

bzf: @jquense: Shouldn't we inject the class only, since people might not want to use Glyphicons?

src/InputBase.js
bzf: Yeah, returning `null` instead will fail a test in `InputSpec.js`.

src/InputBase.js
jquense: nope, that was the point of adding the prop, it'll just work in the 90% case of a user just using a Gylphicon, but if they want to provide some custom icon component that handles form feeback different'y then can provide one

src/InputBase.js
jquense: ok then leave it be, we can clean that wart in a breaking release

src/InputBase.js
bzf: Ah okay, gotcha! :+1:

taion: That actually makes a lot more sense than what I had in mind, heh.

jquense:
LGTM except for the commit messages, can you rewrite them according to the [guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes)

bzf:
@jquense: Like so?

jquense:
yup, just need one more sign off from a @react-bootstrap/collaborators

taion:
:shipit:

AlexKVal:
:+1: :sparkle:

PR #1125, opened 2015-08-08
Dropdown docs and tweaks

jquense:
Adds a doc example highlighting the new child components for customization.

It also "fixed" a few things to get them to work consistently, though its up to ya'll whether those were good ideas. Also let me know if the example styling is too much :P :rainbow: 

cc @mtscout6 

related to: #1096

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -183,7 +183,7 @@ const ComponentsPage = React.createClass({
                   <h3><Anchor id='btn-dropdowns-custom'>Dropdown Customization</Anchor></h3>
                   <p>
                     If the default handling of the dropdown menu and toggle components aren't to your liking, you can
-                    customize the them, by explicitly specifying a Toggle or Menu as children.
+                    customize them, by explicitly specifying a Toggle or Menu as children.
                   </p>
                   <div className='bs-callout bs-callout-info'>
                     <h4>Additional Import Options</h4>
@@ -195,6 +195,10 @@ const ComponentsPage = React.createClass({
                     </p>
                   </div>
                   <ReactPlayground codeText={Samples.DropdownButtonCustom} />
+
+                  <h4>Custom Menus</h4>
+                  <p>You can even create your own menu implementation.</p>
+                  <ReactPlayground codeText={Samples.DropdownButtonCustomMenu} />
 
                   <h3><Anchor id='btn-dropdowns-props'>Props</Anchor></h3>
 
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -180,6 +180,22 @@ const ComponentsPage = React.createClass({
                   <p>Trigger dropdown menus that align to the right of the button using the <code>pullRight</code> prop.</p>
                   <ReactPlayground codeText={Samples.SplitButtonRight} />
 
+                  <h3><Anchor id='btn-dropdowns-custom'>Dropdown Customization</Anchor></h3>
+                  <p>
+                    If the default handling of the dropdown menu and toggle components aren't to your liking, you can
+                    customize the them, by explicitly specifying a Toggle or Menu as children.
+                  </p>
+                  <div className='bs-callout bs-callout-info'>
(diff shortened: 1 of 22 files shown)


docs/src/ComponentsPage.js
```diff
@@ -180,6 +180,22 @@ const ComponentsPage = React.createClass({
                   <p>Trigger dropdown menus that align to the right of the button using the <code>pullRight</code> prop.</p>
                   <ReactPlayground codeText={Samples.SplitButtonRight} />
 
+                  <h3><Anchor id='btn-dropdowns-custom'>Dropdown Customization</Anchor></h3>
+                  <p>
+                    If the default handling of the dropdown menu and toggle components aren't to your liking, you can
+                    customize the them, by explicitly specifying a Toggle or Menu as children.
```
mtscout6: Minor grammar correction here, change to "customize them"

mtscout6:
This looks great! The one customization example that I was envisioning that still isn't there would be a custom Menu. Something like:

```
<DropdownButton id='dropdown-custom-1'>
      <DropdownButton.Toggle>
        <Glyphicon glyph='star' />
        Pow! Zoom!
      </DropdownButton.Toggle>
      <CustomMenu>
        <MenuItem eventKey='1'>Action</MenuItem>
        <MenuItem eventKey='2'>Another action</MenuItem>
        <MenuItem eventKey='3' active>Active Item</MenuItem>
        <MenuItem divider />
        <MenuItem eventKey='4'>Separated link</MenuItem>
      </CustomMenu>
    </DropdownButton>
```

Where you could do something other than an immediate `<ul>`, for example a dropdown with an input that filters the list of items. That's not needed for this PR though.

After the minor grammar correction this LGTM.

jquense:
I wanted to do a custom menu example although (admittedly I didn't try) It didn't seem like it was possible, given how the DropdownBase validates it's children? I was under the impression that in order to be accepted it had to match a type check. Of course I could just be reading it wrong.

mtscout6:
If it doesn't match the type check then it's assumed to be a Custom menu ;)

jquense:
doh, yeah i see it now. great, I will add that as well!

mtscout6:
You're awesome thanks for filling in the holes!

taion:
ref https://github.com/react-bootstrap/react-bootstrap/pull/1096/files#r36700537, https://github.com/react-bootstrap/react-bootstrap/blob/v0.24.4/src/Modal.js#L117

It would be nice to be consistent between how we specify custom modal dialogs and custom dropdown menus, if possible/convenient.

jquense:
Finished tests, and tweaked whatever was needed to make that happen

test/DropdownButtonSpec.js
```diff
@@ -486,19 +517,27 @@ describe('DropdownButton', function() {
       buttonNode.getAttribute('aria-expanded').should.equal('true');
     });
 
-    it('when open and the key "esc" is pressed the menu is closed and focus is returned to the button', function() {
+    // I am not sure why this test has to be so convoluted.
+    // the naive approach would fail _only_ if the entire test suite was run together, but I couldn't isolate it any more
```
mtscout6: I'm not sure I follow this statement.

mtscout6:
Other than that code comment I mentioned this all looks great!

test/DropdownButtonSpec.js
jquense: hehe, sorry, my quick attempt to explain why this test is more complicated looking then it seems like it should be. 

Essentially the test as it was, would sporadically fail, when the entire test suite was run. It would pass with flying colors when just that test, or even all the dropdown tests, were run. I tried to isolate which tests were causing the conflict but didn't have any luck. I reworked the test, so it doesn't sometimes fail (it still makes the same assertions).

I feel fairly confident that the failure was due to a test specific conflict and not an actual bug.

mtscout6: Ok, that makes sense. Could you perhaps take a moment and re-word that comment to reflect that more clearly?

jquense:
ok updated!

PR #1131, opened 2015-08-11
Minor fixes for react-14

ncthis:


taion:
Thanks for the PR. How much breaks if we just switch this to React Router v1.0.0?

taion:
Superseded by #1153.

PR #1133, opened 2015-08-11
Updated alert to restart timer on update

omerts:
After a re-render the dismissTimer did not restart, causing the new alerts to disappear before their time. Also, if a re-render has changed the dissmissAfter, it was ignored.

### DIFF
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -18,6 +18,12 @@ const Alert = React.createClass({
       closeLabel: 'Close Alert'
     };
   },
+  
+  _initDismissTimer() {
+    if (this.props.dismissAfter && this.props.onDismiss) {
+      this.dismissTimer = setTimeout(this.props.onDismiss, this.props.dismissAfter);
+    }
+  },
 
   renderDismissButton() {
     return (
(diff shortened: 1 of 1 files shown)


jquense:
hey there thanks for the PR. I am not quite sure that this is the right way to go here. Restarting on update could put the component in a loop, or keep it visible forever if the parent component is fairly active. 

I am a bit unclear on what use case this is actually solving, if a you want a _new_ alert I would render a completely new alert, not reuse the existing one.

omerts:
@jquense What do you mean by rendering a new alert? 
here is my use case:
```javascript
import React from 'react';
import {Alert} from 'react-bootstrap';

export default class Alerter extends React.Component {
  constructor(props) {
    super(props);
  }

  _getAlerts() {
    if (!this.props.msgs || this.props.msgs.length === 0) {
      return null;
    }

    let msgAlerts = this.props.msgs.map((msg) => {
      return <Alert bsStyle={msg.style}>
              {msg.text}
             </Alert>
    });

    return msgAlerts;
  }

  render() {
    if (this.props.msgs && this.props.msgs.length > 0) {
      return <Alert bsStyle='default' dismissAfter={this.props.dismissAfter} onDismiss={this.props.onHideAlerts}>
              {this._getAlerts()}
             </Alert>;
    }

    return null;
  }
}

Alerter.defaultProps = {dismissAfter: 15000}
```

Even if i re-render my alerter component with new messages every second, it will still close after 15 seconds from the initial render. If a new message arrived I would like to restart the 15 seconds timer, otherwise some messages will just show for a second.

Maybe it should be a property on the alert, if to restart timer on update?

taion:
I'm not sure why we have support for timeouts there at all - they're not present upstream, and I feel like they're better handled by user code.

For this case, I think the right thing would definitely be to not use the supplied timeout management and use your own, anyway.

PR #1146, opened 2015-08-12
Clean up tabs documentation a little bit

taion:
1. The "extends tabbed navigation" bit was just copied from upstream for no particular reason - it doesn't make any sense here since it's not visible to the user at all
2. As far as I can tell we don't actually support dropdowns here, or I don't see how to use them anyway and it's not in the examples
3. The upstream affix sidebar just says "Tab" (while using the full "Togglable tabs" as the section title), so we may as well be consistent-ish and make the section easier to find

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -515,7 +515,7 @@ const ComponentsPage = React.createClass({
                 <div className='bs-docs-section'>
                   <h1 className='page-header'><Anchor id='tabs'>Togglable tabs</Anchor> <small>TabbedArea, TabPane</small></h1>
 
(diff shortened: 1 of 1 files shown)


jquense:
We do theoretically support Dropdown menus, since we support them in navs. But we don't seem to have a good way of doing that in tabs. We should figure out a way to handle that...

http://react-bootstrap.github.io/components.html#navs-dropdown

jquense:
LGTM tho, sorry.

taion:
In principle we could render them in the nav bar, but as set up right now, there's no way to specify this. We'd need to add a `DropdownTabGroup` component or something.

PR #1148, opened 2015-08-12
changed onSelect behavior

jquense:
I kept the changes a localized to the dropdown as possible, this was actually helpful just in that it uncovered a few focus bugs (and made some tests simpler).

changed: onRequestClose -> onToggle

onSelect passes the eventKey directly instead of the selection event.

### DIFF
--- a/src/DropdownBase.js
+++ b/src/DropdownBase.js
@@ -19,8 +19,7 @@ export default class DropdownBase extends React.Component {
     this.toggleOpen = this.toggleOpen.bind(this);
     this.handleClick = this.handleClick.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
-    this.handleRequestClose = this.handleRequestClose.bind(this);
-    this.handleSelect = this.handleSelect.bind(this);
+    this.handleClose = this.handleClose.bind(this);
     this.extractChildren = this.extractChildren.bind(this);
 
     this.refineMenu = this.refineMenu.bind(this);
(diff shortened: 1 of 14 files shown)


jquense:
anyone have any feels about this one?

src/DropdownBase.js
```diff
@@ -288,27 +305,80 @@ export function menuWithMenuItemSiblings(...nonMenuTypes) {
...
+  /**
+   * Whether or not the Dropdown is visible.
+   *
+   * @controllable onToggle
+   */
   open: React.PropTypes.bool,
 
-  onRequestClose: React.PropTypes.func,
+  /**
+   * A callback fired when the Dropdown wishs to change visibility. Called with the requested
```
taion: wishs -> wishes

taion:
I like it. Can we put `uncontrollable` on `DropdownBase` though, or does it have to go on the concrete components?

Also, I wonder if `uncontrollable` could be used as a decorator... heheheh.

jquense:
Given the current setup it can't go on the Base b/c the other components would then be exending the wrapping HoC not the actual dropdown base.

We can totally make it a decorator, i had been meaning to make that work.

taion:
Aah, got it - that's a little annoying though, that anybody who extends this on their own would have to separately use the controller.

If we go with your proposal where we just have a `Dropdown` component with the full children-based syntax, and just keep the special-cased ones as wrappers for `Dropdown`, would we still need each of those to be wrapped with `uncontrollable`?

jquense:
I am almost about to PR that proposal, but no you only need to wrap the base Dropdown component in that case

taion:
Sweet. Is that on top of this? i.e. do we need to merge this first?

jquense:
yup, merging this first would be great

PR #1151, opened 2015-08-13
Remove wrong componentClass='input' Button tests

taion:
These are testing something weird/wrong, because `Button` injects its children as the children of the component used, which is not legal for `input`.

### DIFF
--- a/test/ButtonSpec.js
+++ b/test/ButtonSpec.js
@@ -12,16 +12,6 @@ describe('Button', function () {
     assert.equal(React.findDOMNode(instance).nodeName, 'BUTTON');
   });
 
-  it('Should output a component with button classes', function () {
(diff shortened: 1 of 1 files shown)


taion:
Can someone take a look at this? I'd like to cut v0.24.5 and v0.24.5-react-pre.0 once this is merged.

jquense:
LGTM, though shouldn't we be testing whether the input is created with a `value` vs children? that or validate against using input as a component class? Thats a bit of another issue.

taion:
There's plenty of other component classes that would make no sense, though. I don't really want to enumerate the ones that we can assert do make sense, and `button` and `a` are pretty reasonable for almost all cases, except when the user wants a custom React component or something (in which case they own the problem anyway).

Issue #613, opened 2015-05-04
Remove iosClickHack() from Modal


AlexKVal:
iosClickHack()
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js#L125
I've checked all related info and patches facebook/react#1169.
I've found out that the patch fixing this problem is not present in v0.13.2 yet. 

But it already is in v0.14-alpha and I've checked that alpha with deleted iosClickHack().
All works fine. Therefore it is only a matter of time when we can remove that hack.

taion:
Fixed by #1153.

PR #1153, opened 2015-08-13
Update for React v0.14

taion:
Big things are

- Upgrade to React-Router 1.0.0-beta3 for docs (for React v0.14 compatibility) - most material change was switching `Root` props to statics
- Included changes #1151 
- Rendering into subtree in portals
- Using `ReactDOM` instead of `React` where appropriate
- In many places, DOM components are treated differently from composite components and lead to invariant violations when attempting to e.g. read their props or do `ReactTestUtils.find` operations on them, so treat those as just normal DOM nodes instead

All remaining tests pass, and the docs still look reasonable.


taion:
One more:

- Warnings now go to `console.error`

taion:
I want to release this as `v0.24.4-react-pre.0` once this is reviewed. Given the quantity of changes, this is going to be very annoying to keep in sync, but I wanted to have this for experimenting with Relay.

taion:
This fixes #1153.

jquense:
looks good tho we should not use fbjs

jquense:
fat fingers, sorry about the closure :P

test/CollapsibleMixinSpec.js
```diff
@@ -67,13 +70,13 @@ describe('CollapsibleMixin', function () {
...
     beforeEach(function(){
-      instance = ReactTestUtils.renderIntoDocument(
+      instance = render(
         <Component>Panel content</Component>
       );
     });
 
     it('Should have collapsing class', function () {
-      instance.setProps({expanded:true});
+      instance.renderWithProps({expanded:true});
```
jquense: If this is using my helper I believe it needs to be `instance = instance.renderWithProps(...)` otherwise its working on the old value, interesting that this doesn't fail tho....

mtscout6:
> I want to release this as `v0.24.4-react-pre.0` once this is reviewed.

May I suggest `v0.24.4-react14.0` that way it's clearer what this release contains based on the version number.

test/CollapsibleMixinSpec.js
taion: I guess React is pretty aggressive about re-using component objects, and of course the mount point doesn't change... but I'll fix it anyway.

taion:
@mtscout6 

I think `react14` is a little weird as a pre-release tag because we'd end up re-using it if/when React does get to v14. Also, `react14.0` looks a little weird (i.e. I parse it as being the pre-release for React 14.0, rather than pre-release 0 for React 14).

I think `react-pre` is reasonably unambiguous because we're going to bump our major version with React version updates anyway, and there shouldn't be more than one pre-release React version for any of our breaking-change versions.

taion:
@jquense 

Switched to `warning`. Bound the return value of `renderWithProps` in those cases where we reference the re-rendered instance after re-rendering (but not in others).

taion:
Also handled #613 and the relevant portions here of https://github.com/react-bootstrap/react-router-bootstrap/issues/77 (with an update to the test to make sure that things still work as expected).

jquense:
lgtm, you have a fixup! in there tho

taion:
Yup, will squash before merging.

AlexKVal:
A lot of work. And a lot will be to merge it in the future :sweat_smile: 
Cool :sparkle:

PR #1156, opened 2015-08-13
Improve tabs accessibility

matt-royal:
Add role="tablist" to the nav inside of tabs. Also backfill a test for adding aria-selected to the nav item for the currently active tab.

The Pivotal UI Team
@atomanyih @ctaymor @d-reinhold @kennyw12 @matt-royal @nicw @stubbornella

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -85,7 +85,7 @@ const Tabs = React.createClass({
     }
 
     let nav = (
-      <Nav {...props} activeKey={this.getActiveKey()} onSelect={this.handleSelect} ref="tabs">
+      <Nav {...props} activeKey={this.getActiveKey()} onSelect={this.handleSelect} ref="tabs" role="tablist">
         {ValidComponentChildren.map(this.props.children, renderTabIfSet, this)}
       </Nav>
     );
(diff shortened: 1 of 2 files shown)


jquense:
LGTM, can you squash and set the commit message according to the guidelines here: https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes

taion:
LGTM as well once squashed.

matt-royal:
@jquense We can definitely squash the commits together. Since we don't think we changed any APIs in this PR, we're unsure of what changes you'd like to see in the commit message. Could you please clarify?

taion:
It should be `[changed]` because of the (very minor) change to what the public API does.

matt-royal:
Thanks, @talon. Please let us know if the new commit message isn't what you had in mind.

taion:
Oh `[added]` is fine too.

jquense:
That looks great! Changing ARIA affects the interactions for SR users so its fair to consider it an "api change" in general we are pretty broad on that definition anyway.

PR #1163, opened 2015-08-14
[added] base Dropdown component

jquense:
cc @mtscout6 @taion 

This approach seems to work well, it makes a few things a bit more explicit that the base component approach hid, like `this.Toggle` and the handling of click props, etc.

It does offer slightly less ease of use for some customization cases i think, but more flexibility overall. I am not sure where the basic Toggle should go its really more of the DropdownButton.Toggle, but that would make the general usage of Dropdown a bit uglier (and require a second import) but it would have symmetry with SplitButton.Toggle....


taion:
Don't have a chance to do a full review at the moment, but looking at the examples + the new code for the helper components - those look really sweet. :+1:

jquense:
Another thought: as an alternative to bsRole would could just provide empty wrappers the Dropdown understands like:

```js
<Dropdown>
  <Toggle>
    <Button>hi</Button>
  </Toggle>
</Dropdown>
```

where Toggle is just:

```js
render(){
  return cloneElement(this.props.children, toggleProps)
}
```

I am not sure what we do with the existing Toggle and Menu components because they are still needed and useful. This approach may just create confusion between the Toggle component and the Toggle signaler.

We could alternatively provide some sort of HoC helper that does the same thing

src/DropdownButton.js
```diff
@@ -1,50 +1,46 @@
...
 
-class DropdownButton extends DropdownBase {
   constructor(props) {
     super(props);
   }
 
   render() {
-    if (this.props.navItem) {
-      return <NavDropdown {...this.props} />;
-    }
```
mtscout6: I didn't realize that I hadn't added a deprecation warning inside that If block. May still be worth including?

mtscout6: Ah, just saw that you moved it down. We will want a deprecation warning on it anyhow.

src/NavDropdown.js
```diff
@@ -1,47 +1,32 @@
...
+          useAnchor
+          bsStyle={bsStyle}
+          disabled={props.disabled}
+          noCaret={noCaret}
+        >
+          {title}
+        </Dropdown.Toggle>
+        <Dropdown.Menu>
+          {children}
+        </Dropdown.Menu>
```
mtscout6: By not inspecting the children here you are not allowing outside customization.

mtscout6: Ok, delayed understanding. You are proposing that these methods are not customizable, but if you want to roll your own with the Dropdown component you can. I'm catching your drift now.

test/DropdownSpec.js
```diff
@@ -0,0 +1,447 @@
...
+      </Dropdown>
+    );
+    const node = React.findDOMNode(instance);
+
+    node.tagName.should.equal('DIV');
+    node.className.should.not.match(/\bdropdown\b/);
+    node.className.should.match(/\bdropup\b/);
+  });
+
+  it('renders toggle with DropdownButton.Toggle', function() {
```
mtscout6: `Dropdown.Toggle` now right?

webpack/test.config.js
```diff
@@ -6,5 +6,5 @@ export default _.extend({}, baseConfig, {
     pathinfo: true
   },
 
-  devtool: 'eval'
+  //devtool: 'sourcemap'
```
mtscout6: Doesn't this play weird tricks on failing test stack traces?

src/DropdownButton.js
mtscout6: Never mind don't mind me, I forgot that I did that in the propType validation (Shamefully walking away, I've slept since writing that code :smile:)

mtscout6:
I really like this, it cleans it up and removes that base class that I was uncomfortable with anyways. I think this goes well with @taion's concern of having too many ways to do things with the same component. Very nice Job! I approve 100%

taion:
I like having a role prop - the example where you just set it on a random DOM component is pretty nice.

Instead of a full-fledged HoC, I'd just add a decorator that injects the appropriate role into `defaultProps` for the relevant component (assuming it's an ES6 class component).

test/DropdownSpec.js
jquense: ahh yes, thanks :)

webpack/test.config.js
jquense: sorry that was a test of something i just didn't undo. silly mistake

src/NavDropdown.js
jquense: yeah, I should have have made an attempt to explain this PR in the initial message, instaed of jsut throwing over the wall, but i got pulled away by the baby. Sorry about that

mtscout6: Those darn kids :wink:

jquense:
ok made those few cosmetic tweaks. Anything else substantive we want to do here?

taion:
Sorry, was too busy playing with Relay this weekend - will look at this tomorrow morning.

mtscout6:
Looks great to me!

PR #1169, opened 2015-08-17
[added] #460 ListGroupItem outputs <button> when an onClick handler is set.

apkiernan:


taion:
Is this for http://getbootstrap.com/components/#list-group-buttons?

apkiernan:
Sorry, submitted this late last night and forgot to add a comment. Yes, this is for [http://getbootstrap.com/components/#list-group-buttons] (http://getbootstrap.com/components/#list-group-buttons). We were planning on doing this originally during #467, but Bootstrap didn't support the similar style for buttons. Now that it does, this can be added.

taion:
:shipit:

taion:
Thinking about this a little more, the `isAnchorOrButton` check is really not ideal. Consider the case where I have a custom e.g. `MyListGroupItem` that doesn't take an `href` prop but instead itself injects an `href` prop into a `ListGroupItem`.

With the current implementation, I end up with a `ul` that contains a number of `a`s, which renders fine but makes for very odd markup. The wrapping component really should be a `div` in this case.

I'm not sure what the right answer is here. I'm not sure we can ever tell a priori that using a `ul` is correct here.

/cc @jquense @mtscout6: any thoughts?

mtscout6:
priori?

I can't say that I'm following your example completely.

jquense:
["a priori"](http://www.iep.utm.edu/apriori/) whipping out the philosophy terms :P brings me back.

I think the problem is just that we can't with any sort of certainty know if a child is going to render a li tag or not, and so we can't know whether to render a ul as the parent, since the user can use any custom list item component they want. I sort of think we should just solve this with a componentClass prop on listGroup and let the user decide?

taion:
Makes sense. Should we have special-case handling for the case where `children` actually are instances of `ListGroupItem`, and do not have `onClick` or `href` in `props`?

mtscout6:
Ok, I've slept so I'm following the problem now :smile:. I think it would be ideal if we did the special case handling when the components are known to us because they are from our library. While allowing a propType override for those users doing some custom work.

mtscout6:
In any case I think we should open a new issue and continue the discussion there.

apkiernan:
Opened #1181 to continue discussion

Issue #937, opened 2015-07-03
Use CustomPropTypes from react-prop-types


AlexKVal:
https://github.com/react-bootstrap/react-prop-types

I'm on it, but before `react-prop-types` has to be released on npmjs.

--
backport [new custom validators](https://github.com/react-bootstrap/react-bootstrap/pull/1096/files#diff-6e40506f419cfbb5c93d15e804cc90fdR47) `requiredRoles(...roles)` and `exclusiveRoles(...roles)`

Update: `requiredRoles` `exclusiveRoles` are specific for `R-B` project. (e.g. `bsRole` property), hence it is not necessary to backport them.

AlexKVal:
Because now `utils/CustomPropTypes` is public API [src/utils/index.js#L10](https://github.com/react-bootstrap/react-bootstrap/blob/d105749b0a48fdb8976405f6e0648e85c286d788/src/utils/index.js#L10)
it seems it needs to be first deprecated in 0.24 and only in 0.25 it can be removed and https://github.com/react-bootstrap/react-prop-types could be used instead.
Right ?

mtscout6:
My understanding of the new rule: https://github.com/react-bootstrap/react-bootstrap/blob/v0.24-rc/CONTRIBUTING.md#breaking-changes

The deprecation warning should not land till 0.25, and the removal in 0.26. So essentially patch releases should just be bug fixes and new features. So from what I understand our users would not see new deprecation warnings until the next minor release.

taion:
On second thought, it might make sense to defer that rule - it's actually a decent amount of work to merge in and validate a bunch of changes, quite a lot more than I thought it'd be. Maybe just quietly delete that part until/unless we get better at releasing minor versions more quickly.

mtscout6:
I think that as we start following that on weeks with active development we could push for a minor release every two weeks. Instead of holding off for so long.

mtscout6:
Lean on the fact that Semver describes what's happening. If we reach a minor version over 150 (arbitrary number) before we release 1.0.0 I don't think it matters.

AlexKVal:
Let's continue this in gitter for maintainters.

AlexKVal:
Updated topic about backporting new custom validators introduced in #1096
Just not to forget.

AlexKVal:
This is resolved by #1223

PR #768, opened 2015-05-31
[added] utils namespace

tomatau:
Exposes `domUtils`, `CustomPropTypes` and other useful utilities.

Original PR here https://github.com/react-bootstrap/react-bootstrap/pull/745

This has a nicer commit history

PR #1172, opened 2015-08-17
Dom utils

jquense:
beginning prep work for swapping out overlay stuff with react-overlays. Removes our dom helpers and uses dom-helpers so both packages can share the same helpers. Bonus, we finally remove our jquery dependency for ie8.

Slightly unrelated I removed domUtils from the public API, why where they there in the first place? If folks would like a deprecation we could do that, tho I think it would be cleaner just to point ppl to dom-helpers if they need those utils. In either case i seemed weird to me for RB have them as a part of a public api

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -1,6 +1,8 @@
 /* eslint react/no-did-mount-set-state: 0 */
 
 import React from 'react';
+import getOffset from 'dom-helpers/query/offset';
+import css from 'dom-helpers/style';
 
 import Affix from '../../src/Affix';
 import Nav from '../../src/Nav';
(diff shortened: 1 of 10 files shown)


taion:
For history there see:

https://github.com/react-bootstrap/react-bootstrap/issues/743
https://github.com/react-bootstrap/react-bootstrap/pull/745
https://github.com/react-bootstrap/react-bootstrap/pull/768

Moving them into separate packages is a much better way to go, though.

/cc @tomatau

AlexKVal:
If we are removing [domUtils](https://github.com/react-bootstrap/react-bootstrap/pull/1172/files#diff-1fdf421c05c1140f6d71444ea2b27638L82) without transitional deprecation warning,
then I propose to do the same with `CustomPropTypes`. https://github.com/react-bootstrap/react-bootstrap/pull/1063

And also remove them in favor of https://github.com/react-bootstrap/react-prop-types `npm package`

jquense:
I don't mind adding a deprecation warning, just wondering if it makes sense too. I do think we should do the same thing for propTypes tho :)

AlexKVal:
I have `react-prop-types` local branch prepared to PR after `v0.25`,
but if we decide to proceed w/o transitional deprecation warning, I will PR them against the `v0.25` branch.

taion:
LGTM other than that one comment about the docs webpack config... just want to understand what it's for.

AlexKVal:
@taion should I PR my local branch for https://github.com/react-bootstrap/react-bootstrap/issues/937 against `v0.25-rc` then (instead of `26`) ?
(https://github.com/react-bootstrap/react-bootstrap/pull/1172#issuecomment-131823637)

taion:
@jquense This is good to merge after we strip the docs Webpack config changes.

@AlexKVal Let's keep it with the deprecation warning for now.

AlexKVal:
@taion OK. No problem. :cherries:

PR #1174, opened 2015-08-17
added react-overlays for base overlay components

jquense:
Step 1. just replaces the stuff that required minimal refactoring, and no api changes. The next step is to rewrite the Modal component.

I am also unsure what we should do for components that are straight reexports: Position, Portal, Transition. Does it make sense to just re-export them? Or do we want to pull them out of our public api and point ppl to react overlays?

Then there are components like Overlay with have reallly light additions to the base Overlay component


taion:
For v0.25 I think we'd have to keep them around with deprecation warnings (+ forwarding the old props to the new props) - then delete for v0.26.

jquense:
I am working on proper deprecations but there is the immediate problem of what to do about the documentation for these components. Since they are no longer in the same package the metadata parser is not going to work.

jquense:
Do we want to try and get the modal migration in for 0.25 as well? I good with either, there are a few bugs fixed upstream but nothing serious.

heavy migration work more or less done in: https://github.com/jquense/react-bootstrap-modal so it _shouldn't_ be to much work, but those are famous last words :P

jquense:
Anything else here?

I added a deprecation warning for Transition, Position, and Portal. Does everyone feel comfortable telling folks to go to react-overlays for those components?

taion:
LGTM

taion:
Ah, good - just the two merge commits + my one actual change show up here.

jquense:
Awesome. :shipit: I guess?

Issue #1073, opened 2015-07-27
Affix in IE10 - scrollHeight


MaaikeB:
I know the Affix is not finished for general use yet, but I'm using it in my projects and it works well.

...however... apart from IE10.
Here it gets the 'affix-bottom' class where it is not supposed to, and then the affixed element jumps out of the visible view port. 

This happens because the AffixMixin uses the variable: 
scrollHeight = document.documentElement.offsetHeight
to get the height of the entire page. 

document.documentElement.offsetHeight returns in all the browsers indeed the height of the entire page (in my case 3482), though in IE10 it returns a different value (878).

a StackOverflow question suggests a way to get the height of the entire page in all different browsers is:
pageHeight = Math.max(document.height, document.body.scrollHeight, document.body.offsetHeight)
(http://stackoverflow.com/questions/16384340/internet-explorer-gives-total-webpage-height-as-0)

When I use this to assign the scrollHeight, the Affix indeed works well also in IE10.

burtyish:
Would one of the maintainers please respond on this issue? @MaaikeB has put the time into making a PR, so a response would be nice.

taion:
You need to actually submit the PR - we can't really do much with just a commit on a fork.

MaaikeB:
Thanks, will do.

burtyish:
I mistakenly thought a PR was made. Thanks for the response, @taion

AlexKVal:
Closed by https://github.com/react-bootstrap/react-bootstrap/pull/1180#issuecomment-134212761

PR #1180, opened 2015-08-18
Affix in ie10

MaaikeB:
PR - Affix in IE10 - scrollHeight #1073


taion:
I wonder if we'd want to put this in `domUtils`.

Can you squash your commits?

mtscout6:
I agree with @taion this seems reasonable to live in [`domUtils`](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/domUtils.js)

I'm not sure about the breaking build looks like some eslint problems @AlexKVal are those problems fixed on master? I believe that rebasing on the latest master should resolve that.

mtscout6:
@jquense Are all the dom utils in flight with #1172 does this change make sense to be in `domUtils`?

jquense:
They are in flight, but not all of them are going, just the ones that are covered by `dom-helpers` which this isn't. I would move it to domUtils, I'd also make sure it uses the owner document of the affix node, not just the global document, in case of iframes

AlexKVal:
Eslint warnings are fixed now by https://github.com/react-bootstrap/react-bootstrap/pull/1185
I've restarted both CI-es.

taion:
@MaaikeB 

Per this discussions here, go ahead and make this something like `function getDocumentHeight(document)` in `domUtils`, then squash your commits, and we should be good to go.

src/utils/domUtils.js
```diff
@@ -79,6 +79,15 @@ function getOffset(DOMNode) {
 }
 
 /**
+ * Get the height of the document
+ *
+ * @returns {documentHeight: number}
+ */
+function getDocumentHeight() {
```
taion: You need to actually export this.

taion:
Great - can you squash all of these down into 1 commit? LGTM otherwise. Need one more reviewer from @react-bootstrap/collaborators .

mfunkie:
LGTM!
I'm going to have to start importing domUtils in my own projects, as I've hit this kinda stuff before.

taion:
I don't know how `af6654f` ended up in there. Rather than go back again, I squashed and merged this PR manually.

PR #1183, opened 2015-08-18
Document Button's default type attribute

insin:
I just answered a [Stack Overflow question related to the default `type` of a `<Button>`](http://stackoverflow.com/questions/32074972/reactjs-onsubmit-on-form-doesnt-work) and noticed the default value wasn't documented (since it's assigned in a render method as it doesn't always apply).

This PR adds support for a new `@defaultValue` doclet and uses it to document the default for when `type` applies.

Should this also be mentioned as a prose section in the docs? Defaulting to `'button'` is IMO a nicer default than `'submit'`, but the difference in default behaviour compared to a regular HTML `<button>` is a potential gotcha.

### DIFF
--- a/docs/generate-metadata.js
+++ b/docs/generate-metadata.js
@@ -64,6 +64,11 @@ function applyPropDoclets(props, propName){
   if ( doclets.required) {
     prop.required = true;
   }
+
+  // Use @defaultValue to provide a prop's default value
+  if (doclets.defaultValue) {
+    prop.defaultValue = cleanDocletValue(doclets.defaultValue);
+  }
 }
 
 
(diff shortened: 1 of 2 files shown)


jquense:
hey! thanks for the PR this is cool, and definately helpful, though In this particular case with the button I wonder if its better to just explicitly define a defaultProp instead of the doclet and `|| 'button'`.

I either way the doclet code is helpful for plenty of other cases as well.

insin:
I assumed a default prop wasn't  being used to avoid passing a `type` when spreading this.props in `renderAnchor()` when rendering as a link.

AlexKVal:
Eslint warnings are fixed now by #1185
I've restarted both CI-es.

jquense:
LGTM need one more from a @react-bootstrap/collaborators

taion:
LGTM

Issue #1158, opened 2015-08-13
Use NavbarBrand component for Navbar brand


taion:
There's no real reason to have `brand` be a prop on `Navbar` - it's a sibling of the `Nav` rendered inside it, so it'd make just as much sense as a separate component. The new API should then be e.g.:

```js
<Navbar>
  <NavbarBrand>My Brand</NavbarBrand>
  <Nav>{/* ... */}</Nav>
</Navbar>
```

kwnccc:
Hi @taion , I'm new to this repo. I'd love to start contributing to it. I was thinking of starting this implementation if there's no one else working on it.

taion:
Nobody as far as I'm aware. Go for it!

taion:
Was resolved by #1184. Thanks @kwnccc !

PR #1184, opened 2015-08-19
Added NavBrand component, including docs

kwnccc:
as described in issue #1158: Use NavbarBrand component for Navbar brand

### DIFF
--- a/src/NavBrand.js
+++ b/src/NavBrand.js
@@ -1,0 +1,38 @@
+import React, { cloneElement } from 'react';
+import BootstrapMixin from './BootstrapMixin';
+import classNames from 'classnames';
+
+const NavBrand = React.createClass({
+  mixins: [BootstrapMixin],
+
+  propTypes: {
+    bsRole: React.PropTypes.string,
+    navbar: React.PropTypes.bool
+  },
+
+  getDefaultProps() {
+    return {
+      bsRole: 'brand',
+      navbar: false
+    };
+  },
+
+  render() {
+    let brand;
+
+    if (React.isValidElement(this.props.children)) {
+      brand = cloneElement(this.props.children, {
+        className: classNames(this.props.children.props.className, 'navbar-brand'),
+        bsRole: this.props.bsRole,
+        navbar: this.props.navbar
+      });
+    } else {
+      brand = <span {...this.props} className="navbar-brand">{this.props.children}</span>;
+    }
+
+    return brand;
+  }
+
+});
+
+export default NavBrand;
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -5,6 +5,7 @@ import classNames from 'classnames';
 import ValidComponentChildren from './utils/ValidComponentChildren';
 import createChainedFunction from './utils/createChainedFunction';
 import CustomPropTypes from './utils/CustomPropTypes';
+import deprecationWarning from './utils/deprecationWarning';
 
 const Navbar = React.createClass({
   mixins: [BootstrapMixin],
@@ -57,6 +58,12 @@ const Navbar = React.createClass({
     return !this._isChanging;
   },
 
+  componentDidMount() {
+    if (this.props.brand) {
+      deprecationWarning('Navbar brand attribute', 'NavBrand Component');
+    }
+  },
+
   handleToggle() {
     if (this.props.onToggle) {
       this._isChanging = true;
@@ -73,6 +80,19 @@ const Navbar = React.createClass({
     return this.props.navExpanded != null ? this.props.navExpanded : this.state.navExpanded;
   },
 
+  navbrandChild() {
+    let navChild =
+      ValidComponentChildren.findValidComponents(this.props.children, child => {
+        return child.props.bsRole === 'brand';
+      });
+
+    return navChild;
+  },
+
+  hasNavbrandChild() {
+    return this.navbrandChild().length > 0;
+  },
+
   render() {
     let classes = this.getBsClassSet();
     let ComponentClass = this.props.componentClass;
@@ -82,14 +102,28 @@ const Navbar = React.createClass({
     classes['navbar-static-top'] = this.props.staticTop;
     classes['navbar-inverse'] = this.props.inverse;
 
+    let displayHeader = (this.props.brand || this.props.toggleButton || this.props.toggleNavKey != null) && !this.hasNavbrandChild();
+
     return (
       <ComponentClass {...this.props} className={classNames(this.props.className, classes)}>
         <div className={this.props.fluid ? 'container-fluid' : 'container'}>
-          {(this.props.brand || this.props.toggleButton || this.props.toggleNavKey != null) ? this.renderHeader() : null}
-          {ValidComponentChildren.map(this.props.children, this.renderChild)}
+          {displayHeader ? this.renderHeader() : null}
+          {ValidComponentChildren.map(this.props.children, this.renderChildren)}
         </div>
       </ComponentClass>
     );
+  },
+
+  renderNavBrand(child, index) {
+    let navbrandEl = cloneElement(child, {
+      navbar: true,
+      toggleNavKey: this.props.toggleNavKey,
+      toggleButton: this.props.toggleButton,
+      handleToggle: this.handleToggle,
+      key: child.key ? child.key : index
+    });
+
+    return this.renderHeader(navbrandEl);
   },
 
   renderChild(child, index) {
@@ -101,10 +135,14 @@ const Navbar = React.createClass({
     });
   },
 
-  renderHeader() {
-    let brand;
+  renderChildren(child, index) {
+    return (child.props.navbrand) ? this.renderNavBrand(child, index) : this.renderChild(child, index);
+  },
 
-    if (this.props.brand) {
+  renderHeader(navbrandEl) {
+    let brand = navbrandEl || '';
+
+    if (!brand && this.props.brand) {
       if (React.isValidElement(this.props.brand)) {
         brand = cloneElement(this.props.brand, {
           className: classNames(this.props.brand.props.className, 'navbar-brand')
@@ -146,6 +184,7 @@ const Navbar = React.createClass({
       </button>
     );
   }
+
 });
 
 export default Navbar;
--- a/src/utils/ValidComponentChildren.js
+++ b/src/utils/ValidComponentChildren.js
@@ -94,10 +94,39 @@ function find(children, finder) {
   return child;
 }
 
+/**
+ * Finds children that are typically specified as `props.children`,
+ * but only iterates over children that are "valid components".
+ *
+ * The provided forEachFunc(child, index) will be called for each
+ * leaf child with the index reflecting the position relative to "valid components".
+ *
+ * @param {?*} children Children tree container.
+ * @param {function(*, int)} findFunc.
+ * @param {*} findContext Context for findContext.
+ * @returns {array} of children that meet the findFunc return statement
+ */
+function findValidComponents(children, func, context) {
+  let index = 0;
+  let returnChildren = [];
+
+  React.Children.forEach(children, child => {
+    if (React.isValidElement(child)) {
+      if (func.call(context, child, index)) {
+        returnChildren.push(child);
+      }
+      index++;
+    }
+  });
+
+  return returnChildren;
+}
+
 export default {
   map: mapValidComponents,
   forEach: forEachValidComponents,
   numberOf: numberOfValidComponents,
   find,
+  findValidComponents,
   hasValidComponent
 };
--- a/test/NavbarSpec.js
+++ b/test/NavbarSpec.js
@@ -1,9 +1,11 @@
 import React from 'react';
 import ReactTestUtils from 'react/lib/ReactTestUtils';
 import Navbar from '../src/Navbar';
+import NavBrand from '../src/NavBrand';
 import Nav from '../src/Nav';
+import {shouldWarn} from './helpers';
 
-describe('Nav', () => {
+describe('Navbar', () => {
 
   it('Should create nav element', () => {
     let instance = ReactTestUtils.renderIntoDocument(
@@ -64,6 +66,13 @@ describe('Nav', () => {
     assert.equal(React.findDOMNode(instance).nodeName, 'HEADER');
   });
 
+  it('Should throw a deprecation warning message when brand is passed', () => {
+    ReactTestUtils.renderIntoDocument(
+      <Navbar brand="Brand" />
+    );
+    shouldWarn('deprecated');
+  });
+
   it('Should add header with brand', () => {
     let instance = ReactTestUtils.renderIntoDocument(
       <Navbar brand="Brand" />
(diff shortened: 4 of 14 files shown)


taion:
I think we actually want to leave a deprecation for the old `brand` prop rather than deleting it immediately. Looks good otherwise, though. Were there any tests covering this? Guess separately we need to fix the CI build.

AlexKVal:
Looks like guys from https://github.com/walmartlabs/eslint-config-defaults
doesn't follow semver. `4.1.0` brokes our linting process.
screen shot 2015-08-19 at 10 07 36 am

AlexKVal:
I've restarted `travisci` and `apveyor` after #1185
and now we have the real picture for tests :smile: 
screen shot 2015-08-19 at 10 37 43 am

--
This is here [test/NavbarSpec.js#L67](https://github.com/react-bootstrap/react-bootstrap/blob/daa21f20b03eeecff42fc7358cc9b66a88306f4a/test/NavbarSpec.js#L67)
You need to add 
```js
import {shouldWarn} from './helpers';
...
shouldWarn('deprecated');
```
like here [test/InputSpec.js#L50](https://github.com/react-bootstrap/react-bootstrap/blob/b688014c0857112b03273f0f73ef3545099c537a/test/InputSpec.js#L50)

AlexKVal:
@react-bootstrap/collaborators 
Guys, shouldn't it be done via two PR's ?
One (this) with deprecation warning against the next version `v0.25-rc` branch
and another with old component removing against `v0.26-rc` ?

Or at least this PR should be re-submitted against `v0.25-rc` ?

kwnccc:
@AlexKVal I haven't yet implemented the spec for the new component. The eslint error was an issue that you solved and merged to the branch? Really sorry for the dummy questions, but I'm a new contributor here. What about deprecation warning?

test/NavbarSpec.js
```diff
@@ -81,7 +84,9 @@ describe('Nav', function () {
 
   it('Should add header with brand component', function () {
     let instance = ReactTestUtils.renderIntoDocument(
-      <Navbar brand={<a>Brand</a>} />
+      <Navbar>
+        <NavBrand>{<a>Brand</a>}</NavBrand>
```
taion: Should just be

```js
<NavBrand><a>Brand</a></NavBrand>
```

hopefully?

taion:
@kwnccc 

That looks about right - the idea is to give people still on the old API a full version to migrate over before we make breaking changes, and to show them a deprecation warning so they know.

@AlexKVal 

What do you think of having a special helper for a deprecated prop, so this would be possible through `propTypes`?

test/NavbarSpec.js
kwnccc: You are totally right, will update asap.

kwnccc:
@taion this means that I have to leave also the Navbar specs as it previously were for brand attribute, or is it gonna fail because of the deprecation warning? After the last updates the brand attributes works fine in Navbar but it displays correctly the deprecation warning in console.

AlexKVal:
> @kwnccc The eslint error was an issue that you solved and merged to the branch? 

yeah. It was because of one of dependencies update. It was not because of your PR :wink: 

> Really sorry for the dummy questions, but I'm a new contributor here. 

It's OK. Actually I am too kinda intern in this project :blush: 

> @taion this means that I have to leave also the Navbar specs as it previously were for brand attribute, or is it gonna fail because of the deprecation warning?

In theory we should try to preserve old tests as is, but because now the deprecation warning is thrown, you should add `shouldWarn('deprecated');` to each failing (because of deprecation warning) test.

But in this particular case it maybe better to little refactor those tests.

Your tests:
```js
it('Should throw a warning message when brand is passed', function () {
  ReactTestUtils.renderIntoDocument(
    <Navbar brand="Brand" />
  );
  shouldWarn('deprecated');
});
```
This is OK, though I would make it more explicit: `...throw a deprecation warning message...`

About `Should add header with brand component` this test is for old component and you should not change it,
just add to the end `shouldWarn('deprecated');` to make it green.

About `Should pass navbar prop to navs` I would just remove `brand="Brand"` as you did and wouldn't add `<NavBrand>Brand</NavBrand>`.

--
to @taion about deprecated property helper
I did it straighforward [collapsable => collapsible property](https://github.com/react-bootstrap/react-bootstrap/pull/609/files#diff-a14adc7bab675b9c9e99be10dae7e817R6)
though it seems the time has come to add `a special helper for a deprecated prop`.
I'll try to design one and make another PR.

I believe that the way [deprecation done in this PR is OK](https://github.com/kwnccc/react-bootstrap/commit/cacc7f541304de07bb02d1c13e6839f8da8d085a#diff-9d651e5c92548c4d9e89560e0df67ce9R63)

AlexKVal:
@kwnccc You can squash the commits and change the commit message as prescribed in the [Contributing Guide](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) 
Doing so will allow for our automated changelog tool to document this on the next release.

I would form the commit message something like:
```
[added] NavBar component

...and here you add line by line what you've done.
.. tests..
..deprecation..
..etc..
```

and then you just 
```sh
> git push --force
```

--
At the times of my first contributions to this project, @mtscout6 gave me the link to this amazing video https://vimeo.com/43659036 about git. It helps me a lot :cherries:

taion:
This is LGTM to me once everything is squashed.

kwnccc:
Well, well! First things first, @AlexKVal @taion thank you for all the support and commenting above! You have been pretty descriptive and helped me work things out. Will definitely check this video.

Specs with Navbar brand attribute did fail usins <code>shouldWarn('deprecated')</code>
I have included these specs in order to preserve them as mentioned.

AlexKVal:
Only now I've got a change to look through the code a little more deeply.

According to http://getbootstrap.com/components/#navbar
`NavBar` should be just a dumb component with a class `navbar-brand`
```html
<a class="navbar-brand" href="#">Brand</a>
```
In the `R-B` it is a `span` element: [src/Navbar.js#L113](https://github.com/react-bootstrap/react-bootstrap/blob/c7e3f07cf992067437cf9e020e8d3869bf64c36a/src/Navbar.js#L113)
```js
brand = <span className="navbar-brand">{this.props.brand}</span>;
```

--
If a new component is introduced, then it should be refactored to use it instead of the old code:
particularly at least [this line](https://github.com/react-bootstrap/react-bootstrap/blob/c7e3f07cf992067437cf9e020e8d3869bf64c36a/src/Navbar.js#L113) should become something like this:
```js
brand = <NavBar>{this.props.brand}</NavBar>;
```
instead of the old one:
```js
brand = <span className="navbar-brand">{this.props.brand}</span>;
```

And new tests are for the new component should be in a separate file:
`Input` => `InputSpec`, `Button` => `ButtonSpec` etc :wink: 

--
Because `NavBar` component should be dumb, may I suggest to take a look at the `Row` component and its tests.
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Row.js
https://github.com/react-bootstrap/react-bootstrap/blob/master/test/RowSpec.js

:cherries:

AlexKVal:
BTW: you don't have to write such detailed description in the commit :wink: 
Just a two-three lines should be enough.
Though it is a matter of taste of course.
The main part is `the header`, which will go into https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md, which you did write :sparkle: 
:cherries:

src/NavBrand.js
```diff
@@ -0,0 +1,74 @@
...
+import BootstrapMixin from './BootstrapMixin';
+import classNames from 'classnames';
+
+import createChainedFunction from './utils/createChainedFunction';
+
+const NavBrand = React.createClass({
+  mixins: [BootstrapMixin],
+
+  propTypes: {
+    navbar: React.PropTypes.bool,
```
taion: What is this for?

src/Navbar.js
```diff
@@ -86,12 +105,22 @@ const Navbar = React.createClass({
       <ComponentClass {...this.props} className={classNames(this.props.className, classes)}>
         <div className={this.props.fluid ? 'container-fluid' : 'container'}>
           {(this.props.brand || this.props.toggleButton || this.props.toggleNavKey != null) ? this.renderHeader() : null}
```
taion: Isn't this going to re-render the header toggle bits even when a `NavBrand` is specified?

taion:
So on a closer look, this is actually implementing the entire header rather than just the brand, which makes sense, but means that `NavBrand` can't be a "dumb" component as much any more.

Maybe we can use `bsRole` instead of the `navbrand` prop (per e.g. https://github.com/react-bootstrap/react-bootstrap/blob/9dad63e03fe94d759c83b5fc3468a9e8dfcfa78c/docs/examples/DropdownButtonCustomMenu.js#L62 for the new dropdown API) to mark this.

Sorry this is turning out to be a bigger ordeal than we initially thought - I forgot about the rest of the header stuff.

AlexKVal:
It seems I didn't get the whole picture with the `NavBar` component and must apologize for my misdirecting comments. :cherries: 

It seems the part for the `NavBar` component is bigger that just
```html
<a class="navbar-brand" href="#">Brand</a>
```
and it is
http://getbootstrap.com/components/#navbar-default
```html
<!-- Brand and toggle get grouped for better mobile display -->
<div class="navbar-header">
  <button type="button" class="navbar-toggle collapsed" data-toggle="collapse" data-target="#bs-example-navbar-collapse-1" aria-expanded="false">
    <span class="sr-only">Toggle navigation</span>
    <span class="icon-bar"></span>
    <span class="icon-bar"></span>
    <span class="icon-bar"></span>
  </button>
  <a class="navbar-brand" href="#">Brand</a>
</div>
```
and it should be possible to render `Image`
http://getbootstrap.com/components/#navbar-brand-image
```html
<div class="navbar-header">
  <a class="navbar-brand" href="#">
    <img alt="Brand" src="...">
  </a>
</div>
```
And clicking to the `hamburger` button has to un-collapse navbar (within the mobile screen size).

--
Now I understand it right @taion ?

taion:
Pretty much - which means that contrary to what I initially thought, this is a contextual child component that needs to receive props from the parent `NavBar`.

src/NavBrand.js
kwnccc: I suppose you meaning navbar attribute? I've seen that also other components like <code>NavItem</code> use it being default to false. I suppose that it is needed when you either use it as an independent component or inside a <code>Navbar</code> component.

kwnccc:
Yes, I've seen that it also implements the entire header. Is it a problem if the <code>Navbrand</code> is not a dumb component? Don't we want to implement header inside navbrand? 

So what you are proposing is to replace navbrand entirely using bsRole?

kwnccc:
@taion @AlexKVal any update on this?

taion:
My available time for this project has been going into getting v0.25.0 release sorted out - will look at this once I have that all sorted out.

taion:
I've been pretty swamped lately. I've still been meaning to think about what to do here, but I don't have a great answer yet.

kwnccc:
Is there any suggestion in my comment above that could someway help to figure out a solution? Can you elaborate your thoughts on this, how do you imagine it could be implemented? I don't have a clear view in the whole project as you, but I'd really like to help and contribute.

taion:
@kwnccc 

The problem is https://github.com/react-bootstrap/react-bootstrap/pull/1184/files#r37707058 - basically if the user specifies `toggleButton` or `toggleNavKey` and passes in a `<NavBrand>`, the header will render twice. Unless I'm missing something...

kwnccc:
@taion You are totally right! I think that with this commit we resolve this issue and we also a really dummy <code>NavBrand</code> Component. What do you think?

kwnccc:
@taion @AlexKVal guys did you have the time to review the last commit? Let me know if it is what we need. :)

AlexKVal:
@kwnccc I'm sorry I will not comment on your new realization's variant.
This is @taion's idea https://github.com/react-bootstrap/react-bootstrap/issues/1158#issue-100879382 :smile: 

But nevetheless I will try to help you with merges and pushes :)

After your local changes you should rebase it before next force-push.

At the beginning you've done it like that:
```
> git create <your-branch>
// make your changes
> git add && git commit
> git push <to GitHub>
```
Then time goes by... and after some comments and discussions you've made some fixes,
but the main codebase of the project has gone forward too and you should resolve it like that:
```
// you've made your new fixes / changes
> git add && commit
// and here instead of just `git push --force`, 
// you should `actualize` your local codebase with GitHub repo:
> git checkout master
> git pull
> git checkout <your-working-branch-for-this-pr>
> git rebase master // <-- that is what you need to `actualize`
// and only then you
> git push --force
```
As the result of all of this `"git dancing"` there will no such additional commits:
`Merge remote-tracking branch 'upstream/master' into feature/navbrand`
`resolve eslint issues by TravisCI`

I hope it will help you 
:cherries:

src/NavBrand.js
taion: Well, it's used for something like: https://github.com/react-bootstrap/react-bootstrap/blob/1173725f31d6c1aad34cffab74039fa790cc2042/src/Nav.js#L80. The prop doesn't do anything else otherwise.

src/NavBrand.js
```diff
@@ -0,0 +1,36 @@
+import React, { cloneElement } from 'react';
+import BootstrapMixin from './BootstrapMixin';
+import classNames from 'classnames';
+
+const NavBrand = React.createClass({
+  mixins: [BootstrapMixin],
+
+  propTypes: {
+    navbar: React.PropTypes.bool,
+    navbrand: React.PropTypes.bool
```
taion: For consistency with how we do dropdowns, this should be something like e.g. `bsRole: 'brand'`. cc @jquense 

https://github.com/react-bootstrap/react-bootstrap/blob/09fe7a54cbb1461f1988ec321ef8227830afc698/src/DropdownToggle.js#L40

src/utils/ValidComponentChildren.js
```diff
@@ -82,6 +82,34 @@ function hasValidComponent(children) {
...
+ * @param {?*} children Children tree container.
+ * @param {function(*, int)} findFunc.
+ * @param {*} findContext Context for findContext.
+ * @returns {array} of children that meet the findFunc return statement
+ */
+function findValidComponents(children, func, context) {
+  let index = 0;
+  let returnChildren = [];
+
+  React.Children.forEach(children, function(child) {
```
taion: I think our linting rules require fat arrow here now.

src/Navbar.js
```diff
@@ -85,13 +105,25 @@ const Navbar = React.createClass({
     return (
       <ComponentClass {...this.props} className={classNames(this.props.className, classes)}>
         <div className={this.props.fluid ? 'container-fluid' : 'container'}>
-          {(this.props.brand || this.props.toggleButton || this.props.toggleNavKey != null) ? this.renderHeader() : null}
-          {ValidComponentChildren.map(this.props.children, this.renderChild)}
+          {((this.props.brand || this.props.toggleButton || this.props.toggleNavKey != null) && !this.hasNavbrandChild()) ? this.renderHeader() : null}
```
taion: Minor cosmetic thing, but can we bind this to a variable outside the return? This is quite hard to read.

taion:
Sorry for taking so long to get back to you - I got really swamped with some work projects.

This is looking really good now - I have a few more minor comments, but otherwise we're almost good to merge. Thanks for working on this and sorry it's taken so long.

kwnccc:
@taion thank you very much for reviewing the last commits! No worries at all, I understand there is a a great load of work to be done! Will also proceed with the updates on your comments and update branch. Should you find anything else let me know. I will also update the tests.

Also thanks @AlexKVal for your guidance through the git flow. It was really helpful! :)

taion:
Oops forgot to respond to this somehow. LGTM. Need one more reviewer from @react-bootstrap/collaborators.

AlexKVal:
LGTM
(though it is adding a bit of work to https://github.com/react-bootstrap/react-bootstrap/pull/1257, but this is a simple component and it won't be hard to remove that mixin from it)

@kwnccc Thank you for the contribution :tada: 
:cherries:

kwnccc:
@taion @AlexKVal thank you very much guys for all the help, the patience and the guidance through my first ever open source contribution. :tada:

In case there is something I can help with #1257 let me know. :)

AlexKVal:
Well.. Then you still can help :+1: 
In case you have some free time for it you can remove `BootstrapMixin` dependency from your component.

You can take as an example what I've done recently with another new component in this commit:
https://github.com/AlexKVal/react-bootstrap/commit/03482748b68ebba7de24491e957f705f48e32bd5
and the full story about it https://github.com/react-bootstrap/react-bootstrap/pull/1354

And of course you can read more about it at #1257.
:cherries:

kwnccc:
@AlexKVal I didn't find the time to solve the bug #1384 and remove the `BootstrapMixin`. Glad that @taion had some time to fix the code and update.

AlexKVal:
It's OK.

PR #1189, opened 2015-08-20
Update position if children changed

aparticka:


jquense:
I am pretty sure that change will mean that it will always update, since children is going to be a different object/array on every render

aparticka:
Yes, it seems I was a bit too quick with the pull request. I just need the overlay to update its position whenever the content inside of it changes and this fixed my issue. I'll close this and work on it some more.

PR #1190, opened 2015-08-20
Add manual update position flag

aparticka:


taion:
For consistency with other similar things, should name the prop `shouldUpdatePosition`. Otherwise, we're actually going to be releasing v0.25 soon I believe, which pulls out `Position`, so could you re-submit this against https://github.com/react-bootstrap/react-overlays?

aparticka:
Ok, will do.

Issue #983, opened 2015-07-13
Change Panel and ListGroupItem API


kylehood:
Per the documentation: http://react-bootstrap.github.io/components.html#panels, when using a panel, unless you specifically supply the header prop some heading tags, it will print the title area as normal text within the panel-heading div. If you wrap that Panel in an Accordion, supplying just a string to Panel's header prop will suddenly output a title wrapped in an H4.

I think this is inconsistent behavior. Is there any reason why it specifically creates an H4 tag in this instance? I would like to have an accordion panel where the heading text is just normal text and not an H4.

Panel.js:187 looks like it could be the culprit.

To reproduce create:

<pre>
   Panel header='test title' 
</pre>

Note that the title is rendered as normal text. Then try:

<pre>
   Accordion
        Panel header='test title'
   /Accordion
</pre>

The title now automatically becomes an H4

AlexKVal:
According to `BS3` docs http://getbootstrap.com/components/#panels-heading
there are two additional classes for panel headers
`.panel-heading` class to make `div` as header for panel 
and `.panel-title` class exactly for use with `<h1>-<h6>` headers

`R-B` `Panel` component has just `header` property - which indeed has inconsistent behavior
http://react-bootstrap.github.io/components.html#panels-props

It seems it's need to be re-factored. :sparkle: 

--
Let's think about new / changed API for this.
What is the best way to achive it ? Your ideas guys.

taion:
This API is poorly designed - we should have separate `heading` and `title` props.

jquense:
Been thinking about this, but all api's strike me as verbose or ugly...

```
<Panel>
  <Panel.Header>
    <Panel.Title>hi!</Panel.Title>
  </Panel.Header>
  <Table>...</Table>
</Panel>
```

maybe? we'd loose some of the functionality with making it collapsible. Maybe we jsut drop that as a feature to begin with? Or make it probably controlled all the time, and let the user handle the heading click to collapse?

taion:
I'd actually go with a more minimal change to the current API, where the user can do either

```js
<Panel heading="foo">
  bar
</Panel>
```

or

```js
<Panel title="foo" />
  bar
</Panel>
```

In the former we don't apply any of the `title` styling; in the latter we do. Plus spit out an error if the user tries to specify both `heading` and `title`... otherwise essentially the same as the current API, just slightly more explicit.

AlexKVal:
[...]

PR #1191, opened 2015-08-20
[added] accessibility props for PanelGroup and Panels.

kennyw12:
Adds accessibility, following the guideline specified [here](http://www.w3.org/WAI/PF/aria-practices/#tablist).

The Pivotal UI Team
@atomanyih @ctaymor @d-reinhold @kennyw12 @matt-royal @nicw @stubbornella

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -60,26 +60,32 @@ const Panel = React.createClass({
   },
 
   render() {
+    let {headerRole, panelRole, ...props} = this.props;
     return (
-      <div {...this.props}
+      <div {...props}
         className={classNames(this.props.className, this.getBsClassSet())}
         id={this.props.collapsible ? null : this.props.id} onSelect={null}>
-        {this.renderHeading()}
-        {this.props.collapsible ? this.renderCollapsibleBody() : this.renderBody()}
+        {this.renderHeading(headerRole)}
+        {this.props.collapsible ? this.renderCollapsibleBody(panelRole) : this.renderBody()}
         {this.renderFooter()}
       </div>
     );
   },
 
-  renderCollapsibleBody() {
-    let collapseClass = this.prefixClass('collapse');
+  renderCollapsibleBody(panelRole) {
+    let props = {
+      className: this.prefixClass('collapse'),
+      id: this.props.id,
+      ref: 'panel',
+      'aria-hidden': !this.isExpanded()
+    };
+    if (panelRole) {
+      props.role = panelRole;
+    }
 
     return (
       <Collapse in={this.isExpanded()}>
-        <div
-          className={collapseClass}
-          id={this.props.id}
-          ref='panel'>
+        <div {...props}>
           {this.renderBody()}
 
         </div>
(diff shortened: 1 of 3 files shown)


kennyw12:
Oops. Forgot PropType validation. I'm also not sure if these roles should be exposed.

src/Panel.js
```diff
@@ -166,7 +174,7 @@ const Panel = React.createClass({
       if (this.props.collapsible) {
         header = cloneElement(header, {
           className,
-          children: this.renderAnchor(header.props.children)
+          children: this.renderAnchor(header.props.children, headerRole)
```
taion: Would it be incorrect for the header role to actually get attached to the "heading" element at https://github.com/react-bootstrap/react-bootstrap/pull/1191/files#diff-1e26d9c69579835d9b0137815187fafeR185 rather than the anchor?

src/Panel.js
```diff
@@ -60,26 +62,32 @@ const Panel = React.createClass({
...
+    let {headerRole, panelRole, ...props} = this.props;
     return (
-      <div {...this.props}
+      <div {...props}
         className={classNames(this.props.className, this.getBsClassSet())}
         id={this.props.collapsible ? null : this.props.id} onSelect={null}>
-        {this.renderHeading()}
-        {this.props.collapsible ? this.renderCollapsibleBody() : this.renderBody()}
+        {this.renderHeading(headerRole)}
+        {this.props.collapsible ? this.renderCollapsibleBody(panelRole) : this.renderBody()}
```
taion: Unfortunate that we can't add `panelRole` (should this be `bodyRole`?) in the general case because of the silly stuff I added with the "fill" children. Oops.

taion:
BTW, would appreciate seeing your thoughts on discussion on #983 from https://github.com/react-bootstrap/react-bootstrap/issues/983#issuecomment-130740354 on down, i.e. we might structure `Panel` to look like e.g. `Modal`, and have e.g.

```jsx
<Panel>
  <Panel.Heading>{heading}</Panel.Heading>
  <Panel.Body>{body}</Panel.Body>
</Panel>
```

and then this would just work with the `role` prop set respectively on heading/body/whatever... haven't worked out all the kinks there yet though.

src/Panel.js
ctaymor: The use-case we're anticipating here is using an accordion like a set of tabs, as in the example in the react-bootstrap documentation. Based on our research, the `a` is the correct place to put the role attribute. Here's [one blog post that confirms this](https://www.marcozehe.de/2013/02/02/advanced-aria-tip-1-tabs-in-web-apps/).

ctaymor:
We just amended the last commit to add `role="presentation"` to the h4 when we generate the header for a collapsible panel which was only passed a string for a header. The heading sizes are supposed to semantically describe the outline of the page, and since we don't know the structure of a user's page, we don't know what level the header should actually be. It is a visual aesthetic use of an h4, not a semantic one, thus presentation.

taion:
LGTM. Might make sense to squash these commits. Otherwise :shipit:!

jquense:
LGTM

PR #1201, opened 2015-08-22
[added] restoreFocus prop to Modal

steevel:
Add the ability not to automatically set focus to the last active
element, an example would be displaying a Modal on onFocus, which would
then instantly trigger opening of the same Modal everytime it's closed.

This jsfiddle demonstrates such a situation:
https://jsfiddle.net/zs3ed1ho/1/

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -130,6 +130,11 @@ const Modal = React.createClass({
     enforceFocus: React.PropTypes.bool,
 
     /**
+     * When `true` The modal will automatically restore focus to the last active element when it closes.
+     */
+    restoreFocus: React.PropTypes.bool,
+
+    /**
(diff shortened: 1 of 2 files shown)


jquense:
Hi there, thanks for the PR, I appreciate the time take to help us out.

I can see how the current behavior doesn't suit your particular usecase here. However, I am sort of the opinion that opening a Modal on focus is not good behavior for a modal, akin to navigating to a link when you focus on it.. The Overlay component is really built for that particular use case. More importantly though, restoring focus to the trigger is required for meeting accessibility guidelines. Its already unfortunate that we allow turning off the other focus behavior (and probably something that will be deprecated in the future now that we've refactored the Modal).

steevel:
Fair enough. At least I tried :-)

Issue #1213, opened 2015-08-25
MenuItem ids


kennyw12:
Id props are not passed through on MenuItems. We would like to pass through an id and we think it should be on the anchor tag. Was it an intentional decision to remove it in the [Dropdown rework](https://github.com/react-bootstrap/react-bootstrap/commit/346501e736e280e1b744009412c5285087eb6c90#diff-bb70ed377da60e0f35071b688a96d337L59)?
(@mtscout6)

If not, we'll PR.

The Pivotal UI Team
@atomanyih @ctaymor @d-reinhold @kennyw12 @matt-royal @nicw @stubbornella

mtscout6:
I did not remove them intentionally. Looks like the first bug of the rewrite, sorry about the inconvenience on that.

rstormsf:
I've noticed that in Navbar dropdown menuitem, the first element is always `:focused` :-(

<img width="218" alt="react_github_notetaker" src="https://cloud.githubusercontent.com/assets/9360827/9487802/f734f0f4-4b89-11e5-87a7-b7d717068d24.png">

jquense:
@rstormf yes. that's a feature not a bug. it's particularly one required for Web accessibility

AlexKVal:
It seems there will be more such questions, maybe we should concider documenting of this feature ?

jquense:
its documented in the changelog, and since it mostly matches upstream bootstrap I don't think additional documentation is probably needed. But it doesn't hurt either I guess.

mtscout6:
The focus on open seems to be a different issue from the theme of IDs on the MenuItem so that should probably move to another issue. Though should only need to be focused if opened via keyboard commands, but not by mouse click right?

AlexKVal:
Resolved by https://github.com/react-bootstrap/react-bootstrap/pull/1214

mtscout6:
Did an issue get created for the focus state discussion?

rstormsf:
I've never seen any bootstrap based website if a user clicked with a mouse on a dropdown it would focus on first element.
if was opened with keyboard - true, makes sense.

I apologize if I've driven this ticket from the point of Pivotal Labs UI team regarding 
`Id props are not passed through on MenuItems`

stubbornella:
IMO, it's better not to have divergent behavior between keyboard behavior and mouse behavior unless there is a really good reason. 




I'd want to focus on the first item in either case.

jquense:
[...]

PR #1214, opened 2015-08-25
[fixed] id passthrough for MenuItem

kennyw12:
#1213

### DIFF
--- a/src/MenuItem.js
+++ b/src/MenuItem.js
@@ -47,6 +47,7 @@ export default class MenuItem extends React.Component {
         <SafeAnchor
           role='menuitem'
           tabIndex='-1'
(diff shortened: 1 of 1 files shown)


jquense:
LGTM, though what is the rational behind adding to the anchor vs the li? seems like a good choice, just wondering what ya'll were thinking.

ctaymor:
LGTM.
Looking at it, seems like it's on the anchor instead of the li because the li is presentational. Thus, the li won't "show up" to a screen reader, and won't work properly. (I'm conjecturing based on the [role=presentation docs](http://www.w3.org/TR/wai-aria/roles#presentation) which say an element with a presentational role is removed from the accessibility tree.)

PR #1217, opened 2015-08-25
[fixed] isRequiredForA11y check should be against undefined

kennyw12:
We were getting silent a11y fails and found this bug.

### DIFF
--- a/src/utils/CustomPropTypes.js
+++ b/src/utils/CustomPropTypes.js
@@ -44,7 +44,7 @@ const CustomPropTypes = {
 
   isRequiredForA11y(propType){
     return function(props, propName, componentName){
-      if (props[propName] === null) {
+      if (props[propName] === undefined) {
         return new Error(
           'The prop `' + propName + '` is required to make ' + componentName + ' accessible ' +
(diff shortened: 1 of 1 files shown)


taion:
Can we just do `== null` to be consistent with how `isRequired` works in React, which considers a prop missing if it's either `null` or `undefined`. That should also un-break the test for free.

ctaymor:
Closing because the problem was fixed in PR [#13](https://github.com/react-bootstrap/react-prop-types/pull/13).

AlexKVal:
@ctaymor you are too fast :smile: 
I recovered it as https://github.com/react-bootstrap/react-bootstrap/pull/1224,
though as it turns, it needs more thoughts.

--
https://github.com/react-bootstrap/react-prop-types/pull/13 is against another project 
which `R-B` doesn' use yet.
https://github.com/react-bootstrap/react-bootstrap/pull/1223 is on the way

--

ctaymor:
@AlexKVal Oops, sorry about that. I was wondering why just typing "#13" didn't get the right story. :)

PR #1225, opened 2015-08-26
[fixed] Error on opening dropdown without focusable items

jesenko:
When dropdown without focusable items (e.g. containing only divs to display some content) is opened, exception is raised in `DropdownMenu.focusNext()` method as `items[0]` is not defined.

This commit adds additional check in `focusNext()` method, skipping focusing altogether if there are no focusable items.

### DIFF
--- a/src/DropdownMenu.js
+++ b/src/DropdownMenu.js
@@ -38,6 +38,10 @@ class DropdownMenu extends React.Component {
   focusNext() {
     let { items, activeItemIndex } = this.getItemsAndActiveIndex();
 
+    if (items.length === 0) {
+      return;
+    }
+
     if (activeItemIndex === items.length - 1) {
       items[0].focus();
       return;
--- a/test/DropdownSpec.js
+++ b/test/DropdownSpec.js
@@ -186,6 +186,21 @@ describe('Dropdown', function() {
     buttonNode.getAttribute('aria-expanded').should.equal('false');
   });
 
+  it('opens if dropdown contains no focusable menu item', function() {
+    const instance = ReactTestUtils.renderIntoDocument(
+      <Dropdown title='custom child' id='dropdown'>
+        <Dropdown.Toggle>Toggle</Dropdown.Toggle>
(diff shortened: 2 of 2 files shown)


jquense:
hi there thanks for the PR.

I think the fix is correct here, but I am a bit worried that we are fixing the wrong thing. First there is a philosophical issue, a menu should always have a focusable item in it, otherwise its not a menu! More importantly tho in your test case, you are adding a div to a Menu, which is going to render invalid html (a div inside of a `ul`), I don't think we should be encouraging that, especially when there is a now a "correct" way to create custom menus if you want that sort of behavior

```jsx
<Dropdown title='custom child'>
    <Dropdown.Toggle>Toggle</Dropdown.Toggle>
    <div bsRole='menu'>
       Some custom nonfocusable content
    </div>
</Dropdown>
````

taion:
@jquense I agree that a `div` is not valid there, but what about trivial cases like an empty menu?

jquense:
I think its worth guarding against, I am mostly just wondering aloud if it also encourages bad patterns.

jesenko:
@jquense I agree that the test case is flawed; I actually encountered the error when rendering custom menus inside dropdown, e.g.

      <Dropdown>
        <Dropdown.Toggle>Toggle</Dropdown.Toggle>
        <Dropdown.Menu>
          <li><a href="...">I do not contain 'tabIndex="-1"' attribute</a></li>
          <li class="display-only">Some info</li>
        </Dropdown.Menu>
      </Dropdown>

This results in a valid HTML; Is it considered a valid use case to provide custom html within `react-bootstrap` container?

I have added another commit modifying test case to render valid html (`<div>` -> `<li>`)

taion:
This patch looks good to me, but I haven't really thought about the bad pattern issue. Leaving it to you, @jquense !

jquense:
Looks good to me, can you squash your commits?

jesenko:
@jquense Done

PR #1237, opened 2015-08-27
[fixed] screen-reader accessible dismiss button on alerts

kennyw12:
Before this change, the dismiss button was focusable in VoiceOver, but not in JAWS. This also moves the position of the SR-accessible dismiss button to the end of the alert.

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -623,6 +623,11 @@ const ComponentsPage = React.createClass({
                   <p>just pass in a <code>onDismiss</code> function.</p>
                   <ReactPlayground codeText={Samples.AlertDismissable} />
 
+                  <div className="bs-callout bs-callout-info">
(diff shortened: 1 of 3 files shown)


ctaymor:
Looks like this was failing because of an issue on master, not in this change. (Pulling it down, rebasing on top of master, and running npm install fixed the issue locally).

taion:
Ah, cool, probably makes more sense to have this at the end anyway. LGTM.

I wonder if we should be making a note of all the places where we handle things differently from upstream Bootstrap for a11y reasons (or otherwise).

ctaymor:
It makes sense that we should be taking note of those divergences. Probably a note in the component docs?

jquense:
ya it would be really great to document the accessibility stuff in the docs properly, especially since its a really neat and helpful feature we can provide on top of bootstrap!

ctaymor:
LGTM.

PR #1242, opened 2015-08-28
[fixed] Badge was not shown when it had only a single child.

xsistens:
Fix: Badge was not shown when it had only a single child. Now every tagtree you pass to a badge is shown.

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -16,7 +16,7 @@ const Badge = React.createClass({
 
   hasContent() {
     return ValidComponentChildren.hasValidComponent(this.props.children) ||
-      (React.Children.count(this.props.children) > 1) ||
+      (React.Children.count(this.props.children) > 0) ||
       (typeof this.props.children === 'string') ||
       (typeof this.props.children === 'number');
   },


jquense:
...lgtm though I find the original code to be sort of obtuse... @react-bootstrap/collaborators Anyone else sign off on this/ know what is going on in this method?

mathieumg:
Looks like it only allowed a number or a string when it had a single child, but I have no idea why it would put that constraint. With that change, is it safe to say the last two conditions won't ever be reached and can thus be removed? Perhaps add a test for when a child is a single component? This is also one of the components that will be removed in Bootstrap v4 (in favor of a new round style on `Label`).

taion:
Wouldn't `ValidComponentChildren.hasValidComponent(this.props.children)` handle the cases with non-number/string children though?

Why do we need the `hasContent` check to start with?

jquense:
image

I guess not showing anything when empty is actually a feature, and our check is to make it more robust?

taion:
FWIW, the second check there is a fix-ish for: https://github.com/react-bootstrap/react-bootstrap/issues/532.

I think this is a fairly fragile check, because it won't work for cases where e.g. the `Badge` containers a custom element that doesn't render anything.

My personal preference would be to get rid of this and just match upstream Bootstrap.

@stevoland / @pieterv ? Looks like you guys added the behavior initially. Do you still need it?

My preferred implementation would be to iterate through the children and ensure that at least one is non-false-y, or is `0`, assuming we want to keep this behavior.

taion:
I'm going to close this for now. The current check is not quite right, but this doesn't move us in the right direction.

PR #1243, opened 2015-08-28
[fixed] title is not passed to tab pane DOM node

### LINKED ISSUE #1238: [Regression] Using props "title" for Tabs conflicts with HTML title attribute and causes buggy hover behavior
The switch to using the property name ```title``` is problematic since it's conflicts with the HTML title attribute. This has the side effect of causing the title to display on hover anywhere on the tab's contents. You can see this on the React Bootstrap demo page:

react-bootstrap-tab-title-bug

On my site, I used a component as the title, so you see "[Object object]" whenever I hover anywhere on the tab's contents. I'm on Chrome.

jquense:
fixes #1238 

It effectively removes the ability to add a title to tab panes, but I don't think that is going to to be an issue...

It is a bit odd that that props doesn't do anything unless inside a `tabs` component, but nothing too weird

### DIFF
--- a/src/Tab.js
+++ b/src/Tab.js
@@ -87,6 +87,7 @@ const Tab = React.createClass({
 
     return (
       <div {...this.props}
+        title={undefined}
         role='tabpanel'
         aria-hidden={!this.props.active}
         className={classNames(this.props.className, classes)}


taion:
We could have `Tabs` strip off the `title` prop.

Hoping for @ctaymor or someone else to chime in on whether this was an intentional effect from the prop change, though.

ctaymor:
:+1:  LGTM. That was not an intentional change. I don't see it being a problem not being able to add a title to tab panes. If at some point down the road, someone does have that problem, we can reconsider, but it seems to me to be usually undesirable.

PR #1244, opened 2015-08-28
[fixed] allow null activeKey (empty) selection

### LINKED ISSUE #1227: Tabs behavior when activeKey is null
We currently treat `null` and `undefined` identically for deciding whether a `Tabs` component is controlled: https://github.com/react-bootstrap/react-bootstrap/blob/v0.25.0/src/Tabs.js#L196

Is this what we want, or should we interpret `null` as "controlled, but nothing selected"?

jquense:
fixes #1227 I believe

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -196,7 +196,7 @@ const Tabs = React.createClass({
   },
 
   getActiveKey() {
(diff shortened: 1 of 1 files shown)


taion:
I'll add a test case.

PR #1250, opened 2015-08-28
[changed] tab keyboard navigation to be more inline with ARIA spec

jquense:
http://www.w3.org/TR/wai-aria-practices/#tabpanel

Only the active tab is focus-able, and rather then hitting tab to change the active tab you use the arrow keys. 

I didn't do the ctrl page up stuff b/c that sounds bit more undecided from reading through. It also seems like it would conflict with anything inside the panel that used that key combo...

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -36,6 +36,7 @@ const NavItem = React.createClass({
         title,
         target,
         children,
+        tabIndex, //eslint-disable-line
         'aria-controls': ariaControls,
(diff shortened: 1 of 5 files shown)


taion:
LGTM. I'm going to fix the merge conflicts, then merge.

We'll probably want to eventually handle `tab` appropriately as well.

taion:
:facepunch: AppVeyor

Issue #1220, opened 2015-08-26
Regression: overlaytrigger tooltip does not disappear when used with dropdown button


petejkim:
See:
![overlaytrigger-bug](https://cloud.githubusercontent.com/assets/30173/9492248/ba5893fe-4baa-11e5-8bee-42fb8e956010.gif)

``` jsx
  <OverlayTrigger placement='top' overlay={<Tooltip>Hello World</Tooltip>}>
    <DropdownButton title="Menu">
      <MenuItem>Click Me</MenuItem>
      <MenuItem>Click Me</MenuItem>
    </DropdownButton>
  </OverlayTrigger>
```

Expected behavior:
Tooltip should disappear as soon as user moves mouse out of the button.

Actual behavior:
Tooltip does not disappear when user moves mouse from the button to the menu.
Tooltip does not disappear when user clicks on a menu item.
Tooltip does not disappear until after user moves mouse on to the trigger element and then back out to force mouseleave event.

Used to work correctly in the previous release.

cbeihl:
Is this issue being worked ?  I'm still seeing this bug in the 0.28.1 release.

irealva:
Also having this issue. Is there another way an overlay on a DropdownButton?

untitledboy:
Have someone found any acceptable solutions for this bag?

ron23:
I set the trigger to hover only as I don't need mobile support right now. Not a solution though.

PR #1251, opened 2015-08-28
[fixed] focus returns to the toggle by default onClose

jquense:
Reworks the focus management so focus is properly returned to the toggle when an item is selected from the list. This partially address #1220 fixing the `focus` trigger case. it doesn't address the hover case though.

need to clean up the commits here...


jquense:
something is wrong with eslint...

taion:
:shipit:

PR #1253, opened 2015-08-29
[fixed] OverlayTrigger event handlers are properly maintained

### LINKED ISSUE #1140: onBlur events (maybe others) not passed to inputs with <OverlayTrigger>
It appears that the onBlur event is not firing when the **OverlayTrigger** is applied to an input.

```javascript
<OverlayTrigger placement='top' overlay={<ToolTip>{this.props.errorMsg}</ToolTip>} delayShow={100} delayHide={100}>
    <input disabled={this.props.disable} placeholder={this.props.placeholder} id={this.props.id} data-id={this.props.dataId} className={classes} type={this.props.type} value={this.props.value} onChange={this.props.onChange} onBlur={this.props.onBlur} maxlength={this.props.maxlength} />
</OverlayTrigger>
```

jquense:
fixes #1140

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -175,6 +175,7 @@ const OverlayTrigger = React.createClass({
 
   render() {
     const trigger = React.Children.only(this.props.children);
+    const triggerProps = trigger.props;
 
     const props = {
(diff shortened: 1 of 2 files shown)


taion:
:shipit:

PR #1254, opened 2015-08-29
[fixed] OverlayTrigger hover triggers on mousenter/leave

jquense:
Addresses an inadvertent revert we did switching the hover triggers to use mouse{enter|leave} instead of out/over, since those events also fire on child elements, causing flicker. Unfortunately the React mousenter/leave implentation is [broken](https://github.com/facebook/react/issues/4251), so we need to simulate it ourselves.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -1,6 +1,6 @@
 /*eslint-disable react/prop-types */
 import React, { cloneElement } from 'react';
-
+import contains from 'dom-helpers/query/contains';
 import createChainedFunction from './utils/createChainedFunction';
 import createContextWrapper from './utils/createContextWrapper';
 import Overlay from './Overlay';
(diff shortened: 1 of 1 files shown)


src/OverlayTrigger.js
```diff
@@ -195,8 +200,8 @@ const OverlayTrigger = React.createClass({
         '[react-bootstrap] Specifying only the `"hover"` trigger limits the visibilty of the overlay to just mouse users. ' +
         'Consider also including the `"focus"` trigger so that touch and keyboard only users can see the overlay as well.');
 
-      props.onMouseOver = createChainedFunction(this.handleDelayedShow, this.props.onMouseOver, triggerProps.onMouseOver);
-      props.onMouseOut = createChainedFunction(this.handleDelayedHide, this.props.onMouseOut, triggerProps.onMouseOut);
+      props.onMouseOver = createChainedFunction(this.handleDelayedShow, this.handleMouseOver, triggerProps.onMouseOver);
```
taion: Looks like this replaced the wrong callback? i.e. `this.props.onMouseOver` instead of `this.handleDelayedShow`.

jquense: yup, ooops, thanks

jquense:
fixed!

taion:
LGTM. Mind adding a comment explaining this though?

jquense:
good idea

PR #1255, opened 2015-08-29
pin eslint version because of bug in 1.3.0

jquense:
is there any good reason not to just always pin deps like this to patch versions? That we have to spend any bandwidth on these sorts of breakages is annoying. We rarely need new eslint features, but deal with breakages due to upgrades fairly often it seems


jquense:
merging to get the rest of the PR's to work

taion:
Pinning dev deps works for me. Just need to make sure on occasion that they don't get too stale.

PR #1256, opened 2015-08-29
[remove] previously deprecated Components, props, and features

jquense:
- Transition, Position, Portal
- `navItem` Dropdown prop,
- Fade/Collapse Mixins
- DomUtils


taion:
:shipit:

Issue #988, opened 2015-07-14
Allow totally custom styles via 'bsStyle'


AlexKVal:
--
For the code:
```js
<ButtonGroup>
  <Button bsStyle='primary'>My button</Button>
  <Button bsStyle='myCustom'>My button</Button>
  <Button bsStyle='facebook'>My button</Button>
</ButtonGroup>
```

R-B generates markup: (`'-undefined'` suffix for every `custom` style name)
<img width="620" alt="btn-undefined" src="https://cloud.githubusercontent.com/assets/847572/8683320/aa4ad2ca-2a7a-11e5-8445-0666e9fd29ca.png">

Such will be possible: (for `bsStyle='facebook'` => `'facebook'`)
<img width="618" alt="custom-styles-mappint" src="https://cloud.githubusercontent.com/assets/847572/8683345/d6f0ae94-2a7a-11e5-9818-e12865557bfc.png">

I will document the ability to use custom suffixes such as `'btn-myCustom'` via:
```js
styleMaps.addStyle('myCustom');
```
Now it is possible because of https://github.com/react-bootstrap/react-bootstrap/pull/496, but is not documented.

--
It will address: 
this https://github.com/react-bootstrap/react-bootstrap/issues/537#issue-69286225 
this https://github.com/react-bootstrap/react-bootstrap/issues/543#issue-69510135
and this https://github.com/react-bootstrap/react-bootstrap/issues/306#issuecomment-71010320
> you can't rely on bootstrap draft semantics and call it "rules". 
It was built for user extension. Bootstrap info, warning, etc. were built more for demo purpose than for mindless usage in production code. That's BS developers opinion, not mine. End users may and should create their own sets of statuses etc. The current way of handling this is very uncomfortable, I 100% agree here with @janmarek. You don't need to validate class names against predefined sets at all. Just allow users to pass whatever they want to. That simple.

--
And it will document this: https://github.com/react-bootstrap/react-bootstrap/issues/537#issuecomment-94298524

--
- [ ] Allow set `custom styles with validation warning` through `bsStyle` property. 
  - Solves https://github.com/react-bootstrap/react-bootstrap/issues/537#issue-69286225
  - partially https://github.com/react-bootstrap/react-bootstrap/issues/543#issue-69510135
  - and https://github.com/react-bootstrap/react-bootstrap/issues/306#issuecomment-71010320
- [ ] Document `custom styles with warnings`

--
[...]

PR #1257, opened 2015-08-29
WIP: bootstrap mixin removal

jquense:
DO NOT MERGE: not ready yet.

- [x] remove component dependence on Mixin
- [x] change propType validation (either remove it, or make it actually work)
- [ ] first class api for changing `bsClass` result prefix: `btn` -> `mybtn`, `btn-primary` -> `mybtn-primary`
- [ ] Api for adding custom styles to components
- [x] Allow custom "styles" to work with bsClass prefix
- [x] make sure `bsClass` prefixes are not [hard coded](https://github.com/react-bootstrap/react-bootstrap/blob/63d8e066e4e4578fe234d63d0adc038d719cea3e/src/Button.js#L48)
- [x] documentation strategy for allowed styles and sizes

Am I missing anything else @AlexKVal?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,6 @@
 {
   "stage": 1,
   "optional": ["runtime"],
-  "loose": ["all"]
+  "loose": ["all"],
+  "plugins": ["object-assign"]
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,7 +46,15 @@
     "react/no-did-mount-set-state": 2,
     "react/no-did-update-set-state": 2,
     "react/no-multi-comp": 2,
-    "react/prop-types": [2, { "ignore": [ "children", "className", "style" ] }],
+    "react/prop-types": [2, { "ignore": [
+        "children",
+        "className",
+        "style",
+        "bsStyle",
+        "bsClass",
+        "bsSize"
+        ]
+    }],
     "react/react-in-jsx-scope": 2,
     "react/self-closing-comp": 2,
     "react/sort-comp": 0,
--- a/docs/generate-metadata.js
+++ b/docs/generate-metadata.js
@@ -3,6 +3,7 @@ import glob from 'glob';
 import fsp from 'fs-promise';
 import promisify from '../tools/promisify';
 import marked from 'marked';
+import defaultDescriptions from './src/defaultPropDescriptions';
 
 marked.setOptions({
   xhtml: true
@@ -18,6 +19,7 @@ let cleanDoclets = desc => {
 
 let cleanDocletValue = str => str.trim().replace(/^\{/, '').replace(/\}$/, '');
 
+let quote = str => str && `'${str}'`;
 
 let isLiteral = str => (/^('|")/).test(str.trim());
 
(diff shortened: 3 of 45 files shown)


mtscout6:
I'm excited to see this underway! Can you expound on what you are planning when you say "change propType validation"? Is this the work to not pull all the styles from one location, ie a button should be allowed a pill style?

jquense:
Yeah it encompasses that issue as well as another I think,

We had a bit of a conversation about it in Slack already...I am not sure that validation is a great idea for these props. Ultimately folks need to be able to add their own Styles without much hassle, since bootstrap is built in a way that allows and encourages that. Unfortunately though that means we need to know about all acceptable Styles (like in my case `btn-minimal`) in order to validate `bsStyle`. My question is whether its worth it to require users to go through the hassle of registering  their variants, so we can warn when someone spells "primary" wrong, or put `pills` incorrectly.

I think am of the opinion that the poorly rendered button/tab/etc provides validation enough, especially if it makes extending styles easier for users.

The current behavior of incorrectly validating a bunch of styles is no good either, and of course, however we go, we need to document what builtin styles are allowed so folks are trying to do `btn-pills`.

mtscout6:
I agree that a visual check would definitely tell you something is wrong. The problem with that though is it's not something that can easily be caught on CI. I personally like things that can fail a build because a Dev didn't actually load something on the screen (I know devs should be testing what they write, but let's face it not all devs are good at doing that all the time.)

I see your point as well though that it will take some thought as to what is needed to make such validation still extensible for custom styles.

Perhaps it's worth posting a Survey on the Docs site to see how many folks actually want that validation?

AlexKVal:
@jquense yes. your todo list is right. :sparkle:

jquense:
Ok, now we are in api territory, and that needs some discussion. I've added three functions/decorators for adding propTypes and default values to Components.
```js
@bsClass('alert')
@bsStyle(['danger', 'warning', 'info'], 'default')
class Alert {}
```
The use as a decorator is optional, they work just fine as normal functions calls as well, for folks that don't want to use stage 1 features, or are using `createClass`

```js
class Alert {
}

Alert = bsClass('alert', Alert);
Alert = bsStyle(['danger', 'warning', 'info'], 'default', Alert);
```

Instead of the big global map of styles in `styleMap` I've added some constants instead, which you can see I use instead of strings in the components. Not sure if it makes sense to expose them to users.

jquense:
o and there is an annoying Gotcha with the decorators, when using this pattern:

```js
@bsClass('alert')
@bsStyle(['danger', 'warning', 'info'], 'default')
class Alert {
}

// this overrides the prop types set by the decorator!
Alert.propTypes = {
  anotherProp: PropTypes.string
}
```

instead you need to do (same for defaultProps as well):

```
Alert.propTypes = {
  ...Alert.propTypes,
  anotherProp: PropTypes.string
}
```

taion:
Are static property initializers also stage 0? That's really annoying.

Maybe clearer not to use decorators here and just call the decorators functions?

jquense:
Maybe someone wants to write an eslint rule for just allowing static `propTypes` and `defaultProps`property initializers? As you say the static ones are sort of a no brainer. I would limit it to those tho...

.babelrc
```diff
@@ -1,5 +1,6 @@
 {
   "stage": 1,
   "optional": ["runtime"],
-  "loose": ["all"]
+  "loose": ["all"],
+  "plugins": ["object-assign"]
```
taion: This should be a no-op with `runtime` enabled, no?

jquense: that is a good question, at one point it didn't polyfill Object.assign, may not be true now tho...

taion: I believe it now does polyfill that.

AlexKVal:
I've closed parent issues #988 and #646 in favor to continue those discussions here.
And because those were in `0.26` milestone - I've assigned this PR to `0.26` too.
But it seems this change rather should go to `0.27` milestone, because there are already enough changes in `0.26` to release it.

jquense:
?

taion:
Oh crap. That was an accident. I deleted the `v0.26-rc` branch because we'd already merged it into master.

jquense:
ah ok, we are going to need to PR it again anyway to 27-rc i guess

Issue #1023, opened 2015-07-20
FormControls.Static support for components in value?


eventhough:
I'd like to display a link as the value for FormControls.Static but I get this warning:

```
Warning: Failed propType: Invalid undefined `value` supplied to `Static`. Check the render method of `DetailForm`.
```

If you swap value with children you get a similar warning:

```
Warning: Failed propType: Invalid undefined `children` supplied to `Static`. Check the render method of `DetailForm`.
```

It looks like FormControls.Static currently takes either a value or children and requires that the value be a string.

Is there a way for us to have the option to use a component?

[http://react-bootstrap.github.io/components.html#input-static](http://react-bootstrap.github.io/components.html#input-static)

aabenoja:
At the moment the expected values are numbers and strings. It is, however, very simple to change this to `PropTypes.node` [in this section](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/childrenValueInputValidation.js#L10-L11). A pull request is welcome if you want to hop right on it, though I could get around to it in a few hours.

jontewks:
Ran into this issue as well, so submitted a PR #1269 to fix.

taion:
Fixed by #1269.

PR #1269, opened 2015-09-01
[changed] allow PropTypes.node in valueValidation

jontewks:
Remove PropTypes.number and PropTypes.string in favor of PropTypes.node.

This allows for a number, string, element, or an array of those to pass validation per #1023. Also fixes the ButtonInputSpec test and adds one to FromControlsSpec.

### DIFF
--- a/test/ButtonInputSpec.js
+++ b/test/ButtonInputSpec.js
@@ -42,8 +42,6 @@ describe('ButtonInput', () =>{
     ReactTestUtils.renderIntoDocument(
       <ButtonInput value="button" bsStyle="danger" />
     );
-
-    console.warn.called.should.be.false;
   });
 
   it('throws warning about wrong type for bsStyle=error', function () {
@@ -72,11 +70,9 @@ describe('ButtonInput', () =>{
     assert.notInstanceOf(result, Error);
   });
 
-  it('does not allow elements for children', function () {
+  it('allows elements as children', function () {
     ReactTestUtils.renderIntoDocument(
       <ButtonInput><span>blah</span></ButtonInput>
     );
-
-    shouldWarn('propType: Invalid');
   });
 });
--- a/test/FormControlsSpec.js
+++ b/test/FormControlsSpec.js
@@ -33,5 +33,11 @@ describe('Form Controls', function () {
 
       result.should.be.instanceOf(Error);
     });
+
+    it('allows elements as children', function () {
+      ReactTestUtils.renderIntoDocument(
+        <FormControls.Static><span>blah</span></FormControls.Static>
+      );
+    });
   });
 });
--- a/src/utils/childrenValueInputValidation.js
+++ b/src/utils/childrenValueInputValidation.js
@@ -2,13 +2,13 @@ import React from 'react';
 import { singlePropFrom } from './CustomPropTypes';
 
 const propList = ['children', 'value'];
-const typeList = [React.PropTypes.number, React.PropTypes.string];
 
 export default function valueValidation(props, propName, componentName) {
   let error = singlePropFrom(propList)(props, propName, componentName);
+
   if (!error) {
-    const oneOfType = React.PropTypes.oneOfType(typeList);
-    error = oneOfType(props, propName, componentName);
+    error = React.PropTypes.node(props, propName, componentName);
   }
+
   return error;
 }


taion:
Can you get rid of the `oneOfType` logic below? It's not needed any more if we only support a single type.

jquense:
Wait...doesn't this also allow nodes for  `value` attribute of other, non button, `<input/>`s? Is that ok? That will definitely cause issues if folks are trying that.

taion:
I believe this is only used for `FormControls.Static` and `ButtonInput` right now, unless I'm misreading our code.

jontewks:
Removed the oneOfType logic and added a similar elements test for the FormControls.Static

test/ButtonInputSpec.js
```diff
@@ -72,11 +72,11 @@ describe('ButtonInput', () =>{
...
   });
 
-  it('does not allow elements for children', function () {
+  it('allows elements as children', function () {
     ReactTestUtils.renderIntoDocument(
       <ButtonInput><span>blah</span></ButtonInput>
     );
 
-    shouldWarn('propType: Invalid');
+    console.warn.called.should.be.false;
```
taion: You don't need this - our tests automatically fail on any unexpected warnings.

test/FormControlsSpec.js
```diff
@@ -33,5 +33,13 @@ describe('Form Controls', function () {
 
       result.should.be.instanceOf(Error);
     });
+
+    it('allows elements as children', function () {
+      ReactTestUtils.renderIntoDocument(
+        <FormControls.Static><span>blah</span></FormControls.Static>
+      );
+
+      console.warn.called.should.be.false;
```
taion: https://github.com/react-bootstrap/react-bootstrap/pull/1269/files#r38559067

test/ButtonInputSpec.js
jontewks: Thanks for the info, I'll remove. I picked up that pattern from line 46 of ButtonInputSpec, which I will remove as well.

taion: Thanks!

taion:
LGTM. Need one more reviewer from @react-bootstrap/collaborators. Please squash your changes, though. Thanks!

jontewks:
Squash'd.

jquense:
LGTM

mtscout6:
Can you amend your commit message to match our [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes) for our automated changelog tool?

jontewks:
@mtscout6 changed commit message, sorry I only had that on the PR, not the message. I was also unsure of what to include in the commit message about upgrade paths, since its just allowing more options rather than requiring code changes, so I just included a short summary. Please let me know if it needs further modification.

test/ButtonInputSpec.js
```diff
@@ -42,8 +42,6 @@ describe('ButtonInput', () =>{
     ReactTestUtils.renderIntoDocument(
       <ButtonInput value="button" bsStyle="danger" />
     );
-
-    console.warn.called.should.be.false;
```
aabenoja: Should this test be nixed now that the assertion is removed?

test/ButtonInputSpec.js
```diff
@@ -72,11 +70,9 @@ describe('ButtonInput', () =>{
     assert.notInstanceOf(result, Error);
   });
 
-  it('does not allow elements for children', function () {
+  it('allows elements as children', function () {
     ReactTestUtils.renderIntoDocument(
       <ButtonInput><span>blah</span></ButtonInput>
     );
-
-    shouldWarn('propType: Invalid');
```
aabenoja: I say we just remove this test.

mtscout6:
@jontewks That new commit message looks great thanks!

test/ButtonInputSpec.js
jontewks: @taion let me know that the tests will fail on any unexpected warnings, so the test will still fail appropriately without the assertion.

test/ButtonInputSpec.js
taion: Well, it's still a custom prop type validation for children/value. This does accomplish a semi-goal of asserting that it won't throw a spurious false positive here. I'm okay with leaving it in place.

taion:
I'm going to go ahead and merge this. @aabenoja we can delete the test cases if you want. They're not tests I would have actively added, but I think we get a little bit of marginal value out of having them be there.

aabenoja:
@taion That's fair. It looked like empty tests so I wasn't sure if that was just some missed cleanup.

Issue #56, opened 2014-03-31
Consider implementing simple components


stevoland:
There's a lot of components with no interactivity that would just be wrappers around DOM elements with certain class names: badge, label, well, icon etc.

Should we add these? I'm in 2 minds, it adds a level of abstraction that might be useful if the classes change but creates another, albeit simple, API to learn.

Any comments?

__EDIT:__

In summation of current state:

- [x] Badge
- [x] Label
- [x] Well
- [x] Icon
- [x] Media
- [x] Thumbnail (See #488)
- [x] Breadcrumb (See #1273 and #1354)
- [x] Pager
- [x] Pagination (See #22 PR #786)
- [x] Responsive Embed (See #1359)
- [x] Images (See #1293)
- [x] Affix (See #200)

mtscout6:
In summation of current state:

- [x] Badge
- [x] Label
- [x] Well
- [x] Icon
- [ ] Media
- [ ] Thumbnail
- [ ] Breadcrumb
- [x] Pager
- [ ] Pagination (See #22)

JordanTheriault:
I'd like to do the Thumbnail if that's okay. I'm new to this repo but would love to get involved @mtscout6

mtscout6:
@JordanTheriault That'd be great!

ivan-kleshnin:
@mathieumg media is the worst Bootstrap part ever. It's totally broken in responsive aspect. 
Not recommend.

dozoisch:
@mtscout6  Do we add embeds? http://getbootstrap.com/components/#responsive-embed

mtscout6:
Both of those seem reasonable.

dozoisch:
should add affix to the list

https://github.com/react-bootstrap/react-bootstrap/issues/200

dozoisch:
#488 resolves thumbnails

AlexKVal:
/cc @react-bootstrap/collaborators
Do we really want this issue as a condition for the `Milestone 1.0.0` ?

In my opinion if nobody is interested in those additional components,
so there is no value to make them.

And Steve named topic as 'Consider..', i.e. it doesn't mean that they 'must' be implemented.

I propose to remove it from the `Milestone 1.0.0`. :cherries:

mathieumg:
[...]

PR #1273, opened 2015-09-02
[added] Breadcrumb Component

ycavatars:
I added `Breadcrumb` component for issue #56. Please review this request, thanks :)


src/Breadcrumb.js
```diff
@@ -0,0 +1,47 @@
...
+
+  propTypes: {
+    /**
+     * CSS classes for the wrapper `ol` element
+     */
+    className: React.PropTypes.string,
+    /**
+     * HTML id for the wrapper `ol` element
+     */
+    id: React.PropTypes.string
```
taion: This doesn't seem to be used?

src/Breadcrumb.js
```diff
@@ -0,0 +1,47 @@
...
+import ValidComponentChildren from './utils/ValidComponentChildren';
+
+const Breadcrumb = React.createClass({
+  mixins: [BootstrapMixin],
+
+  propTypes: {
+    /**
+     * CSS classes for the wrapper `ol` element
+     */
+    className: React.PropTypes.string,
```
taion: As a rule, we generally don't document the `className` prop.

src/Breadcrumb.js
```diff
@@ -0,0 +1,47 @@
...
+    return {
+      bsClass: 'breadcrumb'
+    };
+  },
+
+  render() {
+    const classes = this.getBsClassSet();
+
+    return (
+      <ol role="navigation" aria-label="breadcrumbs" className={classNames(this.props.className, classes)}>
```
taion: I think you need to pass the rest of the props to the `ol`.

src/BreadcrumbItem.js
```diff
@@ -0,0 +1,46 @@
...
+
+  render() {
+    const {
+      href,
+      title,
+      target,
+      linkId,
+      children,
+      ...props } = this.props;
+    const active = (href == null);
```
taion: Do we think this is the best way to do things, versus having an explicit `active` prop? I think I generally prefer the more explicit behavior when possible. Not totally sure though.

taion:
Thanks for the PR and the extensive tests! Just left a few comments.

src/BreadcrumbItem.js
ycavatars: Actually I also prefer to set the `active` prop explicitly, but I don't know how to properly handle `active` and `href` in these situations: 
1. if `active` is false and `href` is null: should I remove the `href` prop?
2. if `active` is true and `href` is given as well: [in bootstrap](http://getbootstrap.com/components/#breadcrumbs), an `active <li>` element doesn't have an `anchor` element. Is there any other good place to put the `href` prop, or just add an `anchor` to the `active <li>` element?

src/Breadcrumb.js
ycavatars: I forgot to pass them into the `ol` element. I fixed it and tested if the `id` is set correctly in the `ol` element.

src/Breadcrumb.js
ycavatars: I fixed it by passing `{...props}` into the `ol` element.

taion: What's the point in specifying this as a propType though? You're not using it in the component code, nor attaching any special semantics to it - it's just getting passed through as an HTML attribute. Better to just drop this propType entirely.

src/BreadcrumbItem.js
taion: I don't think we necessarily need to deal with those - just don't render the anchor for active components. Otherwise, maybe it makes more sense to spread the input props onto the anchor rather than the `li`? e.g. if the user wants to specify `onClick` or something.

src/Breadcrumb.js
ycavatars: I notice that the `Nav.js` also defines `className` and `id` propTypes, and pass `{...this.props}` to child component. https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Nav.js#L20-L30. Is there any reason why `Nav.js` define them in propTypes? Thanks.

taion: In that case it's to clarify that `id` and `className` are separate from what's attached to the wrapped `ul` element for documentation. You don't have that here, so it's not relevant.

ycavatars: Thanks for your explanation. I will remove the `propTypes` from the `Breadcrumb.js`.

src/BreadcrumbItem.js
ycavatars: I think spread the props onto the anchor is a better idea since we don't assign props to `li` in most cases. But what if an `anchor` is not rendered in an active component? Do I just ignore the props which are assigned to an active component?

taion: That's a tough one. Maybe just drop them onto a `span` if there are any such props specified? Sorry it took so long to get back to you on this one.

ycavatars: It's ok, and really thanks for your review. Do you mean that I always use a `<span>` to wrap the value of an active `<li>`, and drop the specified props onto it?

taion: On second thought, that's really just overthinking it. If the user wants additional styling on the rendered child, they can pass in a React node instead of a string.

Why don't you just call `warning` if both `href` and `active` are specified, but otherwise keep things as-is?

ycavatars: Now I specify the `active` prop explicitly, and spread props onto the children of `li` element. I add the prop validation for `id` back in order to pass the `eslint` checks. Thanks.

ycavatars:
Sorry I haven't noticed that there're some CI errors. I will fix it by today.

taion:
Very nice. LGTM. Can you squash these down to a single commit?

Need one more reviewer.

ycavatars:
Done! Thanks for your review :+1:

AlexKVal:
Now it looks like:
screen shot 2015-09-25 at 5 25 09 pm
and I would change it to
screen shot 2015-09-25 at 5 29 30 pm
Add `active` into the last "breadcrumb" in the example to make it on par with `Bootstrap`'s example
screen shot 2015-09-25 at 5 26 40 pm


Then I would not use `BootstrapMixin` at all. It is extraneous in this particular case
and generates misleading docs properties, which are not in use with this component:
screen shot 2015-09-25 at 5 21 06 pm
And `BootstrapMixin` will be removed anyway https://github.com/react-bootstrap/react-bootstrap/pull/1257 :sparkle: 

And this PR has to be rebased against the latest `master` codebase because there already new component has been added with the `key={29}`
https://github.com/react-bootstrap/react-bootstrap/pull/1337/files#diff-0a2cd7e5cb40bd61f3d1c7e6f852060dR956
https://github.com/react-bootstrap/react-bootstrap/pull/1293


All other LGTM

P.S. If @ycavatars you want, (and nobody else mind it), I could myself fix all those small "issues" just right after merging.

:cherries:

ycavatars:
@AlexKVal Please help me fix the issues :+1:
In my country, it's Moon Festival now so I won't be able to code for several days. Thanks for your help :)

AlexKVal:
I'm closing this PR because of #1354
If #1354 will be merged, then your @ycavatars commit will be merged as yours too.
screen shot 2015-09-25 at 9 49 42 pm

Continue discussion in #1354

:cherries:

PR #1278, opened 2015-09-03
Trivial linting fix

taion:


taion:
I'm just going to merge this because it's a trivial change and it fixes the build.

PR #1288, opened 2015-09-05
Pass onHide to OverlayTrigger overlay

taion:
`OverlayTrigger` used to pass `onRequestHide` as a prop to the overlay child. At some point during the overlay refactor this behavior got dropped.

Is this reasonable? Maybe we don't want to bring back this feature and I should actually just manage my own overlay state explicitly. I feel a lot less confident that this is the right thing now that I'm submitting this PR.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -168,7 +168,8 @@ const OverlayTrigger = React.createClass({
 
     let overlay = cloneElement(this.props.overlay, {
       placement: overlayProps.placement,
-      container: overlayProps.container
+      container: overlayProps.container,
+      onHide: this.hide
     });
 
     return (
--- a/test/OverlayTriggerSpec.js
+++ b/test/OverlayTriggerSpec.js
@@ -266,4 +266,24 @@ describe('OverlayTrigger', function() {
       });
     });
   });
+
+  it('Should forward onHide to overlay', function() {
+    let props;
+    class PropsSpy extends React.Component {
+      render() {
+        props = this.props;
+        return <div>test</div>;
+      }
+    }
+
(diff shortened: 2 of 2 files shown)


taion:
I re-wrote my code to directly use `Overlay` instead of `OverlayTrigger` - it's not quite the same since I needed to add my own wrapping `span` element to make my own thing.

Still not quite sure whether it's better to just leave things as-is, though.

taion:
@jquense Any thoughts on this?

jquense:
Sorry been super busy, I will take a look in a bit hopefully!

jquense:
This feels a bit like an anti pattern to me...but I can definitely see why you'd want that behavior. I like the practicality of passing it in, since it makes having overlays with dismissal triggers much simpler, tho It feels pretty inconsistent to pass it through to the child in the case of a Trigger, but not the overlay itself...

I don't know, if you need fancy behavior it seems like a better option to just handle it with an Overlay directly. I defer to you on this one, especially since you actually have a real use case.

taion:
Given that I'm the only person who's been affected by this so far, and I've already switched to using `Overlay` directly, I'm okay with not merging this for now.

taion:
And we can't really change the behavior on `Overlay` without potential weird edge cases from people otherwise passing in `onHide` to the child (i.e. we can't necessarily safely chain, but would have to figure out if the `onHide` prop passed into the child is the same, &c.).

taion:
On closer inspection, this would be usef

taion:
Thought of one use case – this avoids the extra `<span>` or whatever that React requires of you if you want to use `<Overlay>` directly.

PR #1293, opened 2015-09-06
[added] Images Component with Sample and Docs

deerawan:
Create an images component as refer to http://getbootstrap.com/css/#images including samples and docs.

### DIFF
--- a/src/Image.js
+++ b/src/Image.js
@@ -1,0 +1,52 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const Image = React.createClass({
+
+  propTypes: {
+
+    /**
+     * Sets image as responsive image
+     */
+    responsive: React.PropTypes.bool,
+
+    /**
+     * Sets image shape as rounded
+     */
+    rounded: React.PropTypes.bool,
+
+    /**
+     * Sets image shape as circle
+     */
+    circle: React.PropTypes.bool,
+
+    /**
+     * Sets image shape as thumbnail
+     */
+    thumbnail: React.PropTypes.bool
+  },
+
+  getDefaultProps() {
+    return {
+      responsive: false,
+      rounded: false,
+      circle: false,
+      thumbnail: false
+    };
+  },
+
+  render() {
+    const classes = {
+      'img-responsive': this.props.responsive,
+      'img-rounded': this.props.rounded,
+      'img-circle': this.props.circle,
+      'img-thumbnail': this.props.thumbnail
+    };
+
+    return (
+      <img {...this.props}  className={classNames(this.props.className, classes)} />
+    );
+  }
+});
+
+export default Image;
--- a/test/ImageSpec.js
+++ b/test/ImageSpec.js
@@ -1,0 +1,61 @@
+import React from 'react';
+import ReactTestUtils from 'react/lib/ReactTestUtils';
+import Image from '../src/Image';
+
+describe('Image', function() {
+
+  it('should be an image', function() {
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Image />
+    );
+    let image = React.findDOMNode(instance);
+
+    image.nodeName.should.equal('IMG');
+  });
+
+  it('should provide src and alt prop', function() {
+    let instance = ReactTestUtils.renderIntoDocument(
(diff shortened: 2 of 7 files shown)


src/Image.js
```diff
@@ -0,0 +1,58 @@
...
+  propTypes: {
+    src: React.PropTypes.string,
+    alt: React.PropTypes.string,
+    responsive: React.PropTypes.bool,
+    rounded: React.PropTypes.bool,
+    circle: React.PropTypes.bool,
+    thumbnail: React.PropTypes.bool
+  },
+
+  getDefaultProps() {
```
taion: I thought our prop tables were generated off prop types, not default props.

src/Image.js
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import classNames from 'classnames';
+
+const Image = React.createClass({
+
+  propTypes: {
+    src: React.PropTypes.string,
```
taion: Do we need to explicit document these? Might be cleaner just to pass them along by spreading props.

src/Image.js
```diff
@@ -0,0 +1,58 @@
...
+
+      /**
+       * Sets image shape as thumbnail
+       */
+      thumbnail: false
+    };
+  },
+
+  render() {
+    let classes = {
```
taion: Should be const.

test/ImageSpec.js
```diff
@@ -0,0 +1,69 @@
...
+    image.nodeName.should.equal('IMG');
+  });
+
+  it('should provide src p', function() {
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Image src="image.jpg" />
+    );
+    let image = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'IMG');
+
+    image.props.src.should.equal('image.jpg');
```
taion: Assert on the actual attribute value, not the prop.

test/ImageSpec.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it('should provide alt prop', function() {
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Image alt="this is alt" />
+    );
+    let image = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'IMG');
+
+    image.props.alt.should.equal('this is alt');
```
taion: As above.

taion:
See inline comments. Also, we'd want this squashed into a single commit with a comment like: `[added] Image component`

src/Image.js
deerawan: Thank you @taion for the review. Does it mean the getDefaultProps() method is not needed here?

src/Image.js
deerawan: @taion noted. I think it also can be applied to *alt*, right?

test/ImageSpec.js
deerawan: I see. So, I should check if the Image component has props named **src** instead of the value, shouldn't I?

jquense: I believe he means actually check the DOM node to see if it rendered a node with a src attribute vs checking the prop. otherwise you are just testing if react sets props instead of if you used it correctly

taion: Exactly.

src/Image.js
taion: That's correct. I meant all of the props that just get forwarded to `img` directly.

AlexKVal:
LGTM. I'm fine to place it to the end of `Components` page.

Need one more sign of @react-bootstrap/collaborators 

--
Docs part looks like:

screen shot 2015-09-19 at 8 00 43 pm
screen shot 2015-09-19 at 8 00 49 pm

taion:
LGTM, but would we be okay with putting this above "Utilities" at least?

AlexKVal:
Ok. Then I merge it and I will move it upper and fix eslint warnings fixing because of #1335.
I'll move it right before `Thumbnail` component. Right ?

deerawan:
@AlexKVal thanks for your fixing :)

AlexKVal:
Np :cherries:

AlexKVal:
Thank you @deerawan for contributing to this project :sparkles: :sparkle:

PR #1295, opened 2015-09-06
Set the disabled css class so that the text is greyed out.

tannewt:


jquense:
is this a feature from bootstrap? could you provide more context please?

tannewt:
I think you could say that. This change makes the inner text of the dropdown match this less rule from Bootstrap: https://github.com/twbs/bootstrap/blob/master/less/navbar.less#L434

taion:
Does this properly prevent the `Dropdown` button from being clicked?

tannewt:
The click is already correctly prevented. This change only affects the appearance so that it looks disabled.

taion:
Ah. Cool. LGTM. Need +1 reviewer.

jquense:
I am a bit concerned that this is too implementation specific to go one the general Dropdown. would this present annoying styling resets for custom menus/toggles? Would it make more sense for the DropdownButton(s) to set it?

taion:
Wouldn't that only be the case if they've defined a `.disabled` rule that would apply? Maybe I'm misunderstanding.

taion:
@jquense ping

AlexKVal:
Actually this is the fix for undiscovered, till now, bug.
@tannewt could you please add `[fixed]` into the beginning of your commit header ?
(it is then will be reflected in the [Changelog](https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md))

Without this fix:
not-fixed

--
And with this fix:
fixed

and
fixed-neg

tannewt:
@AlexKVal Done.

AlexKVal:
I presume we cannot merge it yet because of @jquense concern.
Will wait his comments on this.
:cherries:

jquense:
I think my concern as abated for now. LGTM

AlexKVal:
@tannewt thank you for your contribution :cherries:

tannewt:
Np! Happy to help.

PR #1298, opened 2015-09-07
[fixed] Don't render Grid or Row with Tabs

### LINKED ISSUE #1276: Left Tabs shouldn't use `Grid`
The Grid component in the Left Tabs configuration adds a `container` class which has set width's, effectively means the tabs will overflow their parent if the width is less than the bootstrap page width amount. Using it in things like Modals becomes untenable. I don't think we should be using the Grid component, its too prescriptive, either we add padding to offset the row margins, or just let the user handle it.

taion:
Fixes #1276 

I think this is okay to PR against `master` because, while the emitted markup does change, the net padding applied when using standard markup does not, and the new behavior is more correct/desirable semantically.

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -1,10 +1,9 @@
+import classNames from 'classnames';
 import React, { cloneElement } from 'react';
 
 import Col from './Col';
-import Grid from './Grid';
 import Nav from './Nav';
 import NavItem from './NavItem';
-import Row from './Row';
 import styleMaps from './styleMaps';
 
 import ValidComponentChildren from './utils/ValidComponentChildren';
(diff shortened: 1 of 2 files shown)


jquense:
LGTM, is standalone just a way not render the clearfix? maybe clearfix would be a more descriptive name for the prop.  Just a thought, I think the current name is also fine. removing both the row and grid seems like the right move :+1:

taion:
It's just for the clearfix, so you can use it in an existing grid. I thought it'd be better to have something that could be set with the binary shorthand w/e.g. `<Tabs standalone>`, and we also already use `standalone` with a conceptually similar meaning on `Input` components.

PR #1304, opened 2015-09-08
Remove extraneous Tooltip prop types

taion:
`BootstrapMixin` didn't do anything here, except for adding misleading entries in the props table. The `title` prop doesn't do anything, either.


taion:
Going to just merge this for now since it's purely a change to documentation, not to functionality.

PR #1305, opened 2015-09-09
ListGroup shouldRenderDiv issue

chrishoage:
ListGroup shouldRenderDiv check is not using ValidComponentChildren, causing errors if the child is a falsey value.


taion:
LGTM. Need +1 reviewer.

jquense:
LGTM

Issue #1309, opened 2015-09-10
Update for React v0.14.0-rc1


Guro:
Hi there, im using react-bootstrap@react-pre with react 0.14.0-rc1
and getting this error: Warning: Button(...): React component classes must extend React.Component.
while using Button component from react-bootstrap
any suggestions?

jquense:
nope this is b/c of a breaking change introduced between beta3 and rc

taion:
`Button` extends `React.Component` just fine... or rather it doesn't need to because it's not an ES6 class. I have no idea where your error came from.

https://github.com/react-bootstrap/react-bootstrap/blob/v0.25.100-react-pre.0/src/Button.js#L7

taion:
I'm going to use this issue to track updating for RC1 anyway.

Guro:
I have switched to beta3 and problem resolved

On Saturday, September 12, 2015, Jimmy Jia <notifications@github.com> wrote:

> I'm going to use this issue to track updating for RC1 anyway.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/react-bootstrap/react-bootstrap/issues/1309#issuecomment-139809289>
> .
>


-- 
Sent from Gmail Mobile

taion:
Released.

Kerosene2000:
Hello,
I still get this error, and don't know how you update ReactBootstrap to beta3(or greater) ?
I want to use meteor + react + bootstrap. thx

jquense:
@Kerosene2000 this is a really old issue, the Normal releases of React-bootstrap have supported React 0.14 for like 3 breaking versions now. Just use the latest version

Kerosene2000:
Ok thank you !
Did React-bootstrap can be used with meteor by the way ?
Like [vazco/universe-react-bootstrap](https://github.com/vazco/universe-react-bootstrap) ?

Kerosene2000:
I think the problem can come from here, universe-react-bootstrap use explicitly an [universe-react-bootstrap old 0.24.2  version](https://github.com/vazco/universe-react-bootstrap/blob/master/bower_components/react-bootstrap/bower.json) of your lib. [Your versions](https://github.com/react-bootstrap/react-bootstrap/releases)

And I use this package to simplify the interface between react and meteor, then I got these error messages...
What do you think about that ?

jquense:
[...]

PR #1310, opened 2015-09-10
Updates react dependency to 0.14.0-rc1

mmarcant:


taion:
Per https://github.com/react-bootstrap/react-bootstrap/issues/1309 this won't actually just work.

PR #1323, opened 2015-09-15
Removed unused props from ListGroupItem

taion:
They don't do anything, and they're not used for anything.


AlexKVal:
Yep. Remnants.

PR #1328, opened 2015-09-17
Make Dropdown exports CommonJS-friendly

### LINKED ISSUE #1326: Dropdown default export
We ran into a small issue while trying to make Pivotal UI more specific when requiring react-bootstrap modules.

For most of the modules we use, this works:
```js
var Panel = require('react-bootstrap/lib/Panel');
```

But for Dropdown we have to do:
```js
var Dropdown = require('react-bootstrap/lib/Dropdown').default;
```

When we tried to do it the first way, we got issues invoking `<Dropdown>`.
React was crapping out because the value of the `Dropdown` variable was a module object instead of the class.

Digging in a little, we saw that the transpiled code for `Panel` exported like so:

```js
exports['default'] = Panel;
module.exports = exports['default'];
```

While `Dropdown` was only:
```js
exports['default'] = Dropdown;
```

Not sure if this is something about module exports that I don't understand, but it was kind of confusing that two modules with default exports had to be required in different ways.

 — @atomanyih @clairethompson

taion:
Fixes #1326.

cc @atomanyih @clairethompson


AlexKVal:
There is "lost" `Toggle` property now
screen shot 2015-09-17 at 10 10 30 am

jquense:
Ahh yes the Playground import needs to change, it was doing the "correct" thing and accessing `.default`, so we need to move that back to just using the main export

taion:
Forgot about that. Fixed.

AlexKVal:
LGTM

PR #1334, opened 2015-09-19
[fixed] any props not known by DropdownMenu are passed through to the underlying ul

thealjey:


src/DropdownMenu.js
```diff
@@ -80,37 +80,36 @@ class DropdownMenu extends React.Component {
...
     };
 
     let list = (
       <ul
-        className={classNames(this.props.className, classes)}
+        className={classNames(className, classes)}
         role="menu"
-        aria-labelledby={this.props.labelledBy}
+        aria-labelledby={labelledBy}
+        {...props}
```
jquense: I wouldn't pass the entire props object to the ul you will get issues where  things like id will be on two elements (this and the toggle). instead I would just explicitly apply the style prop like we do className

thealjey: I mean you're the boss and I only care about **`style`** right now, but the menu does not share anything with the toggle, or anything else for that matter, including its children.
Even **`DropdownButton`** is not passing any props down to **`Dropdown.Menu`**.

jquense: ahh yes, sorry, was mostly asleep when I commented, and thought it was on dropdown itself not the menu. yeah, this should be fine then

taion:
LGTM

jquense:
LGTM

thealjey:
great, thanks!
how soon will this land on npm? :smile:

jquense:
sometime this week I'd imagine. depends on when anyone of us get some time

Issue #1339, opened 2015-09-20
Unfortunate misleading documentation about IE8 support


AlexKVal:
As noted in https://github.com/react-bootstrap/react-bootstrap/issues/1338#issuecomment-141775173
there is a link at `R-B` documentation http://react-bootstrap.github.io/getting-started.html#browser-support to `React's` documentation about supporting of browsers http://facebook.github.io/react/docs/working-with-the-browser.html#browser-support-and-polyfills
> At Facebook, we support older browsers, including IE8.

and it is misleading information about the real status about IE8 support https://github.com/react-bootstrap/react-bootstrap/issues/692#issuecomment-106796950.
> only support IE8 so far as folks submit PRs to fix it

It seems it needs clarifying about it at http://react-bootstrap.github.io/getting-started.html#browser-support

`English good text formulation` help wanted
:cherries:

AlexKVal:
resolved by #1340

Issue #1338, opened 2015-09-20
IE8 compatibility issues


dmatteo:
I can't make `react-bootstrap` work in IE8, even while taking all the necessary "precautions" (shims, shams and so forth).
React 0.13 works fine when loaded alone, but when I import anything from `react-bootstrap` things brake.

An exception is thrown in `es5-shams` and it seems to me is due to something trying to use `Object.defineProperty`, but I'm not entirely sure why this is happening.

Is someone else using IE8 without issues?

AlexKVal:
I know that this info won't help at all but just for clarity: It was decided to drop IE8 support https://github.com/react-bootstrap/react-bootstrap/issues/692

There are a lot of `babel` and `babel-runtime` code aside from just `react` codebase.
The error could be in any of those layers.

This project uses only one shim `import 'es5-shim'` here https://github.com/react-bootstrap/react-bootstrap/blob/master/test/index.js#L1 and only for tests to work.

jquense:
I don't think that ie8 support was dropped, we just don't test for it. 

like react it probably won't work in dev mode but the production mode strips out all the defineProperty calls. if ya can't point to where something is broken or want to PR a fix we'd most probably accept it

dmatteo:
That is unfortunate, you probably should have mentioned this in http://react-bootstrap.github.io/getting-started.html#browser-support, where there is a specific mention to IE8.

I'm no big fan of supporting IE8 either, but we specifically chose to go with React + Bootstrap for a major refactor, mostly to have the IE8 compatibility and I was lead to think that `react-bootstrap` was supporting it too.
We're rewriting  Webforms at Podio, which are used not by our customers, but by our customers' customers, and that's the reason why we can't drop IE8 support, while at Podio itself we only support the 2 most recent versions of the major browsers.

I will spend some time trying to pinpoint the issue and see if I can find a solution to PR, but it could be hard if you "stopped caring" about IE8.
I'll keep you posted.

jquense:
Like I said, its not that we have stopped caring, its just that there currently isn't enough resources from maintainers to constantly test, and work on IE8. We do what we can to the extent we can to support it. we've choosen DOM libraries that work with IE8, we do our best to write compatible code as well, but since we don't have the resources to test thoroughly in an IE8 environment its hard to get it write 100%.

We will almost certainly make changes, and accept PR's that fix our oversights, and everyone is working to the extent that is feasibly possible by folks who do this in their spare time to maintain support.

[...]

PR #1340, opened 2015-09-20
IE8 docs update to reflect the current status

dmatteo:
PR addressing the small discussion in issue https://github.com/react-bootstrap/react-bootstrap/issues/1338 and https://github.com/react-bootstrap/react-bootstrap/issues/1339

Proposal for being more specific about the current (and future) support for IE8.
I've split the changes into 2 commits, one addressing IE8 support and one addressing JQuery since I assumed (correct if I'm wrong) that there is no more plan to drop JQuery dependency for IE8

### DIFF
--- a/docs/src/GettingStartedPage.js
+++ b/docs/src/GettingStartedPage.js
@@ -104,6 +104,8 @@ $ bower install react-bootstrap`
                 <div className="bs-docs-section">
                   <h2 className="page-header"><Anchor id="browser-support">Browser support</Anchor></h2>
                   <p>We aim to support all browsers supported by both <a href="http://facebook.github.io/react/docs/working-with-the-browser.html#browser-support-and-polyfills">React</a> and <a href="http://getbootstrap.com/getting-started/#support">Bootstrap</a>.</p>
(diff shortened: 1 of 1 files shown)


jquense:
looks good to me, except the jquery situation. we actually recently switched to the dom-helpers library which gives use ie8 support for all our positioning components without jquery. so we can actually just remove the jquery caveat all together

dmatteo:
That's even better.
I'll remove the whole stuff

dmatteo:
There you go

AlexKVal:
LGTM.

Just a side note about `[docs]` prefix - it is not recognizable by `mt-changelog` package this project uses for `Changelog.md` generation.
Therefore it won't be reflected in `Changelog.md`.
But it is anyway doesn't fall into any of `[added] | [changed] | [fixed] (code)` category because it is just a clarification. 
And there is no harm in this prefix whatsoever :smile: 
I merge it as is.

@dmatteo Thank you ! :cherries:

dmatteo:
Yeah I saw that there was no mention about it in the guidelines, but it felt right to distinguish it :)

Cheers

PR #1341, opened 2015-09-20
[changed] use `lodash-compat` in src/ for IE8 compatibility

dmatteo:
`lodash` needs to be customised in order to be IE8 compatible.
Luckily, the owner is providing a package already customised for that purpose, and that package is `lodash-compat` and this is what this PR is about.

I've replaced every call to `lodash` **in the source code** to `lodash-compat`, while leaving the development part untouched (src, tooling, webpack cfg).

What do you think?

### DIFF
--- a/docs/src/PropTable.js
+++ b/docs/src/PropTable.js
@@ -1,4 +1,4 @@
-import merge from 'lodash/object/merge';
+import merge from 'lodash-compat/object/merge';
 import React from 'react';
 import Glyphicon from '../../src/Glyphicon';
 import Label from '../../src/Label';
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "karma-webpack": "^1.7.0",
     "less": "^2.5.1",
     "less-loader": "^2.2.0",
+    "lodash": "^3.10.1",
     "marked": "^0.3.5",
     "mocha": "^2.2.5",
     "mt-changelog": "^0.6.1",
@@ -113,7 +114,7 @@
     "classnames": "^2.1.3",
     "dom-helpers": "^2.2.4",
     "keycode": "^2.0.0",
-    "lodash": "^3.10.0",
+    "lodash-compat": "^3.10.1",
     "react-overlays": "^0.4.4",
     "uncontrollable": "^3.1.1"
   },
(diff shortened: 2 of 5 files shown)


dmatteo:
Easy enough to be replaced once again when transitioning to Bootstrap 4

AlexKVal:
LGTM.

Though it needs review from senior developers @react-bootstrap/collaborators

taion:
Hold on - this should be entirely controllable via e.g. Webpack externals for either your app bundle (or for our distributed bundles).

What's the benefit of actually making this change in the imports in the source?

dmatteo:
I don't think leaving this to the consumer is a good idea, if we want to be IE8 friendly.

As per changing it in RB bundle, it could be an option, but I'd prefer to be more explicit in the code, so we always know exactly what we're using.

taion:
People supporting IE8 already have to load up a number of shims/&c. for things to work. What's one more configuration option in the webpack config?

dozoisch:
What are the changes in terms of bundle size and etc. ? If lodash compat is bigger than the standard version, it might be a bad idea to use that directly. Especially since people might use the normal version of lodash which could lead to duplicate imports... If that's possible as a webpack config we could generate a react-bootstrap.compat.min.js which would change the lodash used to that one for the distribution. And if people are linking to lib directly with webpack, they can probably alias it?

dmatteo:
@taion true that, but you can't assume people will be using webpack and you can't force people to change bundler/loader because of a library. 

@dozoisch I don't have my laptop with me, I will verify the size tomorrow, if no one else can do it before.

jquense:
why wouldn't we do it? It generally isn't accepted to use webpack to swap out dependencies of your dependencies? I wouldn't expect to have to alias RB's lodash, thats a bit different then providing polyfills no?

taion:
Is it not possible to do the same w/Browserify? I guess it'd be simpler all around to just cut things over. Just a bit messy IMO.

dmatteo:
I guess it is, but there's also Systemjs, requirejs (still) and more in the future, I guess. 
Looks impractical to me. 

Thx and I will keep digging to find more issue with IE8.
It will be nice if we could tag an IE8 tested version so people could fallback to this "old" version, if the real one is not working. 

Ofc, first of all, we need to have a working version :)

PR #1342, opened 2015-09-20
[added] 'onShow' property to Modal.

codykrainock:
I'm currently building a webapp using react-bootstrap and I am using the Modal class. However, I wanted the ability to pass a callback that would be executed when the modal is shown.

Please review and let me know what you think, it's a rather simple change. Thanks!


codykrainock:
Hold on, tests aren't passing. I will open a new request when this is ready.

PR #1343, opened 2015-09-20
[added] 'onShow' property to Modal.

codykrainock:
I'm currently building a webapp using react-bootstrap and I am using the Modal class. However, I wanted the ability to pass a callback that would be executed when the modal is shown.

Please review and let me know what you think, it's a rather simple change. Thanks!

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -117,7 +117,12 @@ const Modal = React.createClass({
     /**
      * When `true` The modal will show itself.
      */
-    show: React.PropTypes.bool
+    show: React.PropTypes.bool,
+
+    /**
+     * When supplied this function will be called upon showing the modal.
+     */
+    onShow: React.PropTypes.func
   },
 
   getDefaultProps() {
@@ -314,6 +319,10 @@ const Modal = React.createClass({
 
     if (this.props.backdrop) {
       this.iosClickHack();
+    }
+
+    if (this.props.onShow) {
+      this.props.onShow();
     }
 
     this.setState(this._getStyles(), () => this.focusModalContent());
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -257,6 +257,17 @@ describe('Modal', () => {
     assert.notInclude(document.body.className, 'modal-open');
   });
 
+  it('Should call onShow if it is supplied as a property', () => {
+    let count = 0;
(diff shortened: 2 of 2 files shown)


taion:
You could have just re-pushed to the old PR.

Is the concern that the `onEntering` and `onEntered` animation lifecycle hooks are insufficient?

codykrainock:
I wasn't aware that these lifecycle hooks were available since it didn't explicitly say so in the Modal docs - that's what led me to implement `onShow`. Upon further investigation I now see that this exists on the underlying Fade element, which is only used when animation == true.

I think it would be smart to have an `onShow` property simply so that the Modal element isn't tightly coupled to the Fade element. Also, I do feel that the language `onShow` is  much clearer, especially given that the `onHide` prop is available for Modal.

Let me know what your thoughts are.

codykrainock:
I have confirmed through some tests that `onEntering` and `onEntered` are only used when `animation === true`. For this reason I think the `onShow` property would be more sufficient than these lifestyle hooks.

taion:
@jquense 

Any thoughts here? This isn't quite parallel to `onHide`, which does the actual hiding.

I feel that since the developer controls the modal being shown (e.g. by doing `this.setState({shown: true})`) that it's not clear that this adds any new capabilities.

jquense:
I have gone back and forth on this myself...generally an onShow is helpful if you previously had stuff in the `{Wii|Did}Mount` hooks.

However it has been hard to fine something that is a reasonable parallel to `onHide`. At work we settled on adding a hook like:

```js
componentWillMount() {
    if (this.props.onShow && this.props.show)
      this.props.onShow();
  }

  componentWillReceiveProps(nextProps) {
    if (nextProps.onShow && nextProps.show && !this.props.show)
      nextProps.onShow();
  }
```

Its been convenient, but it's also easy enough for a user to do themselves. I'd add it if folks don't think it will be confusing that onHide triggers a hide while onShow is an unpreventable hook that's just a convenient alias for checking the `show` prop

taion:
@codykrainock Sorry for letting this languish. I think the right approach here is to let you embed a `ModalDialog` of some form inside a `Modal`, and then use the standard React lifecycle hooks there.

codykrainock:
@taion - thanks for the input. I'll definitely give it a shot!

Issue #1181, opened 2015-08-18
How to determine which element `ListGroup` should output.


apkiernan:
As mentioned in #1169 continuing discussion here.

Based on the discussion, it sounds like the current thinking is:
- Check the children of `ListGroup`, and if they are `ListGroupItem`s then we can render as we do currently
- If they are not (user created custom components), then we would fall back to a componentClass prop that the user would specify whether to render `<ul>` or `<div>` and render the user specified element.

taion:
I think that's right. For the latter, `div` should probably be the default. I think this is the safest way to handle it.

apkiernan:
For the `componentClass` prop, I believe everywhere we use this prop its propType is [elementType](https://github.com/react-bootstrap/react-prop-types/blob/master/src/elementType.js). Should we still use that here? I could see using it for consistency, but `<ListGroup>` should only be `<div>` or `<ul>` so I wonder if `React.PropTypes.oneOf(['div', 'ul'])` would be more accurate.

AlexKVal:
I vote for `React.PropTypes.oneOf(['div', 'ul'])` :sparkle:

taion:
Does it make sense to have a custom container component here?

taion:
I mean a custom React component.

apkiernan:
I can see using a custom component, but I think making it `oneOf(['ul', 'div'])` is the right call, as I think we want to ensure that the correct markup is being used.

taion:
I'm okay with that. It wouldn't be a breaking change to allow custom components down the road anyway.

taion:
Fixed by #1348.

PR #1348, opened 2015-09-23
[added] #1181 ListGroup supports componentClass prop

apkiernan:
As discussed, this keeps the handling for the case where all children are `<ListGroupItem>`, but now users can specify the element `<ListGroup>` should output via the componentClass prop, which defaults to a div.

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -733,6 +733,15 @@ const ComponentsPage = React.createClass({
                   <p>Set the <code>header</code> prop to create a structured item, with a heading and a body area.</p>
                   <ReactPlayground codeText={Samples.ListGroupHeader} />
 
+                  <h3><Anchor id="listgroup-with-custom-children">With custom component children</Anchor></h3>
+                  <p>
+                    When using ListGroupItems directly, ListGroup looks at whether the items have href
+                    or onClick props to determine which DOM elements to emit. However, with custom item
+                    components as children to <code>ListGroup</code>, set the
+                    <code>componentClass</code> prop to specify which element <code>ListGroup</code> should output.
+                  </p>
+                  <ReactPlayground codeText={Samples.ListGroupCustom} />
+
                   <h3><Anchor id="listgroup-props">Props</Anchor></h3>
 
                   <h4><Anchor id="listgroup-props-group">ListGroup</Anchor></h4>
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -1,4 +1,5 @@
 import React, { cloneElement } from 'react';
+import ListGroupItem from './ListGroupItem';
 import classNames from 'classnames';
 import ValidComponentChildren from './utils/ValidComponentChildren';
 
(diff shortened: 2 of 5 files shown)


src/ListGroup.js
```diff
@@ -9,6 +9,11 @@ class ListGroup extends React.Component {
       (item, index) => cloneElement(item, { key: item.key ? item.key : index })
     );
 
+    if (this.areCustomChildren(items)) {
+      let componentClass = this.props.componentClass;
+      return componentClass === 'ul' ? this.renderUL(items) : this.renderDiv(items);
```
taion: This is a little weird - shouldn't this use the actual component class?

taion: I mean, the specified `componentClass` prop.

src/ListGroup.js
```diff
@@ -21,16 +26,25 @@ class ListGroup extends React.Component {
...
 
   isAnchorOrButton(props) {
     return (props.href || props.onClick);
   }
 
+  areCustomChildren(children) {
+    let customChildren = false;
+
+    ValidComponentChildren.forEach(children, (child) => {
+      if (child.type.displayName !== 'ListGroupItem') {
```
taion: Can this just be `child.type === ListGroupItem`?

docs/src/ComponentsPage.js
```diff
@@ -733,6 +733,12 @@ const ComponentsPage = React.createClass({
                   <p>Set the <code>header</code> prop to create a structured item, with a heading and a body area.</p>
                   <ReactPlayground codeText={Samples.ListGroupHeader} />
 
+                  <h3><Anchor id="listgroup-with-custom-children">With custom component children</Anchor></h3>
```
taion: Does this example add a lot? I think it might be sufficient just to clarify when `componentClass` is needed in the prop documentation for `ListGroup`.

src/ListGroup.js
jquense: better yet `bsRole` to be consistent with the newer dropdown api?

taion: I think we want to assert on this being our `ListGroupItem`, because otherwise we don't know a priori whether it's going to emit a `div` or an `li`.

src/ListGroup.js
apkiernan: Tbh I didn't know you could specify the componentClass itself in the oneOf propType, etc. Fixing.

docs/src/ComponentsPage.js
apkiernan: Not sure it adds a ton, but I thought it better to over document then under. I can remove.

taion: Fair enough. I might reword the copy to say something like:

>
When using `ListGroupItem`s directly, `ListGroup` looks at whether the items have `href` or `onClick` props to determine which DOM elements to emit. However, with custom item components ...

src/ListGroup.js
apkiernan: I'm not sure of another way to do this check. Using `child.type` returns the function definition. I agree it is awkward, and if there is a different way to do it, it would probably be preferable.

taion: You can do `child.type === ListGroupItem` though. Try e.g.

```js
class C extends React.Component {
  render() {
    return (
      <div>
        {String(this.props.children[0].type === ListGroupItem)}
        <ListGroup>
          {this.props.children}
        </ListGroup>
      </div>
    );
  }
}

const listgroupInstance = (
  <C>
    <ListGroupItem>Item 1</ListGroupItem>
    <ListGroupItem>Item 2</ListGroupItem>
    <ListGroupItem>...</ListGroupItem>
  </C>
);

React.render(listgroupInstance, mountNode);
```

apkiernan:
I think I got all of the edits in here. Let me know if I missed anything.

Issue #1320, opened 2015-09-13
Can't style for tab component


KhueHuynh:
Hi guys,
I'm using Tabs component in react-bootstrap. But I can not tab style (it's render in tabs, function renderTab). I think you can add className for every tabs ?

Can you help me fix it ? 
Thanks for your contributions on react-bootstrap. It's useful.

abhinavsingi:
'Tabs' component does take className as 'contentContainerClassName' but not the 'Tab' component

mannyyang:
Hey KhueHuynh,

I'll be tackling this sometime this week. Will comment here again when finished and sending the pull request.

AlexKVal:
Closed by #1368

PR #1368, opened 2015-10-01
[added] #1320 allow NavItem class to be set

egauci:
Tabs currently allows setting of the pane className through the className prop. This adds a tabClassName prop which gets merged into the corresponding NavItem's className.

I am unsure how to write this up for the docs. Any hints?

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -88,7 +88,6 @@ const Tabs = React.createClass({
       React.PropTypes.number,
       React.PropTypes.object
     ]),
-    className: React.PropTypes.string,
     /**
      * Render without clearfix if horizontally positioned
      */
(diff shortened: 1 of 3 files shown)


src/Tabs.js
```diff
@@ -88,6 +88,7 @@ const Tabs = React.createClass({
       React.PropTypes.number,
       React.PropTypes.object
     ]),
+    className: React.PropTypes.string,
```
taion: This is unnecessary, we don't validate the `className` prop here.

taion:
I would add this to the `propTypes` for `Tab` with a comment explaining what it does so our prop table generation picks up on it, then perhaps include it in an example (or not).

egauci:
Thank you @taion for your suggestions.

I'm going to take a look at examples next.

taion:
I'm actually okay with this as-is. I don't know that we explicitly need to show this in an example.

AlexKVal:
LGTM as-is. (I believe that it is a no big deal with the second commit and I merge it.)

@egauci  Thank you for the contribution :cherries:

Issue #1233, opened 2015-08-27
Dropdown should only focus item when keyboard is pressed


jquense:
from #1213. CC @rstormsf 

I didn't realize that the menu focusing was happening on mouse clicks as well, that is [probably not the proper behavior](http://www.w3.org/TR/wai-aria-practices/#menu) but ARIA stuff only really covers keyboard interaction. Baring someone informing us otherwise I think we should limit the "open and focus first" to keyboard triggers.

mtscout6:
In #1213 @stubbornella pointed out that she's not a fan of divergent behavior between keyboard and mouse users. While I can agree to a point I think these are vastly different concerns. I think it can be assumed that if a user used the mouse they will continue to do so, thus negating the need to display focus.

One example from some accessibility guidelines describe an invisible ["skip to main content"](http://webaim.org/techniques/skipnav/) link prior to site navigation. You can see such an example on the [WebAIM](http://webaim.org/) page. After the page load and before you click on anything hit the TAB key. You'll notice a new button show up that is specific for keyboard users. I know this is not the same as Dropdowns, but it does highlight that sometimes it makes sense to have varying behavior for keyboard vs. mouse users.

jquense:
So having just checked some implementations I am starting to lean a bit the other way. [jquery UI](https://jqueryui.com/menu/) has our current behavior (focus on open), and it is highly respected in its a11y work. [kendo ui](http://demos.telerik.com/kendo-ui/menu/keyboard-navigation) does not, though i would go with jquery ui over kendo in terms of trusting a11y work

taion:
Is it worth documenting that we're explicitly not matching upstream TWBS here: http://getbootstrap.com/components/#btn-dropdowns-single? My a11y knowledge is too limited to comment on which is more correct - it seems fine to not match TWBS for the purpose of improved a11y, but we should probably document it more clearly.

rstormsf:
jquery ui, makes it focused, but they don't have browser focused behavior(blue one with Chrome) which makes it looks nice. 
so if you want to go that route, it would be great not to have ugly blue highlight in menu in my case with Chrome.

trevorbernard:
Is there any way to override this functionality? If no, should there be a way to?

taion:
[...]

PR #1369, opened 2015-10-02
Dropdown no longer autofocuses first item on open

joshuahhh:
Click on a `NavDropdown` on https://react-bootstrap.github.io/components.html#navs-dropdown, and you'll get something like this:
screen shot 2015-10-01 at 8 16 04 pm

But click on a dropdown on http://getbootstrap.com/components/#nav, and you get:
screen shot 2015-10-01 at 8 16 27 pm

As you can see, `react-bootstrap` autofocuses the first menu item when you click on the button, while non-React Bootstrap does not. I am not sure if this change is intended, but it is certainly not what I want – it is unexpected and disorienting for mouse users.

I believe the code I remove here is responsible. Does this look OK?

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -67,10 +67,6 @@ class Dropdown extends React.Component {
   componentDidUpdate(prevProps) {
     let menu = this.refs.menu;
 
-    if (this.props.open && !prevProps.open && menu.focusNext) {
-      menu.focusNext();
-    }
-
     if (!this.props.open && prevProps.open) {
       // if focus hasn't already moved from the menu lets return it
       // to the toggle


taion:
Not as-is: ARIA unambiguously requires focusing on the first item on keyboard interaction, per http://www.w3.org/TR/wai-aria-practices/#menu, and this is not negotiable; TWBS is just doing the wrong thing here.

Per https://github.com/react-bootstrap/react-bootstrap/issues/1233 I think our preference would be to make the default be to not focus on first item on mouse interaction, but also to make the on-click behavior customizable such that the current behavior is available as an option.

We'd be happy to accept a PR that made such a change, assuming it also updated tests accordingly.

taion:
I'm closing this in favor of a proper fix.

PR #1373, opened 2015-10-03
[fixed] Dropdown focus behavior on click

### LINKED ISSUE #1233: Dropdown should only focus item when keyboard is pressed

taion:
By default, only focus the first dropdown menu item after keydown.

Fixes #1233.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -46,13 +46,13 @@ class Dropdown extends React.Component {
     }];
 
     this.state = {};
+
+    this.lastOpenEventType = null;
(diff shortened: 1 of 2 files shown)


jquense:
LGTM, man focus stuff always gets so messy

PR #1374, opened 2015-10-03
Fix contained popover examples

### LINKED ISSUE #1177: Popover Scrolling example doesn't work on first render.
http://react-bootstrap.github.io/components.html#popovers-positioned-scrolling

Doesn't work on the first render, because mountNode is null, I assume its due to the way @taion is handling the initial render for the examples.

If you trigger a compile on the live editable code it works perfectly

taion:
Fixes #1177 per the recommendation there


jquense:
LGTM

PR #1375, opened 2015-10-03
Use role to determine dropdown open focus behavior

### LINKED ISSUE #1233: Dropdown should only focus item when keyboard is pressed

taion:
Actually fixes #1233.

I think this is a little cleaner - http://www.w3.org/TR/wai-aria-practices/#menubutton specifies explicitly to not change focus on space or enter, which I will interpret as also covering mouse clicks.

### DIFF
--- a/test/DropdownSpec.js
+++ b/test/DropdownSpec.js
@@ -416,7 +416,7 @@ describe('Dropdown', () => {
     // I am fairly confident that the failure is due to a test specific conflict and not an actual bug.
     it('when open and the key "esc" is pressed the menu is closed and focus is returned to the button', () => {
       const instance = React.render(
-        <Dropdown defaultOpen alwaysFocusNextOnOpen id='test-id'>
+        <Dropdown defaultOpen role="menuitem" id="test-id">
           {dropdownChildren}
         </Dropdown>
       , focusableContainer);
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -48,7 +48,6 @@ class Dropdown extends React.Component {
     this.state = {};
 
     this.lastOpenEventType = null;
-    this.isKeyboardClick = false;
   }
 
   componentDidMount() {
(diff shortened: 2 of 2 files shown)


jquense:
I think this is right on the money. :+1:

test/DropdownSpec.js
```diff
@@ -416,7 +416,7 @@ describe('Dropdown', () => {
     // I am fairly confident that the failure is due to a test specific conflict and not an actual bug.
     it('when open and the key "esc" is pressed the menu is closed and focus is returned to the button', () => {
       const instance = React.render(
-        <Dropdown defaultOpen alwaysFocusNextOnOpen id='test-id'>
+        <Dropdown defaultOpen role="menuitem" id="test-id">
```
jquense: is this the right test?

taion:
It also saves us from that horribly named `alwaysFocusNextOnOpen` prop.

test/DropdownSpec.js
taion: Yes, because of https://github.com/react-bootstrap/react-bootstrap/pull/1375/files#diff-7d1cbca6e042a330878fd99a6c9a90b9R427. I need this to make the test pass.

taion: By which I mean that this is the fix I need to make tests pass. None of the other tests seemed to cover this behavior so in some broader sense this is not "correct". ¯\\\_(ツ)_/¯

jquense: ahhh gotcha

AlexKVal:
This is a more evident way and the code looks much cleaner now :+1:

Issue #1371, opened 2015-10-02
[Pagination] how to add text for 'Previous' and 'Next' buttons


lovedota:
Hi friends.

I am using React Bootstrap, and i want ask there are some options for adding text for *previous* and *next* button like this. Thanks !

```
<BootstrapPagination
       bsSize='small'
       prev
       next
       first
       last
       ellipsis={false}
       items={pages}
       maxButtons={Math.min(pages, 5)}
       activePage={this.props.options.currentPage}
       onSelect={this.onPageSelect.bind(this)}
/>
```

![pagination](https://cloud.githubusercontent.com/assets/417914/10244419/56235e54-6929-11e5-88cf-00ac430ac929.png)

taion:
Not yet, but we'd happily take a PR for this; something along the lines of making `first` and the others take a `React.PropTypes.node` as an argument.

chilts:
I'd be keen on tackling this.

Quick question: currently we show (for example) `&laquo;` for first. First is a `React.PropTypes.bool`. If we were to change that to `React.PropTypes.node`, then presumably the defaultProps would be `null`.

When checking if we have nothing in `renderFirst()`, we would obviously show nothing. If we have something, then we render that. However, we lose the ability to have a default *yes* which would just show the `&laquo;`. I'd propose we somehow want to keep that but happy to be guided here.

Of course, you could pass that in as `first` ... but then you gotta remember to do it and it's not automatic like it currently is. :)

taion:
This was actually fixed by https://github.com/react-bootstrap/react-bootstrap/pull/1376 and https://github.com/react-bootstrap/react-bootstrap/pull/1400. Not released yet though.

sedi0861:
Is this functionality released? Could not find proper documentation. Please give me the link if it is there.

PR #1376, opened 2015-10-03
[added] Custom labels for Pagination's special element

mdziekon:
Hi, I've added the ability to use custom labels for Pagination's special buttons (ellipsis, first, last, previous and next). It seems I'm not the only one who would like to see this feature, as it was reported in the issue tracker a couple of days ago here: https://github.com/react-bootstrap/react-bootstrap/issues/1371

I've created new props for this purpose - ``ellipsisLabel`` (and others), without modifying the old ``ellipsis`` boolean. I was considering merging them both into one prop, but there would be a problem with default values - any value other than ``false`` (so, any default value we show right now) would be considered as "display this element for me", so this would break the current behaviour for first, last, previous and next elements.

There is another thing which might be "weird" about this PR - I've used unicode numbers to provide elements' sigils, instead of writing unicode characters directly into the component's file. According to [this page](http://facebook.github.io/react/docs/jsx-gotchas.html) it's a safer alternative, but if that's not compliant with other pieces of the code (I saw unicode characters withing test cases code), please let me know and I will change it.


AlexKVal:
A bit weird looking props-documentation
http://localhost:4000/components.html#pagination
screen shot 2015-10-04 at 10 00 53 am

It can be fixed like that [src/BootstrapMixin.js#L14](https://github.com/react-bootstrap/react-bootstrap/blob/f41090465a54d43b60c988fe0cef3022a53cd7d2/src/BootstrapMixin.js#L14)
```js
/**
 * Style variants
 * @type {("default"|"primary"|"success"|"info"|"warning"|"danger"|"link")}
 */
bsStyle: React.PropTypes.oneOf(styleMaps.STYLES),
```
by explicitly documenting those `<` `>` etc.

otherwise LGTM :sparkle:

mdziekon:
@AlexKVal: I can either display it as a Unicode character (but it will be saved as that character in the code), or as the HTML named entity (in that case it's a regular string in the code). Which one should I choose?

AlexKVal:
It doesn't really matter. 
Any approach that works will suit. 
(that's the question I don't know answer for, because I'm not an author of props-docs' auto-generation code, @jquense is) 

As a result we should have the right simbols / text on the docs page.

You can check it by `npm run docs` - that's for development with hot-reloading.
And `npm run docs-prod`. The latter command we always run to check how it would look on the site after publishing.

mdziekon:
It appears there is a problem with parsing unicode characters in this docs auto-generator:
docu

(When using unicode characters, default values loose their last character, as on the attached screenshot)

 @jquense: Could you take a look at this problem?

AlexKVal:
Then I presume we could use a hack just for now until generator will be fixed up.

I believe it is not crucial to show in the documentation exact unicode simbol.
We safely could use, for example, `<<` and `>>`.
What do you thinkg @react-bootstrap/collaborators guys ?

taion:
@mdziekon I feel like we could do something like `true` means show the default label value, `false` or `null` mean show nothing, and anything else means "show the specified label". Would that not work?

mdziekon:
@taion I think it will, but we still should show what actually is the default label. So it solves the "redundancy" problem, but not the "docs" problem. For now, we could follow @AlexKVal's idea about putting ``<<`` [and other] sigils in place of unicode symbols, but we also could show which HTML named entity it uses by default.

taion:
That makes sense to me. I just think it's a slightly cleaner API, maybe. I don't feel strongly about that though.

taion:
@mdziekon Would you be able to update the PR accordingly? I think it's fine just to show the HTML named entity. The actual documentation example will show what is really rendered anyway.

mdziekon:
@taion: Done - merged both "the flag" and "the label" into one prop, which takes either ``bool`` or ``node`` value now. Hope I didn't make any mistakes in the docu :)

taion:
Sorry, to clarify for me, why do we need the strict `false` check instead of just a falsiness check like on `Input` for the label? I don't think we're going to omit any actual use cases by not showing a label on things like `null` or `''`. Then you can write the comment as something like

```js
// If `true`, the default label ('foo'). Otherwise, the label to display if specified.
```

mdziekon:
@taion TBH, I'm still not sure myself about this strict comparison (we probably should give as much "freedom" as users like, but on the other hand, who would want to render an empty "next" button).

But there will be one "problem" with checking falsiness of values - empty arrays are treated similarly as ``null` (they are rendered by react the same way - as nothing), but ``!![] === true`` (empty arrays are not falsy). And since we probably want to maintain a unified handling of such cases, this would introduce slightly more complicated comparison.
Of course it's probably not a big deal, so I can adjust the code to be less strict about the value (and include the additional array test).

taion:
Do empty arrays satisfy `React.PropTypes.node`?

I guess where I'm coming from is that in other places where we have similar patterns, we just check falsiness: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/InputBase.js#L145.

Granted, those cases don't take booleans, but broadly the behavior is identical except when an explicit `true` is passed in.

taion:
I'll leave it up to you - LGTM either way. Need one more reviewer. (But let's hold off on the the React v0.14 release to actually merge)

taion:
:+1:

Need one more from @react-bootstrap/collaborators.

AlexKVal:
I've rebased it to the latest updates and changes and it throws an error:
screen shot 2015-10-09 at 6 42 12 pm

I'll fix it next PR.

Docs looks OK:
screen shot 2015-10-09 at 6 39 36 pm

I merge it.

AlexKVal:
@mdziekon Thank you for the contribution :tada: 
:cherries:

PR #1381, opened 2015-10-05
Added title for close button

vguzev:
Our QA team requested titles for Close buttons. So, I decided it would be good to patch react-bootstrap instead of creating custom ModalHeader...


taion:
This seems to assign the `aria-label` for the entire `ModalHeader` to the close button. That doesn't seem right to me.

taion:
Closing this for now because this seems like it's not the right thing.

PR #1382, opened 2015-10-05
Explain why to import components from lib/

### LINKED ISSUE #986: Add documentation to describe why you'd want to only import the components you plan to use.
It'd be great to have some kind of "Customize" section just like Bootstrap's, where people can choose to install only some components.

I'd be interested in helping build this, but just because I'm not super familiar with npm: would this be the sort of case where you can add a flag of some kind that installs only certain things, or would that mean having a separate node module for each component?

taion:
Fixes #986


jquense:
LGTM

PR #1391, opened 2015-10-07
Update for React v0.14.0

taion:
This is really two changes.

- Sync the `react-14` branch with `master` and update merge conflicts and tests for everything since the last `react-pre` release
- Use `react-prop-types/deprecated` instead of `deprecationWarning` for deprecated props

Not sure if anybody wants to look at the former - it's entirely formulaic. The latter is worth taking a peek at.

We'll want to update the react-overlays dependency shortly, but it's not actually a blocker.


taion:
Updated the react-overlays dep to the new release. Thanks @jquense!

mathieumg:
Reviewed https://github.com/react-bootstrap/react-bootstrap/commit/5376f887683de380fbb8031ccadee5a7b4fa3908

LGTM, haven't personally tested it however. (or 0.14 itself yet for that matter...)

Issue #1393, opened 2015-10-07
Bundling errors with React 0.14 


insin:
A bunch of internals have been moved to the [fbjs](https://github.com/facebook/fbjs) package:

```
ERROR in ./~/react-bootstrap/lib/OverlayTrigger.js
Module not found: Error: Cannot resolve module 'react/lib/warning'
 @ ./~/react-bootstrap/lib/OverlayTrigger.js 32:23-51

ERROR in ./~/react-bootstrap/lib/utils/deprecationWarning.js
Module not found: Error: Cannot resolve module 'react/lib/warning'
 @ ./~/react-bootstrap/lib/utils/deprecationWarning.js 11:23-51
```

jquense:
Yeah 0.26.0 isn't really compatible, 0.27.0 coming shortly

taion:
Fixed by #1394.

PR #1394, opened 2015-10-07
Use caret range for React dependency

### LINKED ISSUE #1392: React peer dependency is installing react 0.14
I think the peer dependency needs to be changed to ~ or ^ for 0.13. Currently npm will install react twice in my project, v0.13 for my app and v0.14 for react-bootstrap. This causes errors because two incompatible versions are loaded at the same time.

taion:
This release is not really compatible with React 0.14.

Fixes #1392 and #1393 (I guess? There's nothing wrong per se with what we have in `package.json` ¯\\\_(ツ)_/¯ , but we're not really compatible with React 0.14 on this release anyway).

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   "author": "Stephen J. Collings <stevoland@gmail.com>",
   "license": "MIT",
   "peerDependencies": {
-    "react": ">=0.13"
+    "react": "^0.13"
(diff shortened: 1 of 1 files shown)


jquense:
:rainbow: now we know. LGTM

PR #1398, opened 2015-10-08
Don't include react-dom in the bundles

### LINKED ISSUE #1397: Provide pre-built package without React bundled in it
I noticed there is no easy way to use this package in a project not depending on nodejs. Correct me if I'm wrong, but there are no pre-built packages around that I could just add to `<script>` or requirejs config. The closest thing I found is react-bootstrap-bower, however it has a version of react bundled with it, which means it does not work if React is already included in the project

https://raw.githubusercontent.com/react-bootstrap/react-bootstrap-bower/master/react-bootstrap.js

taion:
Fixes #1397 

I'll release this as v0.27.1.

### DIFF
--- a/docs/src/GettingStartedPage.js
+++ b/docs/src/GettingStartedPage.js
@@ -85,13 +85,14 @@ import { Button } from 'react-bootstrap';`
                   </div>
 
                   <h3><Anchor id="browser-globals">Browser globals</Anchor></h3>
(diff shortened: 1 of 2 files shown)


taion:
That AppVeyor failures looks totally spurious.

jquense:
LGTM

PR #1399, opened 2015-10-08
Update dependencies

taion:


AlexKVal:
LGTM.
Generated docs looks good too.

(though I'll report in the according issue that 'undefined' for 'aria-like' properties still is an issue)

PR #1411, opened 2015-10-16
Fix logging from afterEach hook in tests

taion:
Logging was previously messed up because this function tried to use `this.currentTest`.


taion:
Just gonna go ahead and merge this since all it affects is error logging from tests.

PR #1413, opened 2015-10-16
[fixed] Put onClick on correct element on NavItem

taion:
This ignores the `disabled` prop on the `NavItem` BTW... seemed more consistent since e.g. `onClick` on `Button` does the same.

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -1,7 +1,9 @@
-import React from 'react';
 import classNames from 'classnames';
+import React from 'react';
+
 import BootstrapMixin from './BootstrapMixin';
 import SafeAnchor from './SafeAnchor';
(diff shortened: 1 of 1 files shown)


jquense:
LGTM tho there is probably larger discussion be had about the disabled thing, some components do seem to respect it, like dropdown.

taion:
@jquense How so? `MenuItem` doesn't respect `onClick`. `Dropdown` for the toggle respects `disabled` only for the built-in click handler, not for any `onClick` passed in.

jquense:
Ahh yeah missed that distinction. although Button is blocked natively since the underlying a/button is getting a disabled attribute. upstream definitely doesn't do anything to block onClicks in the case of menu and nav items since it's just a cosmetic class. I've always found that annoying but that's not necessarily any reason to differ

PR #1414, opened 2015-10-16
[fixed] Respect onClick on MenuItem

taion:
Fixes https://github.com/react-bootstrap/react-router-bootstrap/issues/112.

Same caveat as on #1413.

### DIFF
--- a/src/MenuItem.js
+++ b/src/MenuItem.js
@@ -1,7 +1,9 @@
-import React from 'react';
 import classnames from 'classnames';
+import React from 'react';
 import all from 'react-prop-types/lib/all';
+
 import SafeAnchor from './SafeAnchor';
+import createChainedFunction from './utils/createChainedFunction';
 
 export default class MenuItem extends React.Component {
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/MenuItem.js
```diff
@@ -35,6 +37,8 @@ export default class MenuItem extends React.Component {
       );
     }
 
+    const {className, style, href, onClick, ...props} = this.props;
```
mathieumg: I imagine `href = '',` could have worked here?

mathieumg:
LGTM

src/MenuItem.js
taion: Not quite - the destructuring default only works on `undefined`, where as the `href || ''` replaces any falsy values. I don't know if that part matters, but it'd be an additional behavior change I didn't want to introduce.

mathieumg: Alright, that's a subtle difference but it makes sense, thanks for clarifying!

jchapple:
Cheers for getting this fixed, @taion. :+1:

PR #1416, opened 2015-10-17
[fixed] DropdownTitle children v. title

### LINKED ISSUE #1410: Dropdown Toggle title prop vs. html global attr title
Using the prop named "title" as the button text disallows the ability to get a html global attribute "title" (default browser tooltip behavior) which is desirable at times for buttons.
usage is here:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/DropdownToggle.js#L26

Might you consider renaming the title prop and allowing the html title attribute to be added to the control?

taion:
Prefer children for label, but default to title otherwise

Fixes #1410

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import classNames from 'classnames';
 import Button from './Button';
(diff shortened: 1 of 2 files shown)


AlexKVal:
LGTM :sparkle: 
(docs looks the same too)

PR #1421, opened 2015-10-19
Fix tooltip in copy example

### LINKED ISSUE #1420: Tooltip "placement" prop not working correctly (even on your example page)
If you navigate to the tooltips section of the component example page (https://react-bootstrap.github.io/components.html#tooltips), the "In Text Copy" example has the following line of code:

````
let tooltip = <Tooltip placement="top">{this.props.tooltip}</Tooltip>;
````

This `placement="top"` statement does not appear to work correctly (tooltip is appearing on the right).

taion:
Fixes #1420


AlexKVal:
:shipit:

PR #1427, opened 2015-10-21
Add table headers

mwanji:
This PR adds a `headers` prop to `Table`. `headers` is an array of strings that generates the column headers in <thead>, as well as an empty <tbody>. It removes a fair bit of boilerplate.

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -7,7 +7,8 @@ const Table = React.createClass({
     bordered: React.PropTypes.bool,
     condensed: React.PropTypes.bool,
     hover: React.PropTypes.bool,
-    responsive: React.PropTypes.bool
+    responsive: React.PropTypes.bool,
+    headers: React.PropTypes.arrayOf(React.PropTypes.string)
   },
 
   getDefaultProps() {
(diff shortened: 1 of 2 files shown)


taion:
I'm inclined against merging this. We've generally been preferring more explicit and flexible APIs.

mwanji:
Would you prefer something like this?
```
<Table>
  <Headers><Header>Col 1</Header></Headers>
  <Body>
    ...
  </Body>
</Table>
```

Right now, Table doesn't seem to add a lot of value.

taion:
That's a good thing! It's exactly like working with a standard `<table>` - we just wrap up some of the Bootstrap styling bits to be a little cleaner.

taion:
I'm going to go ahead and close this for now. Thanks for the PR, and the code looks good, but I don't think this API is where we want to go.

It's slightly simpler/cleaner only for the case where you have one header per column, and all of your headers are strings. It's not super expressive when the headers might be React nodes, and doesn't let you do things like apply `colspan`.

Given that, the convenience added here is insufficient to justify expanding the API surface area, as I think the current API looks good enough given that it's just standard HTML table markup.

mwanji:
I accept your rejection, but my proposal seems quite similar to how Panel has header and footer props and ListGroupItem has a header prop.

PR #1429, opened 2015-10-21
Remove cross import between Button & ButtonInput

### LINKED ISSUE #1428: Circular dependency between Button & ButtonInput?
Hi,

I'm getting the following error having updated to 0.27.2 from 0.23.7.

```
TypeError: Cannot read property 'types' of undefined
    at Object.<anonymous> (/home/mike/root/projects/hive/portal/node_modules/react-bootstrap/lib/Button.js:54:70)
    at Module._compile (module.js:460:26)
    at Module._extensions..js (module.js:478:10)
    at Object.require.extensions.(anonymous function) [as .js] (/home/mike/root/projects/hive/portal/node_modules/babel-core/lib/api/register/node.js:214:7)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/mike/root/projects/hive/portal/node_modules/react-bootstrap/lib/ButtonInput.js:19:15)
    at Module._compile (module.js:460:26)
```

I'm trying to understand what might be going on. I feel like it might be a circular import issue as Button & ButtonInput import each other which can be an issue. That said, I've no idea why others might not be experiencing this problem which makes me feel like it is something in my set up. I am getting this error from a `express-react-views` set up rather than compiling into a client side bundle.

I'm using:
- babel **5.8.23**
- express-react-views **0.9.0**
- react **0.14.0**

I'm on Ubuntu 14.04. 

My understanding of circular imports in node is limited but I feel like they aren't a great idea though in some situations the problems can be mitigated by extra code as explained here: https://coderwall.com/p/myzvmg/circular-dependencies-in-node-js

Might be wrong about the whole thing though. Any help would be much appreciated. Thank you for working on and sharing this project. I've finding it extremely useful.

Cheers,
Michael

michaeljones:
Hi,

This is an attempt to fix #1428 which is where I'm having trouble using `ButtonInput` in a server side React render. I believe this is due to cross imports between `ButtonInput` and `Button` which might not be an issue when using a client bundle?

The details are in the commit message.

There might be a better way to solve the issue but I hope this helps if it is the preferred route.

Cheers,
Michael

### DIFF
--- a/src/ButtonInput.js
+++ b/src/ButtonInput.js
@@ -17,7 +17,7 @@ class ButtonInput extends InputBase {
   }
 }
 
-ButtonInput.types = ['button', 'reset', 'submit'];
+ButtonInput.types = Button.types;
 
 ButtonInput.defaultProps = {
   type: 'button'
--- a/src/Button.js
+++ b/src/Button.js
@@ -2,7 +2,8 @@ import React from 'react';
 import classNames from 'classnames';
 import BootstrapMixin from './BootstrapMixin';
 import elementType from 'react-prop-types/lib/elementType';
-import ButtonInput from './ButtonInput';
+
+const types = ['button', 'reset', 'submit'];
 
 const Button = React.createClass({
   mixins: [BootstrapMixin],
@@ -24,7 +25,7 @@ const Button = React.createClass({
      * @type {("button"|"reset"|"submit")}
      * @defaultValue 'button'
      */
-    type: React.PropTypes.oneOf(ButtonInput.types)
+    type: React.PropTypes.oneOf(types)
   },
 
   getDefaultProps() {
@@ -100,5 +101,7 @@ const Button = React.createClass({
     );
   }
 });
+
+Button.types = types;
 
 export default Button;


src/ButtonInput.js
```diff
@@ -17,14 +17,14 @@ class ButtonInput extends InputBase {
   }
 }
 
-ButtonInput.types = ['button', 'reset', 'submit'];
+ButtonInput.types = Button.types;
```
taion: Possible to just delete this entirely? Wasn't this just for `Button` above?

taion:
:+1: nice catch, 1 quick comment but otherwise LGTM

AlexKVal:
Nice catch. :sparkle: 

Could you also prefix the commit header with `[fixed]` according to our 
[CONTRIBUTING.md#commit-subjects-for-public-api-changes](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes),
then it will be reflected in [CHANGELOG.md](https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md)

src/ButtonInput.js
michaeljones: Thanks for the review. Happy to remove it. It was an effort on my part to preserve the 'interface' as such. If third parties are relying on it being there. I don't know whether you treat that as part of your interface or not.

michaeljones:
Updated. Sorry for failing to do that.

src/ButtonInput.js
taion: Fair enough. I don't know that anybody should be depending on this, but I guess you're being more careful than me (:

Could you modify the `propTypes` validation below to use this `ButtonInput.types` then, just in case someone replaces `ButtonInput.types`?

taion:
Okay one last request... can you squash the 2 commits? Then we'll be good to go. Thank you for fixing this!

michaeljones:
Done! :) Thanks for the super quick feedback and responses.

taion:
I'll merge when checks are green.

michaeljones:
Are you planning a release soon? I'm trying to play it cool but I am relying on this fix and `react-bootstrap` doesn't really seem to install from github-style npm paths so I'm going to have to manage it locally if you don't have a release planned.

Cheers!

taion:
Huh? I'll cut a release but if you're installing from GitHub, why do you are about npm?

michaeljones:
Thanks for the response. I've tried installing from github but it doesn't work as far as I can tell as none of the built files are checked into the repository.

taion:
Never mind - I misread your comment.

If you install from GH, you need to manually install dependencies, then do `npm run build`. Otherwise I'll have a new patch up soon.

michaeljones:
I'll take the lazy option and wait for the patch. Much appreciated!

taion:
v0.27.3 is out

michaeljones:
Thank you!

PR #1433, opened 2015-10-21
[fixed] OverlayTrigger show/hide timeout management

### LINKED ISSUE #795: OverlayTrigger with delays can unintentionally leave overlays visible due to focus events.
If you use a `Tooltip` with an `OverlayTrigger` and specify show and hide delays, there is a way to keep the tool-tip open indefinitely, even after the cursor leaves the OverlayTrigger: Mouse over the trigger, wait for the tool-tip to appear, then rapidly click and drag mouse the cursor off the component. I was able to reproduce this behavior in Chrome, Firefox, and Internet Explorer.

This behavior can be demonstrated using the sample code from the documentation:

```js
const LinkWithTooltip = React.createClass({
  render() {
    let tooltip = <Tooltip>{this.props.tooltip}</Tooltip>;

    return (
      <OverlayTrigger placement='top' overlay={tooltip} delayShow={300} delayHide={150}>
        <a href={this.props.href}>{this.props.children}</a>
      </OverlayTrigger>
    );
  }
});
```
Place this link anywhere, click and drag off the link quickly, and the tooltip will remain open.

I dove into the OverlayTrigger code a bit, and it seems the mouse click triggers a call to `handleDelayedShow()` (since that gives the wrapped component focus), which sets the `_hoverDelay` property. When `handleDelayedHide()` is invoked right after, it will just call `clearTimeout(this._hoverDelay)` and return, thus keeping the tooltip visible.

I think this behavior can be disabled by only specifying `hover` as a trigger. Perhaps `focus` should not be a default trigger?

### LINKED ISSUE #1432: Delayed tooltip stays visible forever on disabled element
Take for example:

```js
const LinkWithTooltip = React.createClass({
  getInitialState() {
    return {
      disabled: false
    };
  },
  
  render() {
    let tooltip = <Tooltip>tooltip</Tooltip>;

    return (
      <OverlayTrigger overlay={tooltip} delayShow={300} delayHide={150}>
        <Button
          disabled={this.state.disabled}
          onClick={() => {this.setState(({disabled}) => ({disabled: !disabled}))}}
        >
          Click me
        </Button>
      </OverlayTrigger>
    );
  }
});

ReactDOM.render(<LinkWithTooltip />, mountNode);
```

If you mouse over the button, then click it, the tooltip never goes away.

taion:
Fixes #795 
Fixes #1432

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -149,7 +149,8 @@ const OverlayTrigger = React.createClass({
   componentWillUnmount() {
     ReactDOM.unmountComponentAtNode(this._mountNode);
     this._mountNode = null;
-    clearTimeout(this._hoverDelay);
+    clearTimeout(this._hoverShowDelay);
+    clearTimeout(this._hoverHideDelay);
   },
 
   componentDidUpdate() {
@@ -226,9 +227,13 @@ const OverlayTrigger = React.createClass({
   },
 
   handleDelayedShow() {
-    if (this._hoverDelay != null) {
-      clearTimeout(this._hoverDelay);
-      this._hoverDelay = null;
+    if (this._hoverHideDelay != null) {
+      clearTimeout(this._hoverHideDelay);
+      this._hoverHideDelay = null;
+      return;
+    }
+
+    if (this.state.isOverlayShown || this._hoverShowDelay != null) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


taion:
Oops, bug - fixed.

AlexKVal:
:+1:  :sparkle:

PR #1434, opened 2015-10-22
Add missing type="button" in modal header

neoziro:
If we have a form in the modal, it triggers the onClick..


taion:
I guess rather having the modal in the form? LGTM.

Need another reviewer from @react-bootstrap/collaborators.

PR #1436, opened 2015-10-22
when overlay animation is false the transition prop passed to base ov…

### LINKED ISSUE #1435: disabling animation in OverlayTrigger displays a React PropType warning
When adding an animation: false prop to the overlay to disable animation if you are using React in development mode it displays the following warning:

`Warning: Failed propType: Invalid prop 'transition' of value 'false' supplied to 'Overlay'. Expected an Element `type` such as a tag name or return value of React.createClass(...) Check the render method of `Overlay`.`

We've the offending code to be in how it assigns the transition prop in the Overlay component,  the overlay expects a non-boolean so when disabling the animation it should use undefined or null instead of passing the original animation boolean through.

This is somewhat annoying since its not a true problem, just an issue with the way that the overlay trigger is passing its props through to the base overlay component. We'll add a a pull request to this issue.

rdy:
when overlay animation is false the transition prop passed to base overlay should be undefined so that it does not trigger a React PropTypes warning

fixes #1435

### DIFF
--- a/src/Overlay.js
+++ b/src/Overlay.js
@@ -19,6 +19,10 @@ class Overlay extends React.Component {
       transition = Fade;
     }
 
+    if (transition === false) {
+      transition = null;
+    }
+
     if (!transition) {
       child = cloneElement(child, {
         className: classNames('in', child.props.className)


taion:
Nice catch. This is the not my preferred fix though. Modify https://github.com/react-bootstrap/react-bootstrap/blob/v0.27.2/src/Overlay.js#L18-L20 to set `transition = null` if `transition === false`.

Also, please add the `[fixed]` prefix to the commit message for our changelog generation.

taion:
@rdy Are you going to update this PR?

rdy:
yes I'll update this once i get a chance.

rdy:
@ taion I redid this PR in the way you specified but I'm not really certain why the AppVeyor build failed in the eslint step

taion:
I have no clue. It's obviously not caused by your PR.

LGTM. Need one more reviewer from @react-bootstrap/collaborators.

jquense:
LGTM

PR #1441, opened 2015-10-23
Clean up Navbar docs

taion:
Why not...

Also the Navbar brand example was almost identical to the basic one so I just killed that section entirely.


jquense:
LGTM

PR #1442, opened 2015-10-24
Update badges on README

taion:
- Add Coveralls badge because we have that information anyway
- Drop HuBoard badge because we've stopped using it
- Add Discord badge because there's active chat on there
- Switch Gitter badge for less ugly one


taion:
Gonna merge this since it's purely cosmetic. I'm not too attached to anything in particular here and would be more than happy to modify things if anybody has concerns.

AlexKVal:
:+1:

PR #1443, opened 2015-10-26
Remove no longer relevant scripts for releases

taion:


taion:
I'm merging this because it doesn't change anything - just removes some stale entries pointing to tools that no longer exist.

PR #1448, opened 2015-10-27
[fixed] Don't clone children in ResponsiveEmbed

### LINKED ISSUE #1447: Error if put iframe in Responsive embed component
My code:
```
<ResponsiveEmbed a16by9>
  <iframe
    src="some url">
  </iframe>
</ResponsiveEmbed>
```

React send error in console, but iframe rendered
```
Uncaught Error: Invariant Violation: findComponentRoot(..., .1.0.$/=11.0.0.0): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID ``.
```

It's normal behavior?

taion:
Fixes #1447

### DIFF
--- a/src/ResponsiveEmbed.js
+++ b/src/ResponsiveEmbed.js
@@ -5,7 +5,7 @@ import warning from 'warning';
 
 class ResponsiveEmbed extends React.Component {
   render() {
-    const { bsClass, className, a16by9, a4by3, ...props } = this.props;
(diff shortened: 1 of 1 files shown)


AlexKVal:
Pff :sweat: 
It is my fault. 
Thank you for the fix :cherries: :smile:

PR #1457, opened 2015-10-29
pass autoFocus to input's

doodlemoonch:
Test is pretty light as react does weird things with autoFocus on mount


taion:
This doesn't do anything though? `<Input>` already just forwards all of its props to the child.

doodlemoonch:
homer-doh

PR #1463, opened 2015-11-01
Suggest wrapping `Collapse` contents in a `div`

WChargin:
It makes a huge difference when you have inline elements.
Compare the following two animations (sorry for poor quality).
The first is with the contents wrapped in a `<div>`;
the second has the contents not wrapped in a `<div>`.
There is no relevant lag;
the stuttering happens because the `Collapse`
can't get smaller than the line height of the alert.

example with `div`
example without `div`

The example already does this, but it's far from obvious
(to ~~a newcomer~~ me, at least!)
that that's what fixes this particular issue.

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -952,6 +952,20 @@ const ComponentsPage = React.createClass({
 
                   <h3><Anchor id="utilities-collapse">Collapse</Anchor></h3>
                   <p>Add a collapse toggle animation to an element or component.</p>
+                  <div className="bs-callout bs-callout-info">
+                    <h4>Smoothing animations</h4>
+                    <p>
+                      If you're noticing choppy animations,
+                      and the component that's being collapsed
+                      has non-zero margin or padding,
+                      try wrapping the contents
+                      of your <code>&lt;Collapse&gt;</code>
+                      {" "}inside a node with no margin or padding,
+                      like the <code>&lt;div&gt;</code> in the example below.
+                      This will allow the height to be computed properly,
+                      so the animation can proceed smoothly.
+                    </p>
+                  </div>
                   <ReactPlayground codeText={Samples.Collapse} />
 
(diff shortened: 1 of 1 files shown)


taion:
For something like the above, how were you structuring it previously (i.e. in the bottom example)?

WChargin:
Here's the commit: wchargin/lc3@344da3aa497204722c9a8c47dccc37a984e5d4e4.

Basically, I had `<Collapse><Alert>{/* … */}</Alert></Collapse>`, which I changed to `<Collapse></div><Alert>{/* … */}</Alert></div></Collapse>`.

I admit that I don't fully understand *why* this works, given that the `Alert`s are also `display: block`. If this is a case of "solves the problem indirectly by triggering something else, which *actually* fixes the problem," then feel free to update the copy text to explain the root cause and why this fix works.

jquense:
I believe its the margins that throw the Collapse off (same upstream) since it messes with height calculations and what you can animate

WChargin:
Ah, yep, that makes sense @jquense. I can "fix" the problem if I remove both the margin *and* the padding (neither alone suffices), but obviously that looks hideous.

Example with margin and padding gone, and a very sad alert

jquense:
I would change the message to reflect the actual situation needed to make it work, ie. a node with no padding or margin, instead of saying div.

WChargin:
Yep—agreed!

taion:
Can you squash your commits? LGTM otherwise.

WChargin:
Sure! Not sure how that works with the pull request flow—should I squash and then force-push to my branch @taion?

taion:
@WChargin Exactly.

WChargin:
Okay, done; waiting for Travis and friends.

taion:
LGTM. @jquense, is this what you meant with your comment?

jquense:
LGTM

PR #1465, opened 2015-11-02
remove bootstrap mixin

jquense:
No deprecation since it was mostly unfeasible, but the upgrade path is straightforward.

Not quite done here, still messing with the API (and it needs documentation) but I think most of guts are done.

previous: #1257


AlexKVal:
Yay. 
- high order functions
- and decorators
- and modal's components API's consistency with others: `modalClassName` => `bsClass`
- default prop descriptions
- prop types inheritance `bsStyle: Button.propTypes.bsStyle` :+1: 

Now new props look in docs like that:
new look

Though when they are inherited, they look like that: (I think it is for another PR. not for this)
inherited


A lot of work has been done here :sparkle:

src/utils/bootstrapUtils.js
```diff
@@ -0,0 +1,152 @@
...
+
+  if (defaultSize !== undefined) {
+    let defaultProps = Component.defaultProps || (Component.defaultProps = {});
+    defaultProps.bsSize = defaultSize;
+  }
+
+  return Component;
+});
+
+export default {
```
taion: Why are some of these named exports v. keys on the default export?

jquense: it was easier for grepping stuff from the mixin...and the named exports seemed like some sort of Idea, I don't know.

taion:
:+1: This is really great - I don't envy your having had to go through and remove all uses of `BootstrapMixin`.

What do you think is still incomplete here?

jquense:
I think its complete, just not sure its clean enough, In particular the `addStyle` api thing, there is also nothing for bsSize, but that seemed less helpful anyway.

we can always iterate later

taion:
I see - you mean in the sense that there's no way to e.g. globally add a style, and no way to add sizes at all?

jquense:
yeah, it probably makes sense that style variants can't be added globally. There is no way to add sizes, or aliases for sizes ('large' -> 'lg') at all, and perhaps there should be a way to change the default bsClass on a component other than: `Button.defaultProps.bsClass = 'mybtn'`? I don't know at this point I am a bit separated from the use cases.

taion:
You wouldn't do that would you? It's be easier just to make your own `CustomButton` maybe. Guess it clogs up the VDOM a bit. I don't really have a ton of use cases for this either.

taion:
@jquense Are the docs things that @AlexKVal noted a blocker? If not, I'm okay with merging this.

AlexKVal:
I'm okay with it too o/c. :sparkle:

jquense:
ok then!

PR #1479, opened 2015-11-06
MenuItem allow PropType.object for eventKey

bryanmartin82:


taion:
Why? We don't accept objects for `eventKey` anywhere else - in general `eventKey` needs to be compared against e.g. `activeKey`, which won't work for objects.

bryanmartin82:
For more flexibility, but didn't consider activeKey comparisons. Thanks, closing PR

PR #1481, opened 2015-11-09
Added CollapsibleNav to Navbars component list

dozoisch:
Very small change but it should help this component visibility in the docs as I had totally missed the third example! (I probably would've gone further in the list if I had seen there was a component called like that!


taion:
LGTM

PR #1487, opened 2015-11-13
Cancel slide when passing in an activeIndex prop

dbellizzi:
Previous code used getActiveIndex to compare to nextProps, but getActiveIndex falls back to looking at state when props.activeIndex is null.  state always has an activeIndex.  But we want to cancel the slide timeout when we're moving from a prop of activeIndex null to a prop of activeIndex that is not null (container specifying an active index).

The symptom of this bug is that you get one more slide after you've passed in an active index.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -57,7 +57,7 @@ const Carousel = React.createClass({
   componentWillReceiveProps(nextProps) {
     let activeIndex = this.getActiveIndex();
 
-    if (nextProps.activeIndex != null && nextProps.activeIndex !== activeIndex) {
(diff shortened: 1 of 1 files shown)


taion:
I don't think this works for the case where `activeIndex` is uncontrolled.

dbellizzi:
If ``activeIndex`` is uncontrolled (I assume that means not passed in by the container in this case) ``nextProps.activeIndex != null`` will never pass, right?  If it moves from uncontrolled to controlled, I think this PR still applies.  I'll try an uncontrolled one out and see what happens.

taion:
Right, so `nextProps.activeIndex !== this.props.activeIndex` will always pass in that case, but `nextProps.activeIndex !== activeIndex` may fail.

PR #1494, opened 2015-11-14
[changed] remove extra wrapping `<nav>` element in Nav components

jquense:
The API between Nav and Navbar is a mess...

I am a bit worried that this change will break a lot of styles for folks, but that extra element seemed inane. Perhaps it was an attempt at better a11y? at least in a navbar though its unnecessary since its already in a wrapping `<nav>`... in any case I think adding a nav element should be up to the user


taion:
This is consistent with upstream anyway. Nice!

PR #1496, opened 2015-11-15
Navbar rework

jquense:
The amount of complexity that deprecations would have added was exceptional so I just changed the major chunk of the API ~~without deprecations~~ with a good idea from @taion.

- gutted the `Navbar` and split everything into sub components
- The Collapse aspect of the `<Nav>` has been removed and left up to `Navbar.Collapse`
- All the toggleKey weirdness is gone
- deprecated CollapsibleNav
- use `uncontrollable` for controlled/uncontrolled nav collapsing


src/Navbar.js
```diff
@@ -107,103 +116,55 @@ const Navbar = React.createClass({
...
+  bsClasses('navbar',
+    uncontrollable(Navbar, { navExpanded: 'onToggle' })
+  )
+);
 
-    if (child.props.bsRole === 'brand') {
-      return React.cloneElement(this.renderHeader(child), {key});
-    }
+function createSimpleWrapper(tag, suffix, displayName) {
+  let wrapper = (
```
taion: Is there a risk of breaking people due to https://github.com/facebook/react/issues/4936?

test/NavSpec.js
```diff
@@ -65,7 +65,7 @@ describe('Nav', () => {
 
   it('Should add navbar-right class', () => {
     let instance = ReactTestUtils.renderIntoDocument(
-          <Nav bsStyle="tabs" right activeKey={1}>
+          <Nav bsStyle="tabs" navbar right activeKey={1}>
```
taion: Should be `pullRight`

docs/examples/NavbarCollapsible.js
```diff
@@ -1,17 +1,28 @@
...
+        <NavItem eventKey={2} href="#">Link</NavItem>
+        <NavDropdown eventKey={3} title="Dropdown" id="collapsible-navbar-dropdown">
+          <MenuItem eventKey="1">Action</MenuItem>
+          <MenuItem eventKey="2">Another action</MenuItem>
+          <MenuItem eventKey="3">Something else here</MenuItem>
+          <MenuItem divider />
+          <MenuItem eventKey="4">Separated link</MenuItem>
+        </NavDropdown>
+      </Nav>
+      <Nav right>
```
taion: `pullRight`?

docs/examples/NavbarCollapsible.js
```diff
@@ -1,17 +1,28 @@
 const navbarInstance = (
-  <Navbar inverse toggleNavKey={0}>
-    <NavBrand><a href="#">React-Bootstrap</a></NavBrand>
-    <Nav right eventKey={0}> {/* This is the eventKey referenced */}
-      <NavItem eventKey={1} href="#">Link</NavItem>
-      <NavItem eventKey={2} href="#">Link</NavItem>
-      <NavDropdown eventKey={3} title="Dropdown" id="collapsible-navbar-dropdown">
-        <MenuItem eventKey="1">Action</MenuItem>
```
taion: LOL I can't believe we thought this was a good idea. `"1"` v. `1`.

docs/examples/NavbarForm.js
```diff
@@ -0,0 +1,21 @@
...
+    <Navbar.Header>
+      <Navbar.Brand>
+        <a href="#">Brand</a>
+      </Navbar.Brand>
+      <Navbar.Toggle />
+    </Navbar.Header>
+    <Navbar.Collapse>
+      <Navbar.Form pullLeft>
+        <div className="form-group">
+          <Input type="text" placeholder="Search"/>
```
taion: Doesn't `<Input>` already render a `div.form-group`?

docs/src/ComponentsPage.js
```diff
@@ -501,51 +501,72 @@ const ComponentsPage = React.createClass({
...
+                  <p>
+                    They also supports all the different Bootstrap classes as properties. Just camelCase
+                    the css class and remove navbar from it.
+
+                    For example <code>navbar-fixed-top</code> becomes the property <code>fixedTop</code>.
+                    The different properties are <code>fixedTop</code>, <code>fixedBottom</code>, <code>staticTop</code>
+                    , <code>inverse</code>, <code>fluid</code>.
+                  </p>
+                  <p>
+                    You can also align elements to the right by specifying the <code>right</code> prop on
```
taion: `pullRight` again, maybe? Dunno - I feel like maybe `right` actually feels slightly more natural here? Not sure.

src/Nav.js
```diff
@@ -113,24 +116,25 @@ Nav.propTypes = {
    * HTML id for the inner `ul` element
    */
   ulId: React.PropTypes.string,
-  expanded: React.PropTypes.bool,
   navbar: React.PropTypes.bool,
   eventKey: React.PropTypes.any,
   pullRight: React.PropTypes.bool,
-  right: React.PropTypes.bool
+  right: deprecated(React.PropTypes.bool, 'Use the `pullRight` prop instead')
```
taion: Do we need to expose `pullRight` as a special prop on `<Nav>` at all? Maybe just have something like `navbarRight`? Or only support `right` when used with `navbar` and fail on prop type validation otherwise? Users can always add `.pull-right` themselves if they need.

taion:
Here's a _really_ hacky way to give a deprecation warning.

Keep around the old components in e.g. `lib/deprecated`, then instead render a `DeprecatedNavbar` if `toggleNavKey`, `brand`, &c. are specified. And then sniff `context` on `<Nav>` to do the same potentially.

`Header` is a little annoying \= Like... either all it does is wrap the `Brand`, or when it contains both `Brand` and `Toggle`, the toggle isn't even logically part of the `Header`. I don't have a good answer there though. idk

src/Navbar.js
jquense: Not in these cases since none of these components exist yet...I can't think of a reasonable use case for why you'd want to ref these, In anycase I only did it b/c it was terser, not cost in changing it

docs/examples/NavbarForm.js
jquense: Not without a label prop I think...I can double check but the markup looked good when I looked at it originally

docs/src/ComponentsPage.js
jquense: the change to pullRight happened after i wrote all the docs, I do need to change it. 

in terms of natural, I agree `right` is more semantic feeling (`navbar right`), but everywhere else in the code base uses `pullRight` so I went with the more consistent of the two props

src/Nav.js
jquense: A lot of the navbar examples do explicitly add `navbar-right|left` to the sub components, of which Nav is sort of in a middle space there. I figure since other components (dropdown) have `pullRight` why not just go with that and prefix it correctly in the navbar context.

taion: Fair enough. Makes sense.

docs/src/ComponentsPage.js
taion: :+1:

docs/examples/NavbarForm.js
taion: This looks unconditional: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/InputBase.js#L206

We need to rewrite `<Input>` though... it's a bit much right now.

jquense:
:raised_hands: good idea on the deprecation path, I was able to fairly minimally provide both API's in _I think_ a non breaking way!

docs/examples/NavbarForm.js
taion: I just checked that this still looks fine without the extra `div.form-group`.

taion:
One tiny nit above on the form example. Otherwise :shipit:, this is really fantastic; thank you!

jquense:
fixed the form example, but I have one more wart I want to remove and just thought of a way to do it cleanly..

taion:
I take it `navExpanded` -> `expanded` was not the wart in question?

jquense:
not was,  just didn't get a moment to merge!

PR #1499, opened 2015-11-15
Added a local js mode for CodeMirror to properly handle jsx

jquense:
take from that PR

also solarized is a terrible theme


taion:
:shipit:

PR #1501, opened 2015-11-15
Clean up use of Affix

### LINKED ISSUE #51: Implement  ScrollSpy
Not sure how this should be implemented so here's some thinking out loud.

Could be a component that takes one or more `Nav`s as children and a `target` CSS selector prop:

```
<ScrollSpy target="body">
    <Nav>
        <NavItem href="#1" key={1}>Item 1</NavItem>
        <NavItem href="#2" key={2}>Item 2</NavItem>
    </Nav>
</ScrollSpy>
```

This would recurse through the nav props finding hrefs, work out which is the closest to the current scroll position then pass `activeHref` prop to `cloneWithProps` nav.

`ScrollSpy` could also be controlled with a `onScrollToItem(href)` callback prop.

Actually this could be built straight into `Nav`: `<Nav scrollTarget="body">...</Nav>`.

Any thoughts?

### LINKED ISSUE #200: Implement affix
Hey guys,

First, thanks for all your hard work. From a random dev, I really appreciate this project!

Second, I'm just curious why this project doesn't implement the bootstraps affix component. Is there any particular reason why/why not? Or maybe you don't have time?

Thanks for any and all info.

taion:
- Removed vestigial Affix component
- Explain what to do in docs instead
- Switch docs to use react-overlays AutoAffix
- Move SubNav into docs, where it belongs
- Explain why we're missing scrollspy as well

Closes #200
Closes #51


taion:
Fixed now, I think. Went through a few too many iterations on this between adding and not adding `<Affix>` to R-B and got confused.

jquense:
LGTM!

PR #1506, opened 2015-11-16
[changed] dropdown props are passed through to the dropdown Button

### LINKED ISSUE #1497: DropdownButton applies classname to both button group and button
I am using the Dropdown component, however I noticed an odd behaviour at the HTML output level. 

I have a set of 4 buttons, with second one being a Dropdown component. I'm also using ionicons via css style sheet to place an icon to the left of the button name. Normal buttons appear fine, however the Dropdown inserts 2 icons. (see image below)

image

Inspecting the HTML output via the inspector, it appears the initial Dropdown component className I've set is applied twice, one for the encapsulating DIV and one for the BUTTON that follows (encircled in red). Clearly that's not my desired effect I was after. When removing the style inline via inspector from the DIV element while leaving the style for the BUTTON, the final effect is exactly how I'd want it. (see image below) 

image

Is this a bug? If so, can it be squashed please?

jquense:
fixes #1497

### DIFF
--- a/src/DropdownButton.js
+++ b/src/DropdownButton.js
@@ -1,22 +1,31 @@
 import React from 'react';
 import Dropdown from './Dropdown';
 import omit from 'lodash-compat/object/omit';
+import pick from 'lodash-compat/object/pick';
 import Button from './Button';
 
 class DropdownButton extends React.Component {
 
   render() {
-    let { title, ...props } = this.props;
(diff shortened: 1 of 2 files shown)


taion:
:shipit:

jquense:
lol found a bug

jquense:
how did any of this ever work...wat was I even doing

PR #1507, opened 2015-11-16
Require lodash-compat instead of lodash from Modal.js

ide:
lodash-compat is what's listed under package.json. This matches how `pick` is imported from other source files in react-bootstrap too.


taion:
Going to merge this as a straightforward bugfix and cut v0.28.1.

Issue #1418, opened 2015-10-18
Improve Tabs API


taion:
I'm consolidating a number of other issues here:
- [ ] Support dropdowns in tabs (#100)
- [ ] Do not mount hidden tabs (#129)
- [x] Tab panel disappears when switching between tabs quickly (#557)
- [x] Setting `href` on tab headers (#1332)
- [x] Tabs with no initially selected tab and animation don't work (#1352)
- [x] Allow setting `className` on `<nav>` element (#1405)
- [x] Allow setting `className` on individual nav items (#1456)

taion:
I think the way to go about this is to add a `Panes`/`Pane` pair of components, something that looks like:

```jsx
<Panes activeEventKey={2}>
  <Pane eventKey={1}>{content1}</Pane>
  <Pane eventKey={2}>{content2}</Pane>
  <Pane eventKey={3}>{content3}</Pane>
</Panes>
```

For the simple case, we can have `Tabs`/`Tab` (wondering if the old `TabbedArea`/`TabPane` names were better though...) wrap these components - probably even simplify it further by dropping the horizontal tabs support.

For the more complex cases with custom tabs, dropdowns, &c., just let users wire together a `Nav` and a `Panes` together themselves.

@jquense @mtscout6 This is a bit thematically similar to what you did with `Modal` and `Dropdown`. Do you think it'd be reasonable here?

jquense:
we already have Tab, which I assume maps to Pane above...I assume Panes is sort of some incrementally helpful thing between wiring a bunch of Tab's to a Nav, and using the `Tabs` component? I like that approach i think, the current api is pretty hands off in an unhelpful way, this seems like a good middle ground, I with naming wasn't so hard...renaming anything in tab API would suck since we already did that fairly recently

taion:
Yup that's the idea - so in the more complex scenarios you render a component with a `Nav` and a `Panes`, something like:

```jsx
onSelect = (e, eventKey) => {
  this.setState({activeEventKey: eventKey});
};

render() {
  const {activeEventKey} = this.state;

  return (
    <div>
      <Nav activeEventKey={activeEventKey} onSelect={this.onSelect}>
        {this.renderNavItems()}
      </Nav>
      <Panes activeEventKey={activeEventKey}>
        {this.renderPanes()}
      </Panes>
    </div>
  );
}
```

[...]

PR #1525, opened 2015-11-19
Add lower level Tab APIs

jquense:
not ready, but structurally sound. damn animations are brittle I hate them and I am tired of working on this right now.

also the way I hacked in a11y support with context, is terrifying, looking for alternatives


taion:
This looks really cool BTW. Will take a closer look over the weekend.

taion:
ref #1418

src/TabContainer.js
```diff
@@ -0,0 +1,87 @@
...
+    let ids = this._ids;
+    let old = ids[type].get(key);
+
+    if (old !== id) {
+      ids[type].set(key, id);
+      this._needsUpdate = true;
+    }
+
+    clearTimeout(this._timer);
+    this._timer = setTimeout(()=> {
```
taion: Why not put `ids` on `state`?

taion:
- It's a little odd to have the tab-specific stuff in `<Nav>`. If nothing else, probably we don't want the weight in bundles for users who aren't using `<Tabs>`? Maybe we need `<TabsNav>` or something... ick.
- That a11y support... we could just not support automatically setting `aria-labelledby` if the user is using his own `id`s on the `<NavItem>`s.

jquense:
yeah I felt that way about the tab stuff in Nav initially although as I did it,  it actually seemed like it is the right place. Nav is, in large part, there specifically for tabs (or similarly pills), it started to not feel weird the focus stuff was there. 

I am a bit ambivalent though to it, adding yet another Component also felt confusing. this seemed like the better tradeoff at the time. 

the a11y stuff is really ugly tho I am generally ok with that if it's not leaky, but I'm not sure it isn't :p

src/TabContainer.js
jquense: it was an attempt to not over render since it would fire for each tab pane. no other reason tho

taion: Well, `setState` would coalesce all of those anyway, no?

taion:
Hmm, on the `<Nav>`... I don't know... the navbar stuff is pretty lightweight, so it seems like it fits in nicely. The tab handling seems to have meaningful additional complexity.

The a11y stuff is a bit leaky, though, in that it won't work for server-side rendering, since the initial render won't pick up the custom `id`s. Do you use that feature? If not, I feel like the complexity and the SSR edge-casiness suggests that we might want to drop that bit.

jquense:
closing to reopen against the rc branch

Issue #1526, opened 2015-11-19
Cannot pass `style` and `className` to `react-overlays\Modal`


jesenko:
While `className` and `style` can be set for `container` and `backdrop`, they cannot be set for the `<div>` that contains them.

Setting these props seems to be enabled in `react-overlays\Modal`, however they are not passed on to the `BaseModal`; `className` and `style` are not explicitly listed in `react-overlays\Modal` `propTypes`, nor explicitly passed on. Not sure though which one of these two options would be more appropriate way to fix this...

taion:
The portal root isn't a React component - it's just a mount node and can't be styled that way anyway.

jesenko:
@taion Are you sure about this? Actually, the [bootstrap-overlay example](http://react-bootstrap.github.io/react-overlays/examples/#modals) uses this to set `z-index` of both modal and backdrop... I think that currently, rendered output for modal can be summarized as 
```
<div><!-- portal div, unstylable -->
  <div><!-- base modal div, stylable in react-overlay/Modal, currently unstylable in react-bootstrap -->
    <div><!-- modal backdrop, stylable in react-overlay and react-bootstrap --></div>
    <div><!-- modal container, stylable in react-overlay and react-bootstrap --></div>
  </div>
</div>
```
I have opened related PR that passes `wrapperClassName` and `wrapperStyle` to BaseModal (`className` is already passed to one of Modal children, so it cannot be used)

taion:
That's fine, but this is already handled by Bootstrap's CSS, no?

In general if you want to deeply customize your modals, we'd prefer you to use the react-overlays components directly, rather than massively expanding the API surface area of React-Bootstrap to offer that customizability.

jesenko:
As I needed this for overriding `z-index` for stacked modals, using [react-bootstrap-modal](https://github.com/jquense/react-bootstrap-modal) is probably a better way to go...

PR #1527, opened 2015-11-20
Pass wrapperStyle and wrapperClassName to react-overlay/Modal

jesenko:
Related to #1526.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -146,6 +146,8 @@ const Modal = React.createClass({
           this._wrapper = (ref && ref.refs.modal);
           this._backdrop = (ref && ref.refs.backdrop);
         }}
+        style={props.wrapperStyle}
+        className={props.wrapperClassName}
         onEntering={this._onShow}
         onExited={this._onHide}
         backdropClassName={classNames(tbsUtils.prefix(props, 'backdrop'), inClass)}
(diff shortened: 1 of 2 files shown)


taion:
This is failing on Lint.

I'm inclined against merging this - seems like at this point you're better off just using react-overlays directly. @jquense?

jesenko:
Fixed Lint error. I considered just using react-overlay directly, however I actually use all styling / functionality from bootstrap _except_ for setting `z-index` - bootstrap version does not handle stacked modals. Setting `z-index` on wrapper solves the issue.

Considering that `backdropClassName`/`backdropStyle` and `containerClassName`/`containerStyle` are already passed to underlying react-overlay Modal, would it not be more consistent to do the same for the wrapper?

jquense:
yeah...I'd like to hear the use case where you need 99% of the bootstrap model but also this hook. My thought is is the same, that if you need to style or add classes to the outer wrapping element that you might want to just use the raw Modal.

If you are looking for stackable modals: i'd use: https://github.com/jquense/react-bootstrap-modal

its a more complete solution to that.

jesenko:
@jquense Great, thanks for the link, I will try switching to `react-bootstrap-modal`, seems like an ideal match. I will close this PR/and related issue...

PR #1545, opened 2015-12-04
a11y: fixing the tab-index for the non-screenreader version of the Alert dismiss button

carlyle:
The version of the Alert close button that is supposed to be hidden from screen readers has `aria-hidden="true"`, as expected, but did not have `tab-index="-1"` so it may not be skipped when tabbing through the interface

### DIFF
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -23,7 +23,8 @@ let Alert = React.createClass({
         type="button"
         className="close"
         onClick={this.props.onDismiss}
-        aria-hidden="true">
+        aria-hidden="true"
+        tabIndex="-1">
         <span>&times;</span>
       </button>
     );


taion:
The correct prop name is in fact `aria-hidden`. Is it generally correct to disable tabbing to the close button? Seems like it might be useful w/kb nav.

carlyle:
@taion good catch on the property naming thing with `aria-hidden`, I'll push that change up in a sec.

Disabling the tabbing for elements with `aria-hidden="true"` is recommended by react-a11y. The full text of the warning is: "You have `aria-hidden="true"` applied to an interactive element but have not removed it from the tab flow. This could result in a hidden tab stop for users of screen readers."

There is already a screen reader-facing alternative for the close button being rendered that would be accessible via keyboard navigation

taion:
I'm okay with that, then. @jquense?

Also, please squash your commits.

jquense:
LGTM aside from the squashing

carlyle:
Great! Thanks for reviewing the change, @taion and @jquense. The commits have been squashed together and look good to merge on my end

taion:
Thanks for the PR!

PR #1548, opened 2015-12-07
Fix slow OverlayTrigger + enforce react-overlays version

ninja-:


jquense:
I'm not really comfortable merging this without some info about why this would make a difference. I haven't had any issues with calling clearTimeout ever before, nor heard anything about performance issues with clearTimeout before.

If possible could you provide some rational or evidence that this is helpful beyond "it made a difference in my app"?

ninja-:
@jquense clearTimeout is seriously slow for me on chrome, I don't know if it should...if it is for example 1ms per call it sums up to pretty big overhead.

jquense:
that's fine in all but my hunch is that there is something else wrong with your code that isn't affecting others, like you are starving the event loop, or are getting behind on your timers.

My problem with code like this is it is very unclear what it is doing and why, presenting a maintenance burden, Initial searches on my part have not turned up any issues with clearTimeout taking a long time, which is why I am reticent to accept that thats' the root cause of the issue.

ninja-:
@jquense this code avoids clearInterval(0) for tasks that were not scheduled. Is this more clear now?

taion:
I'm going to replace this with a fixed version.

PR #1557, opened 2015-12-10
Overlays should stay shown when we hover over them

dustingetz:
HI guys, I don't expect you to accept this PR, because I know regular bootstrap forbids this behavior. I am just leaving this here as documentation for the next person.

Do you have any suggestions for how I might implement this that would not require modifying react-bootstrap?

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -177,9 +177,13 @@ const OverlayTrigger = React.createClass({
       onEntered: this.props.onEntered
     };
 
+    let hoverOrFocus = isOneOf('focus', this.props.trigger) || isOneOf('hover', this.props.trigger);
+
     let overlay = cloneElement(this.props.overlay, {
       placement: overlayProps.placement,
-      container: overlayProps.container
+      container: overlayProps.container,
+      onMouseOver: hoverOrFocus ? this.handleDelayedShow : undefined,
+      onMouseLeave: hoverOrFocus ? this.handleDelayedHide : undefined
     });
 
     return (


taion:
You'd use `<Overlay>` directly instead of `<OverlayTrigger>`.

dustingetz:
Is there a way to avoid copy/pasting all 294 lines of OverlayTrigger to add three lines? I tried to monkey patch but I couln't get it to work, maybe you guys have seen some tricks I can use?

taion:
Just don't do it – look at the components page example that's tailored to the specific use case.

dustingetz:
Forgive me, but which quote are you referring to? "It's inadvisable to use "hover" or "focus" triggers for popovers, because they have poor accessibility" ? or " Creating custom overlays is as easy as wrapping some markup in an Overlay component"?

I understand how to use Overlay, but I also want all the behaviors in the OverlayTrigger abstraction, with my small patch to the hover behavior.

taion:
This section - http://react-bootstrap.github.io/components.html#custom-overlays. If you just use what's specific to your "hover" use case, it's quite a lot simpler - just track the enter/leave events you want. `<OverlayTrigger>` is intended to mimic the TWBS API, which is why we keep the meaning of "hover" (e.g. what you have is mostly not appropriate for something like a tooltip).

mezod:
hey @dustingetz did you manage to pull this behaviour in a nice way? i'd be interested to see how to do it without copying the complete OverlayTrigger into my project, thanks!

dustingetz:
Yes, I am happily running with the commit attached to this pull request

mezod:
you created a new OverlayTrigger into your project and changed the paths or?

mezod:
got it working that way, but feels clunky :p

PR #1560, opened 2015-12-12
responsively hiding columns

csabapalfi:
Although the `hidden-*` classes are not just for columns they are really frequently used on them.

### DIFF
--- a/src/Col.js
+++ b/src/Col.js
@@ -38,6 +38,38 @@ const Col = React.createClass({
      */
     lg: React.PropTypes.number,
     /**
+     * Hide column
+     *
+     * on Extra small devices Phones
+     *
+     * adds class `hidden-xs`
+     */
+    xsHidden: React.PropTypes.bool,
+    /**
+     * Hide column
+     *
+     * on Small devices Tablets
+     *
+     * adds class `hidden-sm`
+     */
+    smHidden: React.PropTypes.bool,
+    /**
+     * Hide column
+     *
+     * on Medium devices Desktops
+     *
+     * adds class `hidden-md`
+     */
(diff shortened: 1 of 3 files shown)


taion:
Can you add a quick test for this?

AlexKVal:
It seems it would be nice to have commits "squashed".

Also the header of the resulted commit should begin with `[added] ...` tag. [Here is more about it.](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md#commit-subjects-for-public-api-changes)
That way this change will be reflected in the [Changelog](https://github.com/react-bootstrap/react-bootstrap/blob/master/CHANGELOG.md)

:cherries:

csabapalfi:
Done. Thanks for reviewing!

Should've started by reading CONTRIBUTING.md. :blush:

AlexKVal:
LGTM :sparkle: 
Though I believe it woud be nice to have an additional review by @taion or @jquense to get merged.

src/Col.js
```diff
@@ -38,6 +38,38 @@ const Col = React.createClass({
      */
     lg: React.PropTypes.number,
     /**
+     * Hide column
+     *
+     * on Extra small devices Phones
+     *
+     * adds class `col-xs-hidden`
```
jquense: This isn't actually correct, right? the classes are just [size]-hidden

csabapalfi: Ouch, fixing comment now.

csabapalfi:
Now I also fixed propTypes comments. Thanks @jquense!

taion:
:shipit:

AlexKVal:
@csabapalfi Thank you for the contribution :wink: :cherries:

PR #1578, opened 2015-12-24
Check for hover timeout before cancelling it

taion:


jquense:
why?

taion:
We do it https://github.com/taion/react-bootstrap/blob/OverlayTrigger-clearTimeout/src/OverlayTrigger.js#L237 and https://github.com/taion/react-bootstrap/blob/OverlayTrigger-clearTimeout/src/OverlayTrigger.js#L262.

I dunno - think it's clearer not to do this check at all?

jquense:
Ahh, I see consistency. I'm not sure I've never checked for a timeout before trying to clear it...I defer to you on that one.

taion:
Actually forget it, that check is because we need to do something different in the case that the timeout is set.

PR #1579, opened 2015-12-24
Ensure Modal cleanup

### LINKED ISSUE #1538: Modal doesn't guarantee listener clean up
The window resize handler for a modal is controlled by [Modal._onHide](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js#L175),  but there is no guarantee that this function will be called by the child base modal component.

This leads to a bug where after a modal is closed, a window resize event can throw an error due to the component no longer existing.

I'm unsure of the best way to ensure cleanup other than adding it to componentWillUnmount (I am currently doing this as a workaround).

### LINKED ISSUE #1563: A regression for #1078 "Modal (and probably others) should use componentWillUnmount"
There is a regression for https://github.com/react-bootstrap/react-bootstrap/issues/1078

In the https://github.com/react-bootstrap/react-bootstrap/commit/214bc772e414050779bd37cbfb326c699f6f8be9 commit there was both the fix and the corresponding test removed which brought the initial problem back.

Currently there is no (?) code that invokes `_onHide` on `Modal` unmount event which causes `handleWindowResize` to still be invoked on a window resize, whereas the modal element does not exist in the DOM which causes an exception on `let scrollHt = node.scrollHeight;` line.

@jquense @taion could you guys please comment on that?

taion:
Fixes #1538
Fixes #1563

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -106,6 +106,10 @@ const Modal = React.createClass({
     };
   },
 
+  componentWillUnmount() {
+    events.off(window, 'resize', this.handleWindowResize);
+  },
+
   render() {
     let {
         className
--- a/test/ModalSpec.js
+++ b/test/ModalSpec.js
@@ -1,3 +1,4 @@
+import events from 'dom-helpers/events';
 import React from 'react';
 import ReactTestUtils from 'react/lib/ReactTestUtils';
 import ReactDOM from 'react-dom';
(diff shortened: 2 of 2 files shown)


jquense:
LGTM

frankleng:
:+1: please cut a new release.

PR #1584, opened 2016-01-04
inverse order of class names

Ma27:
with this change the custom class name is the last one, so it is able to override the styles from the bootstrap classes


taion:
That's not how CSS works.

PR #1607, opened 2016-01-17
[added] lower level tab components

jquense:
also deprecates horizontal tab positions in Tabs


docs/examples/.eslintrc
```diff
@@ -58,6 +58,9 @@
     "SplitButton",
     "Tab",
     "Table",
+    "TabPane",
```
taion: lol alphabet

docs/examples/LeftTabs.js
```diff
@@ -1,9 +1,24 @@
 const tabsInstance = (
-  <Tabs defaultActiveKey={2} position="left" tabWidth={3}>
-    <Tab eventKey={1} title="Tab 1">Tab 1 content</Tab>
-    <Tab eventKey={2} title="Tab 2">Tab 2 content</Tab>
-    <Tab eventKey={3} title="Tab 3" disabled>Tab 3 content</Tab>
-  </Tabs>
+  <TabContainer id="left-tabs-example" defaultActiveKey="first">
```
taion: Can this example use `<Row componentClass>` and `<Col componentClass>`?

docs/src/sections/TabsSection.js
```diff
@@ -26,8 +26,17 @@ export default function TabsSection() {
       <p>Set the <code>animation</code> prop to <code>false</code></p>
       <ReactPlayground codeText={Samples.TabsNoAnimation} exampleClassName="bs-example-tabs" />
 
-      <h3><Anchor id="left-tabs">Left tabs</Anchor></h3>
-      <p>Set <code>position</code> to <code>"left"</code>. Optionally, <code>tabWidth</code> can be passed the number of columns for the tabs.</p>
+      <h3><Anchor id="left-tabs">Custom Tab layout</Anchor></h3>
```
taion: Probably worth changing this hash :stuck_out_tongue:

src/TabContainer.js
```diff
@@ -0,0 +1,89 @@
...
+     * is specified.
+     */
+    id(props, ...args) {
+      let error = null;
+
+      if (!props.generateChildId) {
+        error = idPropType(props, ...args);
+
+        if (!error && !props.id) {
+          error = new Error('You must provide an either an `id` or a `generateChildId` prop to TabContainer');
```
taion: Is this a hard requirement now, or only for a11y? If the latter, might be worth noting.

docs/examples/LeftTabs.js
jquense: I can, I didn't know if maybe that was to much noise over the API. I guess the using the plain classes is also a bit confusing

src/TabContainer.js
jquense: not a hard requirement, good to mention you are right.

docs/examples/LeftTabs.js
taion: Well, the clearfix is a bit confusing – the rest is just consistency.

src/Nav.js
```diff
@@ -1,16 +1,38 @@
...
 
 import ValidComponentChildren from './utils/ValidComponentChildren';
-import createChainedFunction from './utils/createChainedFunction';
+import chain from './utils/createChainedFunction';
 
 import Collapse from './Collapse';
 
 class Nav extends React.Component {
 
+  componentDidUpdate() {
```
taion: I feel like a lot of this logic (and below) might be extractable into a separate `TabNav` component. Maybe expose something like `navItemProps` or `itemProps` on `Nav`, and pass down the tab props from above?

src/Nav.js
```diff
@@ -77,19 +143,80 @@ class Nav extends React.Component {
...
+      'aria-controls': controls,
+      tabIndex: isActive ? tabIndex : -1,
+      onKeyDown: chain(
+        this.handleTabKeyDown.bind(this, onSelect || (()=>{})),
+        onKeyDown
+      )
+    };
+  }
+
+  handleTabKeyDown(onSelect, event) {
```
jquense: So this is a problem with moving the context stuff down to a TabNavItem, the focus behavior needs to be able to trigger `onSelect`, either this here _still_ needs to listen to tab container context, or some sort of ugly calling down to the child method.

taion: Oops. I guess this is the best way to go for now, then.

jquense:
ready to go I think

PR #1608, opened 2016-01-17
[removed] Previously deprecated API's

jquense:
- old style Navbar
- collapsible Nav
- NavBrand alias
- duration prop on Transitions


taion:
Ch-ch-ch-ch-changes

PR #1614, opened 2016-01-19
Use actual ellipsis in pagination

taion:


taion:
Fixed the tests.

AlexKVal:
It really looks better :+1: 

I am just a bit concerned wouldn't it break existing designs / layouts ?
What do you think about shipping this as a `[changed]` feature in the next minor version ?
I am not an expert at all in this area. Just a thought.

In general it LGTM :sparkle:

taion:
I'm not planning on releasing a v0.28.3 anyway – v0.29.0 or bust.

Personally I think this is more like a "fix"; there's no logic here, it's purely fixing display.

PR #1620, opened 2016-01-20
Add Modal propTypes for transition callbacks so it will be documented.

Data-Meister:
Currently these props are not documented. This is causing confusion. See this SO thread for reference: http://stackoverflow.com/questions/31836812/trigger-modal-shown-for-react-bootstrap/34901027#34901027


jquense:
LGTM until we have a better story for spread propTypes from other modules

Data-Meister:
react-overlays Modal ( BaseModal ) doesn't have these in propTypes either. Maybe it should. 

I'll submit a PR for that too.

PR #1634, opened 2016-01-28
[added] noButton in Alert

jcazor:
Allow creating auto closeable alerts with no buttons.

### DIFF
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -8,12 +8,14 @@ let Alert = React.createClass({
   propTypes: {
     onDismiss: React.PropTypes.func,
     dismissAfter: React.PropTypes.number,
-    closeLabel: React.PropTypes.string
+    closeLabel: React.PropTypes.string,
+    noButton: React.PropTypes.bool
   },
 
   getDefaultProps() {
     return {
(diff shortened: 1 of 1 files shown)


jquense:
LGTM, if there isn't some sort of a11y concern I am not thinking of...

taion:
@jquense Do you like having the timer-based dismiss thing as part of React-Bootstrap? I'm half-tempted to just delete that part.

jquense:
o yeah that's not upstream is it? If not then yeah I don't think we should support it, since its trivial to implement, and doesn't cause issues like needing to reset timers when the alter changes or something.

taion:
TWBS3 has no timer support, just dismiss support: http://getbootstrap.com/javascript/#alerts

My personal inclination is to drop the timer support; it's not awful to have but I don't see a great reason to have it.

taion:
https://github.com/react-bootstrap/react-bootstrap/pull/1636 would deprecate `<Alert dismissAfter>` entirely.

taion:
No longer relevant now that https://github.com/react-bootstrap/react-bootstrap/pull/1636 is merged, since we're killing `dismissAfter` entirely.

PR #1639, opened 2016-01-29
Added minWidth to progressbars as requested in #1574

WrathZA:
Added test and updated docs
- New property of minWidth on ProgressBar (defaulted to Zero)

original issue: https://github.com/react-bootstrap/react-bootstrap/issues/1574

### DIFF
--- a/docs/examples/ProgressBarMinimumWidth.js
+++ b/docs/examples/ProgressBarMinimumWidth.js
@@ -1,0 +1,8 @@
+const progressInstance = (
+  <div>
+    <ProgressBar now={0} label="%(percent)s%" minWidth={15} />
+    <ProgressBar now={2} label="%(percent)s%" minWidth={15} />
+  </div>
+);
+
+ReactDOM.render(progressInstance, mountNode);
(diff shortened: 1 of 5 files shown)


taion:
Wow why is this spreading all `props` to both the inner and outer divs? Should this not at least merge incoming `styles` on the inner `div`? Maybe not...

taion:
I think this is actually the wrong way to go – we want to clean up how we set styling on the internal and external divs, and we don't want to have this new prop just for this thing going forward.

sorahn:
@taion if you don't want another prop to support this feature (as discussed in #1574) what would you prefer?

jquense:
@sorahn instead of individual props for styling, the component should just hande the `style` prop correctly so you can arbitrarily style how you want

WrathZA:
What about a style prop instead of the minWidth prop and then merging the width: percentage in with Object.assign() in the render method?

Should I do that and submit another pull request?

sorahn:
@jquense I agree that it should handle the style prop directly, but I think the request we're looking at here is "fill the bar to at least the width of the label", and I think **that** should be a prop.

sorahn:
You could also take the approach that material-ui does, which is add the params like `containerStyle` and `barStyle`.  And those objects get merged down into the styles of the divs down lower.  That aligns a little better with what @WrathZA suggested.

WrathZA:
In looking at this again.... @sorahn makes a good point. How do you want to handle the nested divs? individual params like he said?

Issue #1490, opened 2015-11-13
possible to override Dropdown toggle onClick behavior


perrydt:
Is there a way to override the onClick behavior in Dropdown if you are using fully custom toggle and menu?

The use case here is that I am using an input as the bsRole='toggle' and a custom Dropdown.Menu as a 'search term completion' similar to what search engines do.  I want to keep all the behavior in the Dropdown component, keyboard handling, etc, but do not want the onClick handler to ever close the dropdown. 

Unfortunately Dropdown installs its own click handler after the Input (toggle)'s click handler, and event.preventDefault/stopPropagation don't work.    I already have a custom open prop and onToggle, but onToggle can't intervene because it contains no information about which event triggered the toggle, and we want to keep all non-click behavior (ESC key, onBlur, etc).    Other than copying all the behavior into a custom component I don't see a way to do this unless I'm missing something?

taion:
IMO the right fix would be for `Dropdown.handleClick` and others to check for the event having its default prevented. Care to PR this?

jquense:
would it make more sense to just wrap the input and not do anything with the injected onClick?

taion:
Ah, that would be easier :p

perrydt:
Aha, I tried various ways to intercept the onClick handler, but didn't think of using a wrapper for the Input itself -- it works perfectly.  Thanks for the suggestion @jquense! This is what I ended up using:

```javaScript
class InputWrapper extends Component {
  constructor(props) {
    super(props)
    this.handleClick = this.handleClick.bind(this)
    this.wrappedClick = this.props.onClick
  }

  render() {
    let otherProps = _.omit(this.props, ['onClick', 'bsRole'])
    return (
      <Input onClick={this.handleClick} {...otherProps} />
    )
  }

  handleClick(evt) {
    // Prevent the root click handler from closing the dropdown
    evt.nativeEvent.stopImmediatePropagation()
    // Note have to handle anything that needs to happen on "click when closed" in here,
    // not in the wrapped handler.
    if (!this.props.open) {
      this.wrappedClick(evt)
    }
  }
}
```
The issue with stopping propagation is that Drodown hooks into the onClick by replacing it with a chained function that just calls the original handler and then immediately calls its handler; I'm not sure how one would replace that with something that could check that the event was stopped.  I'm still not fully used to React yet -- I think I was tackling the problem from the wrong direction :)

jquense:
[...]

PR #1642, opened 2016-02-01
[added] closeOnSelect prop for Dropdown

fsi:
Allow making dropdown menus that persist clicks, e.g. for dropdown lists of toggle switches, for issue #1490

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -218,7 +218,7 @@ class Dropdown extends React.Component {
     menuProps.onSelect = createChainedFunction(
       menu.props.onSelect,
       this.props.onSelect,
-      this.props.closeOnSelect && this.handleClose
+      this.props.closeOnSelect ? this.handleClose : undefined
     );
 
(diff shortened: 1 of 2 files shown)


jquense:
Thanks for the PR! I appreciate you taking the time. However, If someone wanted to keep any clicks from closing they dropdown then I think just listening to `onToggle` and not flushing the value to to `open` makes more sense, and doesn't require a change in the api here.

If we are to add anything I think the right thing to add is an extra param to the onToggle handler that says what triggered the toggle. That way a user can differentiate between whether an item was selected, or clicked outside the menu.

BryceHamilton:
Hey @jquense would you mind clarifying: 

> not flushing the value to to `open`

Much appreciated :)

BryceHamilton:
Hey @jquense could you elaborate on what you mean by "not flushing the value to to `open`" ? Thank you!

twall:
See https://stackoverflow.com/questions/32510976/react-bootstrap-dropdown-with-input-wont-stay-open

Issue #1657, opened 2016-02-14
Improve MenuItem docs


jakebuhler:
The documentation lists the arguments passed to `MenuItem`'s `onSelect` callback incorrectly. It states that the callback is given `eventKey`, `href`, and `target`:

<img width="552" alt="screen shot 2016-02-13 at 5 23 53 pm" src="https://cloud.githubusercontent.com/assets/259811/13031149/ab9c8110-d276-11e5-817f-4bc39c06d770.png">

But really, `onSelect` gets passed `event` and `eventKey`.

Later in this section, `MenuItem`'s props are listed, but most of them don't have any explanation about what they do. It would be helpful if each one had a short description, similar to how the props of other components are documented. I'm working on a quick PR with these improvements.

AlexKVal:
Closed by #1658

PR #1658, opened 2016-02-14
Correct and improve MenuItem docs

jakebuhler:
I just started using `react-bootstrap` and have really liked it so far. I stumbled for a little while over a mistake in the documentation, so I thought I'd take a stab at fixing it.

The section for `MenuItem` lists the wrong arguments for the `onSelect` callback (#1657). While fixing this issue, I also added blurbs for `MenuItem`'s props. Hope this is helpful.

### DIFF
--- a/src/MenuItem.js
+++ b/src/MenuItem.js
@@ -76,7 +76,8 @@ MenuItem.propTypes = {
   disabled: React.PropTypes.bool,
 
   /**
-   * Make the menu item a divider.
+   * Styles the menu item as a horizontal rule, providing visual separation between
+   * groups of menu items.
    */
   divider: all(
     React.PropTypes.bool,
@@ -93,7 +94,7 @@ MenuItem.propTypes = {
   eventKey: React.PropTypes.any,
 
   /**
-   * Make the menu item a header.
+   * Styles the menu item as a header label, useful for describing a group of menu items.
    */
   header: React.PropTypes.bool,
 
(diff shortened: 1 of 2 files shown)


src/MenuItem.js
```diff
@@ -64,8 +64,20 @@ class MenuItem extends React.Component {
...
+   */
   active: React.PropTypes.bool,
+
+  /**
+   * Disable the menu item, making it unselectable.
+   */
   disabled: React.PropTypes.bool,
+
+  /**
+   * Make the menu item a divider.
```
jquense: I might explain this a bit more, since using the word "divider" to describe `divider` doesn't add a lot of clarity. Perhaps something like: "Styles a menu item as a horizontal rule, providing visual separation between groupings of menu items"

src/MenuItem.js
```diff
@@ -74,14 +86,51 @@ MenuItem.propTypes = {
...
     }
   ),
+
+  /**
+   * Value passed to the `onSelect` handler, useful for identifying the selected menu item.
+   */
   eventKey: React.PropTypes.any,
+
+  /**
+   * Make the menu item a header.
```
jquense: ditto here as `divider`

jquense:
thanks! just a few comments and then LGTM

taion:
LGTM with the same caveats

jakebuhler:
Thanks for the feedback! I've updated the descriptions for `divider` and `header` as suggested.

AlexKVal:
It looks like: :sparkle: :sparkles: 
screen shot 2016-02-19 at 22 18 04

Thank you @jakebuhler
:cherries:

PR #1662, opened 2016-02-21
[added] Allow overriding Dropdown CSS classes

jonathaningram:
(In case it's not obvious), the intent of this is to be able to use something like CSS modules to pass custom Bootstrap class names like:

```js
import bsStyles from 'bootstrap.css'

<Dropdown openClassName={bsStyles.open} />
```

Any issues, let me know.
Cheers, Jon.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -92,10 +92,10 @@ class Dropdown extends React.Component {
     let className = bootstrapUtils.prefix(this.props);
 
     const rootClasses = {
-      open: this.props.open,
-      disabled: this.props.disabled,
+      [this.props.openClassName]: this.props.open,
+      [this.props.disabledClassName]: this.props.disabled,
       [className]: !this.props.dropup,
-      dropup: this.props.dropup
+      [this.props.dropupClassName]: this.props.dropup
     };
 
     return (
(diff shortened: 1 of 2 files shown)


taion:
I like this in principle, but this seems like a vaguely dangerous path to go down. Would it make sense to extract this version of `<Dropdown>` into react-overlays?

cc @jquense

jquense:
Yeah, I think react-overlay's is a better fit for this sort of thing generally...that being said I don't really like the tbs dropdown implementation as a general one. If I had my way i'd all be built on top of Position, or something like it for more general absolute positioning. 

In any case though I think the complication here is that most of the functionality of the dropdown is aggressively coupled to the backing css, including assumptions about markup position, etc...thats a lot of rambling to say

1. I don't know if a properly abstracted-from-tbs Dropdown is helpful
2. I don't want to add more of these configuration hooks to react-bootstrap, they get out of control super quickly and we can't support the same for other components

If I had to choose I think sticking this in react-overlays is the better of the options.

jonathaningram:
For the record, my aim is not to abstract Dropdown away from Bootstrap's impl, but it's just to provide the (hashed/shortened) class names that my build/bundle produces.

I can understand that this kind of thing would get out of control when you consider the multitude of components - adding custom class props everywhere that is. But, without doing this, does that mean that one can't really use react-bootstrap in the way I'm trying to and you'd be forced to stick with the original class names?

I had a quick look at react-overlays, but I'm not sure what it means practically for me that this should go in react-overlays?

taion:
The idea is that we move `<Dropdown>` entirely (or a base version of the same) into react-overlays.

taion:
I'm closing this out for now. We'd be happy to get a PR that moves some of this logic into react-overlays in a way that you can get this functionality, though.

PR #1663, opened 2016-02-22
[added] contentClassName in Modal Component

flqw:
I needed to add custom classes to the modal-content element in my current project, so I added the possibility with a new ``contentClassName`` property analougous to ``dialogClassName``.

Example:
```jsx
<Modal show={show} onHide={this.close} 
             className="m__modal" 
             contentClassName="m__modal__content">
...
</Modal>
```

Results in:
```html
<div class="m__modal fade in modal" ...>
	<div class="modal-sm modal-dialog" ...>
		<div class="m__modal__content modal-content" ...>
		...
		</div>
	</div>
</div>
```


taion:
I'm pretty sure we'd rather you just use a custom modal dialog if you wanted this.

flqw:
Why then the ``dialogClassName`` and ``className`` properties? Wouldn't it be more consistent to choose one approach? All three div's are within the same [component](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ModalDialog.js) and on only two can receive custom classes?

taion:
As noted, please use the lower-level primitives in react-overlays if you need additional customizability.

PR #1664, opened 2016-02-22
[added] Collapsible Panel transition hooks

### LINKED ISSUE #1540: onExited not called when set in Accordion or Panel component
I'm using Accordion component to create a set of collapsable panels and I needed to assign a call back when the component is collapsed. I did so by setting the onExited prop on Accordion and even tried setting this on the Panel component but, the callback is never getting called. Here's my render method:
```
  render() {
    return (
      <Accordion defaultActiveKey='1' onExited={this.handleExited}>
        <Panel onExited={this.handleExited} >
          <span>Some text</span>
        </Panel>
      </Accordion>
    )
  }
```
I can see that the Panel component is getting the prop onExited but, the generated Transition component never gets the prop onExited set from the parent Panel component. Is this a bug or am I missing something here? Help is much appreciated.

I'm using react-bootstrap 0.25.1

martijnvermaat:
This should fix #1540 

I found a number of different strategies for passing props to children in the codebase. Here I tried to minimize duplication of proptype definitions and their documentation.

For some reason the docs fail to build with a `TypeError: Converting circular structure to JSON` and this seems to be due to how I extended the proptypes. I don't yet understand why, any hints are appreciated.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -19,7 +19,14 @@ let Panel = React.createClass({
     expanded: React.PropTypes.bool,
     eventKey: React.PropTypes.any,
     headerRole: React.PropTypes.string,
-    panelRole: React.PropTypes.string
+    panelRole: React.PropTypes.string,
+
+    onEnter: Collapse.propTypes.onEnter,
+    onEntering: Collapse.propTypes.onEntering,
+    onEntered: Collapse.propTypes.onEntered,
+    onExit: Collapse.propTypes.onExit,
+    onExiting: Collapse.propTypes.onExiting,
+    onExited: Collapse.propTypes.onExited
   },
 
   getDefaultProps() {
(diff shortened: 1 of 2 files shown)


martijnvermaat:
Sorry, I just realized I forgot to push some changes to this branch. Will fix when I have access to that machine later today.

martijnvermaat:
This is now working. The passed props are now just enumerated explicitely.

Please let me know if you would like me to change anything.

jquense:
LGTM, need one more @react-bootstrap/collaborators to ok it

mathieumg:
I don't write my empty arrows without spacing (`()=>`) but I don't think it's something we enforce! (Otherwise I imagine we'd use the ESLint rule for it...) LGTM.

martijnvermaat:
@mathieumg Me neither, I copied that code from another similar test and didn't notice. It occurs quite a lot in the tests.

Thanks all for having a look and merging!

PR #1672, opened 2016-02-29
Clean up "getting started guide"

jquense:


taion:
Hah – you use `npm install <foo> --save` instead of `npm install --save foo`.

A lot neater now. LGTM.

jquense:
--SAVE LAST 4 LIFE!

PR #1604, opened 2016-01-16
[changed] onSelect events consistent signature

jquense:
The various onSelect events will now consistently be called with:
eventKey, event (if any)

PR #1676, opened 2016-03-02
Pass event to PanelGroup onSelect handler and cancel expanding if the custom handler returns false

jkrehm:
This changes the public API for the onSelect handler from `function handleSelect(eventKey)` to `function handleSelect(event, eventKey)`.

Because the onSelect handler receives the DOM event, it can check what the target element was. Combined with the new behavior when "false" is returned from the handler, clicks on specific elements can be ignored, e.g. a button with onClick logic is included in the panel and clicking it should not toggle the panel.

### DIFF
--- a/docs/examples/PanelGroupControlled.js
+++ b/docs/examples/PanelGroupControlled.js
@@ -5,7 +5,7 @@ const ControlledPanelGroup = React.createClass({
     };
   },
 
-  handleSelect(activeKey) {
+  handleSelect(e, activeKey) {
     this.setState({ activeKey });
   },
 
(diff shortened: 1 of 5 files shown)


taion:
For consistency with https://github.com/react-bootstrap/react-bootstrap/pull/1604 (which looks like it missed `<Panel>`?) the signature should be the other way around. Please re-submit against the v0.29-rc branch.

jkrehm:
Good deal, I wasn't sure if that branch was active or not since master has been committed to quite a lot since the last time that branch was committed to. I'll update to match the function signature that's in `0.29-rc` and re-submit.

PR #1677, opened 2016-03-02
[changed] Make Panel expansion cancelable

jkrehm:
Pass event to PanelGroup onSelect handler and cancel expanding if the custom handler returns false

This changes the public API for the onSelect handler from `function handleSelect(eventKey)` to `function handleSelect(eventKey, event)`.

Because the onSelect handler receives the DOM event, it can check what the target element was. Combined with the new behavior when "false" is returned from the handler, clicks on specific elements can be ignored, e.g. a button with onClick logic is included in the panel and clicking it should not toggle the panel.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -38,9 +38,7 @@ let Panel = React.createClass({
     e.selected = true;
 
     if (this.props.onSelect) {
-      if (this.props.onSelect(this.props.eventKey, e) === false) {
-        e.selected = false;
-      }
+      this.props.onSelect(this.props.eventKey, e);
     } else {
       e.preventDefault();
     }
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -38,7 +38,9 @@ let Panel = React.createClass({
     e.selected = true;
 
     if (this.props.onSelect) {
-      this.props.onSelect(e, this.props.eventKey);
+      if (this.props.onSelect(this.props.eventKey, e) === false) {
+        e.selected = false;
+      }
     } else {
       e.preventDefault();
     }
--- a/src/PanelGroup.js
+++ b/src/PanelGroup.js
@@ -70,13 +70,17 @@ const PanelGroup = React.createClass({
     return !this._isChanging;
   },
 
-  handleSelect(e, key) {
+  handleSelect(key, e) {
     e.preventDefault();
 
     if (this.props.onSelect) {
       this._isChanging = true;
-      this.props.onSelect(key);
+      let setActiveKey = this.props.onSelect(key, e);
       this._isChanging = false;
+
+      if (setActiveKey === false) {
+        return false;
+      }
     }
 
     if (this.state.activeKey === key) {
--- a/src/PanelGroup.js
+++ b/src/PanelGroup.js
@@ -75,12 +75,8 @@ const PanelGroup = React.createClass({
 
     if (this.props.onSelect) {
       this._isChanging = true;
-      let setActiveKey = this.props.onSelect(key, e);
+      this.props.onSelect(key, e);
       this._isChanging = false;
-
-      if (setActiveKey === false) {
-        return false;
-      }
     }
 
     if (this.state.activeKey === key) {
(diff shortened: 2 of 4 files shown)


jkrehm:
Looks like the build failed while npm installed the dependencies, but it's not related to my changes.

jquense:
Thanks for making the `onSelect` consistent, that's helpful. However I dont think adding cancellation logic this way is the right approach, and is not consistent with how other the components work.

Good news though, right now selections can be already be cancelled by controlling `expanded` and `onSelect`. You can decide not to flush the change or not. In terms of your use-case, of stopping propagation for certain header elements I think this needs to be more elegantly addressed.

cc: @taion

taion:
Agreed on both points.

jquense:
see: https://github.com/react-bootstrap/react-bootstrap/issues/983

jkrehm:
I'm fine with whatever makes it so I can prevent expanding/collapsing if a specific element is selected. The solution I provided is pretty free-format so the developer could use whatever logic they want and as long as they return `false` from the `onSelect` hook, they'll prevent the expansion/collapse. So for that reason I still think my approach has merit.

taion:
I understand what you want to do, but in the interest of minimizing API surface area, we don't want to have this `return false` behavior – it's not broadly consistent with the other `onSelect` callbacks, and it's also something that can be handled fairly adequately in user-space.

jkrehm:
I guess as long as the PanelGroup select handler receives the event, a developer (me) can use `e.selected = false;` (since objects are passed by reference in JavaScript). So then the most important part of the commit is getting the event to the PanelGroup.

Want me to update the commit to revert the `return false` stuff and simply fix Panel's select handler signature and add the event to PanelGroup's `onSelect` call?

jquense:
> Want me to update the commit to revert the return false stuff and simply fix Panel's select handler signature and add the event to PanelGroup's onSelect call?

yes please.

And like I noted, you can already cancel selection with the current API by controlling `expanded` prop with `onSelect`, with your change you can do based on the target as well.

jkrehm:
I'm not sure I understand what your method would look like.

Panel.js's `handleSelect` function:
```
  handleSelect(e) {
    e.selected = true;

    if (this.props.onSelect) {
      this.props.onSelect(this.props.eventKey, e);
    } else {
      e.preventDefault();
    }

    if (e.selected) {
      this.handleToggle();
    }
  },
```

During the user-land `onSelect` call, I don't see how modifying the props (which should be treated as immutable, so we shouldn't be changing it at all) for the `Panel` component (which won't be available via `this`, so I'm not sure how we'd do it anyway) would stop the `this.handleToggle();` call from firing. So best case scenario, it toggles and then the dev finds a way to toggle it back before the DOM renders. I must be missing something.

jquense:
take a look at [the controlled panel examples](http://react-bootstrap.github.io/components.html#panels-collapsible) if you provide an `expanded` prop the panel will _aways_ us that value for whether its expanded or not. Just like a controlled input. You can decide when to change it.

```js
handleSelect(key, event) {
   let current = this.state.expanded;
   // only toggle panel if an input wasn't clicked
   if (event.target.tagName !== 'INPUT')
     this.setState({ expanded: !current })
},

render() {
  return (
    <Panel
      header={myHeader}
      expanded={this.state.expanded}
      onSelect={this.handleSelect}
    >
      content
    </Panel>
  )
}
``

jkrehm:
I guess I don't fully grok how React deals with multiple states getting set then, because that example looks like it will set the expanded state and `handleToggle` in Panel.js will set the expanded state a second time unless you set `e.selected = false`. To me it looks like a race condition.

jquense:
it does set both but it only uses one at a time. look at the isExpanded method

jkrehm:
All righty. What would you like me to do with the pull request?

taion:
Sorry for the delay – it looks like the code on this PR is a pure win, so I'm going to merge it by hand for now (with an amended commit message that describes what it actually does).

We can resolve other issues later.

Thanks!

taion:
Ah, I see the commit message details actually do describe the change made. Sorry, misread the subject.

Issue #1660, opened 2016-02-17
MenuItem divider should render className prop


mjvestal:
The MenuItem returns here when the ```divider``` prop is included:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/MenuItem.js#L33-L35

I think it should include any classes the consumer provides, such that if I have:
```
<MenuItem divider className="visible-xs visible-sm" />
```

It should render
```
<li role="separator" class="divider visible-xs visible-sm"></li>
```

mjvestal:
It occurred to me, why don't I just:
```
<Dropdown.Menu>
    <MenuItem ..>...</MenuItem>
    <li role="separator" className="divider visible-xs visible-sm" />
    ...
</Dropdown.Menu>
```

Besides the un-DRY-ness of having ```role="separator"``` for dividers, why not just have consumers do that instead?

taion:
Fixed by https://github.com/react-bootstrap/react-bootstrap/pull/1682.

PR #1682, opened 2016-03-04
[fixed] #1660 MenuItem divider className

granteagon:
<MenuItem divider className="foo" />

rendered:

<li role="separator" class="divider" />

now renders:

<li role="separator" class="divider foo" />


jquense:
LGTM

PR #1687, opened 2016-03-08
Improving user experience on panels

estshy:
Currently user has really bad experience while using `Accordion` component. He have to click directly on title to switch between tabs. He should be able to do so by clicking on header instead of anchor.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -192,7 +192,7 @@ let Panel = React.createClass({
     }
 
     return (
-      <div className={bootstrapUtils.prefix(this.props, 'heading')}>
+      <div className={bootstrapUtils.prefix(this.props, 'heading')} onClick={this.handleSelect}>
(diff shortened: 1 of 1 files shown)


taion:
This exactly matches the upstream TWBS behavior as-is.

estshy:
> Just add data-toggle="collapse" and a data-target to the element to automatically assign control of a collapsible element. The data-target attribute accepts a CSS selector to apply the collapse to. Be sure to add the class collapse to the collapsible element. If you'd like it to default open, add the additional class in.

The upstream behavior or example from documentation? Please see how easily you can customize it: http://jsfiddle.net/R6EAW/3447/

jquense:
vanilla Bootstrap is more customizable in this case. We are working on that in #983. For now though we are going to not differ from the default use case,

Issue #1680, opened 2016-03-03
DropdownButton doen't supports block property.


khakulov:
DropdownButton doen't supports block property.

taion:
Seems like an easy fix. Care to put together a PR?

hitchcockwill:
I took a stab at this today. It's a pretty simple fix, but leads to an edge case with styling that's not covered in the bootstrap library styling.

Line 22 in `src/ButtonGroup.js` requires that if the `block` prop is present that `vertical` must also be present.

The bootstrap styling for ButtonGroup expects that all children of `.btn-group-vertical` are instances of `.btn`. In the case of `DropdownButton`, we generate the following HTML where `.btn-group-vertical` has multiple children but only one `.btn`:

```
<div class="dropdown btn-group-vertical btn-block">
  <button id="dropdown-size-medium-block" role="button" aria-haspopup="true" aria-expanded="false" type="button" class="dropdown-toggle btn btn-default btn-block">Default Button</button>
  
  <ul role="menu" class="dropdown-menu" aria-labelledby="dropdown-size-medium-block">...</ul>
</div>
```

That, combined with this styling from bootstrap renders a button with no border radius on the bottom corners each time.

```
.btn-group-vertical > .btn:first-child:not(:last-child) {
    border-top-right-radius: 4px;
    border-top-left-radius: 4px;
    border-bottom-right-radius: 0;
    border-bottom-left-radius: 0;
}
```

![screenshot 2017-10-11 18 37 51](https://user-images.githubusercontent.com/1471313/31473703-783d6774-aeb3-11e7-89d3-7355d055c350.png)

One option here would be to make it so `block` does not require `vertical` in `ButtonGroup.js`, but I'm reluctant to make that change because I'm assuming that's there for a reason. 

Another option is to add some inline styling, but that seems hacky.

Looking for some other suggestions here. Any ideas?

taion:
What does upstream TWBS show in their demos, if anything? Does `block` do anything for dropdowns if `vertical` isn't set?

We can just enable the feature and not update the docs if it's not something that's actually supported by default TWBS styling.

hitchcockwill:
TWBS doesn't show a block level button dropdown in the docs. Adding the `btn-block` class to `.btn-group` and `.btn` gives the correct styling but when `.btn-group-vertical` is added the border radius is removed.

[...]

PR #1688, opened 2016-03-08
[added] DropdownButton now supports block attribute

jguffey:
Hi there, Issue #1680 requested block property on Dropdown buttons, this PR attempts to incorporate that feature.

I've updated the documentation, and wrote a minimal test.

### DIFF
--- a/test/DropdownButtonSpec.js
+++ b/test/DropdownButtonSpec.js
@@ -63,6 +63,21 @@ describe('DropdownButton', () => {
     menu.props.pullRight.should.be.true;
   });
 
+  it('renders block level with block prop', () => {
+    const instance = ReactTestUtils.renderIntoDocument(
+      <DropdownButton title="blah" id="test-id" block vertical>
+        <MenuItem eventKey="1">Action</MenuItem>
+      </DropdownButton>
+    );
+    const menu = ReactTestUtils.findRenderedComponentWithType(instance, DropdownButton);
+    const node = ReactDOM.findDOMNode(instance);
+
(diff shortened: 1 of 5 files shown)


test/DropdownButtonSpec.js
```diff
@@ -63,6 +63,21 @@ describe('DropdownButton', () => {
...
+      <DropdownButton title="blah" id="test-id" block vertical>
+        <MenuItem eventKey="1">Action</MenuItem>
+      </DropdownButton>
+    );
+    const menu = ReactTestUtils.findRenderedComponentWithType(instance, DropdownButton);
+    const node = ReactDOM.findDOMNode(instance);
+
+    menu.props.vertical.should.be.true;
+    menu.props.block.should.be.true;
+    console.log(node.className);
```
taion: delete

jguffey:
How's this look now?

jquense:
`vertical` is only required when the dropdown `componentClass` is a `ButtonGroup`, which is the default case but not the only one. To most ppl its not going to make sense _why_ vertical is required since the rendering of a ButtonGroup is opaque to a user of DropdownButton, and just semantically its a button group of one. The other problem is that this won't work with SplitButton since there are two buttons in that component. I have no idea what the right behavior of NavDropdown would be either.

Which is all to say, I think this prop belongs in the DropdownButton component instead, be only the `block` prop and add `vertical` internally.

jquense:
Gonna close for now, If you want to take a stab at moving this to DropdownButton @jguffey I'd be happy to review and merge

PR #1692, opened 2016-03-11
Fix for a responsive and collapsive navbar on item select event

### LINKED ISSUE #1301: Add option to collapse Navbar when item is selected
How do you make the menu collapse after item is selected?

    import React from 'react';
    import {Navbar, Nav, NavItem, NavDropdown, DropdownButton, MenuItem, CollapsibleNav} from 'react-bootstrap';

    export default class App extends React.Component {

    constructor(props) {
      super(props);
      this.onSelect = this.onSelect.bind(this);
      this.toggleNav = this.toggleNav.bind(this);
      // this.state = {navExpanded: false};
    }

    onSelect(e){
        console.log('OnSelect')
        // console.log(this.state.navExpanded);
        // this.setState({navExpanded: false});
    }

    toggleNav(){console.log('toggle...')};

    // <Navbar inverse fixedTop toggleNavKey={0} navExpanded={this.state.navExpanded} onToggle={() => this.toggleNav()}>
    // <Navbar inverse fixedTop toggleNavKey={0} navExpanded={this.state.navExpanded} >

    render() {
        return (

          <Navbar inverse fixedTop toggleNavKey={0} >
            <Nav right eventKey={0} onSelect={this.onSelect} > {/* This is the eventKey referenced */}
              <NavItem eventKey={1} href="#">Link</NavItem>
              <NavItem eventKey={2} href="#">Link</NavItem>
            </Nav>
          </Navbar>     

      )

    }

    componentDidMount() {

    }
    }

    React.render(<App />, document.getElementById('example'));

zirho:
This is not perfect since dropdown menu in nav not collapsing all the way up to main nav but hope this help for someone to use or continue to fix this issue
fix #1301

### DIFF
--- a/docs/examples/NavbarCollapsible.js
+++ b/docs/examples/NavbarCollapsible.js
@@ -9,7 +9,7 @@ const navbarInstance = (
     <Navbar.Collapse>
       <Nav>
         <NavItem eventKey={1} href="#">Link</NavItem>
-        <NavItem eventKey={2} href="#">Link</NavItem>
+        <NavItem eventKey={2} autoCollapse href="#">Link (autoCollapse on)</NavItem>
         <NavDropdown eventKey={3} title="Dropdown" id="basic-nav-dropdown">
           <MenuItem eventKey={3.1}>Action</MenuItem>
           <MenuItem eventKey={3.2}>Another action</MenuItem>
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -16,12 +16,14 @@ const NavItem = React.createClass({
     title: React.PropTypes.node,
     eventKey: React.PropTypes.any,
     target: React.PropTypes.string,
+    autoCollapse: React.PropTypes.bool,
     'aria-controls': React.PropTypes.string
   },
 
   contextTypes: {
     $bs_navbar_onToggle: PropTypes.func,
-    $bs_navbar_expanded: PropTypes.bool
+    $bs_navbar_expanded: PropTypes.bool,
+    $bs_navbar_auto_collapse: PropTypes.bool
   },
 
   getDefaultProps() {
@@ -52,13 +54,17 @@ const NavItem = React.createClass({
 
     let {
       $bs_navbar_onToggle: onToggle,
-      $bs_navbar_expanded: expended
+      $bs_navbar_auto_collapse: autoCollapse
     } = this.context;
+
+    if (this.props.autoCollapse !== undefined) {
+      autoCollapse = this.props.autoCollapse;
+    }
 
     let linkProps = {
       role,
       href,
-      onClick: createChainedFunction(onClick, this.handleClick, () => { if (expended) onToggle(); }),
+      onClick: createChainedFunction(onClick, this.handleClick, () => { if (autoCollapse) onToggle(false); }),
       title,
       target,
       tabIndex,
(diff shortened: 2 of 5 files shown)


taion:
Implementation concerns aside (and there are those), we don't want this to be the default behavior – upstream TWBS doesn't work this way.

taion:
The right API here is to give `NavItem` something like a prop that opts into this behavior on the `NavItem` level.

zirho:
To make clicking NavItems collapse NavbarCollapse, put "autoCollapse"
attribute on NavbarCollapse or directly to NavItems to overwrite the
behavior.
or just simply put it on only one NavItem to collapse Navbar.

taion:
It stands for Twitter Bootstrap.

zirho:
I found something wrong in action.
Hold this for a moment 

thanks

olalonde:
+1, also need this, can we merge :)

@taion Not sure it's fair to say TWBS doesn't work this way as TWBS nav links are not typically intercepted in Javascript and to the user, it looks as if the menu collapsed after clicking a link. I'm fine if it's an optional attribute though.

jquense:
OK the diff here is a bit of a mess so its tough to see what's going one but I don't think this is the right approach...

If I was writing this I think I would not worry about "auto collapsing" at all and just make sure to surface any onSelect events from Nav's or Dropdown's up to the Navbar, then the user can decide to do whatever they want with that, such as toggling the collapse. 

My problem with codifying this behavior as a feature is that we can't reasonably cover all the general cases you'd want the navbar to collapse, like Buttons...actually this is fairly straightforward to do in user land I think we should just stay out entirely.

```jsx
render() {
  return (
    <Navbar 
       expanded={this.state.expanded}
       onToggle={expanded => this.setState({ expanded })}
    >
      <Nav onSelect={() => this.setState({ expanded: false })}>
         {/* whatever */}
      </Nav>
    </Navbar>
  )
}
```

zirho:
Hi, Thanks for all feedbacks.

@jquense I tried to apply your approach on react-tutorial. (using CDNs)
Here is the code that I was working, but I couldn't make it work. (it may be due to lack of basic for react)
https://github.com/reactjs/react-tutorial
Can you advise me to make this work, if it's not too much trouble? or anybody?
I'd be very much appreciated if you can.

Thank you in advance.

taion:
@jquense Do you think there's any value in having a simplified version for the case where the navbar only contains links?

jquense:
perhaps? I am worried about the complication tho, and "contexting all the things" adding too much hidden complexity. 

in this case I'd be worried that there will be to many edge cases and we'll get issues about it. I don't feel super strongly tho, especially since I don't tend to need this behavior. what do you think?

taion:
I'd probably use this feature if it were available.

At a guess, most users of `<Navbar>` probably use it in a very simple form with just plain `<NavItem>`s, so in practice this should give users a good way to get what they need.

olalonde:
I'm not sure I understand what the work around is. I tried setting collapsed attribute on my NavBar but once the toggle is pushed once, it ignores the collapsed attribute.

abloomston:
I found @jquense user land implementation to work fine for me. I would, however, find it quite useful to have that implementation available via simply setting a prop.

zirho:
@abloomston Can you please give a link to it?

thanks

abloomston:
@zirho here is the commit where I used @jquense user land implementation: https://github.com/abloomston/learning-bootstrap-3/commit/275940373d3877ece2c3e4851ee117cdccb000a3

zirho:
@abloomston Thank you!

taion:
Replaced by https://github.com/react-bootstrap/react-bootstrap/pull/2280

PR #1707, opened 2016-03-18
[added] Media object component

alphashuro:
Added media objects and media list.


docs/examples/.eslintrc
```diff
@@ -42,6 +42,12 @@
     "NavDropdown",
     "NavItem",
     "MenuItem",
+    "Media",
```
taion: Alphabetize these

docs/src/ComponentsPage.js
```diff
@@ -76,6 +77,7 @@ const sections = {
     thumbnails: '#thumbnail',
     embed: '#responsive-embed',
     carousels: '#carousels',
+    'media-objects': '#media-objects',
```
taion: Make the key here `mediaObjects` for consistency.

docs/src/ReactPlayground.js
```diff
@@ -33,6 +33,12 @@ const Jumbotron = require('../../src/Jumbotron');
 const Label = require('../../src/Label');
 const ListGroup = require('../../src/ListGroup');
 const ListGroupItem = require('../../src/ListGroupItem');
+const Media = require('../../src/Media/Media');
```
taion: Alphabetize these

src/Media/MediaLeft.js
```diff
@@ -0,0 +1,31 @@
...
+    href: '#',
+    align: 'top'
+  }
+
+  render() {
+    const {href, align, children, className} = this.props;
+
+    return (
+      <div className={`media-left media-${align} ${className}`}>
+        <a href={href}>
```
taion: We don't need this – if the user needs a link, he or she can pass in an `<a>` as children to this component.

src/Media/MediaLeft.js
```diff
@@ -0,0 +1,31 @@
...
+  static defaultProps = {
+    href: '#',
+    align: 'top'
+  }
+
+  render() {
+    const {href, align, children, className} = this.props;
+
+    return (
+      <div className={`media-left media-${align} ${className}`}>
```
taion: Use `classNames`.

taion: Do not pass in an `align` by default – per the upstream examples.

src/Media/MediaLeft.js
```diff
@@ -0,0 +1,31 @@
+import React, {Component, PropTypes} from 'react';
```
taion: We just do `import React from 'react'`; don't do the faux named imports here.

src/Media/MediaBody.js
```diff
@@ -0,0 +1,20 @@
...
+     * The heading of the media object
+     */
+    heading: PropTypes.string
+  }
+
+  render() {
+    const {heading, children, className} = this.props;
+    return (
+      <div className={`media-body ${className}`}>
+        <h4 className="media-heading">{heading}</h4>
```
taion: This heading should be optional. `<h4>` is also probably not generally semantically correct.

src/Media/Media.js
```diff
@@ -0,0 +1,12 @@
+import React, {Component} from 'react';
+
+export default class Media extends Component {
+  render() {
+    const {className, children} = this.props;
```
taion: You should be forwarding the remaining props here.

src/Media/index.js
```diff
@@ -0,0 +1,6 @@
+export Media from './Media';
```
taion: Why nest these in a subdirectory? We don't do that elsewhere, I don't think.

src/Media/Media.js
```diff
@@ -0,0 +1,12 @@
+import React, {Component} from 'react';
+
+export default class Media extends Component {
```
taion: I think for consistency, we'd want to re-export e.g. `Media.Body`, `Media.Left`, &c. from here, and use those by preference in examples. @jquense WDYT?

jquense: yup. I also think Media.Header probably should be a seperate component, otherwise we will have folks complaining about the `<h4>` component hard code

src/Media/MediaLeft.js
alphashuro: Thank you for the feedback, I'm working on the changes now but I'm not certain what you mean in this statement?

alphashuro: Are you saying that if align isn't passed by the user, I should not add a ```media-${align}``` class at all?

jquense: yup I believe that's his point :)

alphashuro:
@taion and @jquense I have updated the code, please take a look and tell me what you think

src/Media.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+
+  getDefaultProps() {
+    return {
+      componentClass: 'div'
+    };
+  },
+
+  render() {
+    const ComponentClass = this.props.componentClass;
```
taion: You can simplify this whole block as:

```js
const { ComponentClass, className, ...props } = this.props;

return (
  <ComponentClass {...props} className={classNames(className, 'media')} />
);
```

`children` is implicitly spread through as a prop.

taion: Fixed a syntax error in my code snippet.

src/Media.js
```diff
@@ -0,0 +1,38 @@
...
+
+import Heading from './MediaHeading';
+import Body from './MediaBody';
+import Left from './MediaLeft';
+import Right from './MediaRight';
+import List from './MediaList';
+import ListItem from './MediaListItem';
+
+const Media = React.createClass({
+  displayName: 'Media.Media',
```
taion: Probably call this `Media` and make this the default export. In other words, do e.g.

```js
Object.assign(Media, { Heading, Body, ... });

export default Media;
export { Heading, Body, ... }; // Optional.
```

taion:
One nit about exports – otherwise this is looking good.

alphashuro:
@taion I have updated again :D

taion:
Thanks! LGTM.

jquense:
thanks!

alphashuro:
awesome!

mathieumg:
Thank you for this @alphashuro !

Issue #1703, opened 2016-03-17
Add linkRel prop to NavItem


IT521:
Add linkRel prop to NavItem to allow rel attribute to be set on anchor tag.

taion:
Subset of https://github.com/react-bootstrap/react-bootstrap/issues/629

Issue #629, opened 2015-05-08
Specifying props for nested anchor components


derflocki:
Edited (14 aug 2015):
These look to be the components that currently lack this feature:

- [x] `MenuItem`
- [ ] `NavItem`
- [ ] `PageItem`
- [ ] `Pagination`

--
The MenuItem currently only passes a few props (`href`, `target`, `title`) to the containing `<a>`, the bulk is applied to the `<li>`. I suggest passing the props `download`, `media`, `ping`, `rel`, `type` (I'm specifically trying to add the `download`-attribute) to the `<a>`.

What do you think?

mtscout6:
Related to #526 

I hadn't though about that, but this sounds reasonable.

AlexKVal:
The same `bad API` problem as in https://github.com/react-bootstrap/react-bootstrap/pull/764

I propose to resolve these both issues by just adding `propsForAnchor` "multi-property" ?

What do you think guys ?
@react-bootstrap/collaborators

chrisvfritz:
I'd personally love a `propsForAnchor`. Then there's no need to pass a huge list of props, which could never be sufficient anyway, because data attributes would be excluded.

taion:
Or `anchorProps`? I like being able to save typing for the commonly-used props, though.

chrisvfritz:
:+1: for anchorProps. While I also like saving typing in general, I also worry that implicitly passing props to different components might get tricky. For example, in a component containing a SafeAnchor where you're also allowed to change the element type, a user might change it to something where props normally only passed to the anchor would be valid. While it saved users typing, I worry that it could be a source of bugs and a maintenance headache for maintainers. This way, we also don't have to deal with scoping the names of *some* props, like linkId.

taion:
These look to be the components that currently lack this feature:

- [ ] `MenuItem`
- [ ] `NavItem`
- [ ] `PageItem`
- [ ] `Pagination`
- [ ] `SubNav` (what is this? it's not documented and I only see it in use in our docs)

AlexKVal:
@taion I've moved your list to the first message for this:
<img width="936" alt="screen shot 2015-08-14 at 11 22 51 am" src="https://cloud.githubusercontent.com/assets/847572/9270006/e6c13064-4276-11e5-8121-8323beae7e2e.png">

legomind:
Any status on this? I am really needing style prop for the anchor element in the NavItem

illing2005:
[...]

PR #1714, opened 2016-03-23
[added] add rel attribute to NavItem

tehkaiyu:
Issue: https://github.com/react-bootstrap/react-bootstrap/issues/1703


taion:
I don't like this – per https://github.com/react-bootstrap/react-bootstrap/issues/629 we should be using something like `anchorProps` for this and other anchor props, and we should be consistent across all wrapped anchor components.

tehkaiyu:
Ah yes, that makes sense. I'll close this PR then.

PR #1717, opened 2016-03-24
Minor updates to build process

taion:


package.json
```diff
@@ -25,8 +25,6 @@
     "lib": "lib/"
   },
   "files": [
-    "LICENSE",
```
taion: These don't need to be explicitly specified.

package.json
```diff
@@ -89,7 +87,6 @@
     "less-loader": "^2.2.1",
     "marked": "^0.3.5",
     "mocha": "^2.3.3",
-    "mt-changelog": "^0.6.2",
```
taion: Per https://github.com/AlexKVal/release-script/blob/05ab23673b040549bdece1802a262f3744c4beb6/src/release.js#L27-L32, this is enough to disable automated changelog generation.

taion:
Note: until https://github.com/npm/npm/pull/11995 lands, we'll need to use npm@2 to publish releases.

mtscout6:
Should that be added to the Contributing guide then for posterity? It'll be hard to find if it's hiding in this PR comment.

taion:
I'll update `MAINTAINING.md`.

taion:
Actually, I take that back, it looks like that PR will land any minute now, so the caveat won't be necessary. I'll still update the maintaining guide regarding changelog procedures though.

taion:
Updated.

taion:
I want to cut v0.28.4 soon. Are we okay with this?

PR #1719, opened 2016-03-25
Fix style prop on nested <ProgressBar>

taion:


taion:
Oops, too quick on the merge button. @react-bootstrap/collaborators are we okay with this change?

jquense:
LGTM

PR #1720, opened 2016-03-25
Fix ComponentsPage ScrollSpy - last section active

jason-whitted:
Fix for minor bug on ComponentsPage with the ScrollSpy.  It was unable to set the last section/Anchor as active.

## Steps to reproduce

1. Navigate to: [Components Page](http://react-bootstrap.github.io/components.html)
1. Resize browser so that the **ScrollSpy** section should be marked as active in the side nav.
image

### Expected
**ScrollSpy** section to be active.

### Actual
**Affix** section is active.

## Fix Details
The `setActiveNavItem` function loops through the sections until it finds the first `afterSection` that is false.  In the case where all sections are true it did nothing.  This change breaks out of the for loop if/when a false value is encountered and then sets the state of the `activeNavItemHref`.

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -139,12 +139,13 @@ const ComponentsPage = React.createClass({
 
     for (const href of Object.keys(this.afterSections)) {
       if (!this.afterSections[href]) {
-        this.setState({ activeNavItemHref });
-        return;
+        break;
       }
 
       activeNavItemHref = href;
     }
+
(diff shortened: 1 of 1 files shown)


taion:
We probably should just replace this scrollspy impl with https://gist.github.com/jquense/a9364605bb3fa4668d77 or something like it. I was wrong in suggesting we direct people toward React Waypoint – it's not the right API for building scroll spy.

taion:
I'll merge this for now, since this is a real improvement pending a rewrite.

PR #1722, opened 2016-03-25
Add BreadcrumbItem as static prop on Breadcrumb

jquense:


taion:
:shipit:

taion:
Should we update the examples to use this?

Issue #1271, opened 2015-09-01
Allow other HTML components in ListGroupItem than <p>


binarykitchen:
ListGroupItem currently pastes a `<p>` inside for any contents.

But I would like to try something like this:

```
<ListGroupItem
  header   = {title}
>
  {
    <ul>
      <li>{archivedCount} archived</li>
      <li>{nonArchivedCount} non-archived</li>
    </ul>
  }
</ListGroupItem>
```

Have a bulleted list of some stats inside the list item. This instead of a paragraph. Is that possible?

apkiernan:
This is currently not possible (technically it renders, but you get that empty `<p>` tag in the markup which isn't ideal). Bootstrap doesn't mention if this is supported, but after messing around with it a bit, it doesn't appear to break anything, and I think it could be implemented.

taion:
Related to discussions RE:idea API for panels.

danieljuhl:
Any news on this? The problem right now, is that though it works it produces invalid HTML.

danieljuhl:
If I look at the Bootstrap documentation, it states nowhere that it has to be a `<p>`, it is just an example of custom content, where they have a `<h4>` combined with a `<p>`.

IMO we could release the API, and only render the `<p>` in case the `children` is pure text.

taion:
PRs welcome.

danieljuhl:
@taion but before I start working on it, I would like to know, whether you have something against the core concept of releasing the API, in regards to `Related to discussions RE:ideal API for panels.`

PiTrem:
Hi,
the `<p></p>` comes from using the props 'header'. [https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ListGroupItem.js#L66](url)
You can avoid using it and define your header yourself, eg:
  `<ListGroupItem>
     <h4 className="list-group-item-heading" > {title} </h4>
     ....
  </ListGroupItem> `

danieljuhl:
Have you actually made it work, @PiTrem ? The way I see the code, that would render a `<span>` as the container, which should only contain inline elements.

taion:
We'll probably support this via https://github.com/react-bootstrap/react-bootstrap/issues/983#issuecomment-203594158.

If we can find a way to do so that's not too breaking, it might be nice to merge #1731, though.

PR #1731, opened 2016-03-30
fix(ListGroupItem): Allowing for div children to be rendered.

TheIronDeveloper:
Attempting to fix https://github.com/react-bootstrap/react-bootstrap/issues/1271

This commit updates `ListGroupItem` to allow passing in block-level
elements (like div) by checking the `typeof` the props.children.

If its a string, then we'll wrap the content in the existing `<p>`
and render the content the same way as before.

If the children is not a string, then simply render out `props.children`
in place of where the `<p>` tag would be.

Example usage:

```javascript
let header = (<div>Header 5</div>);

const listgroupInstance = (
  <ListGroup>
    <ListGroupItem header="Header 4"><div>Item 4</div></ListGroupItem>
    <ListGroupItem header={header}><div>Item 5</div></ListGroupItem>
  </ListGroup>
);
```

will output something similar to:

```
<ul class="list-group">
  <li class="list-group-item">
    <h4 class="list-group-item-heading">Header 4</h4>
    <div>Item 4</div>
  </li>
  <li class="list-group-item">
    <div class="list-group-item-heading">Header 5</div>
    <div>Item 5</div>
  </li>
</ul>
```

screen shot 2016-03-30 at 12 13 46 am

### DIFF
--- a/docs/examples/ListGroupDefault.js
+++ b/docs/examples/ListGroupDefault.js
@@ -1,8 +1,13 @@
+let header = (<div>Header 5</div>);
+
 const listgroupInstance = (
   <ListGroup>
     <ListGroupItem>Item 1</ListGroupItem>
     <ListGroupItem>Item 2</ListGroupItem>
     <ListGroupItem>...</ListGroupItem>
+    <ListGroupItem header="Header 4"><div>Item 4</div></ListGroupItem>
+    <ListGroupItem header={header}><div>Item 5</div></ListGroupItem>
+    <ListGroupItem header="Header 6"><ul><li>Item 6</li></ul></ListGroupItem>
   </ListGroup>
 );
 
--- a/src/ListGroupItem.js
+++ b/src/ListGroupItem.js
@@ -80,6 +80,18 @@ class ListGroupItem extends React.Component {
       );
     }
 
+    /**
+     * If (this.props.children) matches an invalid descendent of the <p>, skip wrapping the children with a <p>
+     * and instead render the children as-is. See validateDOMNesting for more details.
+     */
+    let invalidPTagDescendants = {
+      ul: true,
+      div: true
+    };
+    if (this.props.children && invalidPTagDescendants[this.props.children.type]) {
+      return [header, this.props.children];
+    }
+
     let content = (
       <p key="content" className={bootstrapUtils.prefix(this.props, 'text')}>
         {this.props.children}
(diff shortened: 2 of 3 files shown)


taion:
This is breaking though.

I wonder if we want to just refactor `<ListGroupItem>` the same way we're planning on doing for `<Panel>`, &c.? cc @jquense

TheIronDeveloper:
@taion So for `<ListGroupItem>text content</ListGroupItem>` scenario, everything should be the same.

Things potentially break if the there are elements as children:

```javascript
<ListGroupItem header=""><div>text content</div></ListGroupItem>
```

I'd argue that the above line would have been broken beforehand too, but the following should have continue working:

```javascript
<ListGroupItem header=""><span>text content</span></ListGroupItem>
```

---

I could change things so that output looks similar:

```html
<li class="list-group-item">
    <div class="list-group-item-heading">Header 5</div>
    <div class="list-group-item-text">
        <div>Item 5</div>
    </div>
</li>
```

Which, would be *less* breaking, but can still provide unexpected results converting `<p>` => `<div>`

taion:
With respect to a rewrite, I'm talking about something like https://github.com/react-bootstrap/react-bootstrap/issues/983, consistent with which the API might be:

```js
// http://getbootstrap.com/components/#list-group-basic
const basicListGroupItem = (
  <ListGroup.Item>
    Cras justo odio
  </ListGroup.Item>
);

// http://getbootstrap.com/components/#list-group-custom-content
const customListGroupItem = (
  <ListGroup.Item>
    <ListGroup.Item.Heading>
      List group item heading
    </ListGroup.Item.Heading>
    <ListGroup.Item.Text>
      Donec id elit non mi porta gravida at eget metus. Maecenas sed diam eget risus varius blandit.
    </ListGroup.Item.Text>
  </ListGroup.Item>
);

const moreCustomListGroupItem = (
  <ListGroup.Item>
    <ListGroup.Item.Heading>
      List group item heading
    </ListGroup.Item.Heading>
    <WhateverComponent />
  </ListGroup.Item>
);
```

taion:
Given that, my concern is that we don't want to release two breaking changes to list group items.

TheIronDeveloper:
@taion so you're talking about ditching the `title` attribute and simply rendering `this.props.children`?  I'm cool with trying to build that out.

--

As for this PR, should I continue trying to make it less breaking?  I can inversely render out `this.props.children` iff `this.props.children` is of the type div.

TheIronDeveloper:
Closing this pr and will be working on https://github.com/react-bootstrap/react-bootstrap/pull/1760 instead.

Issue #1645, opened 2016-02-08
Add 'collapsed' class on NavbarToggle


Liz303:
Bootstrap has a `collapsed` class that is removed from the `navbar-toggle` button when the `navbar-collapse` div is opened. In this way you can customize the navbar toggle styling when the `navbar-collapse` is opened ( specifically I want to  turn the spans inside the navbar-toggle from a hamburger to an x when the nav is closed/open). Here's the bootstrap for reference: http://getbootstrap.com/components/#navbar-default. You'll see that when `<div class="navbar-collapse collapse in">` has the class `in` then `<button type="button" class="navbar-toggle" data-toggle="collapse" data-target="#bs-example-navbar-collapse-1" aria-expanded="true"> ` doesn't have the class `collapsed` which is has by default.

olalonde:
+1

olalonde:
Think it's a dupe of #1301

taion:
It's not a dupe of #1301, BTW.

mikatuo:
I believe this was resolved as of #1733

PR #1733, opened 2016-03-30
Add 'collapsed' class on NavbarToggle

hhtran:
Added the 'collapsed' class/state to the NavbarToggle component by default. Removes and adds the 'collapsed' class according the NavbarToggle's collapse state to match the default bootstrap behavior as requested in Issue #1645 .

This is my first time trying to contribute to open source. Please let me know if I'm doing anything wrong/my style is incorrect!

### DIFF
--- a/src/NavbarToggle.js
+++ b/src/NavbarToggle.js
@@ -1,6 +1,5 @@
 import React, { PropTypes } from 'react';
 import tbsUtils from './utils/bootstrapUtils';
-import createChainedFunction from './utils/createChainedFunction';
 import classNames from 'classnames';
 
 
@@ -29,7 +28,7 @@ let NavbarToggle = React.createClass({
 
     return (
       <button type="button"
-        onClick={createChainedFunction(this.handleCollapsedState, onToggle)}
+        onClick={onToggle}
         className={ classNames(tbsUtils.prefix({ bsClass }, 'toggle'), {'collapsed': !expanded}) }
       >
         { children || [
--- a/src/NavbarToggle.js
+++ b/src/NavbarToggle.js
@@ -19,25 +19,18 @@ let NavbarToggle = React.createClass({
     $bs_navbar_expanded: PropTypes.bool
   },
 
-  getInitialState() {
-    return {collapsed: true};
-  },
-
-  handleCollapsedState() {
-    this.setState({collapsed: !this.state.collapsed});
-  },
-
   render() {
     let { children, ...props } = this.props;
     let {
       $bs_navbar_bsClass: bsClass = 'navbar',
-      $bs_navbar_onToggle: onToggle
+      $bs_navbar_onToggle: onToggle,
+      $bs_navbar_expanded: expanded
     } = this.context;
 
     return (
       <button type="button"
         onClick={createChainedFunction(this.handleCollapsedState, onToggle)}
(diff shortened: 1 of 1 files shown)


src/NavbarToggle.js
```diff
@@ -15,6 +18,14 @@ let NavbarToggle = React.createClass({
     $bs_navbar_onToggle: PropTypes.func,
   },
 
+  getInitialState() {
+    return {collapsed: true};
+  },
+
+  handleCollapsedState() {
+    this.setState({collapsed: !this.state.collapsed});
```
jquense: so the potential problem with this approach is that we are tracking the collapsed state in two places now, this Toggle and the parent Navbar, they could get out of sync, or the API change, etc.
I'd be better, since the Navbar is already exposing `expanded` on context to just use that value instead of the additional state logic here.

src/NavbarToggle.js
```diff
@@ -15,6 +18,14 @@ let NavbarToggle = React.createClass({
     $bs_navbar_onToggle: PropTypes.func,
```
jquense: adding 
```js
$bs_navbar_expanded: PropTypes.bool
```
here will also give you access to the current expanded state of the parent navbar.

hhtran:
@jquense I did not know that this was being exposed in the context before! I fixed it in the above commits to use only the context to determine whether or not to have the 'collapsed' class.

jquense:
Thanks for the update. It looks good.

the only thing i'm a bit concerned about here is that `expanded` is toggled at the start of collapse animations, and not at the end of them. but It looks like upstream bootstrap also works like that so no problem.

Only other thing (while we are working here) is that I think the toggle also needs an `aria-expanded` attribute for accessibility. @taion upstream does this, but also adds `aria-controls` I am not sure the `aria-expanded` makes any sense to a SR with out the `controls`...should we try to solve that?

src/NavbarToggle.js
```diff
@@ -13,19 +16,21 @@ let NavbarToggle = React.createClass({
...
-      $bs_navbar_onToggle: onToggle
+      $bs_navbar_onToggle: onToggle,
+      $bs_navbar_expanded: expanded
     } = this.context;
 
     return (
       <button type="button"
-        onClick={onToggle}
-        className={tbsUtils.prefix({ bsClass }, 'toggle')}
+        onClick={createChainedFunction(this.handleCollapsedState, onToggle)}
```
taion: `this.handledCollapsedState` no longer exists, does it?

hhtran: That is true. I thought I had returned that line to the original when I removed the redundant state lines. Will remove now

jquense:
This LGTM, can you rebase against the most recent master to resolve any conflicts and force push @hhtran?

@react-bootstrap/collaborators need one more sign off, I think we can address potential a11y fixes in a different PR.

taion:
LGTM

taion:
Merging by hand.

taion:
Merged as https://github.com/react-bootstrap/react-bootstrap/commit/74d93a5c83154e3514b97bc2e1cdce62147268a0. Added a quick test case in https://github.com/react-bootstrap/react-bootstrap/commit/00ad64225473417cd8990796bba07c4db656dc87.

PR #1734, opened 2016-03-30
introduced new component Carousel.Caption

abloomston:
Carousel.Caption replaces the following bootstrap code:

```<div className="carousel-caption">```

cf. Media.Body from https://github.com/react-bootstrap/react-bootstrap/pull/1707 for a similar change.

### CODE REVIEW

**Comment by abloomston** [comment]
If you are including Carousel as follows before this change:

```const Carousel = require('../../src/Carousel');```

You will receive the following error when e.g. using webpack:

```Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.```

To fix this, change the require to this:

```const Carousel = require('../../src/Carousel').default;```

File: src/Carousel.js [hunk 1]
```diff
@@ -301,4 +303,8 @@ const Carousel = React.createClass({
   }
 });
 
+Carousel = Object.assign(Carousel, { Caption });
+
 export default Carousel;
+
+export { Caption };
```
jquense: this amounts to a breaking change, so i'd either need to go in the 0.29 branch or (preferablly) just remove the named export since the Static property on Carousel is fine

abloomston: The downside is a breaking change, the upside is consistency with e.g. Media.

That being said, I defer to you. Let me know which you'd like me to do, PR to 0.29 branch or modify the code.

taion: That's a mistake on `<Media>`. We'll fix that.

abloomston: Fixed

taion:
LGTM

jquense:
LGTM

taion:
I wonder if we should have `<Carousel.Item>`.

taion:
https://github.com/react-bootstrap/react-bootstrap/pull/1740

PR #1736, opened 2016-03-30
introduced new component Clearfix

abloomston:
Clearfix replaces the following bootstrap code:

```<div className="clearfix">```

### CODE REVIEW

File: src/Tabs.js [hunk 1]
```diff
@@ -1,6 +1,6 @@
-import classNames from 'classnames';
```
abloomston: I'm not sure you'll want fd2148942e73c9b7307871d0badae7cec6039bbe, as it is a change to the internals of react-bootstrap that introduces an additional element in the DOM.

Contrast that to d9cf16131c2b82c25abc07e7da93d102a1e54ee7- which is user-facing (where user is the developer using react-bootstrap) with the goal of nudging the user to use ```<Clearfix>``` rather than thinking directly in terms of bootstrap classes.

taion:
This doesn't seem to match the way this is demonstrated upstream. It should be `visible-${size}-block`.

abloomston:
Good point @taion- I had not been following upstream convention when using clearfix.

I've pushed a commit that changes the properties on `Clearfix` from `xyHidden` to `visibleXyBlock`. Let me know if you'd prefer a different naming convention.

src/Tabs.js
taion: I don't think we want this because we'll be dropping support for horizontal tabs in the next release anyway – no reason to make any changes to deprecated code.

taion:
Yes, please drop https://github.com/react-bootstrap/react-bootstrap/pull/1736/commits/fd2148942e73c9b7307871d0badae7cec6039bbe for now – that code path is deprecated anyway, so we'd rather not change it.

taion:
Err, I mean, will soon be deprecated.

taion:
LGTM

taion:
Sorry, merge conflict.

abloomston:
Rebased, but requires a bit more testing with the visibleXyBlock change, closing PR for now. I will reopen shortly.

taion:
Okay, thank you.

abloomston:
Reopened with fixes. Apologies for that.

Let me know if you'd like me to squash all the commits into one clean commit.

taion:
LGTM

jquense:
Squash and then LGTM

taion:
I am going to try out the new "squash and merge" button.

taion:
Interesting – no merge commit.

krzysiek1507:
@taion https://github.com/blog/2141-squash-your-commits ;)

PR #1740, opened 2016-03-30
Be more consistent with subcomponents

taion:


jquense:
LGTM

taion:
We're not particularly consistent about top-level exports here, though.

jquense:
Yeah at this point maybe we do one big break and remove sub components from the top export?

taion:
I'd be okay with that. It's really weird that some of them are available but some aren't right now.

jquense:
I agree, I take it neither of us import the top level RB ever.

taion:
At some point I'd like to move to top-level imports, once webpack 2 is really out and stable and we have an ES module build.

PR #1742, opened 2016-03-30
Pass onHide to OverlayTrigger overlay

taion:
Recreated from https://github.com/react-bootstrap/react-bootstrap/pull/1288.

I thought it wasn't necessary, but it turns out that https://github.com/react-bootstrap/react-bootstrap/pull/1288#issuecomment-200866021 means it is useful to have this feature.


taion:
On fourth thought, this is still just a bad idea.

jquense:
lol

PR #1743, opened 2016-03-31
Update contributing and maintaining guidelines

taion:
This mostly removes cruft. It'll also be easier to keep around a `next` branch than to keep rolling rc branches.


jquense:
LGTM

PR #1751, opened 2016-03-31
Deprecate interpolation in <ProgressBar label>

taion:
This is a silly feature and we shouldn't have it.


AlexKVal:
Finally it'll be removed :+1:

PR #1754, opened 2016-03-31
Support extra props for Navbar subcomponents

taion:


src/NavbarToggle.js
```diff
@@ -16,24 +20,33 @@ let NavbarToggle = React.createClass({
...
+    const { onClick, className, children, ...props } = this.props;
+    const {
       $bs_navbar_bsClass: bsClass = 'navbar',
       $bs_navbar_onToggle: onToggle
     } = this.context;
 
+    const buttonProps = {
+      type: 'button',
+      ...props,
+      onClick: createChainedFunction(onClick, onToggle),
```
taion: I'm not using either. They're more like "why not".

docs/src/NavMain.js
```diff
@@ -42,9 +41,12 @@ const NavMain = React.createClass({
         role="banner"
       >
         <Navbar.Header>
-          {brand}
+          <Navbar.Brand>
+            <Link to="/">React-Bootstrap</Link>
+          </Navbar.Brand>
+          <Navbar.Toggle />
```
taion: We lost the toggle button here on mobile views somehow.

PR #1756, opened 2016-04-01
Fix remaining onSelect signatures

taion:
- `<Carousel>`
- `<PageItem>`
- `<PaginationButton>`

Also fix examples.


src/Carousel.js
```diff
@@ -280,7 +280,7 @@ const Carousel = React.createClass({
       direction = direction || this.getDirection(previousActiveIndex, index);
 
       if (this.props.onSelect) {
-        this.props.onSelect(index, direction);
+        this.props.onSelect(index, e, direction);
```
taion: This is only compatible with the rest rather than absolutely consistent.

I feel like there ought to be some value in preserving the `direction` argument, though.

Then again, I don't use `<Carousel>`.

jquense: I am wondering if we should make that second argument like an object ball expando, so we can add stuff on to it indefinately?

taion: `event`? I thought about assigning `direction` there, but React actually pools synthetic events, and the pooling code doesn't clear unknown properties, so it'd pollute the object pool in a weird way, which seemed bad.

We could make the third callback argument an object, but I don't think there's much reason to do that – the meaning would be specific to the individual component, and I don't see any use cases other than `direction` for `<Carousel>`.

Alternatively, we could just drop `direction` in the callback altogether.

taion:
Rebased and updated to set `direction` on the event for `<Carousel>` `onSelect`.

jquense:
looks good, only thing I'd add is a doc comment on  the propType explaining the event object addition

taion:
For `<Carousel>` you mean?

taion:
Updated.

PR #1759, opened 2016-04-01
Don't check for prefixed bsStyle in bootstrapUtils

taion:
Plus some code cleanups.

### DIFF
--- a/src/utils/bootstrapUtils.js
+++ b/src/utils/bootstrapUtils.js
@@ -1,7 +1,6 @@
 import { PropTypes } from 'react';
 import styleMaps from '../styleMaps';
 import invariant from 'invariant';
-import warning from 'warning';
 
 function curry(fn) {
   return (...args) => {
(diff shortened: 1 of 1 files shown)


src/utils/bootstrapUtils.js
```diff
@@ -96,45 +100,30 @@ export let bsSizes = curry((sizes, defaultSize, Component) => {
...
-            'bsStyle will automatically prefix custom values with the bsClass, so there is no ' +
-            'need to append it manually. (bsStyle: ' + props.bsStyle + ', bsClass: ' + prefix(props) + ')'
-          );
-          classes[props.bsStyle] = true;
-        } else {
-          classes[prefix(props, props.bsStyle)] = true;
-        }
-      }
+    const classes = {
+      [prefix(props)]: true,
```
taion: `prefix(props)` must return non-falsy; else the `invariant` would fail.

src/utils/bootstrapUtils.js
```diff
@@ -76,15 +78,17 @@ export let bsSizes = curry((sizes, defaultSize, Component) => {
     }
   });
 
-  let values = existing.reduce((result, size) => {
-    if (styleMaps.SIZES[size] && styleMaps.SIZES[size] !== size) {
-      result.push(styleMaps.SIZES[size]);
+  const values = [];
+  existing.forEach(size => {
```
taion: Just do this all in place instead of making a bunch of temporary arrays.

PR #1760, opened 2016-04-01
feat(ListGroupItem): Adding a new API

TheIronDev:
Addressing: https://github.com/react-bootstrap/react-bootstrap/issues/983

This commit adds a non-breaking update to the `<ListGroupItem>`
element.

If the `listItem` propType is added, then use an adjusted react
component that does not inject an `<h4>` header.

Future updates will include adding an `ListGroupItemHeading` (or
`ListGroupItem.Heading`) component that will automatically add the
h4 tag with appropriate classNames.

@taion, let me know if you think https://github.com/react-bootstrap/react-bootstrap/pull/1731 should be merged, but this component will start to deviate from the older api, probably..


src/deprecated/ListGroupItem.js
```diff
@@ -0,0 +1,113 @@
+import React, { cloneElement } from 'react';
```
TheIronDeveloper: This really should be showing `src/ListGroupItem.js` -> `src/deprecated/ListGroupItem.js` :panda_face:

src/ListGroupItem.js
```diff
@@ -1,97 +1,43 @@
-import React, { cloneElement } from 'react';
```
TheIronDeveloper: Probably easier to simply view this file than look at the diff, its now just a simple render function.

taion:
1. We always call this `componentClass`
2. There's no benefit to deprecating all the old code – you already get this behavior if you just don't set `header`

jquense:
My general concern is the value of an incremental change like this, I think we should jsut go for the "correct" API if this is going to get refactored otherwise we are gonna break folks twice.

@TheIronDeveloper would you be up for drafting something like outlined in #983?

TheIronDeveloper:
Yeah I can update this to be closer to the Panel API, and remove the backward compatibility attempt too.

jquense:
Actually for the new new API I think a back compat approach would be good if you can swing it, It seems like the way you are doing it now should work, if you sniff for header props. @taion?

taion:
Yeah – really all we have to do is to add components for the heading and content, add a `componentClass` prop, then deprecate the `header` prop.

The current implementation does the right thing w/r/t `children` if you don't specify `header`.

TheIronDeveloper:
I'm seeing this:

```
FAILED TESTS:
  ✖ "after each" hook: checkNoUnexpectedWarnings for "Should support "header" prop as a string"
    Chrome 49.0.2623 (Mac OS X 10.10.5)
  AssertionError: Warning: "header" property of "ListGroupItem" has been deprecated.
  Use the `ListGroupItemHeading` component 
  In 'ListGroupItem Should support "header" prop as a string'
      at Context.checkNoUnexpectedWarnings (webpack:///./test/index.js?:13:5)
```

I did add a deprecated warning on the header prop.... Any thoughts on what I should to fix this?

TheIronDeveloper:
Alright, @taion / @jquense , I've updated the PR to closer match https://github.com/react-bootstrap/react-bootstrap/issues/983, let me know what you think / what I should change.

Thanks!

jquense:
LGTM, but I am not super familiar with the issues this is fixing...@taion, @apkiernan?

src/ListGroupItem.js
```diff
@@ -1,100 +1,48 @@
...
-    classes.disabled = this.props.disabled;
-
-    if (this.props.href) {
-      return this.renderAnchor(classes);
-    } else if (this.props.onClick) {
-      return this.renderButton(classes);
-    } else if (this.props.listItem) {
-      return this.renderLi(classes);
+    if (header) {
+      return <OldListGroupItem {...this.props} />;
```
taion: There's no reason to deprecate the old list group item; this is strictly additive, and we don't do deprecations except on major bumps.

src/ListGroupItem.js
```diff
@@ -1,100 +1,48 @@
...
-        {this.props.header ? this.renderStructuredContent() : this.props.children}
-      </span>
-    );
-  }
+    let className = classNames(this.props.className, classes);
 
-  renderStructuredContent() {
-    let header;
-    let headingClass = bootstrapUtils.prefix(this.props, 'heading');
+    const componentClasses = {
```
taion: The component class should be the 1st arg to `createElement` directly. See other instances of `componentClass` in our code.

docs/examples/ListGroupDefault.js
```diff
@@ -2,7 +2,9 @@ const listgroupInstance = (
   <ListGroup>
     <ListGroupItem>Item 1</ListGroupItem>
     <ListGroupItem>Item 2</ListGroupItem>
-    <ListGroupItem>...</ListGroupItem>
+    <ListGroupItem itemType="li">3</ListGroupItem>
```
taion: This isn't `itemType` any more, is it?

docs/examples/.eslintrc
```diff
@@ -31,6 +31,7 @@
     "Label",
     "ListGroup",
     "ListGroupItem",
+    "ListGroupItemHeading",
```
taion: There needs to be a `ListGroupItemText` to go with this, no?

TheIronDeveloper: Does there?  The `ListGroupItemHeading` made sense because it wraps the content in an `h4` and adds appropriate classes. Part of the problem before was the ListGroupItem's children were receiving a wrapper span that broke block elements.

docs/examples/ListGroupDefault.js
TheIronDeveloper: Good catch! I'll fix that in just a moment! 👍

docs/examples/.eslintrc
taion: You need it to build a structured list group item.

src/ListGroupItem.js
TheIronDeveloper: Well its going to be a breaking change if the user passes in a header prop... since the magic wrap elements in `h4` and `spans` happens

docs/examples/.eslintrc
TheIronDeveloper: What should happen if other children get passed in?

```
<ListGroupItem>
    <ListGroupItemHeading>Title</ListGroupItemHeading>
    <div>4</div>
</ListGroupItem>
```

src/ListGroupItem.js
TheIronDeveloper: Alright, I updated it to what I *think* is closer to what you were talking about.

docs/examples/.eslintrc
TheIronDeveloper: Added `ListGroupItemText`, so its now up to the user to decide to use `< ListGroupItemText />` or use plain divs.

src/ListGroupItem.js
TheIronDeveloper: So I guess my question, is it alright that I remove the previous `header` prop behavior? It will be a breaking change.

- [ ] Yes (Tyler will update the PR)
- [ ] No (Tyler will update the PR)

:tada:

bpas247:
This PR seems pretty stale... Are there any future plans for it?

TheIronDev:
Wow its been a few years. If I recall I made updates per the comments, the PR didn't gain any traction and I kind of forgot about it. Closing the PR since I lack time to continue the updates (and its super stale by now..)

Issue #1283, opened 2015-09-03
input action popover is covered up


imjared:
https://react-bootstrap.github.io/components.html#input-addons
![image](https://cloud.githubusercontent.com/assets/525011/9667283/eb252598-5248-11e5-9fab-52b50861d810.png)

Looks like if you bump the z-index from 2 to 3 on the rule `.input-group-btn:last-child>.btn, .input-group-btn:last-child>.btn-group` it will be resolved.

screencast: http://cl.ly/192j0B0R181F

taion:
The issue here is actually a bit more subtle than that. This breaks because we nest the menu inside of the `button` element; upstream TWBS doesn't do that, and actually puts the menu next to the `button`.

/cc @mtscout6 @jquense

jquense:
I can't think of a good way to fix this without changing the API to explicitly handle dropdowns like: `dropdownAfter` or something. The issue is that the `input-group-btn` div should be the dropdown parent and not have a nested `btn-group`...

taion:
Oops, miswrote above. @jquense has it right.

taion:
This is fixed with the revamped form control API. See https://github.com/react-bootstrap/react-bootstrap/pull/1797.

PR #1765, opened 2016-04-01
Rewrite form and form control API

taion:


src/Radio.js
```diff
@@ -0,0 +1,59 @@
...
+    if (inline) {
+      const classes = {
+        [bootstrapUtils.prefix(this.props, 'inline')]: true,
+        disabled,
+      };
+
+      return (
+        <label className={classNames(className, classes)} style={style}>
+          <input {...props} type="radio" disabled={disabled} />
+          {' '}
```
jquense: whats with the extra `&nbsp` ?

src/Form.js
```diff
@@ -0,0 +1,38 @@
...
+    const classes = [];
+    if (horizontal) {
+      classes.push(bootstrapUtils.prefix(this.props, 'horizontal'));
+    }
+    if (inline) {
+      classes.push(bootstrapUtils.prefix(this.props, 'inline'));
+    }
+
+    return (
+      <form {...props} className={classNames(className, classes)} />
```
jquense: I mean besides the obvious semantic reason, here isn't any need for this to be an actual `<form>` component, perhaps support `componentClass`? I find I mix and match inline and normal "forms" often enough that I wouldn't mind it, but maybe that's too ugly?

src/Checkbox.js
```diff
@@ -0,0 +1,59 @@
...
+  disabled: false,
+};
+
+class Checkbox extends React.Component {
+  render() {
+    const {
+      inline, disabled, className, style, children, ...props,
+    } = this.props;
+
+    delete props.bsClass;
```
jquense: why delete this? Seems like you do it a bunch, it makes sense for components with custom componentClass's but is there another reason I'm missing?

taion: I don't want to pass it down to the created component, because it's consumed/used by this component for `bootstrapUtils.getClassSet`. I'd pull it out of `this.prop`, but that gives me a variable shadowing error.

src/Form.js
taion: Oh, I didn't think of that. Makes sense to add `componentClass` here.

src/Radio.js
taion: Didn't realize it was unneeded here. Will remove, thanks.

taion:
This looks a bit awkward:

```js
<Form componentClass="fieldset" inline>
```

If this bothers me enough, I might add e.g. `<FieldSet inline>` or something.

This is mostly feature-complete now on the API side. I don't have support for `.input-sm` or `.input-lg`, but that can be replicated with `bsSize` on `<FormGroup>` or `<InputGroup>`.

Only real missing thing is validation state support on `<Checkbox>` and `<Radio>`.

Thoughts? I think the API is clearer, but in more complex cases (e.g. the horizontal form with input group and validation), it's a lot more verbose.

jquense:
I'll take a look at the validation stuff, but in the past I haven't used it at all since it's annoying to have the class on the form-group vs the input... tbh I haven't found a good way to handle it with react-formal aside from custom Form-group component... never used the icon stuff tho

src/FormControlFeedback.js
```diff
@@ -0,0 +1,64 @@
...
+  $bs_formGroup: React.PropTypes.object,
+};
+
+class FormControlFeedback extends React.Component {
+  getGlyph(validationState) {
+    switch (validationState) {
+    case 'success': return 'ok';
+    case 'warning': return 'warning-sign';
+    case 'error': return 'remove';
+    default: return null;
```
jquense: I think the icons used are just "suggested" or examples, I don't think there is any reason why they need to be these three. perhaps  `this.props.glyph` should be used used if it exists?

docs/examples/FormInputAddons.js
```diff
@@ -0,0 +1,71 @@
...
+          <Button>Before</Button>
+        </InputGroup.Button>
+        <FormControl type="text" />
+      </InputGroup>
+    </FormGroup>
+    <FormGroup>
+      <InputGroup>
+        <FormControl type="text" />
+        {/* Set z-index here to work around some layout quirks. */}
+        <InputGroup.Button style={{ zIndex: 1000 }}>
```
jquense: the `zIndex` is unfortunate... is there a better API option here other then requiring these nested button groups? `InputGroup.Dropdown`, or sniff the dropdown child and avoid the nesting?

docs/examples/FormHorizontal.js
```diff
@@ -0,0 +1,37 @@
...
+      <Col componentClass={ControlLabel} xs={2}>
+        Email
+      </Col>
+      <Col xs={10}>
+        <FormControl type="email" placeholder="Email" />
+      </Col>
+    </FormGroup>
+
+    <FormGroup controlId="formHorizontalPassword">
+      <Col componentClass={ControlLabel} xs={2}>
```
jquense: this does kinda suck :/ I wish we had gone with an api for Col more like Collapse, where you had to provide a single child, that it cloned with the right classes...that's a whole other can of worms tho.

Perhaps it makes sense allow specifying Col props on the control-label since its such a common use case? I am not sure that isn't jsut confusing tho.

src/FormGroup.js
```diff
@@ -1,44 +1,107 @@
...
+   */
+  controlId: React.PropTypes.string,
+  /**
+   * @private
+   */
+  standalone: deprecated(
+    React.PropTypes.bool,
+    'Use a `<FormControl>` or `<InputGroup>` directly.'
+  ),
+  validationState: React.PropTypes.oneOf(['success', 'warning', 'error']),
```
jquense: does `bsStyle` make more sense as the prop name? that's in keeping with Button anyway...I guess `error != danger` though.

src/FormGroup.js
```diff
@@ -1,44 +1,107 @@
...
+  /**
+   * @private
+   */
+  bsStyle: deprecated(
```
jquense: ha, oops missed that. rational tho? just semantic clarity?

docs/examples/FormInputAddons.js
taion: This is just https://github.com/react-bootstrap/react-bootstrap/issues/1283. I thought it was pretty gnarly to deal with properly. This is at least a workaround (that's now possible – previously it wasn't). Maybe I'll look at this a bit more later.

docs/examples/FormHorizontal.js
taion: I'm okay with this API I think. Doesn't this come up just as much for building e.g. left tabs? You could write this as

```js
<Col xs={2}>
  <ControlLabel>Password</ControlLabel>
</Col>
```

which just adds an extra `<div>` that doesn't do any harm; this is just closer to the upstream TWBS example.

src/FormGroup.js
taion: Yeah, exactly. The class is `.has-error` instead of `.form-group-error`, so it feels semantically different from e.g. `.btn-danger` or whatever. I'm okay with saying that validation state and styling are somewhat orthogonal.

src/FormControlFeedback.js
taion: Do you think that's necessary if we already have support for doing a fully custom feedback thing? Like

```js
<FormControl.Feedback>
  <Glyphicon glyph="music" />
</FormControl.Feedback>
```

from the examples?

I feel like I don't want to have two APIs for custom feedback icons, and it seems like it'd be better to only have the more general one.

jquense: aww yeah that's fine, sorry I missed the code branch for when there are children.

taion:
This is ready to go now.

I also broke `ValidComponentChildren`, but only a little, and only to bring it more in line with `React.Children` and `Array.prototype` (`numberOf` -> `count`, `findValidComponents` -> `filter`).

src/FormControl.js
```diff
@@ -0,0 +1,70 @@
...
+class FormControl extends React.Component {
+  getControlId(formGroup) {
+    return formGroup && formGroup.controlId;
+  }
+
+  render() {
+    const {
+      componentClass: Component,
+      type,
+      id = this.getControlId(this.context.$bs_formGroup),
```
jquense: if `controlID` exists does it makes more sense to use it over the specified `props.id` so that there isn't a potential label with an `htmlFor` pointing to nothing by mistake?

Or maybe just warn?

src/Radio.js
```diff
@@ -0,0 +1,70 @@
...
+
+    if (inline) {
+      const classes = {
+        [bootstrapUtils.prefix(this.props, 'inline')]: true,
+        disabled,
+      };
+
+      warning(
+        !validationState,
+        '`validationState` is invalid on `<Radio inline>`.'
```
jquense: Maybe the warning should suggest wrapping inline checkboxes in a FormGrop with validation state instead?

jquense:
LGTM, up to you if you want to take any of my comments, they are super minor. 

I do want to make sure we loop back around on that annoying Dropdown addon issue at some point, but there is no need to solve it in this PR

src/FormControl.js
taion: A warning makes sense to me. I'll add one.

src/Radio.js
taion: Makes sense. Will do.

taion:
@jquense Addressed the PR feedback.

jquense:
:shipit:

Anahkiasen:
I like new API, it's powerful and flexible and allows a lot of edge cases, but is deprecating the `Input` API necessary? It offers simplicity and ease of use for most use cases, in particular face to the new API which is _so verbose_ in comparaison:

Before
```js
<Input type="text" label="Username" labelClassName="col-lg-2" wrapperClassName="col-lg-10" addonAfter="foo" />
```

After
```js
<FormGroup>
    <Col lg={2}>
        <ControlLabel>Username</ControlLabel>
    </Col>
    <Col lg={10}>
        <InputGroup>
            <FormControl type="text" />
            <InputGroup.Addon>
                foo
            </InputGroup.Addon>
        </InputGroup>
    </Col>
</FormGroup>
```

I need 13 lines to express what I could do in one, on long forms that quickly accumulates to gigantic components. Again, I like the new API, and I'll probably need it at times, but can't the old one be kept as a shortcut to it, like it's the case for some other components (Dropdown IIRC)?

taion:
@Anahkiasen The idea is that you wrap this up into your own "field group" components, as relevant for your forms. The problem is that `<Input>` is too prescriptive, and was awkward for exactly your case. I'd wrap up what you have into something like `<FieldGroup>`.

Anahkiasen:
Yes that's already what I ended up doing actually, and seen in that light it makes a lot more sense. Guess it just takes some getting used to – it's just the initial fear of the massive rewrite that kind of panicked me.

taion:
We should probably document that part better.

patsissons:
How does one apply validation to a _stand alone_ `FormControl` outside of a `FormGroup` with these changes?

taion:
Bootstrap doesn't have validation states for standalone form controls. They only apply to groups. You've never been able to apply validation states to standalone form controls.

patsissons:
I guess i was just hacking it together before, i had previously just done something like `<Input standalone bsStyle='error' ... />` to accomplish this. It's not a big deal to transition, just means a bit more verbosity.

taion:
If we're missing something I'd like to know it, but there is no e.g. `.form-control.has-success` styling in Bootstrap.

patsissons:
I don't think you're missing something, I'm pretty sure i was just using a non-standard method of input validation.

idolize:
I totally agree the new API is nice and more flexible, but it'd be nice to have a bit more of an upgrade guide/path for large codebases with a lot of `<Input>` elements. Right now I'd like to essentially make my own `<Input>` wrapper that (mostly) replicates the old behavior (I know `Input` still exists, but it's on the chopping block).

taion:
@idolize We need to write this up, but the recommended upgrade path is to write your own `<Input>`. The idea behind this API is that it offers a more flexible low-level abstraction to build the _actual_ field group the user wants.

idolize:
@taion Totally get that, and I agree it's a better API. I just think the upgrade path for existing codebases using `Input` is a little difficult right now, and the changelog only really points to a PR. When I go through the upgrade process I can write up a more in-depth guide we can add to CHANGELOG or wherever if it'd be helpful.

taion:
Might work in the component docs too – more visible that way.

zerkms:
The `<Input>` component deprecation looks like a step back: the releases prior to v0.29 encapsulated the dirty details about bootstrap, which was great. You take a library to do exactly that: to get a generic components that render into complicated markup.

Now everything is exposed back: now a developer needs to care about form groups again, about separated `ControlLabel`s and so on.

taion:
The idea is that you wrap that up yourself – the old API was awful for things like horizontal and inline form groups.

frontsideair:
Any suggestions on migrating a codebase with a lot of `<Input>`s from 0.28.* to 0.29.*?

taion:
Write our own equivalent of the old `<Input>`, tailored to your needs.

Bazze:
People mention writing your own wrapper class since updating a large code base will take a lot of time, however I haven't seen any actual implementations. I thought I'd share my drop-in replacement which works fine for all the usages of Input in my app: https://gist.github.com/Bazze/81cf778d8843948977421bea5c8cdfee

taion:
We have a minimal example in http://react-bootstrap.github.io/components.html#forms-controls as well.

Point being that (1) this is very little code and (2) it's hard to make it properly one-size-fits-all without making it have like two dozen props like the old one, so we're going to hold off on trying to ship anything like that for now.

Sigfried:
Sorry to be dense. In @taion 's reference to http://react-bootstrap.github.io/components.html#forms-controls above, the radio buttons in the doc example don't act like a radio group, more than one can be selected at a time. So, for people like me having a hard time following the whole conversation, is that because we have to implement only-one-at-a-time group behavior ourselves, or is it an oversight in the doc example? Thanks.

Sigfried:
The answer, I believe, is here: http://stackoverflow.com/questions/27784212/how-to-use-radio-buttons-in-reactjs

jchapple:
Thanks @Bazze for posting your `<Input />` drop-in. ❤️ 

I slightly updated it to 'iron out' a few quirks and add a few missing props that existed previously. 

Here's the gist: [`<Input />` drop-in](https://gist.github.com/jchapple/fa054b9c9e372a458b31d94b317bfebe).
_(Note: Depends on [`classnames`](https://www.npmjs.com/package/classnames)) package._

PR #1769, opened 2016-04-03
Rework Panel API

jquense:
First pass at fixing #983. 

I started with the most verbose version of the API. I really hope we can offer something closer to the current API for simple cases, because this is _really_ verbose. In particular the accordion code feels silly, verbose.

Things I don't like about this currently:
- needs a less verbose option for common use cases, I want to keep some of the body inference stuff, with the option to drop down to the verbose one.
- A11y is ugly again, essentially copies the TabContainer
- `panel-title` feels dumb, but an extra component feels worse. If there was anyway to guess the right component to use I would just use that...perhaps we just default it to a div and add titleComponentClass?


taion:
Worth noting, BTW, that `.panel` is dead in Bootstrap 4.

jquense:
true tho Card is philosophically similar...I am hoping it lends itself to a better API more tho

taion:
Can we do something with `bsRole` here?

I'm thinking we have `bsRole` of `heading`, `body`, `footer`.

If no `bsRole` is set for anything, automatically wrap in `Panel.Body`. If a user doesn't want this behavior, he or she can just set his or her own `<CustomPanelBody bsRole="body">` and do whatever in there.

For a `<Panel collapse>`, automatically wrap the element with `bsRole="body"` in `<Panel.Collapse>`.

As for the heading API, maybe `<Panel.Title toggle>` could automatically set up the `Panel.Toggle`? Then for users who want the entire heading to toggle can do a `<Panel.Heading toggle>`. For the accordion case, it'd maybe look like:

```js
<Panel collapse>
  <Panel.Heading>
    <Panel.Title toggle>
      Panel title
    </Panel.Title>
  </Panel.Heading>
  <Panel.Body>
    Panel body
  </Panel.Body>
</Panel>
```

jquense:
> I'm thinking we have bsRole of heading, body, footer.

that seems like a good idea, though we still need a way to say "don't wrap this table or ListGroup", maybe just have them add `bsRole='body'? am I over thinking that?`

> As for the heading API, maybe <Panel.Title toggle> could automatically set up the Panel.Toggle?

I like that, however there is no good default component for `Panel.Title` since the `<hX>` ones are so semantic...maybe just `<header>`?

taion:
I do think

>If no `bsRole` is set for anything, automatically wrap in `Panel.Body`. If a user doesn't want this behavior, he or she can just set his or her own `<CustomPanelBody bsRole="body">` and do whatever in there.

is reasonable and will probably be good enough in practice.

That title thing is messy, I dunno.

jquense:
how do we feel about this APi as it stands? the Title thing is a bit annoying but seems like defaulting to a `<div>` is the least problematic approach. 

I will fix up docs and examples if ya'll feel good about it.

ctxhou:
The new panel feature will add at next version?

I like this new customized panel title and really need it! 😄

timkendrick:
Really nice work on this PR! I've been banging my head against the wall trying to inject an icon and some controls into accordion panel headers, this makes it a piece of cake.

As somebody brand new to react-bootstrap, I personally don't find the API overly verbose – everything does exactly what you'd expect, especially if you're coming from the HTML version of Bootstrap.

Do you know if this is going to make it into master?

BowlingX:
What's the status for this pull-request? Is there any help needed? Right now you can't customize the toggle element in the header and this looks quite complete :).

jquense:
I should rebase this...

taion:
:+1:

Siyfion:
@jquense & @taion It's a shame that the Bootstrap API is more flexible than `react-bootstrap` currently. This is awesome, exactly what is needed in my mind. Any idea of if/when this is likely to get merged in?

zelias500:
any update on this?

jquense:
Anyone want to fix the tests?

Siyfion:
Wow, good work @jquense but I'm afraid I'm not going to be much help, I don't understand it all well enough. All I know is that I'd like to be able to style my panel headers to make them clickable in their entirety rather than just the heading titles!

chilijung:
@jquense I've been working on the tests! https://github.com/react-bootstrap/react-bootstrap/pull/2217

Extra-lightwill:
What's the progress on this?!

Laggii:
Same question, any updates? I see the panel branch has updated docs just the tests are failing.

Siyfion:
@jquense Any idea if/when this is going to get merged, the underlying issue is responsible for about 90% of my UI "bugs" list at the moment, it's getting annoying! :/

jquense:
@Siyfion it will, thanks for the ping on this.

Siyfion:
@jquense Cheers, like I said, I think this is really my only "main" issue with React-Bootstrap at the moment, so it'd be great to get it all merged in. 👍

jquense:
this should be goood to go for review @react-bootstrap/collaborators .

Me and @taion don't actually use the Panel components often (one reason why this is been so slow) so any other input is greatly appreciated on the API and on the code quality. @Siyfion if you want to jump in as well and make sure it addresses your use cases and bugs :+1:

ramarivera:
Awesome!! 
Can´t wait for this making it to master!

src/NavDropdown.js
```diff
@@ -12,6 +12,8 @@ const propTypes = {
   title: React.PropTypes.node.isRequired,
   noCaret: React.PropTypes.bool,
   active: React.PropTypes.bool,
+  activeKey: React.PropTypes.any,
```
taion: unrelated?

docs/examples/PanelCollapsible.js
```diff
@@ -12,7 +12,11 @@ class Example extends React.Component {
         <Button onClick={ ()=> this.setState({ open: !this.state.open })}>
           click
         </Button>
-        <Panel collapsible expanded={this.state.open}>
+        <Panel
```
taion: this example is so weird lol

docs/examples/PanelCollapsible.js
```diff
@@ -12,7 +12,11 @@ class Example extends React.Component {
...
+          collapsible
+          id="collapsible-panel-example"
+          expanded={this.state.open}
+        >
           Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid.
```
taion: it seems like there's a bit of branch-y special-casing just dealing with handling the case of omitting `<Panel.Body>`. do you think it'd be a huge loss just to always require `<Panel.Body>`? modals sort of work in this way of always requiring `<Modal.Body>`, no?

taion:
and then maybe instead of `<Panel.Collapse>`, could we just have `<Panel.Body collapse>`? or would that do something weird for panels that contain e.g. body + other things?

docs/examples/PanelCollapsible.js
jquense: panel-body is different because it adds padding for text content. You can't , for instance, use a table or listgroup inside a panel-body component and have it styled correctly. Tis kinda like Card.Layout vs Card.Content

jquense:
>  or would that do something weird for panels that contain e.g. body + other things?

yeah I think so. My compromise there was having `<Panel collapsible>` which is a bit odd in it's own right

docs/examples/PanelCollapsible.js
jquense: took a quick stab at it. You don't end up saving much unless you also want to get rid of `<Panel collapsible>` as well, b/c you still need to loop through the child to find the non header/footer elements to wrap in PanelCollapse.

This: `<Panel.Body collapse>` is an alternative but it doesn't work with mixed body elements like a table and panel-body. I'm not sure which use-case is more common but the worst case looks like this now:

```
<Panel>
  <Panel.Collapse>
    <Panel.Body>fooobar</Panel.Body>
  </Panel.Collapse> 
<Panel>
```

which sucks

taion: I would keep both `<Panel.Collapse>` and `<Panel.Body>`, but make `<Panel.Body collapse>` just be shorthand for a single body element with `<Panel.Collapse>` around it.

It is having two ways to do the same thing, but I think it's a pragmatic compromise.

jquense: my concern with that is if you have 2+ Bodies it's gonna just awkwardly do the wrong thing.

taion: Well I'd do something like

```js
<Panel>
  <Panel.Body collapsible>
    foo
  </Panel.Body>
</Panel>
```

or

```js
<Panel>
  <Panel.Collapse>
    <Panel.Body>
      foo
    </Panel.Body>
    <Table />
    <Panel.Body>
      bar
    </Panel.Body>
  </Panel.Collapse>
</Panel>
```

It's maybe a little counterintuitive that you can't use two `<Panel.Body collapsible>`, but I don't think it's be the worst thing ever.

jquense:
I remember now why we have the `collapsible` option on Panel. It makes implementing PanelGroup muuuch easier

jquense:
ok Here's where we are:

I agree with @taion that the better API is not having `<Panel collapsible>` and instead requiring Panels to specify a `<Panel.Body collapsible>` component as a shorthand for `Panel.Collapse > Panel.Body`. However, that makes implementation harder in PanelGroup, which passes `collapsible` down to it's children. 

The choice is between convenience API's, upstream bootstrap just makes you specify _everything_ all the time. We need to find the right mix of convenient and flexible, without being confusing. Otherwise the API isn't much better than the current version

jquense:
I feel pretty done with this. Lets just get this in, the new API is better and we will have to revisist this own thing for v4 anyway, perfect being the enemy of good and all that.

docs/examples/PanelCollapsible.js
```diff
@@ -10,11 +10,48 @@ class Example extends React.Component {
...
         </Button>
-        <Panel collapsible expanded={this.state.open}>
-          Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid.
-          Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident.
+        <br />
+        <Panel
+          id="collapsible-panel-example-1"
+          expanded={this.state.open}
+        >
+          <Panel.Collapse>
```
taion: use `<Panel.Body collapsible>`?

docs/examples/PanelGroupControlled.js
```diff
@@ -11,9 +11,28 @@ const ControlledPanelGroup = React.createClass({
...
+        accordion
+        id="accordion-controlled-example"
+        activeKey={this.state.activeKey}
+        onSelect={this.handleSelect}
+      >
+        <Panel eventKey="1">
+          <Panel.Heading>
+            <Panel.Title toggle>Panel heading 1</Panel.Title>
+          </Panel.Heading>
+          <Panel.Body collapsible >
```
taion: extra space

src/Panel.js
```diff
@@ -1,236 +1,143 @@
...
-  onSelect: PropTypes.func,
-  header: PropTypes.node,
-  id: PropTypes.oneOfType([
-    PropTypes.string, PropTypes.number,
-  ]),
-  footer: PropTypes.node,
-  defaultExpanded: PropTypes.bool,
+  /**
+   * Controls the collapsed/expanded state ofthe Panel. Requires a `Panel.Collapse` or `<Panel.Body collapsible>` child component
+   * in order to actually animate out or in,
```
taion: this line ends with a comma?

src/Panel.js
```diff
@@ -1,236 +1,143 @@
...
+  })
+};
 
-    if (e.selected) {
-      this.setState({ expanded: !this.state.expanded });
-    }
+class Panel extends React.Component {
+  constructor(...args) {
+    super(...args);
+    this.handleToggle = this.handleToggle.bind(this);
```
taion: we should probably enable arrow functions one of these days :p

src/Panel.js
```diff
@@ -1,236 +1,143 @@
...
-      if (!React.isValidElement(header)) {
-        return header;
-      }
-
-      return cloneElement(header, {
-        className: classNames(header.props.className, titleClassName),
+    let getIds = null;
+    if (getId || id) {
+      getIds = () => ({
+        headingId: getId(eventKey || id, 'HEADING'),
```
taion: is it normal to use all caps for the generated type here?

src/Panel.js
```diff
@@ -1,236 +1,143 @@
...
-    if (!collapsible) {
-      if (!React.isValidElement(header)) {
-        return header;
-      }
-
-      return cloneElement(header, {
-        className: classNames(header.props.className, titleClassName),
+    let getIds = null;
+    if (getId || id) {
+      getIds = () => ({
```
taion: why is this a thunk?

src/Panel.js
```diff
@@ -1,236 +1,143 @@
...
-          {bodyChildren}
-        </div>
-      );
+  handleToggle(e) {
+    const { onToggle } = this.context.$bs_panelGroup || {};
+    const expanded = !this.getExpanded();
 
-      bodyChildren = [];
+    this.props.onToggle(expanded, e);
+    if (onToggle) {
```
taion: should we check for prevented default here?

src/PanelBody.js
```diff
@@ -0,0 +1,61 @@
...
+    let body = (
+      <div
+        {...elementProps}
+        className={cn(className, prefix(bsProps, 'body'))}
+      >
+        { children }
+      </div>
+    );
+
+    if (collapsible) {
```
taion: yay

src/utils/bootstrapUtils.js
```diff
@@ -15,12 +15,18 @@ function curry(fn) {
   };
 }
 
-export function prefix(props, variant) {
+export function prefix(props, context = {}, variant) {
```
taion: the default value for `context` can never be relevant here, can it?

src/Panel.js
```diff
@@ -1,236 +1,143 @@
...
-    this.handleClickTitle = this.handleClickTitle.bind(this);
-
-    this.state = {
-      expanded: this.props.defaultExpanded,
-    };
-  }
+const contextTypes = {
+  $bs_panelGroup: PropTypes.shape({
+    getId: PropTypes.func,
+    activeKey: PropTypes.any,
```
taion: do you generically prefer this context-based pattern now for passing through active state? should we be using this for `<Nav>` instead of having `<Nav>` clone all its children with an `active` (and `activeKey` and `activeHref`) prop?

aberonni:
Hi, is there any update on this? 

Would be great to get this fixed. I haven't looked at the source code of this project before but I'd be happy to do so if I can help speed this up.

jquense:
just need someone to review it

src/Panel.js
jquense: avant garde

src/utils/bootstrapUtils.js
jquense: i don't know its been a year since i wrote that :P

src/Panel.js
jquense: I don't think so this is controllable so you can just not flush the value as the preventDefault, as we do elsewhere i believe

src/Panel.js
jquense: i think i did it the tabs as well. it's really just a string constant

src/Panel.js
jquense: I don't generically prefer it, but it fits better in some places. In this case I think the direct children aren't really the consumers its the PanelCollapse/Body (I think?) so it was a bit of a protection against prop passing issues, and avoids the Panel needing to sniff out types of it's children.

Moreso tho I was thinking about how many issues we get about "broken" things when folks wrap and don't spread through props.

src/Panel.js
jquense: yeah idk, at one point it probably took an argument

src/utils/bootstrapUtils.js
taion: it'd only be relevant if only `props` were passed in, but that would make no sense. i think you can drop the `= {}` here.

src/Panel.js
taion: i think just make it not a thunk then

jquense:
done

docs/examples/PanelCollapsible.js
taion: ping?

docs/examples/PanelGroupControlled.js
taion: ping

test/utils/bootstrapUtilsSpec.js
```diff
@@ -120,9 +120,9 @@ describe('bootstrapUtils', () => {
       shouldWarn('expected one of ["minimal","boss","plaid","tweed"]');
 
       const Component = bsStyles(['minimal', 'boss', 'plaid', 'tweed'], 'plaid')(
-        class extends React.Component {
```
taion: huh?

src/Panel.js
taion: :p

src/PanelHeading.js
```diff
@@ -0,0 +1,52 @@
...
+      </Component>
+    );
+  }
+}
+
+PanelHeading.propTypes = propTypes;
+PanelHeading.defaultProps = defaultProps;
+PanelHeading.contextTypes = contextTypes;
+
+export default bsClass('panel', PanelHeading);
```
taion: arguably this (and the others) should be like `<TabPane>` where the `bsClass` is `tab-pane`

src/utils/bootstrapUtils.js
```diff
@@ -15,12 +15,18 @@ function curry(fn) {
   };
 }
 
-export function prefix(props, variant) {
+export function prefix(props, context, variant) {
```
taion: i don't see our using the 3-arg form of this anywhere any more

src/PanelHeading.js
jquense: lets defer that, we should do do big rethink of this api anyway, and reach some consistent policy for child components

test/utils/bootstrapUtilsSpec.js
jquense: this was a bad refactor somewhere, just fixing it

taion:
I've been swamped. Will try to look over the weekend.

test/helpers.js
```diff
@@ -1,5 +1,11 @@
 import { cloneElement } from 'react';
 import ReactDOM from 'react-dom';
+import tsp from 'teaspoon';
+
+tsp.fn.log = function log() {
```
taion: @jquense Was this addition intentional?

hitchcockwill:
What's the latest on this? Anything I can do to help out here?

shawnbissell:
I know I'm late to the game here ... but this was a nasty breaking change. We have over 300 usages of <Panel> in our app so it looks like we are stuck of version 0.31.5 for the foreseeable future. How are other people upgrading to v0.32.0? Is there a jscodeshift script for this change?

PR #1774, opened 2016-04-04
Reset deprecation warnings after each test

taion:


taion:
I'm going to merge this as a test-only change. Essentially this lets me check the same deprecation in multiple tests; previously if the same deprecation was exercised in two tests, the deduplication would prevent anything but the first from happening.

PR #1775, opened 2016-04-05
Stricter checking for test warnings

taion:


taion:
1. Fail immediately on warnings so we get a useful traceback
2. Require exactly all warnings to be caught by assertions to avoid accidentally missing any

jquense:
:shipit:

PR #1776, opened 2016-04-05
Don't require duplicate shouldWarn declarations

taion:
I thought this was more explicit but actually it's just dumb.


taion:
Because I'm lazy, I also took away the ability to use regexes here, but that ability wasn't particularly important.

PR #1782, opened 2016-04-06
Lazy-mount tab's children

### LINKED ISSUE #129: Support unmountOnExit for tabs
Currently we are adding all [TabbedPanes to the DOM](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/TabbedArea.jsx#L83) regardless of whether they are active or not. If we were to change this to only mount the active child to the DOM (so only render the active child) it would allow the DOM tree to be smaller and child components would have their React lifecycle methods called at more appropriate times.

Is there any reason that i'm missing for doing it way it currently is?

This came about because of this google groups posting: https://groups.google.com/forum/#!msg/reactjs/fl0z5CZg-BY/tpE5s0-4r_kJ

edmofro:
Tab panes now only mount their children after they’ve been activated
Fixes #129 (mostly)

### DIFF
--- a/src/Tab.js
+++ b/src/Tab.js
@@ -34,7 +34,8 @@ const Tab = React.createClass({
   getInitialState() {
     return {
       animateIn: false,
-      animateOut: false
+      animateOut: false,
+      hasBeenActivated: this.props.active
     };
   },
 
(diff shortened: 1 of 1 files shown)


taion:
1. This shouldn't be the default behavior
2. The entire tab API has been replaced on the `next` branch
3. This is not the correct implementation w/r/t animation anyway – see https://github.com/react-bootstrap/react-bootstrap/issues/129#issuecomment-204030541

PR #1784, opened 2016-04-07
Fix handling of tab props in <Nav>

taion:


src/Nav.js
```diff
@@ -154,6 +156,7 @@ class Nav extends React.Component {
         this.handleTabKeyDown.bind(this, onSelect || (()=>{})),
         onKeyDown
       );
+      tabIndex = isActive ? tabIndex : -1;
```
taion: This is the only part that really matters. Otherwise everything gets a weird `tabIndex`.

src/Nav.js
```diff
@@ -80,20 +80,17 @@ class Nav extends React.Component {
     );
   }
 
-  getActiveKey(props = this.props, context = this.context) {
```
taion: Stuff that got refactored partway through?

Issue #1785, opened 2016-04-07
Modal dialog closes when we drag inside modal and release mouse button on dialog


akiran:
Modal dialog is closing if we drag inside modal and release mouse button on dialog

This bug was also present in bootstrap previously https://github.com/twbs/bootstrap/issues/13816.  I think, it is fixed now

taion:
This was the actual fix in TWBS upstream: https://github.com/twbs/bootstrap/pull/14724

kaiyoma:
Will this be getting fixed anytime soon? I'm running into this issue with an overlay that has a scrollbar, so we hit this problem often when grabbing the scrollbar and releasing the mouse button.

jgrisafe:
Created pull request to fix this, @kaiyoma - https://github.com/react-bootstrap/react-bootstrap/pull/2738

If anyone has any problems using onMouseDown instead of onClick please let me know why. It seems to work fine in my application.

qtbit:
It seems all the sub-issues are closed after a year. Is it fixed to be ready to release?

Rupinderthind:
try 

.modal{
  pointer-events: none;
  .modal-dialog{
    pointer-events: none;
    .modal-content{
      pointer-events: auto;
    }
  }
}

Rupinderthind:
try
`.modal{
  pointer-events: none;
  .modal-dialog{
    pointer-events: none;
    .modal-content{
      pointer-events: auto;
    }
  }
}`

justingolden21:
Coming here from not using Bootstrap and using Svelte, but in case this helps anyone using any CSS framework and any JS framework (since this page ranks high on google)

```
<div class="modal" on:mousedown={close}>
		<div class="modal-content" on:mousedown|stopPropagation>
```

Changing my listener from on click to mousedown solved it for me :)

enesekremergunesh:
I have figure out different way to solve the problem, idk if it will cause a problem later but anyway it works, so basically, I put 'modal-dialog' to another <div> object (I call it 'modal-helper') and then put it to 'modal'. The 'modal-helper' element width and height are inherited (100%) as default but there is small space on top so you can use some margin and padding to close it.

`<div class="modal fade" id="login-modal" tabindex="-1" aria-labelledby="loginModalLabel" style="display: none;" aria-hidden="true">
     <div id="modal-helper" style="pointer-events: auto;">
          <div class="modal-dialog">
               ...
          </div>
     </div>
</div>
`

[...]

PR #1786, opened 2016-04-07
Fix for #1785

akiran:
If we drag on modal and release mouse on dialog, click event is triggered and modal is closed.
This is unexpected behaviour when modal has drag & drop or slider components.

To fix this issue, I change the modal close trigger from click to mouseDown

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -166,7 +166,7 @@ const Modal = React.createClass({
         style={modalStyles}
         className={classNames(className, inClass)}
         dialogClassName={dialogClassName}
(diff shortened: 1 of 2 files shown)


taion:
This code looks fine, but I'll defer to @jquense as to whether these semantics are right.

jquense:
I'm not sure its a good idea, the linked bootstrap issue, did go this route, but quickly reverted it b/c it caused other bugs.

akiran:
@jquense  I will look more closely into final fix that landed in bootstrap and update this PR

taion:
I'm closing this PR for now, as this looks to not be the correct fix here.

Issue #1728, opened 2016-03-30
Add import statement at example code


ev1stensberg:
Add: `import {Button, ButtonToolbar} from 'react-bootstrap;'` in the code snippet. People tend to copy paste and they don't care about what's in the previous section. Also, that is in an own sub-url, which means that people may ignore/oversee it. 

[First Code snippet](https://react-bootstrap.github.io/components.html)

ev1stensberg:
This generally goes for all the sample code

taion:
Not appropriate given that we support multiple import styles.

ev1stensberg:
Maybe just add a comment at top of each snippet? :smile:

Issue #1721, opened 2016-03-25
turn off navbar-default on Navbar Component.


lyquocnam:
i only need to change navbar style, but react-bootstrap only support 2 style: default + inverse
i set bsStyle="home" and receive:

`Warning: Failed propType: Invalid prop `bsStyle` of value `home` supplied to `Uncontrolled(Navbar)`, expected one of ["default","inverse"]. Check the render method of `Component`.`

i think we should remove this warning for free to set custom style.

taion:
You can modify the list of available `bsStyle`s yourself.

lyquocnam:
how can i do that @taion ?

taion:
Use `bootstrapUtils`.

lyquocnam:
what is bootstrapUtils ?

jquense:
It is resolved, we aren't removing the warning because in most cases its helpful to users. If you want to add a custom or additional style to a Component, the way to do it is to use the `bootstrapUtils` methods in `/lib/utils`

lyquocnam:
have any docs or demo @jquense  ?
thanks so much !

jquense:
no docs (PR welcome if you'd like to contribute) but its fairly simple

```js
import Button from 'react-bootstrap/lib/Button'
import bUtils from 'react-bootstrap/lib/utils/bootstrapUtils'

bUtils.addStyle(Button, 'minimal') // renders `btn-minimal` when you do <Button bsStyle='minimal' />
```

lyquocnam:
i tried with navbar:
```
import bUtils from 'react-bootstrap/lib/utils/bootstrapUtils';
bUtils.addStyle(Navbar, 'home');
```
this error:
`Uncaught TypeError: styles.forEach is not a function`
any missing ?

jquense:
oops slight bug, it needs to be an array: `bUtils.addStyle(Navbar, ['home']);`

lyquocnam:
it worked ! thanks @jamesakers awesome !

PR #1788, opened 2016-04-07
bsStyle custom style documentation

abloomston:
Documentation per request in #1721 

* The `require` statement in `ReactPlayground.js` for `bootstrapUtils` needed `.default`; that area of JavaScript is still unfamiliar to me- advice on a better approach is welcome.
* I am not providing an import hint for `bootstrapUtils` per #1728 comment by @taion. I suspect this will be difficult for new users to figure out since the location of bootstrapUtils looks to require a path (e.g. ```import bootstrapUtils from 'react-bootstrap/lib/utils/bootstrapUtils';```). Feedback/suggestions welcome.

### DIFF
--- a/docs/src/ReactPlayground.js
+++ b/docs/src/ReactPlayground.js
@@ -6,6 +6,8 @@ const classNames = require('classnames');
 const React = require('react');
 const ReactDOM = require('react-dom');
 
+const bootstrapUtils = require('../../src/utils/bootstrapUtils.js').default;
+
 const Accordion = require('../../src/Accordion');
 const Alert = require('../../src/Alert');
 const Badge = require('../../src/Badge');
--- a/docs/examples/.eslintrc
+++ b/docs/examples/.eslintrc
@@ -8,6 +8,7 @@
     "React",
     "ReactDOM",
     "classNames",
+    "bootstrapUtils",
     "Accordion",
     "Alert",
     "Badge",
--- a/docs/examples/CustomButtonStyle.js
+++ b/docs/examples/CustomButtonStyle.js
@@ -1,0 +1,15 @@
+bootstrapUtils.addStyle(Button, ['custom']);
+
+const customButtonStyle = (
+  <div>
+    <style type="text/css">{"\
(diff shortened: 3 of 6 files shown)


jquense:
LGTM, @react-bootstrap/collaborators anyone else?

docs/src/ReactPlayground.js
```diff
@@ -6,6 +6,8 @@ const classNames = require('classnames');
 const React = require('react');
 const ReactDOM = require('react-dom');
 
+const bootstrapUtils = require('../../src/utils/bootstrapUtils.js').default;
```
taion: @jquense Should we fix the module exports here to be more CJS friendly?

taion:
LGTM, thanks.

docs/src/ReactPlayground.js
jquense: yeah probably.

taion:
w/r/t bootstrapUtils, my preference would be to just fix the exports there to remove the requirement of `.default` – or be consistent either way.

docs/src/ReactPlayground.js
taion: Okay let's get that fixed before we release this.

taion: How should we handle this? Everything on a default export, or a bunch of named exports? i.e. when using ES module shims, should it be

```js
import { * as bootstrapUtils } from './utils/bootstrapUtils';
```

or

```js
import bootstrapUtils from './utils/bootstrapUtils';
```

abloomston: Hi @taion I noticed this is merged but I did not make any changes to this import statement. Would you like me to submit an additional PR with any changes to it?

taion: No need (:

We need to first figure out how we want to fix the exports from `bootstrapUtils`, since there's two ways to go that produce identical results for CJS users, but different ones for users of polyfilled ES2015 modules.

Thanks for these docs, BTW!

PR #1789, opened 2016-04-07
React 15 compatibility

taion:


taion:
Let's see if this passes CI... (it may well not)

taion:
I'm going to merge this as a trivial dev dep bump, then updating the docs so they're built against the new hotness.

PR #1794, opened 2016-04-08
disable clicks on anchor Buttons

### LINKED ISSUE #1415: How to deal with onClick and disabled
Follow-up from https://github.com/react-bootstrap/react-bootstrap/pull/1413#issuecomment-148876400.

For components that have a `disabled` prop but also accept `onClick`, should we block the `onClick` handler when `disabled` is set? In general we do block `onSelect` handlers.

### LINKED ISSUE #1780: Disabling button should disable onClick handler
Hi. I see the the [doc section](https://react-bootstrap.github.io/components.html#buttons-disabled) on "Event handler functionality not impacted." I humbly propose that 99% of the time if a `Button` is disabled, the developer will also want to disable it's `onClick` handler. So, to better address the common case, `disabled=true` should also disable `onClick`.

It seems odd to have to do the following with Buttons and related classes:
```jsx
<PageItem
  disabled={!hasPreviousPage}
  onClick={!hasPreviousPage ? NO_OP : previousClick}
  previous>
```

Perhaps I'm missing something or there's a previous discussion on this? Neither Google nor GitHub Issues turned up anything.

Thanks. Overall react-bootstrap is excellent.

jquense:
closes #1415

### DIFF
--- a/docs/src/sections/ButtonSection.js
+++ b/docs/src/sections/ButtonSection.js
@@ -33,21 +33,26 @@ export default function ButtonSection() {
       <p>To set a buttons active state simply set the components <code>active</code> prop.</p>
       <ReactPlayground codeText={Samples.ButtonActive} />
 
+      <h3><Anchor id="buttons-tags">Button tags</Anchor></h3>
+      <p>The DOM element tag is choosen automatically for you based on the props you supply. Passing
+        a <code>href</code> will result in the button using a <code>{"<a />"}</code> element otherwise
+        a <code>{"<button />"}</code> element will be used.
+      </p>
+      <ReactPlayground codeText={Samples.ButtonTagTypes} />
+
       <h3><Anchor id="buttons-disabled">Disabled state</Anchor></h3>
       <p>Make buttons look unclickable by fading them back 50%. To do this add the <code>disabled</code> attribute to buttons.</p>
       <ReactPlayground codeText={Samples.ButtonDisabled} />
 
       <div className="bs-callout bs-callout-warning">
-        <h4>Event handler functionality not impacted</h4>
-        <p>This prop will only change the <code>{"<Button />"}</code>&#8217;s appearance, not its
-          functionality. Use custom logic to disable the effect of the <code>onClick</code> handlers.</p>
+        <h4>Cross-Browser compatibility</h4>
+        <p>
+          Because <code>{"<Button />"}</code> will render an <code>{'<a>'}</code> tag in certain situations.
+          Since anchor tags can't be <code>disabled</code>, the behavior is emulated as best it can be.
+          Specifically, <code>pointer-events: none;</code> style is added to the anchor to prevent focusing, which
+          is only supported in newer browser versions.
+        </p>
       </div>
-
-      <h3><Anchor id="buttons-tags">Button tags</Anchor></h3>
-      <p>The DOM element tag is choosen automatically for you based on the props you supply. Passing
-        a <code>href</code> will result in the button using a <code>{"<a />"}</code> element otherwise
-        a <code>{"<button />"}</code> element will be used.</p>
-      <ReactPlayground codeText={Samples.ButtonTagTypes} />
 
       <h3><Anchor id="buttons-loading">Button loading state</Anchor></h3>
       <p>When activating an asynchronous action from a button it is a good UX pattern to give the user
--- a/src/Button.js
+++ b/src/Button.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import classNames from 'classnames';
 import elementType from 'react-prop-types/lib/elementType';
+import SafeAnchor from './SafeAnchor';
 
 const types = ['button', 'reset', 'submit'];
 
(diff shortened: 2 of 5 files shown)


taion:
Can we move this down to `<SafeAnchor>`? See also https://github.com/taion/react-bootstrap/tree/anchorProps – we might want to pass through `disabled` to `<SafeAnchor>` in these cases.

jquense:
sure, tho I have to admit I don't really understand what SafeAnchor is doing?

taion:
I have no idea.

taion:
Pointed at the original reported issue.

jquense:
OK...so It looks like the point is to address exactly these sort of issues where `href`-less anchors are treated like buttons by default?

It seems like it would make sense to put logic there, except that in the case of a disabled `Button` with a `href` you'd want to prevent navigation and clicks whereas `SafeAnchor` only handles the cases where href is `undefined`? Looking at our uses tho it seems like if a disabled was passed to SafeAnchor, not firing anything would probably be the desired behavior

taion:
Yeah, that's the point I'm trying to make. If we make `onClick` on `SafeAnchor` do something like:

```js
onClick = (e) => {
  if (this.props.disabled) {
    e.preventDefault();
    return;
  }

  if (this.props.onClick) {
    this.props.onClick(e);
  }
}
```

jquense:
Just updated with a more robust solution. It may be overkill tho. Also Buttons don't use SafeAnchor hence the new component used by both

src/ButtonAnchor.js
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  handleClick(event) {
+    let { disabled, onClick } = this.props;
+
+    event.preventDefault();
+
+    if (disabled) {
+      event.stopPropagation();
+      return;
```
jquense: the shotgun approach to preventing/hiding click events. In general `pointer-events: none` handles everything, these are in case its not supported or someone (most likely Safari) decides that its better to ignore that css property for links,

taion:
This doesn't actually solve the problem. The issue arises when `href` _is_ set. For e.g. `<Button>`s without an `href`, the element actually is really disabled.

jquense:
right hence the seperate ButtonAnchor component used directly by Button...did I break something in SafeAnchor along the way?

src/ButtonAnchor.js
```diff
@@ -0,0 +1,55 @@
...
+        tabIndex={tabIndex || 0}
+        onClick={this.handleClick}
+      />
+    );
+  }
+
+  handleClick(event) {
+    let { disabled, onClick } = this.props;
+
+    event.preventDefault();
```
taion: You can't unconditionally prevent the default here – this will prevent the normal navigation from working.

src/ButtonAnchor.js
```diff
@@ -0,0 +1,55 @@
...
+export default class ButtonAnchor extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.handleClick = this.handleClick.bind(this);
+  }
+  render() {
+    let { href, role, tabIndex, disabled, style } = this.props;
+
+    if (disabled) {
```
taion: Can't we let the TWBS CSS handle this?

taion:
Oh, I see. Do we actively not want this behavior for other components (for both `onClick` and `href`, I mean).

src/ButtonAnchor.js
jquense: yeah that was the intention, if then link is really a link then it shouldn't use this component.

In other words this is trying to encapsulate "Making a `<button>` but with an `<a>` tag for reasons of browsers being weird"

In the case of SafeAnchor use it will never prevent actual navigation because this only gets rendered when there is no `href`

src/ButtonAnchor.js
jquense: The reason for it is that even tho the css will make it looked disabled it won't prevent focus, for disabled anchors. The `pointer-event: none` will be set on `a.btn.disabled` components by bootstrap css but not for anything else like MenuItems, where we'd also want to keep disabled ones from being navigatable by arrow keys and clicks.

src/ButtonAnchor.js
taion: Ahh I see – right, doesn't hit this code. What do you think of merging this logic into `<SafeAnchor>`? Might be nice to remove one level of nesting for checking things out in devtools.

taion: Wait, no, doesn't this get used whenever we have a `<Button href="whatever" />`?

jquense: bah yeah...so there are two cases to address:

1. links that are really buttons
2. real links that just look like buttons

this addresses 1. but kills 2...

I originally had this in SafeAnchor but Button can't use SafeAnchor...or maybe it can. I'll take another look

taion: Just use SafeAnchor in the render anchor code path. The "has no href" branch will be irrelevant there.

jquense: no yeah SafeAnchor is deciding if a link should act like a button which is based on the `href` whereas Button needs it to always treat it like a button regardless of the href.

let me just fix the navigation thing first and see what makes sense then

taion: I'm clearly a bit confused here but I think that's what we want. If a button has a non-trivial `href`, then per http://getbootstrap.com/css/#callout-buttons-anchor-accessibility it should just be a link with button styling and not otherwise get `role="button"` or whatever.

jquense: yeah. The question for whether it should be inside SafeAnchor comes down to this:

For everything else: NavItems, MenuItems, etc should the disabled logic be the same as `<button>` even when there is an `href`? or should that only be the behavior for all Buttons, and links with no hrefs?

taion: We're obnoxiously inconsistent here.

`<MenuItem>` prevents default whenever `props.disabled` is set: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/MenuItem.js#L17-L19 (not sure what preventing the default accomplishes when there's no `href`, though)

`<NavItem>` unconditionally prevents the default so long as `onSelect` is set: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/NavItem.js#L73-L74

jquense: yeah I think I'm for just letting SafeAnchor handle all that...Since there are no disabled semantics for anchors we might as well just follow the input semantics as best we can and call it a day. I can't think of a use case where you'd want navigation to continue will styling something as "disabled"

taion: Makes sense. I have some code earlier for making link handling be consistent on https://github.com/taion/react-bootstrap/commit/67a45851cb6d472f42996c45e16d403e212c15c0 – I think it'd make more sense to build on top of that.

taion: Although we'd also have to amend that implementation to pass along `disabled` to the `<SafeAnchor>` as well.

jquense:
updated the tests and moved everything to SafeAnchor..I think its not doing the right thing.

We should, as you said also remove that functionality from places where its otherwise specified

taion:
I think this is right – what do you think might be wrong?

jquense:
just a typoe, sorry i DO think its working correctly

taion:
Ah, okay. LGTM then.

PR #1796, opened 2016-04-08
Fix <SubNav> on docs page

taion:


taion:
I'm not 100% sure this is correct, but it seems harmless enough. This was the old behavior and was required for the side menu to work correctly on the components page.

PR #1797, opened 2016-04-08
Fix example of dropdown in input add-on

taion:


taion:
This fixes all the `z-index` crap. I'm taking this as a sign that we made correct API choices in letting us fix this purely on the docs side, without writing new code.

jquense:
:+1:

taion:
image

PR #1801, opened 2016-04-11
Make anchor prop handling consistent

### LINKED ISSUE #629: Specifying props for nested anchor components

taion:
Closes #629

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -47,6 +47,7 @@ const NavItem = React.createClass({
       >
         <SafeAnchor
           {...props}
+          disabled={disabled}
           role={role}
           href={href}
           onClick={createChainedFunction(onClick, this.handleClick)}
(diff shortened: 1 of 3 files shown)


jquense:
this is smart, tho I am a bit concerned about breakage as its gonna be hard to know when folks are relying on the current behavior.

Does it make sense to just warn for a version if props besides `className` and style pass passed to these components? I don't love that since its gonna cause unfixable noise in their console, but I'm a bit worried about subtle breakage :/

taion:
Maybe the `id` prop?

We're unavoidable changing the semantics of `onSelect`, and this is a breaking release anyway, so I'm okay with causing minor breakage.

I don't see a better option here, at least.

PR #1807, opened 2016-04-13
Class names are out of order on Button.js

parterburn:
I should be able to add a class and have those styles override, but instead the default Bootstrap classes are overriding my custom style.


taion:
CSS precedence doesn't work that way. It's class declaration order, not class use order.

parterburn:
@taion my bad. You are 100% correct. Figured out my issue.

PR #1812, opened 2016-04-17
Merge v0.29 changes for release

taion:


taion:
Need to resolve a few more things before cutting a release, but I think this is about right.

PR #1815, opened 2016-04-18
Allow using `.value` property

johnwun:
This more closely mimics the behavior of inputs in the DOM.


taion:
This component is deprecated.

PR #1816, opened 2016-04-18
add getter for value

johnwun:
To more closely mimic browser input elements.


taion:
This component is deprecated. You also shouldn't be doing this.

PR #1823, opened 2016-04-19
Added unmountOnExit support to Tabs

### LINKED ISSUE #129: Support unmountOnExit for tabs

Lavoaster:
Resolves #129 

I didn't add tests for this as there weren't any existing ones for `animate` or any of the new transition ones that I could reference.

### DIFF
--- a/src/TabPane.js
+++ b/src/TabPane.js
@@ -65,7 +65,7 @@ let TabPane = React.createClass({
     onExited: PropTypes.func,
 
     /**
-     * Unmounts the tab (remove it from the DOM) when animation is not `false` and it is faded out
+     * Unmount the tab (remove it from the DOM) when it is no longer visible
      */
     unmountOnExit: PropTypes.bool
   },
@@ -140,10 +140,10 @@ let TabPane = React.createClass({
       : context.animation;
   },
 
-  getUnmountOnExit(props = this.props, context = this.context) {
-    context = this.getContext('$bs_tabcontent', context);
-    return props.unmountOnExit != null
-      ? props.unmountOnExit
+  getUnmountOnExit() {
+    let context = this.getContext('$bs_tabcontent', this.context);
+    return this.props.unmountOnExit != null
+      ? this.props.unmountOnExit
       : context.unmountOnExit;
   },
 
@@ -158,6 +158,10 @@ let TabPane = React.createClass({
     let bsClass = this.props.bsClass || this.getContext('$bs_tabcontent').bsClass;
 
     let Transition = this.getTransition();
+
+    if (!visible && !Transition && this.getUnmountOnExit()) {
+      return null;
+    }
 
     let classes = {
       active: visible,
--- a/src/TabPane.js
+++ b/src/TabPane.js
@@ -62,12 +62,18 @@ let TabPane = React.createClass({
     /**
      * Transition onExited callback when animation is not `false`
      */
-    onExited: PropTypes.func
+    onExited: PropTypes.func,
+
+    /**
+     * Unmounts the tab (remove it from the DOM) when animation is not `false` and it is faded out
+     */
+    unmountOnExit: PropTypes.bool
   },
 
   contextTypes: {
     $bs_tabcontainer: PropTypes.shape({
-      getId: PropTypes.func
+      getId: PropTypes.func,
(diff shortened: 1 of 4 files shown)


jquense:
this looks great, thanks. My only thought is that we should also cover the case where animation is false as well. @taion thoughts? We could either handle that in TabContent, or just render `null` from the TabPane for none active ones...

taion:
I'd do it in `<TabPane>`, since that's what knows about the transition.

I'd also like to see a test just asserting that stuff gets mounted and unmounted appropriately, and maybe an example as well.

taion:
This is exactly the right approach, though.

jquense:
@Lavoaster ya a test would be great, they are missing from RB because the Transition ones are in react-overlays, but something jsut asserting that components are actually not mounted would be great.

Lavoaster:
Ah, I agree, now that RB is going to be handling the unmounting of the component itself it will need a test here.

Lavoaster:
Would there be an expectation of the other animation callbacks to work such as onExit and onEnter?

taion:
@Lavoaster I think that would be ideal. I suppose it's not required, though.

Lavoaster:
@taion Can the fade transition be instantaneous? I'm just wondering if passing all props to it would be better in terms of reducing the code duplication between the two projects.

taion:
We'd rather not have the fade at all if the animation isn't requested, I think, just for consistency with other components.

Lavoaster:
@taion The idea would be that the duration of the animation would be 0 with no animation.

Lavoaster:
I've submitted an update TabPane which wraps everything in a Transition component if there is no Transition specified.

I've spent quite a while trying to get my tests to work when un-mounting the component here but I'm having a bit of trouble doing it. If you would still prefer to implement this in this package then I'll be bowing out from this for now until I'm more familiar with testing in JS.

taion:
I don't think it makes sense to include a transition component when transitions are disabled. Why not just not render the tab when not mounted if not using a transition?

Lavoaster:
It is a transition, a transition doesn't need an animation.

The transition component deals with all of the callbacks such as onExit and onEnter that you only get with the animation, by doing it this way we're not re-implementing the transition component in this component.

taion:
Right, but those don't and shouldn't fire in the case that a transition is not specified, plus the animation semantics aren't relevant.

Lavoaster:
So, I'm going to go ahead and close this. It appears as if I've hit a wall with @taion and his coupling of a transition requiring an animation. I'll maintain a fork with this on instead, feel free to pull in the code yourself from here and add in the extra code and tests required for this.

src/TabPane.js
```diff
@@ -170,6 +184,10 @@ let TabPane = React.createClass({
       labelledBy = getId(eventKey, TAB) || null;
     }
 
+    if (!visible && !Transition && this.getUnmountOnExit()) {
```
taion: Maybe move this up to before line 162? No reason to run through the logic above when it doesn't have any effect.

src/TabPane.js
```diff
@@ -133,6 +140,13 @@ let TabPane = React.createClass({
       : context.animation;
   },
 
+  getUnmountOnExit(props = this.props, context = this.context) {
```
taion: I don't think we need the arguments here? They don't seem to ever be used.

taion:
This is looking great, thanks!

LGTM other than those 2 nits.

src/TabPane.js
Lavoaster: I just took the example from getTransition.

taion: `getTransition` needs it because it gets called from `componentWillUpdate` with `nextProps` and `nextState`. It's not really needed here, though.

PR #1825, opened 2016-04-19
Don't incorrectly pass onSelect past TabContainer

### LINKED ISSUE #1822: Tab collapses on FormControl focus
In my application I have three `Tab` components inside of a `Tabs` component. Each `Tab` has a set of `FormControl` elements. I just upgraded to the latest react-bootstrap and with the previous implementation of the above with `Input` all worked fine.

Now, however, whenever a input field is clicked or otherwise given focus, the whole Tab collapses.
I'm not getting any errors or warnings in the console.

Here's a minimal and verifiable code that I tried on its own, which also produces the above issue:

    <Tabs activeKey={this.state.tabsKey} onSelect={this._handleSelect} animation={false} id="sub-task-tabs">
      <Tab eventKey={1} title="Extern länk">
        <FormGroup>
          <ControlLabel>Titel</ControlLabel>
          <FormControl type="text" onChange={this._subTaskInputChange0} value={this.state.subTaskInput0} />
        </FormGroup>
      </Tab>
    </Tabs>

Here's two screenshots I took.

_Before click on input field (Titel):_

img1

_After:_

img2

taion:
Fixes #1822

I don't have a test case here because I'm not sure I could write one without its just looking sorta dumb.

### DIFF
--- a/src/TabContainer.js
+++ b/src/TabContainer.js
@@ -22,7 +22,7 @@ let TabContainer = React.createClass({
 
         if (!error && !props.id) {
           error = new Error(
-            'In order to properly inialize Tabs in a way that is accessible to assistive technologies ' +
+            'In order to properly initialize Tabs in a way that is accessible to assistive technologies ' +
             '(such as screen readers) an `id` or a `generateChildId` prop to TabContainer is required');
         }
       }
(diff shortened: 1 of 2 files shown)


src/TabContainer.js
```diff
@@ -64,25 +64,25 @@ let TabContainer = React.createClass({
     })
   },
 
-  getDefaultProps() {
-    return {
-      componentClass: 'div'
```
taion: This was not a prop on this component.

taion:
@jquense Does this look okay to you?

Issue #1829, opened 2016-04-20
Accordion triggers a synthetic event warning in React v15


same31:
I copied the following example from the doc: http://react-bootstrap.github.io/components.html#panels-accordion in a fiddle with react v15.0.1 and react-bootstrap v0.29.0 : https://jsfiddle.net/same31/5v143xf3/7/

On the first click on a collapsible panel, this warning is triggered :

> Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're adding a new property in the synthetic event object. The property is never released. See https://fb.me/react-event-pooling for more information.

Is this possible to get rid of this warning ?

Thanks

vincent-dm:
I also have this and additionally no synthetic event is passed to my `onSelect`-handler. Because of this the browser windows scrolls to the top of every page because I cannot call `preventDefault()` on the event...

nickpalmer:
For me this is triggered by the Panel.js handleSelect function.

The function is setting e.selected = true and then checking it down below. I think @vincent-dm may be confused about the API, since the first argument to onSelect is the key and the second is the event.

Presumably e.selected is designed to be used by props.onSelect handlers to allow them to prevent handleToggle from being called. It maybe better to use a return value from onSelect() instead?

```
handleSelect: function handleSelect(e) {
    e.selected = true;

    if (this.props.onSelect) {
      this.props.onSelect(this.props.eventKey, e);
    } else {
      e.preventDefault();
    }

    if (e.selected) {
      this.handleToggle();
    }
  },
```

taion:
We'd accept a PR to fix this. We're looking to replace this full API by the next breaking release, so maintaining the current one is not a high priority.

PR #1864, opened 2016-04-29
Fix issue #1829.

nickpalmer:
So this fixes the issue, but I am not sure how to test it.

If I run ```npm test``` I don't see the react error show up in the console, and I am not sure why.

Suggestions on how to test the warning doesn't happen are welcome.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -46,6 +46,7 @@ let Panel = React.createClass({
   },
 
   handleSelect(e) {
+    e.persist();
     e.selected = true;
 
     if (this.props.onSelect) {


taion:
LGTM

broadhu2:
any estimate on when this change will be available for release?

taion:
This was released like a year and a half ago.

broadhu2:
My mistake, I had searched for the handleSelect() function in newer releases but I think that function name has changed. Can you tell me the earliest version where that fix is available?

taion:
https://github.com/react-bootstrap/react-bootstrap/pull/1864/commits/0f2c3f13f6247512d1c9930a625573ee3db73920 shows the releases that include this commit.

PR #1865, opened 2016-04-30
Add inputRef to <Checkbox> and <Radio>

### LINKED ISSUE #1863: Checkbox getChecked
Hi All,

I can't work out how to get the checked value from a Checkbox component. Is there any way to do this? findDOMNode doesn't seem to help because it doesn't get the input elements.

Thanks

taion:
Closes #1863

### DIFF
--- a/src/Radio.js
+++ b/src/Radio.js
@@ -11,6 +11,14 @@ const propTypes = {
    * Only valid if `inline` is not set.
    */
   validationState: React.PropTypes.oneOf(['success', 'warning', 'error']),
+  /**
+   * Attaches a ref to the `<input>` element. Only functions can be used here.
+   *
+   * ```js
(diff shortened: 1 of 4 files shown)


taion:
@jquense Is this the right way to test stuff with refs using teaspoon?

src/Radio.js
```diff
@@ -11,6 +11,10 @@ const propTypes = {
    * Only valid if `inline` is not set.
    */
   validationState: React.PropTypes.oneOf(['success', 'warning', 'error']),
+  /**
+   * Attaches a ref to the `<input>` element.
```
jquense: this may be confusing to folks maybe a code block with an example?

jquense:
Ya that's how I'd test it. could probably use a helper method for that... 

LGTM, optional on the doc nit

Issue #1866, opened 2016-05-02
Missing eventKey prop for custom ButtonComponentClass in PaginationButton.js (v.0.29.x)


northerneyes:
I need information about which page is rendered. It is impossible to create SEO wise links now.

smashercosmo:
Could someone explain why evenKey property is removed for buttonComponentClass? How can we render links with real urls and not just onSelect callback?

taion:
Go ahead and PR adding it back for now.

smashercosmo:
Cool) Was doubting may be it was done for reason)

taion:
It was elsewhere, but I agree that you need it here.

AndreiRailean:
The code is in with a test showing how to generate urls. Waiting for next release to roll this into my app.

taion:
v0.29.4

PR #1876, opened 2016-05-09
allow anchor renderer access to target page number #1866

AndreiRailean:
Makes it possible to provide an anchor renderer component that makes use of page number information.
Ideally `eventKey` should be renamed to something that makes more sense like `targetPage` to keep it pagination-centric.

https://github.com/react-bootstrap/react-bootstrap/issues/1866

### DIFF
--- a/src/PaginationButton.js
+++ b/src/PaginationButton.js
@@ -48,7 +48,6 @@ const PaginationButton = React.createClass({
     } = this.props;
 
     delete props.onSelect;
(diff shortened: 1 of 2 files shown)


taion:
Can you add a regression test?

AndreiRailean:
added a test to exhibit my requirements for isomorphic rendering

taion:
LGTM

Issue #1894, opened 2016-05-13
Remove onClose on <Dropdown>


bdwain:
the onClose handler for a dropdown only seems to be called when you click outside of the dropdown. not if it closes because you clicked on the toggle.

i tested out something locally that fixes it. i'll open a PR

PR #1895, opened 2016-05-13
call dropdown onClose if the dropdown is closed by clicking the toggle

bdwain:
always call onClose when the dropdown is being closed. not just if it is closed by clicking off of it.

#1894

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -118,6 +118,10 @@ class Dropdown extends React.Component {
     if (this.props.onToggle) {
       this.props.onToggle(open);
     }
+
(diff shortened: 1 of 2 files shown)


jquense:
thanks for the PR...I'm not even sure why we have an `onClose` handler, @taion do you remember? Seems like this is already covered by `onToggle` and there is a duplication.

taion:
I'd be okay with removing it on `<Dropdown>`.

jquense:
gonna close per the above consensus. On toggle is ready doing the right thing and the more correct handler here

PR #1919, opened 2016-05-21
Render children in FormControl.Feedback component

headdetect:
```
<FormControl.Feedback>
   <MyComponent className="oink" />
</FormControl.Feedback>
```

Will produce: `<MyComponent className="form-control-feedback oink" />`

While this is what would be produced if following the exact example from the bootstrap docs... it would be more structurally logical to render it like:

```
<span className="form-control-feedback">
  <MyComponent className="oink" />
</span>

```

Right now, trying to render a `<FormControl.Feedback />` element will work perfectly if only using bootstrap glyphicons. If you use something else like FontAwesome, it will render out like 

```
<span name="check" class="fa fa-check form-control-feedback"></span>
```

which makes everything all funky. 

### Before

| Glyphicon | Other Font Icon |
|------------------|-----------------|-----------------|
| Glyphicon| Glyphicon |

I imagine different types of font/vector icons would have similar problems as well.

### DIFF
--- a/docs/src/sections/FormSection.js
+++ b/docs/src/sections/FormSection.js
@@ -12,7 +12,7 @@ export default function FormSection() {
         <Anchor id="forms">Forms</Anchor> <small>FormGroup, FormControl, ControlLabel</small>
       </h1>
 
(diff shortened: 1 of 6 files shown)


taion:
Well, great, but this breaks it for Glyphicons...

headdetect:
It does? I've tested it on several machines and it looks fine to me. Do you mean it breaks some sort of convention or it visually is unappealing?

image

This is what I get on the docs after making this change.

taion:
I'm closing this out for now given that it affects how the Glyphicons get rendered. I think it should be possible to add CSS rules on your end to render FA icons appropriately.

PR #1920, opened 2016-05-22
Fix tests.

ruiaraujo:
Fixes related with locale and using multiple deprecation libraries (one internally and another from react-prop-types)


taion:
Thanks

PR #1924, opened 2016-05-22
Update lodash

ruiaraujo:
The goal is to reduce bundle size for people who are already consuming lodash in their apps.

As for those who don't, I will push a new PR updating to Babel 6 and https://github.com/lodash/babel-plugin-lodash to reduce the library size.

This is based on the #1920 so that the tests pass.


jquense:
Hi there, two things. First, we are using lodash compat for specific reason, as we support more browsers than modern lodash covers, so swapping it out is not OK here. Second, I'd much more prefer to just cherry pick the modules ourselves, adding another babel plugin doesn't gain us anything except a level of indirection.

ruiaraujo:
@jquense I can change the import style if there an interest .
As for browser support, is IE8 still supported? Even React has dropped it early this year, and 'normal' lodash support IE9 and above.

My original goal was to update the compat version but then I realised it had been discontinued.

ruiaraujo:
I already did the change.
Also why not just have the function written internally? This library only uses pick, omit and find which are easy to write oneself.

jquense:
because there isn't a lot of value in rewriting them ourselves, it amounts to practically the same amount of code and handwritten utils eventually bloat.

React doesn't support ie8 but it's also not removing support that is already there, and bootstrap v3 does support ie8 and users of this library also have needs for ie8 support. I a bit annoying, I agree but such is the deal with maintaining widely used software. In any case I can't just make the call to not support it, the bytes saved here but updating are really minimal anyway.

That being said there is probably no chance that the lodash v4 methods don't work with ie8 with es5-shim so I don't know. @react-bootstrap/collaborators?

taion:
Well, if that was the case, why did we cut them over in the first place? I recall there were some issues there.

jquense:
well there is no guarantee, just musing from info I saw on the lodash issue-tracker, but yeah I believe someone PR'd it because it did cause an issue now that I think about it...

ruiaraujo:
How about doing something a bit more wild? 
First, I would refactor the imports to be from an utils_lodash file.
Then I move lodash-compat to a peer depedency and use require.resolveWeak to check if lodash compat api is available. If not, I try the lodash v4 api. If none is available, explode.

That way I as a consumer of react bootstrap can alias lodash compat to normal lodash and react bootstrap can live.

Another way would be just having that file and depending on some build time constant build different version of the npm package and publish react-bootstrap-ie8 which ships lodash compat and react-bootstrap which relies on normal lodash.

If there is interesting in having one of these solutions, I can do a PR with the one preferred.
Otherwise I guess I can just maintain my own fork for our company since I don't want to send two lodash libs in our bundle. I am also looking into forcing the consumer to provide the polyfill, which will harder for react-bootstrap to accept so maybe I just keep these changes for myself. ;-)

taion:
We're not sending down all of lodash though... just the parts needed for pick/omit/&c. Same as if we were to vendor those, or close enough at any rate.

ruiaraujo:
Ok, got it. I guess IE8 will be dropped during the update to bootstrap 4. ;-)

PR #1925, opened 2016-05-24
Add 'contextMenu' trigger to OverlayTrigger

Zanadar:
This PR aims to allow an OverlayTrigger to accepts a 'contextmenu' trigger event and accompanying event handler. This would allow a menu/overlay to be triggered by a right-click if so desired.

I've added a couple tests so far, but haven't added any examples or anything like that in order to get feedback from the team first. 

Thanks so much for all the great work on the project! Looking forward to your thoughts.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -33,8 +33,8 @@ const OverlayTrigger = React.createClass({
      * Specify which action or actions trigger Overlay visibility
      */
     trigger: React.PropTypes.oneOfType([
-      React.PropTypes.oneOf(['click', 'hover', 'focus']),
-      React.PropTypes.arrayOf(React.PropTypes.oneOf(['click', 'hover', 'focus']))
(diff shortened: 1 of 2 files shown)


taion:
Sounds reasonable to me. I'd just add something to the list of trigger types.

taion:
I mean, the example with the list.

jquense:
This adding features not upstream does that make sense here...I have to say it seems like contextmenu as trigger feels like a weird use case for an overlay trigger.

Zanadar:
Upstream meaning master Bootstrap I'm assuming? I guess you're right that they don't really use this kind of trigger.
In the end its up to you whether this makes sense.

jquense:
Could you take a bit about what your use case is here? Seems like an absolutely positioned overlay is generally not what you want for a context menu replacement,but I might just be thinking about it wrong

taion:
Oh fair point.

Zanadar:
Nope, you're thinking correctly. But I was wasn't. Apologies for the coffee fueled, ill considered PR. I'll close this one....

PR #1946, opened 2016-06-06
Provided basic accessibility to button component

pearl1991:
Using [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role](url) and [https://www.marcozehe.de/2013/04/24/easy-aria-tip-6-making-clickables-accessible/](url) as guides.

### DIFF
--- a/src/Button.js
+++ b/src/Button.js
@@ -51,6 +51,10 @@ class Button extends React.Component {
         {...this.props}
         href={href || '#'}
         className={classNames(this.props.className, classes)}
+        tabIndex="0"
+        role="button"
+        ariaLabel={this.props.ariaLabel}
+        onKeyUp={this.handleBtnKeyUp}
       >
         {this.props.children}
       </SafeAnchor>
@@ -64,7 +68,11 @@ class Button extends React.Component {
       <Component
         {...this.props}
         type={this.props.type || 'button'}
-        className={classNames(this.props.className, classes)}>
+        className={classNames(this.props.className, classes)}
+        tabIndex="0"
+        role="button"
+        ariaLabel={this.props.ariaLabel}
+      >
         {this.props.children}
       </Component>
     );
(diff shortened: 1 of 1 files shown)


jquense:
Hey there thanks for the PR, we appreciate folks contributing especially a11y stuff.

There are a few things here thatI think you may have missed that might make this a bit unnecessary. See my notes on the diff

src/Button.js
```diff
@@ -51,6 +51,10 @@ class Button extends React.Component {
         {...this.props}
         href={href || '#'}
         className={classNames(this.props.className, classes)}
+        tabIndex="0"
+        role="button"
+        ariaLabel={this.props.ariaLabel}
+        onKeyUp={this.handleBtnKeyUp}
```
jquense: few things here. SafeAnchor already handles adding the button role, and tabIndex (its job is to ensure that "link" buttons are accessible). `onKeyUp` is also attached to a handler that doesn't exist so I'm not sure what the goal with that is.

src/Button.js
```diff
@@ -51,6 +51,10 @@ class Button extends React.Component {
         {...this.props}
         href={href || '#'}
         className={classNames(this.props.className, classes)}
+        tabIndex="0"
+        role="button"
+        ariaLabel={this.props.ariaLabel}
```
taion: This is redundant as well, no?

src/Button.js
```diff
@@ -64,7 +68,11 @@ class Button extends React.Component {
       <Component
         {...this.props}
         type={this.props.type || 'button'}
-        className={classNames(this.props.className, classes)}>
+        className={classNames(this.props.className, classes)}
+        tabIndex="0"
+        role="button"
+        ariaLabel={this.props.ariaLabel}
```
jquense: The default component here is a real `button` so the additions aren't actually needed in that case. A user _could_ specify a div, or some other component and then these would be really helpful, however we tend to leave that up to users to make the right choice, since adding the redundant roles and attributes (especially tabIndex) can have unintended side effects even for screen readers (at least in our experience). I'd say we could conditionally render them, but its not actually possible to know for sure what component the user will eventually render.

src/Button.js
jquense: ya, I forgot to note it originally

taion:
Can we close this out?

jquense:
I think so, @pearl1991 let us know if you think we are still missing anything here.

josephdburdick:
What if we want to control the tabindex via props?

taion:
Then just pass it in.

josephdburdick:
Oh. For some reason looking at the proposed code above and seeing this PR closed, not merged, I thought that tabIndex wasn't a possibility. Now I see {...this.props} handling everything, thanks.

PR #1952, opened 2016-06-09
Removed alert

### LINKED ISSUE #1842: Alert when moving away from the page of components to the other documentation page
When I left [the page of components](http://react-bootstrap.github.io/components.html) through the main menu, I get the alert `selected undefined`

sitebug

It is not critical. Just want to let you know.

Regards.

joshuahenson:
Fixes #1842


jquense:
hi there, thanks for the PR. Actually, the alert is supposed to be there for the sake of the example, the bug is that somehow that event leaks outside the tab

PR #1957, opened 2016-06-10
feat: add dialog style property

deser:


taion:
Don't do that. Just merge in `this.props.style` on https://github.com/react-bootstrap/react-bootstrap/blob/c0e02fed8d0138fcf55f09cfbbb1dbb91c8e4f84/src/Modal.js#L170.

deser:
To my mind style prop should be passed to root component which is BaseModal. Therefore I introduced new property (like it done for CSS class - dialogClassName) .  However  don't see the full picture. If you insist on merging I'll do that

taion:
This is not styling the dialog.

jquense:
yeah look at the markup of the bootstrap dialog. the actual visible dialog is not the BaseModal, that is a larger more abstract div that you don't want to style.

deser:
Done

taion:
I think what @jquense is saying is that you don't actually want to style the outermost modal component; instead you do want to style the dialog.

deser:
But what have been done matches my needs. I think if you are OK, guys, we can merge

jquense:
LGTM? I can't keep track of which component is which here anymore

taion:
We should be using the ES7 object spread syntax here.

Also, I thought you said it should be e.g. `dialogStyle` like on https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ModalDialog.js#L37?

deser:
>>I can't keep track of which component is which here anymore
What do you mean?

>>We should be using the ES7 object spread syntax here.
Where? 

>>Also, I thought you said it should be e.g. dialogStyle
Could you explain what you mean?

src/Modal.js
```diff
@@ -150,10 +151,15 @@ const Modal = React.createClass({
       , children
       , dialogClassName
       , animation
+      , style
       , ...props } = this.props;
 
     let { modalStyles } = this.state;
 
+    if (style) {
+      modalStyles = merge(modalStyles, style);
```
taion:
```js
modalStyles = { ...modalStyles, ...style }
```

PR #1962, opened 2016-06-15
Update React peer dependency

mull:
Looking at devDepencies there doesn't seem to be a reason that >= 0.15.0 isn't accepted. (Couldn't upgrade to React 0.15 with npm2)


jquense:
Thanks for the PR, `>=` accepts react 15 already actually. You may have another issue with your installs if you are running into peer issues.

mull:
Ouch, sorry about that! Also noticed that this wouldn't build anyhow.

jquense:
np!

PR #1965, opened 2016-06-17
Add required id to tooltip examples

joshuahenson:
I'm torn on whether to also include id's on tooltips in examples focusing on other components such as the Modal so I left those out. Thoughts?


taion:
I think it's always good to have IDs. Doesn't help for our examples to have a11y issues.

LGTM.

dozoisch:
LGTM!

PR #1986, opened 2016-06-28
FormGroup required property

donteatyellowsnow:
Added required property to automatically append a required class to the form group element and the required attribute to the form control.

### DIFF
--- a/src/FormControl.js
+++ b/src/FormControl.js
@@ -32,6 +32,7 @@ class FormControl extends React.Component {
   render() {
     const formGroup = this.context.$bs_formGroup;
     const controlId = formGroup && formGroup.controlId;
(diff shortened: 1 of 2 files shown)


taion:
This is not part of Bootstrap.

donteatyellowsnow:
I'm sorry I misunderstood the main intention to adopt bootstrap's specs but these lines would really simplify my life if they were part of this patch.

jquense:
not sure how this actually simplifies anything for you...this just means you can put the `required` on FormGroup vs `FormControl` its the same amount of typing though considering that there should generally be only one FormControl in a FormGroup.

We aren't going to add it here, but you can create your own versions of FormGroup and FormControl that wrap the current ones but jsut add this bit of code if its helpful to you.

PR #2001, opened 2016-07-02
Fix arbitrary prop passthrough to DOM nodes. React 15.2.0 !

pointnet:
My 2 cents!

### DIFF
--- a/src/utils/ensureDomProps.js
+++ b/src/utils/ensureDomProps.js
@@ -1,7 +1,5 @@
-import React from 'react';
 import DOMProperty from 'react/lib/DOMProperty';
 import EventPluginRegistry from 'react/lib/EventPluginRegistry';
-import ReactComponentTreeDevtool from 'react/lib/ReactComponentTreeDevtool';
 
 /*
  * Inspired from https://github.com/facebook/react/blob/master/src/renderers/dom/shared/devtools/ReactDOMUnknownPropertyDevtool.js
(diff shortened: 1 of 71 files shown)


pointnet:
**Travis:** 
TabContainer
    ✖ should clean up unmounted tab Sstate

This could not work, the spec is looking for [eventKey=2].active but eventKey is not allowed on a div element, need to change the spec.

timdorr:
I don't know if a whitelisting approach makes sense. I'd rather just blacklist this library's known properties. If the user tries to pass in an unknown prop, that's something that *should* give a warning. 

Also, this digs into React private API, making it brittle against future upgrades (and possibly limits which version are supported). I don't know if copying the pattern designed for the dev tools makes the most sense to apply to this library. It's rather complex where it doesn't really have to be.

pointnet:
blacklisting is a better approach, this one is a quick fix to stop spawning my console while developing.

Hope you guys will fix it somehow asap!

src/utils/ensureDomProps.js
```diff
@@ -0,0 +1,84 @@
+import DOMProperty from 'react/lib/DOMProperty';
```
gaearon: This is unsupported and will break very soon.
Please don’t use internal React modules.
Anything inside `react/lib` is internal.

gaearon:
The correct fix would be for every component to remove *its* props.
Alternatively you can add `domProps` to every component for props that *should* be passed through as is.

gaearon:
Please see my explanation in https://github.com/Hacker0x01/react-datepicker/issues/517#issuecomment-230119718.

taion:
That's correct – the reason it's going to be work is because we want components to remove props they use that should not be passed down. You can see that we do this for more recently implemented components like `<FormControl>`: https://github.com/react-bootstrap/react-bootstrap/blob/v0.29.5/src/FormControl.js.

Thanks for submitting this PR, but I'm going to close this for the time being, as this is not the approach we want to take to resolve these issues.

Issue #1994, opened 2016-06-30
Fix arbitrary prop passthrough to DOM nodes.


jquense:
@jimfb and @gaearon were kinda enough to reach out and warn us that RB will have a lot of warnings in React v15.2.0, of which, a release is imminent.

The problem is that in preparation for removing the DOM attribute whitelist, React has started to warn when props that are _not_ DOM attributes are passed through to DOM nodes. This (AFAIK) to help for folks to clean up code before those attributes are suddenly applied to the DOM nodes in a later version when the whitelist is removed.

@react-bootstrap/collaborators If anyone has time to start talking this in part (or whole) this might be less of a chore if a bunch of folks work on it. We also need to decide what the best way to ensure that we do not break existing API's by being too restrictive in what we pass through.

### Concerns

I'm not sure if any of these cases exist (here) but I am concerned about API's that require carte-blanche passthrough of props to a child that may or may not a DOM node. Presumably in those cases we can either defer to the User to use a component, or make sure to remove "own" props before passing through.

This also effectively means that API's like `dropdown` which allow for using the throwaway `bsRole` prop to indicate _type_ to the parent no longer work for DOM nodes. That sort of sucks, that was a really great side-effect feature of that API.

jquense:
as an aside this will also break, a common (for me) testing pattern of using throwaway props as selector targets `<div _testID='foo' />` and `find([_testID='foo').assert()`

mxstbr:
Is the problem with carte Blanche something we should fix on our side?

gaearon:
For more context, you can reproduce the warnings by:

1. Cloning this repo, running npm install and running the docs website (see instructions in docs folder)
2. Installing react and react-dom 15.2.0-rc.1
3. Opening "Components" page

You will see many warnings in console and you can start fixing them. If you work on this make sure to comment here so others don't do duplicate work. You may comment about particular 5 components you're taking to split the work more efficiently.

jquense:
hah in this case I was just using the phrase (albeit with somewhat liberal license with its actual meaning :p) 

I have been playing with Carter Blanche the proper now recently and it's :+1:

mxstbr:
Ah now that sentence makes more sense :grin: I guess that's what we get for choosing that name…

Glad to hear!

moaazsidat:
[...]

PR #2003, opened 2016-07-03
Stop passing non-dom attributes in Alert

pgilad:
See https://github.com/react-bootstrap/react-bootstrap/issues/1994

### DIFF
--- a/src/utils/sanitizeProps.js
+++ b/src/utils/sanitizeProps.js
@@ -1,0 +1,15 @@
+export const omit = (obj, keysToRemove) => {
+  if (!obj) {
+    throw new Error('Expecting an object');
+  }
+  if (!keysToRemove || keysToRemove.length === 0) {
+    return obj;
+  }
+  return Object.keys(obj).reduce((accumlator, key) => {
+    if (keysToRemove.indexOf(key) === -1) {
+      // shallow copy
+      accumlator[key] = obj[key];
+    }
+    return accumlator;
+  }, {});
+};
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -53,12 +53,13 @@ let Alert = React.createClass({
   render() {
     let classes = getClassSet(this.props);
     let isDismissable = !!this.props.onDismiss;
-
+    const { onDismiss, dismissAfter, closeLabel, ...rest } = this.props;
+    
     classes[prefix(this.props, 'dismissable')] = isDismissable;
 
     return (
       <div
-        {...this.props}
+        {...rest}
         role="alert"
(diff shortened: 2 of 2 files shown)


jquense:
Thanks for the PR and helping us fix this issue! I'm surprised that this passes lint since this leaves a bunch of unused variables...

This is certainly a pattern i've used before myself but I think in this project we want to be a bit clearer about what is going on so that others don't accidentally remove these lines thinking they aren't needed.

An approach we use elsewhere: https://github.com/react-bootstrap/react-bootstrap/blob/v0.29.5/src/FormControl.js#L49 is to delete the props off of the remaining props.

Alternatively we can use lodash/omit

pgilad:
@jquense How 'bout now? I added an `omit` function (shallow copies and excludes provided keys)

taion:
We have an ESLint plugin to suppress the warning. We should have a global `omitBsProps` helper though, then consume props separately per-component.

Actually I think we need these to be against `next` anyway, since I think we need to switch from `bsRole` to `data-bsRole` (`data-bs-role`? how does camelcase/hyphenation conversion work there?).

Gimme a bit.

pgilad:
data attributes are by convention `hyphen-case`. But I don't think removing only `bsRole` props is sufficient. You need to remove all non-dom props.

jquense:
I think we should fix all the props we can in master. we can handle bsRole or dataBsRole for non breaking for folks stuck on 29.0 for now

taion:
We have a bunch of deprecations on master though. I'd really rather not fix things twice.

jquense:
the bsRole bit is a super minor part to what is broken here. it's easy enough to fix this without requiring folks upgrade which I think is the better path for addressing new non-breaking warnings from react. 

plus we don't even need to fix anything in our components using bsRole, just not pass it down to the DOM node. we don't need to fix it twice I think supporting both props is fine for now

src/utils/sanitizeProps.js
```diff
@@ -0,0 +1,15 @@
+export const omit = (obj, keysToRemove) => {
```
jquense: we already have `lodash-compat` in the project so let's use `lodash-compat/object/omit` instead.

We __do__ want a bootstrap specific helper though, that can go in `uitls/bootstrapUtils`. It can simply wrap lodash omit, but also remove the boostrap specific props.

```js
function omitProps(props, keys) {
  return omit(props, [...keys, 'bsRole', 'bsStyle', 'bsClass'])
}
```

taion:
Replaced by https://github.com/react-bootstrap/react-bootstrap/pull/2026.

PR #2004, opened 2016-07-03
Stop passing non-dom attributes in Badge

liorbentov:
See https://github.com/react-bootstrap/react-bootstrap/issues/1994

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -1,8 +1,10 @@
 import classNames from 'classnames';
 import React from 'react';
 
-import { prefix } from './utils/bootstrapUtils';
+import { prefix, omitProps } from './utils/bootstrapUtils';
 import ValidComponentChildren from './utils/ValidComponentChildren';
+
+const PROPS_TO_OMIT = ['pullRight'];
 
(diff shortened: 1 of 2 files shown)


jquense:
Thanks for the PR and helping us fix this issue! I'm surprised that this passes lint since this leaves a bunch of unused variables...@taion?

This is certainly a pattern i've used before myself but I think in this project we want to be a bit clearer about what is going on so that others don't accidentally remove these lines thinking they aren't needed.

An approach we use elsewhere: https://github.com/react-bootstrap/react-bootstrap/blob/v0.29.5/src/FormControl.js#L49 is to `delete` the props off of the remaining props.

Alternatively we can use `lodash/omit`

jquense:
@liorbentov see https://github.com/react-bootstrap/react-bootstrap/pull/2003 for the conversation about how to structure these PR's

liorbentov:
@jquense, how about now?

taion:
Replaced by https://github.com/react-bootstrap/react-bootstrap/pull/2026

PR #2015, opened 2016-07-04
Add 'active' prop to NavDropdown component

mrijke:
This commit adds the 'active' prop to the NavDropdown component, which
when true adds the 'active' class to the component.

Note that this does not automatically fix https://github.com/react-bootstrap/react-router-bootstrap/issues/167, because the `LinkContainer` does not wrap the entire NavDropdown, but only the individual items. 
Personally I fixed this by using a wrapper component wrapping the NavDropdown which checks in the render() if the path of any of its children currently is active (the same check the LinkContainer performs) like so:

```js
  isAnyChildActive() {
    let anyChildActive = false;
    React.Children.forEach(this.props.menuItems, (child) => {
      if (!anyChildActive)
        anyChildActive = this.context.router.isActive(child.props.path, false);
    });
    return anyChildActive;
  }
```
(in the above code `this.props.menuItems` is a list of MenuItem wrappers which have the `path` prop which is passed to the LinkContainer as the `to` prop)

### DIFF
--- a/src/NavDropdown.js
+++ b/src/NavDropdown.js
@@ -1,13 +1,15 @@
 import React from 'react';
+import classNames from 'classnames';
 import Dropdown from './Dropdown';
 
 class NavDropdown extends React.Component {
 
   render() {
-    let { children, title, noCaret, ...props } = this.props;
+    let { children, title, noCaret, active, className, ...props } = this.props;
+    const classes = classNames(className, { active });
 
     return (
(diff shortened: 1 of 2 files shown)


taion:
This should go on NavDropdown.

mrijke:
OK, moved the prop to the NavDropdown component.

jquense:
LGTM

PR #2016, opened 2016-07-04
Add ValidComponentChildren.toArray

taion:
This replaces the inconsistent childrenToArray.


src/utils/ValidComponentChildren.js
```diff
@@ -137,11 +135,26 @@ function some(children, func, context) {
   return result;
 }
 
+function toArray(children) {
```
jquense: does it make sense to leverage `React.Children.toArray` here now that it exists?

taion: Like `React.Children.toArray(children).filter(React.isValidElement)`? I feel like they're about equivalent.

jquense: Yeah I don't know, toArray does some bit of logic, but we may already have it covered

taion: Not really – it's just `map` with the identity function as the predicate: https://github.com/facebook/react/blob/24dfb56113a214d98f29f69e2c26f67d7e947067/src/isomorphic/children/ReactChildren.js#L196-L205

PR #2026, opened 2016-07-05
Fix unknown prop passthrough, phase 1

taion:
#1994


taion:
This makes additional cleanups as we go through the components.

src/utils/bootstrapUtils.js
```diff
@@ -126,6 +127,16 @@ export function getClassSet(props) {
   return classes;
 }
 
+export function omitBsProps(props) {
+  props = { ...props };
+
+  delete props.bsClass;
```
jquense: this should probably just use lodash omit, deleting properties deoptimizes code paths. Its extremely minor but not reason to not take advantage of lodash here since we have it.

modosc:
fwiw i'm seeing these errors in a lot of other components - i'm happy to help out with prs, is this the master pr to reference? is there a list of components already in motion?

src/utils/bootstrapUtils.js
taion: Good point! Fixed.

CHANGELOG.md
```diff
@@ -4,15 +4,23 @@
 - **Breaking:** Remove `linkId` prop from `<Breadcrumb.Item>`, for consistency with other wrapped link components ([#2011])
 - **Breaking:** Remove deprecated functionality from v0.29.x ([#2017])
 - **Breaking:** Remove redundant `caption` prop from `<Carousel.Item>`, as this functionality is available from `<Carousel.Caption>` ([#2018])
+- **Breaking:** Remove pointless `navItem` and `navDropdown` props from `<Button>` ([#2026])
```
jquense: slightly related... do we cover `navbar-btn` anywhere? that's definitely a thing I wonder if this was a bad attempt at that?

taion: That would be like a `<Navbar.Button>`, right? Per http://getbootstrap.com/components/#navbar-buttons

The existing support on v0.29.x is just broken though.

jquense: Yeah, a subcomponent is a good idea here. This isn't needed in this PR just made me think of it

taion: Yeah, `navItem` was doing something totally different/weird/wrong in embedding the button in an `<li>`. So weird.

.babelrc
```diff
@@ -1,5 +1,5 @@
 {
-  "stage": 1,
+  "stage": 0,
```
taion: This is weird but the Babel class property bind thing is faster than `func.bind`, so why not? I'm going through all the components anyway argh

test/MenuItemSpec.js
```diff
@@ -2,11 +2,12 @@ import React from 'react';
 import ReactTestUtils from 'react/lib/ReactTestUtils';
 import ReactDOM from 'react-dom';
 import tsp from 'teaspoon';
+
 import MenuItem from '../src/MenuItem';
 
 import { shouldWarn } from './helpers';
 
-describe('MenuItem', () => {
+describe('<MenuItem>', () => {
```
taion: `.only()` does a grep, which ends up matching other tests... I don't know, maybe I'll back this out and just hack it to e.g. `describe('XXXMenuItem')` when testing anyway

https://github.com/mochajs/mocha/pull/1807

.babelrc
jquense: I thought property initializers was stage 1 now? I dunno, I do feel bit weird about using those here.

CHANGELOG.md
jquense: that is really odd :P

.babelrc
taion: They are in Babel 6.

src/utils/bootstrapUtils.js
```diff
@@ -126,6 +130,10 @@ export function getClassSet(props) {
   return classes;
 }
 
+export function omitBsProps(props) {
```
jharris4: This only takes care of the basic bs props. I'm curious, are you thinking of expanding on this concept to have something like omitProps(props, tagName) ?

taion: No – each component should own deleting the props it knows about. This is just a convenience method for the shared stuff.

taion:
This PR is already getting ridiculously long.

@react-bootstrap/collaborators Do you want to take a look at the current PR? I think it'd be more manageable as-is. I can then submit the rest in a follow-up.

src/Badge.js
```diff
@@ -1,46 +1,56 @@
...
+class Badge extends React.Component {
+  hasContent(children) {
+    let result = false;
+
+    React.Children.forEach(children, child => {
+      if (result) {
+        return;
+      }
+
+      if (child != null && child !== false) {
```
dozoisch: `false != null` is already true. Did you mean to write `!==`? Else I think you can use only the first conditions

taion: `false !== false` => `false`, though, which fails this condition. That's the logic I was trying to express here.

src/Clearfix.js
```diff
@@ -1,71 +1,79 @@
...
-      let size = styleMaps.SIZES[key];
+    DEVICE_SIZES.forEach(size => {
+      const propName = `visible${capitalize(size)}Block`;
+      if (props[propName]) {
+        classes[`visible-${size}-block`] = true;
+      }
 
-      classes['visible-' + size + '-block'] = this.props['visible' + size.charAt(0).toUpperCase() + size.slice(1) + 'Block'];
-    }, this);
+      delete props[propName];
```
dozoisch: should we use lodash for this? 
we can do a map with device sizes and then `_.omit(props, mappedDeviceProps)`. We could also pre calculate the prop list instead of recalculating each time.

taion: We could – I think it's neater to delete props as they're used, though. Makes it easier to keep things in sync.

src/Collapse.js
```diff
@@ -190,17 +100,118 @@ Collapse.propTypes = {
...
+    return `${elem[`scroll${capitalize(dimension)}`]}px`;
+  }
+
+  render() {
+    const {
+      onEnter, onEntering, onEntered, onExit, onExiting, className, ...props,
+    } = this.props;
+
+    delete props.dimension;
+    delete props.getDimensionValue;
```
dozoisch: Should we spread these two props out? Instead of deleting them.

src/ListGroup.js
```diff
@@ -1,98 +1,69 @@
...
+   * If not specified, it will be treated as `'li'` if every child is a non-
+   * actionable `<ListGroupItem>`, and `'div'` otherwise.
+   */
+  componentClass: elementType,
+};
 
-    return shouldRenderDiv ? this.renderDiv(items) : this.renderUL(items);
+function getDefaultComponent(children) {
+  if (!children) {
+    // FIXME: This is the old behavior. Is this right?
```
dozoisch: this seems right!

src/MenuItem.js
```diff
@@ -143,20 +58,97 @@ MenuItem.propTypes = {
...
+      disabled,
+      divider,
+      header,
+      onClick,
+      className,
+      style,
+      ...props,
+    } = this.props;
+
+    delete props.eventKey;
```
dozoisch: should we just spread those props out, instead of delete?

src/PanelGroup.js
```diff
@@ -1,92 +1,98 @@
...
+    const {
+      accordion,
+      activeKey: activeKeyProp,
+      className,
+      children,
+      ...props,
+    } = this.props;
+
+    delete props.defaultActiveKey;
+    delete props.onSelect;
```
dozoisch: same, as before (just noting all of them) should we delete or spread

src/Collapse.js
taion: I think it makes more sense to delete props that we aren't using – it's more explicit and doesn't leave around an unused binding.

dozoisch:
Awesome work @taion :)

taion:
How's this looking for merge? Getting ready with phase 2 now.

src/Collapse.js
jquense: why not accept keys in the omit function as a week as removing the bs props? it conflates things a bit, but avoids an extra spread, which might matter in ppls apps depending on how many of a thing they have

src/Badge.js
jquense: `child || child === 0`? I think zero is the only falsely value that would actually render content here yeah?

src/Badge.js
```diff
@@ -1,46 +1,56 @@
...
+  }
 
   render() {
-    let classes = {
-      'pull-right': this.props.pullRight,
-      [prefix(this.props)]: this.hasContent()
+    const { pullRight, className, children, ...props } = this.props;
+
+    const classes = {
+      ...getClassSet(props),
```
jquense: there are no variants of badge, size or style. I don't know that using this here makes sense...

src/Breadcrumb.js
```diff
@@ -1,42 +1,35 @@
...
-        {...props}
+        {...omitBsProps(props)}
         role="navigation"
         aria-label="breadcrumbs"
-        className={classNames(className, this.props.bsClass)}>
-        {ValidComponentChildren.map(this.props.children, this.renderBreadcrumbItem)}
+        className={classNames(className, classes)}
+      >
+        {ValidComponentChildren.map(children, (child, index) => (
+          this.renderBreadcrumbItem(child, index))
```
jquense: maybe just inline this?

src/Button.js
```diff
@@ -2,120 +2,80 @@ import classNames from 'classnames';
...
-        {this.props.children}
-      </Component>
+        {...omitBsProps(props)}
+        type={props.type || 'button'}
+        className={className}
+      />
     );
   }
 
-  renderNavItem(classes) {
```
jquense: what even was this...

src/Checkbox.js
```diff
@@ -52,7 +51,7 @@ class Checkbox extends React.Component {
 
     if (inline) {
       const classes = {
-        [prefix(this.props, 'inline')]: true,
+        [prefix(props, 'inline')]: true,
```
jquense: i'm slightly worried about this pattern, if someone refactors above to spread out bsClass this will subtly break.

src/Clearfix.js
```diff
@@ -1,71 +1,79 @@
...
+
+const propTypes = {
+  componentClass: elementType,
 
-  getDefaultProps() {
-    return {
-      componentClass: 'div'
-    };
-  },
+  // TODO: These should be `visibleSz`, not `visibleSzBlock`.
```
jquense: should they be? the classes without display stlye are deprecated. If anything we should allow `inline` and `inline-block` yeah?

src/Breadcrumb.js
taion: This is fixed on the next branch... the whole mapping thing shouldn't be there

src/Button.js
taion: You could render a `<Button>` under a `<Nav>`, and something dumb would happen

src/Clearfix.js
taion: Yeah you're right

src/Badge.js
taion: Right that's more correct anyway

src/Badge.js
taion: It doesn't per se, but I just want to use this wherever possible – it's slightly more consistent and if someone were to add styles for some reason, they would "just work".

src/Collapse.js
taion: wdyt of doing

```js
const { active, disabled, ...props } = this.props;
const { bsProps, componentProps } = splitBsProps(props, ['eventKey', 'onSelect']);

// Or maybe:
const [bsProps, componentProps] = splitBsProps(props, ['eventKey', 'onSelect']);

const classes = getClassSet(bsProps);
```

taion: This makes it more obvious that the `bsProps` should not be spread out, and e.g. `prefix(bsProps, 'inline')` looks more explicit.

src/Checkbox.js
taion: With the split out `bsProps`, I think this should be more okay.

taion:
@jquense Does that `splitBsProps` API sound reasonable? I want to defer that to a later stage PR, and get this merged so I can throw up the next one.

taion:
Pulled in a few more small irrelevant changes from my other branch.

taion:
I'm going to merge this for now so I can proceed with the next phase.

PR #2027, opened 2016-07-05
Clean up tooling a bit

taion:
Rename the package.json scripts to match up with names elsewhere


jquense:
:thumbsup: you on a roll

PR #2036, opened 2016-07-08
Fix unknown prop warnings, phase 2

taion:


taion:
This is good to go now. I cleaned up a bit of the state management in `<Nav>` and `<TabPane>` as well.

taion:
Just the overlays left!

test/TabContainerSpec.js
```diff
@@ -308,6 +314,46 @@ describe('<TabContainer>', () => {
     instance.single('[eventKey=0]').single('.active');
   });
 
+  [
+    [[1, 2], [2, 1]],
+    [[2, 1], [1, 2]],
+  ].forEach(([order1, order2]) => {
+    it('should handle event key swaps', () => {
```
taion: @jquense 

Never mind, I found another tab animation edge case (the logic in the previous commit would drop the animation if this happened... the logic on master would instead fail the duplicate `eventKey` invariant)

basically doing this type of imperative animation end callback handling in React is the worst thing ever. I don't think even not using context would simplify this appreciably.

Nishchit14:
Any estimation for next release  please?, this warning messages feels like a mad dragon in the heaven while consoling.

taion:
Use the documented warning filtering functionality

src/Carousel.js
```diff
@@ -122,214 +123,224 @@ let Carousel = React.createClass({
...
-  },
-
-  pause() {
-    this.isPaused = true;
-    clearTimeout(this.timeout);
-  },
+    this.select(index, e, 'prev');
+  }
 
-  play() {
```
jquense: these are missing...my hunch is that they are public methods.

taion: You think so? They're not documented anywhere. They'll break as soon as we move this to use uncontrollable anyway. IMO if someone complains we'd be better off just adding a prop to control this.

src/DropdownButton.js
```diff
@@ -1,57 +1,57 @@
...
-          disabled={disabled}
+          bsClass={prefix(props, 'toggle')}
+          bsSize={bsSize}
+          bsStyle={bsStyle}
         >
           {title}
         </Dropdown.Toggle>
-        <Dropdown.Menu>
+
+        <Dropdown.Menu bsClass={prefix(props, 'menu')}>
```
jquense: is this how we treat other subcomponents with bsClass?

src/DropdownMenu.js
```diff
@@ -2,22 +2,32 @@ import classNames from 'classnames';
...
+const defaultProps = {
+  bsRole: 'menu',
+  pullRight: false,
+};
 
 class DropdownMenu extends React.Component {
   constructor(props) {
     super(props);
 
-    this.focusNext = this.focusNext.bind(this);
```
jquense: these are "public" in that they are called by other components (dropdown) we should bind them as a defensive measure I think.

src/DropdownToggle.js
```diff
@@ -3,56 +3,54 @@ import classNames from 'classnames';
...
+};
 
 class DropdownToggle extends React.Component {
   render() {
     const {
-      noCaret,
-      useAnchor,
-      bsClass: bootstrapClass,
-      ...props
+      noCaret, open, useAnchor, bsClass, className, children, ...props,
```
jquense: in general I think if we are gonna push properties to a new line we should jsut format them all on a new line...I find this style harder to read personally, not sure why.

src/DropdownButton.js
taion: No – this is new. I think this is the right way to handle it. Previously `bsClass` would get passed down from `<Dropdown>`, which made it impossible to directly set a custom `bsClass` on the subcomponents.

This just seemed to be the most intuitive way to go – if you set a `bsClass` on `<DropdownButton>`, it propagates through to the subcomponents.

src/DropdownToggle.js
```diff
@@ -3,56 +3,54 @@ import classNames from 'classnames';
...
-        className={classNames(classes, this.props.className)}
         role="button"
+        className={classNames(className, bsClass)}
         aria-haspopup
-        aria-expanded={this.props.open}>
-        {this.props.children || this.props.title}{caret}
+        aria-expanded={open}
+      >
+        {children || props.title}
+        {useCaret && ' '}
```
jquense: should we be making this assumption for a user? I think the extra space probably isn't our job

src/Nav.js
```diff
@@ -5,165 +5,262 @@ import ReactDOM from 'react-dom';
...
 
-  componentDidUpdate() {
-    if (this._needsRefocus) {
-      let ul = this.refs.ul && ReactDOM.findDOMNode(this.refs.ul);
-      let tabs = ul ? ul.children || [] : [];
-      let tabIdx = this.eventKeys().indexOf(this.getActiveKey());
+// TODO: This `bsStyle` is very unlike the others. Should we rename it?
 
-      this._needsRefocus = false;
+// TODO: `pullRight` and `pullLeft` don't render right outside of `navbar`.
```
jquense: not sure these TODO's are gonna be actionable here, maybe a new issue it better.

on them tho, I think the ship has probably sailed for renaming the style prob (maybe not)

src/DropdownToggle.js
taion: This was what the old logic did. I'm just writing it this way to take advantage of the `span` elision in React 15.

src/Nav.js
jquense: Not sure that note is right about the navbar...they are simply given the pull-right/left class in the non-navbar instance, there isn't actually an expectation for how that should look in those cases, but i've found them useful regardless.

src/DropdownMenu.js
taion: DropdownMenu calls it as `this.menu.focusNext()` though, which doesn't require binding. It's a pretty standard caveat with class-like objects in JS, so I figure why make it any different here?

src/DropdownButton.js
jquense: It also effectively makes setting bsClass on the subcomponents (or changing the defaults) a noop though right? Since Dropdown will always have a default for this.

taion: I took out the logic for `<Dropdown>` itself forwarding through `bsClass`, since it seemed unnecessary.

jquense:
as far as i'm possible, it looks good to me

menems:
Hello, is there a possibility to merge this pr on 0.29.6 ? or a way to use next branch directily with npm ? maybe publish a next tag on npm ?

thanks for the works

PR #2039, opened 2016-07-09
Clean up test tooling

taion:
I dropped the Phantom stuff because the test suite would sporadically crash Phantom


taion:
Wow our test suite doesn't pass on Firefox... okay...

AlexKVal:
LGTM

> Wow our test suite doesn't pass on Firefox... okay...
It is the "Pandora's box" 😄

PR #2042, opened 2016-07-12
Fix link to affixes example

JobLeonard:
From http://react-bootstrap.github.io/react-overlays/examples/#affixes  to http://react-bootstrap.github.io/react-overlays/#affixes

... actually, maybe this is the wrong fix, and the issue is that `/examples/` went missing? Not sure what is supposed to be the situation on the website.

### DIFF
--- a/docs/src/ComponentsPage.js
+++ b/docs/src/ComponentsPage.js
@@ -355,7 +355,7 @@ const ComponentsPage = React.createClass({
                     <Anchor id="affix">Affix</Anchor>
                   </h2>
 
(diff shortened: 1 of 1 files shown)


plemarquand:
@JobLeonard This link is also broken, could you fix it as well? https://github.com/react-bootstrap/react-bootstrap/blob/c0279cff10f20de5bc1f433baaa92ac39f1388b4/docs/src/ComponentsPage.js#L218

taion:
Thanks for noticing this. This was intentional. I updated the URL because reasons. I'll pull this in separately onto `next`.

taion:
Replacing with https://github.com/react-bootstrap/react-bootstrap/pull/2044

PR #2045, opened 2016-07-12
Fix and run tests on Firefox

taion:


jquense:
so literally it was jsut the tests at fault here?

taion:
Yup

jquense:
LGTM

PR #2052, opened 2016-07-14
Replace omitBsProps with splitBsProps

taion:


taion:
This is the last bit of follow-up to #1994.

I guess I also dropped `lodash-compat` as a dependency here and just added `lodash` as a dev dependency.

BorelTung:
When will we merge this PR and cut a new release? I really need it since I want to use `react-bootstrap` with `react@15.2.1`.

src/utils/bootstrapUtils.js
```diff
@@ -130,8 +129,45 @@ export function getClassSet(props) {
...
+  Object.entries(props).forEach(([propName, propValue]) => {
+    if (!isBsProp(propName)) {
+      elementProps[propName] = propValue;
+    }
+  });
+
+  return [getBsProps(props), elementProps];
+}
+
+export function splitBsPropsAndOmit(props, omittedPropNames) {
```
jquense: why two utils here? seems like you could just roll this into splitBsProps above yeah?

src/DropdownMenu.js
```diff
@@ -97,16 +97,16 @@ class DropdownMenu extends React.Component {
       ...props,
     } = this.props;
 
-    delete props.bsRole;
+    const [bsProps, elementProps] = splitBsPropsAndOmit(props, ['bsRole']);
```
jquense: should it maybe count bsRole as a bsProp?

src/utils/bootstrapUtils.js
taion: This should be a slight optimization... just two statically separate functions rather than a branch in the same function. Figure it's worth it since we call this literally everywhere.

src/DropdownMenu.js
taion: We could... I don't know. I didn't add it there because nothing else in `bootstrapUtils` dealt with `bsRole` at all. Do you think it makes sense there?

jquense: I think so, if only to help prevent it being accidentally passed through. and for folks using these utils for their custom components as well

taion:
Updated.

taion:
Good to go now?

jquense:
:shipit:

BarryOSeven:
Who is the hero who will save everybody?

hdrdavies:
so tense....

lynn:
Whoo, thank you! 🎊🎉

marceloemanoel:
Hi guys, first of all thank you all. How can I use this update? I did install "react-bootstrap/react-bootstrap#next" but now I can't import anything from 'react-bootstrap'. :(

davidfurlong:
New release: https://github.com/react-bootstrap/react-bootstrap/releases/tag/v0.30.0-rc.1

marceloemanoel:
Thanks! :D

PR #2072, opened 2016-07-19
Update overlayPositionUtils.js

jquense:
why are we re-exporting this?


taion:
I took this out completely on `next`. Forgot that this was being pointlessly reexported.

PR #2077, opened 2016-07-21
Improve Forms docs

taion:


chalda:
i feel like it would help to not just mention "use findDOMNode" in the docs but actually provide an example. I am thinking print out the input value into a separate div. Just so that there is a like of code findDOMNode(ref).value somewhere.

taion:
I'm sympathetic to that, but the problem is that it would be a bad example. Note the upstream React docs here: https://facebook.github.io/react/docs/forms.html#uncontrolled-components. You _can_ use an uncontrolled component, but you really shouldn't.

chalda:
they actually don't say it one way or the other. Everyone uses react with redux/flux (which forces your hand on some of this stuff) but you dont have to. I am actually not a big fan of having parent control the value of the input. Especially if you use redux you need a TON of wiring just to be able to update the value of your input through parent with redux. This is something inputs can very well do on their own.

if your component is a collection of inputs I very much prefer to just let them handle their own state, grab the value through the ref to do any kind of validation etc. I think this is very common usecase.

taion:
They specifically don't give an example of how to use uncontrolled form inputs, though.

jquense:
uncontrolled inputs are definately not the recommended or encouraged way to handle inputs, from all i've read and per convo' I had with various React team members. @chalda I agree that the naive case of controlling inpubts involves a ton of wiring, however a better solution is to find better ways of doing it "right" instead of falling back to the imperative escape hatch. You might like projects like redux-form, react-formal, or simpler: https://github.com/jquense/topeka

PR #2079, opened 2016-07-21
Fix default bsStyle on Panel

taion:


taion:
Oops. Guess we should cut rc.2.

Issue #2065, opened 2016-07-19
Navbar dropdown link issue on Microsoft Edge


cperera-aud:
Hi,

I've noticed an issue with Microsoft Edge for react-bootstrap navbar dropdowns. The issue is, if you click on a dropdown menu to view the menu then the page is directed to the homepage. This behavior is not present in Chrome or Firefox. I have attached two screenshots from a Microsoft Edge browser and Chrome browser which highlights this issue. 
![react-bootstrap-navbar-edge](https://cloud.githubusercontent.com/assets/16928715/16935422/a48e8f6a-4da2-11e6-99d4-7e0459f79198.png)
![react-bootstrap-navbar-chrome](https://cloud.githubusercontent.com/assets/16928715/16935423/a48efb44-4da2-11e6-8690-124d47aefe65.png)

The screenshot shows a navbar dropdown and I've hovered my mouse over it. On the bottom left corner of the screen the link that being pointed to is shown. 

The non-react bootstrap does not have this issue on Edge. 

So I was wondering is this something that I can fix on my end?

MiguelCastillo:
I am also running this issue. It is actually causing us lots of grief in the application I am working on. The problem is how IE is actually processing anchors with hrefs that are empty strings.  Quirky behavior I must say.

The issue can be resolved in react-bootstrap though, so that's very good news.  The flow is that NavDropdowns use `DropdownToggle`, which in turn uses `SafeAnchor` as the toggle component.  NavDropdown does not pass in an href to `SafeAnchor` which means that `SafeAnchor` will just add an href of empty string and that causes IE to misbehave.

1. Proposed solution is to change `SafeAnchor` to instead of am empty string, we use `javascript:void(0)` so that we make sure we address this issue at the root.

The change needs to happen right [here](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/SafeAnchor.js#L32)

2. Alternatively, we can just have `NavDropdown` pass in `javascript:void(0)` as the href for the DropdownToggle, which is less risky I guess...  But this issue might come up in the future again.

I can make a PR for either one.  I personally think that addressing this with `SafeAnchor` is the better choice...  But I will defer to other people with more knowledge.

taion:
I'm confused. Is `event.preventDefault` not doing anything here?

MiguelCastillo:
Where?

jquense:
[...]

PR #2080, opened 2016-07-21
Changed default href in SafeAnchor to '#' to make IE happy

MiguelCastillo:
This PR makes the default href of an anchor `#` instead of an empty string to make IE happy.

Discussion: https://github.com/react-bootstrap/react-bootstrap/issues/2065

### DIFF
--- a/src/SafeAnchor.js
+++ b/src/SafeAnchor.js
@@ -61,7 +61,7 @@ class SafeAnchor extends React.Component {
       props.role = props.role || 'button';
       // we want to make sure there is a href attribute on the node
       // otherwise, the cursor incorrectly styled (except with role='button')
-      props.href = props.href || '';
+      props.href = props.href || '#';
     }
 
     if (disabled) {
--- a/test/PaginationSpec.js
+++ b/test/PaginationSpec.js
@@ -181,8 +181,8 @@ describe('<Pagination>', () => {
     assert.equal(pageButtons[0].children[0].tagName, 'A');
     assert.equal(pageButtons[1].children[0].tagName, 'A');
 
-    assert.equal(pageButtons[0].children[0].getAttribute('href'), '');
-    assert.equal(pageButtons[1].children[0].getAttribute('href'), '');
+    assert.equal(pageButtons[0].children[0].getAttribute('href'), '#');
+    assert.equal(pageButtons[1].children[0].getAttribute('href'), '#');
   });
 
   it('should wrap each button in a buttonComponentClass when it is present', () => {


src/SafeAnchor.js
```diff
@@ -61,7 +61,7 @@ class SafeAnchor extends React.Component {
       props.role = props.role || 'button';
       // we want to make sure there is a href attribute on the node
       // otherwise, the cursor incorrectly styled (except with role='button')
-      props.href = props.href || '';
+      props.href = props.href || '#';
```
MiguelCastillo: What do you guys think about adding `role` and `href` to the deafultProps instead of manually setting it here?

taion: They only get set on this branch – they shouldn't on the other.

MiguelCastillo: Which branch? next?

https://github.com/react-bootstrap/react-bootstrap/blob/next/src/SafeAnchor.js#L64

taion: No, the logic branch where the `href` is trivial.

taion: https://en.wikipedia.org/wiki/Branch_(computer_science)

MiguelCastillo: k.

taion:
@jquense What are your thoughts here?

taion:
@jquense I'm tempted to say we should merge this just so people stop complaining.

jquense:
I suppose it can't hurt... I really dislike this sort of stuff :/

cperera-aud:
Hi guys,

Thank you for taking time to address the issue in [#2065](https://github.com/react-bootstrap/react-bootstrap/issues/2065). I am noticing the same issue in IE 11. 

I was wondering does this PR address the issue (#2065) seen in IE 11? Please pardon my ignorance on the low level aspects of the details in this issue. 

Thanks!

BrianSHenderson:
I'm still getting this problem on MS Edge, has this fix been deployed to the main release?

taion:
Wait, I think we might need to back this out. Because we use `<SafeAnchor>` everywhere, this means we end up showing links to `#` for `href`-less `<Button>`s, which is not desirable.

I think we actually need to use `<button>` where relevant to do this correctly.

kevinzwhuang:
@taion I just did some manual testing locally on this to confirm the behavior, here are my findings:

- `href`-less `<Button>`'s seem to be safe. They do not show `#` links due to `this.renderButton` being called instead of `this.renderAnchor` in `Button.js` due to the lack of a `href` prop (source: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Button.js#L66)

- However, all other components that do render `<SafeAnchor>` by default are now affected by an appended `#` to their links - including `<Pagination>`, `<Carousel>`, `<Thumbnail>`, `<Breadcrumb>` ( if no custom href provided, `/` will become `/#`), and possibly some other subcomponents.

I agree - we should revert for now and find a better solution in the future.

jquense:
So we I actually just rang into an interesting complication here. Chrome now exhibits the same behavior as IE i think. I ran into this with our work app, the empty href was showing as a link to itself now.

That being the case I don't think there is anymore harm here using `'#'` since it seems like browsers are converging on that behavior?

jquense:
confirm this seems to be the case in latest FF as well

taion:
I'm okay with that. I'll cut a release, then.

PR #2094, opened 2016-07-29
validationState prop should allow `false` on Form Components

cocomode18:
When using validation states in Form Components, `validationState` propType only accepts `'success'` `'warning'` and `'error'`. This is inconvenient if we want the default un-styled form controls.

To avoid the following code producing warnings to our console, I want the Components to accept `false` as the propType for `validationState`.

```es6
getValidationState() {
  const length = this.state.value.length;
  if (length > 10) return false; // using 'success' is too annoying!
  else if (length > 0) return 'error';
  return false;
}

render() {
  <FormGroup validationState={this.getValidationState()}>
    <span>something</span>
  </FormGroup>
}
```

### DIFF
--- a/docs/examples/FormBasic.js
+++ b/docs/examples/FormBasic.js
@@ -10,6 +10,7 @@ const FormExample = React.createClass({
     if (length > 10) return 'success';
     else if (length > 5) return 'warning';
     else if (length > 0) return 'error';
+    return false;
   },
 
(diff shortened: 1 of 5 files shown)


taion:
What's wrong with just using `null`?

cocomode18:
I'm sorry, I wasn't aware null is acceptable as a prop value...

PR #2095, opened 2016-07-29
Update for React 15.3

taion:

PR #2099, opened 2016-08-01
Fix <OverlayTrigger> rootClose regression

### LINKED ISSUE #233: Popover dismiss on next click

taion:
Fixes #233

Requires #2095 for clean build


jquense:
I'm surprised a test didn't already catch this!

jquense:
LGTM tho

taion:
We didn't have any test coverage for the case of clicking on the trigger when `rootClose` was enabled.

taion:
Still don't know why we wrote `rootClose` that way.

PR #1802, opened 2016-04-11
Update to Babel 6 (needs further work!)

bjrn:
So, I am aware that this PR is doing _way_ too much at the moment ... and the component documentation is broken, likely due to that `react-component-metadata` needs work in order to be compatible with Babel 6.

I worked around the issue by wrapping the metadata call with a try/catch in `docs/generate-metadata.js`
This results in that `window.PROP_DATA` in `docs/src/Root.js` is empty, thus the component page is disfunctional. 

Besides the update to use Babel 6, there's fixes for updated react-router, linting of a few files to fix the build etc. 

I would really appreciate if someone could give some pointers on how to move this along! Am I right in the assumption that `react-component-metadata` needs to be patched as well?

Issue #2101, opened 2016-08-01
Wrap the propTypes with NODE_ENV !== "production"?


oliviertassinari:
### The issue

When consuming `react-boostrap` from the npm registry we are most likely shipping a bundle full of `propTypes` definitions to our end users.
The new [warning](https://github.com/facebook/react/issues/7131) introduced in React v15.3.0 make it clear. Those `propTypes` are only here for the development mode. Sending them down the wire is wasteful.
Empirically, I think that this is responsible for 20% of the `react-boostrap` size.
### The proposed solution

For the npm released version, we could be wrapping all the `propTypes` definitions around the following block:

``` js
process.env.NODE_ENV !== "production"
```

However, that should probably be considered as a breaking change. Users might use them internally even though, I don't think that it's not supposed to be a public API.
### In practice

This is what https://github.com/callemall/material-ui/pull/4872 is trying to achieve. I thought it would be great to share it here too.

taion:
We actually use them internally in e.g. https://github.com/react-bootstrap/react-bootstrap/blob/v0.30.1/src/DropdownButton.js#L26-L27, so ATM this isn't possible for all the components – not via the Babel transform anyway (and e.g. that's why create-react-app doesn't strip propTypes either).

It'd be nice to get rid of them but it's not clear to me that there's a good path forward that doesn't violate DRY.

oliviertassinari:
> that doesn't violate DRY.
> We actually use them internally in e.g.

I see, you are using the `propTypes` definitions to forward the properties to the right components.

> It's not clear to me that there's a good path forward that doesn't violate DRY.

Wouldn't changing the signature of [`splitComponentProps`](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/splitComponentProps.js#L1) from:
```js
export default function splitComponentProps(props, Component) {
```
to
```js
export default function splitComponentProps(props, ComponentPropTypes) {
```
solve the isssue?
(We would have to export and import the `propTypes`, which would also be more explict)

taion:
It's possible – it'd need to not break how we detect propTypes for the auto-generated docs though.

oliviertassinari:
I guess you are using `react-docgen`. We are using it on Material-UI too. To be confirmed but I'm pretty sure he is supporting it.

jquense:
[...]

PR #2108, opened 2016-08-03
Babel 6

tgriesser:
I decided to start a fresh PR for converting to Babel 6, since #1802 seems to be a little stale. I've got the docs working in this branch, and all of the tests should be passing.

Things this PR includes:
- Babel 6
- Updated React Router
- Functioning Docs
- Removed decorator use in tests
- Some tweaks removing dead code / extraneous exports in `buildBabel.js`
- ES6 Build

The "ES6 Build" is like [react-router's](https://github.com/reactjs/react-router/blob/f0a7d1410923f90c1cd998d4bd50fee7e024d412/package.json#L13), building the source files except for any `import` / `export` statements. This, combined with a `"jsnext:main"` entry in `package.json`, and a ES6 module aware bundler like rollup or webpack 2 can significantly decrease build size.

Also, a few of the tests are slightly tweaked because it appeared that React was not `console.error`'ing multiple times for the same error message, so a test would pass in isolation but  not when the same `console.error` had been emitted in a prior test in the same run.


taion:
Two quick notes (I haven't read through this PR in detail yet):

- We'll want to use `module` in addition to `jsnext:main` – it's what webpack 2 supports, and Rollup now supports it as well https://github.com/rollup/rollup-plugin-node-resolve/issues/48. Node will probably never implement this, but it's pretty valuable for web stuff.
- Because of what we do with the `bsStyles` bits, &c. in `bootStrapUtils`, tree-shaking won't actually do anything useful with React-Bootstrap. The best way I'm aware of for nicely using the terser import style is `babel-plugin-lodash`, but I'm concerned it might just confuse people if we talk about it in React-Bootstrap docs.

taion:
Looks like tools/es6 didn't get checked in. Probably want the `.gitignore` to hit `/es6` instead of `es6`.

Also, it should probably just be `/es` instead of `/es6`. At least, that's what Redux uses, and it's what I cut the React Router alpha to use as well.

tgriesser:
> The best way I'm aware of for nicely using the terser import style is babel-plugin-lodash

I've been using that as well and it's been really nice. Agreed that it might be confusing to add in the docs, folks who know what to look for will find it on their own. I was thinking this might also be a nice plugin to use internally here to clean up the need for full paths for `react-overlays/lib/*` imports.

The other thing I wanted to propose but thought too much to include in this PR was to add `src` to the files list in `package.json`. This would allow those who wanted to modify their webpack settings to build directly from the `src`, enabling them to configure their own babel parsing presets. 

For example this would be a good solution to #2101, by allowing you to optionally add [`react-remove-prop-types`](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) to the build process, without any additional build steps for this project.

----

> We'll want to use module in addition to jsnext:main

Added

> Looks like tools/es6 didn't get checked in

Fixed

> Also, it should probably just be /es instead of /es6

Updated

.babelrc
```diff
@@ -1,6 +1,9 @@
 {
-  "stage": 1,
-  "optional": ["runtime"],
-  "loose": ["all"],
-  "plugins": ["dev-expression"]
```
taion: Still need `babel-plugin-dev-expression`

tgriesser: Ah interesting! I didn't realize this did more than replace `__DEV__`, looks pretty useful!

taion:
It's a bit annoying to get babel-plugin-istanbul working well. I'd just upgrade isparta loader to v2. That's been working fine in most of my repos.

taion:
Huh, it just works?

tgriesser:
@taion just swapped it out, but yeah it seemed to work - what sort of issues were you seeing?

taion:
Huh maybe there weren't any. You'd still need to restrict the coverage calculation to just the contents of `src/` though, and exclude `tests/`.

taion:
Thanks! LGTM

tgriesser:
@taion Cool, I'll just stick with `isparta-loader` for now and then upgrading to `babel-plugin-istanbul` can be a separate PR

AlexKVal:
Great job 👍

taion:
Thanks! Released in v0.30.2.

Issue #100, opened 2014-06-10
Tabs with dropdowns in nav


mathieumg:
Is this feature on the to-do list (if at all)? http://getbootstrap.com/components/#nav-dropdowns

I see https://react-bootstrap.github.io/components.html#navbars, but I didn't find a way to use a `<DropdownButton>` as one of the `<TabPane>`s in a `<TabbedArea>`.

Thanks!

mtscout6:
It looks like the docs are demonstrating this use case today: https://react-bootstrap.github.io/components.html#navbars. Is this still an issue for you?

mathieumg:
That's navbars, no? This issue is about a similar behavior on tabs. Unless one has just become a specific case of the other. If that is the case, it might be relevant to add an example to do the same with tabs. (As is supported by Bootstrap: http://getbootstrap.com/javascript/#tabs)

mtscout6:
Ah, I see tabs not navs.... Ya, we need to fix this.

![](http://cl.ly/image/2l400H2N1A2G/Image%202015-03-30%20at%208.07.33%20PM.png)

AlexKVal:
Superceded by #526

mtscout6:
Not fixed by #526

taion:
Moved to https://github.com/react-bootstrap/react-bootstrap/issues/1418.

mathieumg:
This is now possible by combining https://react-bootstrap.github.io/components.html#tabs-custom-layout and https://react-bootstrap.github.io/components.html#navs-dropdown.

@react-bootstrap/collaborators Are you fine with me closing this? Do you think we would need to add an example of that use case in the docs?

taion:
Does that work? We should add it to the example if so.

mathieumg:
Yeah I tested it quickly directly in the docs' editor. Could try to find the time to document it and write some tests around it even.

taion:
Would you mind if we keep this open until we have an example showing this, then?

mathieumg:
Not at all. I've been pretty busy, just got around upgrading to 0.29 and realized it was now possible... :no_mouth: I feel terrible, I'd like to invest a bit more time into the project, but I lift my hat off to @jquense and yourself for all the awesome work you've been doing. :+1: :grin:

taion:
Let's track this on #2115 instead.

auregnamian:
uhjhjj

lailaafrin:
### _****_

PR #2115, opened 2016-08-04
Feature/tabs with dropdown

mark2685:
Created example of tabs with dropdown.

Discussion: #100

### DIFF
--- a/docs/examples/TabsWithDropdown.js
+++ b/docs/examples/TabsWithDropdown.js
@@ -1,0 +1,47 @@
+const tabsInstance = (
+  <Tab.Container id="tabs-with-dropdown" defaultActiveKey="first">
+    <Row className="clearfix">
+      <Col sm={12}>
+        <Nav bsStyle="tabs">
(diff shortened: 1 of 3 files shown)


taion:
Huh, it looks like this actually doesn't just work – the dropdown tab, while styled correctly, does not properly receive "active" styling. Huh. I wonder if `NavDropdown` should check whether any of its children are active. Probably?

jquense:
probably, that would give it more value as a component anyway. it is a bit annoying to recreate tho if you need to use Dropdown directly in this context... does normal bootstrap handle this correctly?  I don't remember seeing extra code this this (I'll check)

taion:
They special-case it in a way that we can't really here: https://github.com/twbs/bootstrap/blob/master/js/tab.js#L93-L100

taion:
Active state management here is actually completely non-functional. Let's see if there's an easy way to fix.

taion:
Replaced with https://github.com/react-bootstrap/react-bootstrap/pull/2134

PR #2120, opened 2016-08-05
Carousel now has unfocusable option

ansonkao:
I'm building a very unorthodox web-app that is critically dependent on hotkeys, and it is imperative that anchors are not used, except for when making actual links. This is because as soon as an anchor gets focused, hotkeys do not propogate, and the web app becomes unusable.

If there is interest in merging this in, let me know and I'm happy to write some tests to go with the feature.

Cheers!

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -38,6 +38,7 @@ const propTypes = {
   direction: React.PropTypes.oneOf(['prev', 'next']),
   prevIcon: React.PropTypes.node,
   nextIcon: React.PropTypes.node,
+  unfocusable: React.PropTypes.bool,
 };
 
 const defaultProps = {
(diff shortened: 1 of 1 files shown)


taion:
I feel like the right way to handle this would be to allow the user to pass custom elements for the next/previous controls... and probably for you to use tabIndex rather than anchor v div.

This API is a bit of a mess though.

taion:
I mean, the existing Carousel API.

jquense:
I'm not really that doing this is even the right approach...admittedly I don't know your app, but not having links or buttons to avoid hotkey interference I think is the wrong way to do that. Making these unfocusable will effectively make this unusable by anyone use the keyboard or SR.

ansonkao:
Just to give you a sense of what my app is like, pretend it's a video game. It's not, but similarly to games, screen readers and accessibility are not a concern for us. Hotkeys are the primer concern.

`tabIndex="-1"` does not work because while it takes the control out of the tab sequence, it doesn't prevent it from gaining focus via click.

In any case, I found a work around. I'm just blurring the focused control in the `onSelect` callback. So unless anyone else sees this being a useful addition, I'm closing this. Cheers

PR #2133, opened 2016-08-10
Always pass expanded to renderAnchor in Panel

taion:
Fixes #2132

PR #2137, opened 2016-08-11
Fixes #2132

### LINKED ISSUE #2132: i can't use aria-expanded in Panel
Hello,
when i update react-boostrap, the icon disappeared.
image
image

in v0.30.2:
```  javascript
if (!React.isValidElement(header)) {
      return (
        <h4 role="presentation" className={titleClassName}>
          {this.renderAnchor(header, id, role, expanded)}
        </h4>
      );
    }
```
why don't you passing expanded  parameter to renderAnchor function？

tomlelouch:
in #2133 only add 'expanded' argument in the end, but you don't add 'id'
argument. that lead to the argument not matching

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -89,7 +89,7 @@ class Panel extends React.Component {
 
     return cloneElement(header, {
       className: classNames(header.props.className, titleClassName),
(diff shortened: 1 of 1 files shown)


taion:
LGTM

PR #2158, opened 2016-08-19
create bsCustomClass to allow downstream customization

bcullman:


taion:
Why? What does this do that the current `className` doesn't?

bcullman:
yup, my bad. Closing.

PR #2186, opened 2016-09-01
Don't send unknown prop 'closeButton' to div

liorbentov:


taion:
This code doesn't do anything. We already spread out `closeButton` on https://github.com/liorbentov/react-bootstrap/blob/e7fbfc8c83abd9894047f924b75d963cd13b1494/src/ModalHeader.js#L45.

Issue #2159, opened 2016-08-19
Pagination component bug


intpp:
![](https://api.monosnap.com/rpc/file/download?id=jfqQPtze5tAHI3gWPRYe2In1LgMp1I)

Total pages: 18, current page: 15.
Component renders [ 15 ] [ 16 ] [ 17 ] ... [ 18 ]

Same issue when current page is 4, component renders: [ 1 ] ... [ 2 ] [ 3 ] [ 4 ]

React bootstrap version: 0.30.2

kevinzwh:
Found is causing this! The calculation for both`startpage` & `hasHiddenPagesAfter` have off by one errors that cause the ellipses to display when they shouldn't be.

I'll open a PR for the solution to this issue in a moment

Issue #1758, opened 2016-04-01
Revamp <Pagination> API


taion:
Possibilities:
- `<PaginationButton>` -> `<Pagination.Item>`
- An actual sane API for customizing pagination links (maybe based on `children`?)

bebraw:
If you want inspiration, see http://sapegin.github.io/react-pagify-preset-bootstrap/ . That's a little Bootstrap preset we did on top of react-pagify. Maybe you can mimic the API from there. It's very flexible.

taion:
My current thinking is just to have a `renderItem()` function as the customization hook. I think `<Pagination>` has enough valuable logic inside that we can't just rip it apart (without forcing users to do a lot more work).

This is mostly just a placeholder so I don't forget, though.

AndreiRailean:
You need to allow for more granular and smarter buttonComponentClass-like overrides.
I'm working on isomorphic stuff and need to have the component render the href link. I would be able to do this in `buttonComponentClass` except PaginationButton doesn't pass the page number down (by explicitly unsetting it in https://github.com/react-bootstrap/react-bootstrap/blob/ae48d58afc2bcfcc36cabec136bba33929387513/src/PaginationButton.js#L51)

As far as react-pagify goes - it's got a very confusing API - i wouldn't follow that unless you want to scare people away. I couldn't do isomorphic stuff with react-pagify either because href generator cannot be overridden to make it smarter.

One thing react-pagify has is more granular control over what tags are used to render pagination, but there's no need for it here because bootstrap prescribes a particular structure to pagination component. Rendering without ULs and LIs means it's no longer a bootstrap component but something that looks like one if you invest time into making it look (and act) that way. In that case you don't need react-bootstrap.

taion:
That's a bug per https://github.com/react-bootstrap/react-bootstrap/issues/1866.

The API should be something like a `renderItem` callback.

taion:
This was resolved in #2587.

PR #2189, opened 2016-09-02
Fix Pagination ellipsis placement bug

kevinzwh:
Fixes https://github.com/react-bootstrap/react-bootstrap/issues/2159.
There's an off-by-1 bug in both `startPage` & `hasHiddenPagesAfter` calculations which cause the Pagination component to incorrectly place ellipses & group numbers excluding the first & last numbers.

This error occurs such that the first two numbers [1] ... [2] or the last two [19] ... [20] display with an ellipsis between them, when they should be connected ([1][2] & [19][20]).

It's caused by the `startPage` calculation mistakenly excluding the page `1` away from the active buttongroup when it should be included.
The same goes for `hasHiddenPagesAfter` mistakenly becoming `true` and excluding the last page button away from the button group, causing an ellipsis between the last page button & the group.

**Broken first ellipsis**
screen shot 2016-09-01 at 9 33 04 pm
**Broken second ellipsis**
screen shot 2016-09-01 at 9 33 16 pm

This fix allows Pagination to display properly, with the first & last PaginationButtons included with the current active button group:

**Fixed First Group**
screen shot 2016-09-01 at 9 16 58 pm
**Fixed Last Group**
screen shot 2016-09-01 at 9 18 55 pm

For reference, the component used in the examples above:
```es6
          <Pagination
                  prev
                  next
                  first
                  last
                  ellipsis
                  boundaryLinks
                  items={20}
                  maxButtons={5}
                  activePage={this.state.activePage}
                  onSelect={this.handleSelect} 
          />
```

### DIFF
--- a/src/Pagination.js
+++ b/src/Pagination.js
@@ -87,7 +87,7 @@ class Pagination extends React.Component {
 
     if (maxButtons) {
       let hiddenPagesBefore = activePage - parseInt(maxButtons / 2, 10);
-      startPage = hiddenPagesBefore > 1 ? hiddenPagesBefore : 1;
+      startPage = hiddenPagesBefore > 2 ? hiddenPagesBefore : 1;
       hasHiddenPagesAfter = startPage + maxButtons <= items;
 
       if (!hasHiddenPagesAfter) {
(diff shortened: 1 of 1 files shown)


kevinzwh:
Thoughts on this @jquense, @taion?

taion:
Looks fine to me. Note that we'll might revisit this API more heavily at some point for https://github.com/react-bootstrap/react-bootstrap/issues/1758. IMO it might be better for this logic to live in user space than be configured by a ton of different options here.

jquense:
lgtm, with the same caveat

PR #2190, opened 2016-09-02
Fix dependency security issue / Upgrade 'marked' -> v0.3.6

kevinzwh:
Noticed that VersionEye was reporting a security issue with marked v0.3.5 when I opened a PR for another bug.

[VersionEye report](https://www.versioneye.com/pullrequests/57c9053112b5260012202580)

Seems like there was a significant security bug in that version, so it's worth upgrading to 0.3.6 for the fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "less": "^2.7.1",
     "less-loader": "^2.2.3",
     "lodash": "^4.13.1",
-    "marked": "^0.3.5",
+    "marked": "^0.3.6",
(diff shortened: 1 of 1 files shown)


kevinzwh:
Not sure why VersionEye check is still failing after upgrading marked to 0.3.6...even the VersionEye page even says it's been patched for versions >= 0.3.6
https://www.versioneye.com/Node.JS/marked/0.3.6

taion:
Maybe merging this will make the problem go away.

PR #2195, opened 2016-09-05
Expose event prop from RootCloseWrapper in Dropdown

glenjamin:
This makes the prop added in https://github.com/react-bootstrap/react-overlays/pull/95 available to the dropdown component.

Please let me know if I've overlooked anything in the PR :)

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -94,6 +94,11 @@ const propTypes = {
    * a menu button.
    */
   role: React.PropTypes.string,
+
+  /**
+   * Which event when fired outside the component will cause it to be closed
+   */
+  rootCloseEvent: React.PropTypes.oneOf(['click', 'mousedown']),
 };
 
 const defaultProps = {
@@ -241,7 +246,7 @@ class Dropdown extends React.Component {
     });
   }
 
-  renderMenu(child, { id, onClose, onSelect, ...props }) {
+  renderMenu(child, { id, onClose, onSelect, rootCloseEvent, ...props }) {
     let ref = c => { this.menu = c; };
 
     if (typeof child.ref === 'string') {
@@ -265,6 +270,7 @@ class Dropdown extends React.Component {
       onSelect: createChainedFunction(
         child.props.onSelect, onSelect, this.handleClose,
       ),
+      rootCloseEvent
     });
   }
 
@@ -281,6 +287,7 @@ class Dropdown extends React.Component {
       role,
       bsClass,
       className,
+      rootCloseEvent,
       children,
       ...props
     } = this.props;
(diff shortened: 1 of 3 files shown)


jquense:
sorry I don't quite remember the conversation around this one. Can I get the TLDR on why this should get exposed up in react-bootstrap? :)

taion:
It's to allow hiding the dropdown on mousedown on backdrop rather than click.

jquense:
sure but why expose that here? I don't care that much just want to make sure we have a rational for adding an PAI that bootstrap doesn't have down somewhere

taion:
This is the only component in R-B that uses RootCloseWrapper directly.

taion:
@glenjamin 

I guess you'd also need to add this to `<Overlay>` in react-overlays/&c.? Or are you only concerned w/dropdowns for the time being?

This change LGTM to me. I'd have written the tests to just assert that the `event` prop is passed correctly to `<RootCloseWrapper>`, but this is better I guess though at the cost of more verbosity.

glenjamin:
You don't get the same issue with Overlay, because it covers everything and is the only thing you can click on.

With RootCloseWrapper, the mousedown can be triggering another action on a different component - and then it looks odd when the other action begins onmousedown but the dropdown doesn't hide until mouseup when the click completes. It's especially noticeable if the action is a drag!

taion:
Ah – I guess you're not using popovers with rootClose?

glenjamin:
I'm not, but happy to expand the PR if that's preferred.

taion:
That wouldn't actually be in this repo anyway, mostly.

glenjamin:
I've resolved the merge conflict on this now - if there's something else blocking a merge please let me know.

jquense:
still kinda wondering why here. we are saying no to exposing modal backdrop classes, is this different?

taion:
I think the right answer would be to pull the base logic for `<Dropdown>` into react-overlays. I don't really want to do that, though.

As it is, to do this, I think the user would have to re-implement quite a lot more code.

It's not fundamentally different, but I think the tradeoff is in favor of adding this here, but less so for `<Modal>`.

glenjamin:
Modal is different because the Overlay covers everything - there's only a difference between mouse click and mousedown when it is possible to interact with some other component onmousedown

taion:
@jquense was talking about why we rejected adding `backdropClassName` to `<Modal>`.

jquense:
Ok I'm :thumbsup: on this. I do agree the more correct way to handle this is a generic dropdown in RO but that's tough for a bunch of reasons leaving here the only reasonable place to expose stuff

src/Dropdown.js
```diff
@@ -94,6 +94,11 @@ const propTypes = {
    * a menu button.
    */
   role: React.PropTypes.string,
+
+  /**
+   * Which event when fired outside the component will cause it to be closed
+   */
+  rootCloseEvent: React.PropTypes.oneOf(['click', 'mousedown']),
```
taion: one of these days i'm going to rename this to `clickOutside` or something... anything really

PR #2212, opened 2016-09-13
Fix logic for determining child active state

### LINKED ISSUE #2201: 0.30.3 breaks react-router-bootstrap LinkContainers inside NavDropdown
I have `<LinkContainer>` elements under `<NavDropdown>`s. When I updated to 0.30.3, the `<LinkContainers>`s will have `active={false}` props passed on them. This breaks them because they will never show up as active now.

Rolling back to 0.30.2 fixes the problem.

taion:
Fixes #2201

### DIFF
--- a/docs/src/SubNav.js
+++ b/docs/src/SubNav.js
@@ -52,10 +52,10 @@ class SubNav extends React.Component {
       return true;
     }
 
-    if (props.children) {
-      return ValidComponentChildren.some(props.children, child => (
-        this.isActive(child, activeKey, activeHref)
-      ));
+    if (ValidComponentChildren.some(props.children, (child) => (
+      this.isActive(child, activeKey, activeHref)
+    ))) {
+      return true;
     }
 
     return props.active;
--- a/src/NavDropdown.js
+++ b/src/NavDropdown.js
@@ -30,10 +30,10 @@ class NavDropdown extends React.Component {
       return true;
     }
 
-    if (props.children) {
-      return ValidComponentChildren.some(props.children, child => (
-        this.isActive(child, activeKey, activeHref)
-      ));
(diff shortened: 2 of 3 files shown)


jquense:
Wait what's wrong with `some` here?

jquense:
also what is the deal with  CI recently

taion:
That `some` call will return `false` if all the `children` have `props.active === undefined`.

docs/src/SubNav.js
```diff
@@ -53,9 +53,16 @@ class SubNav extends React.Component {
...
-        this.isActive(child, activeKey, activeHref)
-      ));
+      let active;
+      ValidComponentChildren.forEach(props.children, (child) => {
+        if (active === undefined) {
+          active = this.isActive(child, activeKey, activeHref);
+        }
+      });
+
+      if (active !== undefined) {
```
jquense: if you already have this check maybe use `some` above? that or just add a note to the nuance of this. it wasn't obvious to me why this was needed, re: undefined vs false

taion: I'm not sure what you mean – how would I use `some`?

taion: The issue is that `[undefined].some(Boolean)` is `false`.

jquense: you can use to check if any child is active is so return true otherwise undefined? seems like all you need is the if block guarding against returning false right?

taion: I want the first defined value if there is one. If any child has active false, this function should return false.

taion: Oh, actually, I see what you mean. I don't actually need that.

taion:
Updated.

jquense:
lgtm

PR #2229, opened 2016-09-22
Allow setting carousel interval option to false

eddies:
According to the [docs](http://getbootstrap.com/javascript/#carousel-options), the `interval` option for Carousel can be a number or `false`. If false, "...carousel will not automatically cycle".

However, interval is currently set to `React.PropTypes.number`, preventing the user from setting interval to false.

This PR sets interval to `React.PropTypes.oneOfType([React.PropTypes.number, React.PropTypes.bool])`, and adds a test confirming that interval can be set to false.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -16,7 +16,7 @@ import ValidComponentChildren from './utils/ValidComponentChildren';
 const propTypes = {
   slide: React.PropTypes.bool,
   indicators: React.PropTypes.bool,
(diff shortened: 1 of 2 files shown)


taion:
I think you're just supposed to set it to `null` here.

eddies:
@taion ah, `null` works. It's a slight mismatch w/ the Bootstrap docs I linked to, but works for me.

PR #2241, opened 2016-09-30
use RCW disabled to avoid unnecessary unmount

jquense:


jquense:
can we kill this version eye thing??

taion:
I deleted the web hook... let's see if that works.

saxicek:
This change broke preact compatibility. Not sure why but 0.30.3 DropdownMenu works in preact while 0.30.4 does not.

taion:
0.30.4 is broken; it was untagged and the current release is 0.30.5.

ngduc:
@saxicek just curious if the latest version works with Preact?

saxicek:
@ngduc Tested with react-bootstrap@0.30.7, preact@7.1.0 and preact-compat@3.13.0 and it still does not work. Also tried with 0.30.3 which does not work either so now I am not sure what causes the issue.

My test is described at https://github.com/developit/preact-compat/issues/146 if you want to replicate and dive deeper.

jquense:
this definitely works with React, because we have behavior depending on the components saying mounted. Not sure about Preact, but there isn't much we can do on that front.

saxicek:
Yes, I can confirm it works with React. I was only discussing preact-compat compatibility where dropdown menu does not work.

jquense:
sure :) just wanted to add that i don't think we had a regression somewhere, so soundly in Preacts compat layer

PR #2242, opened 2016-09-30
let initial click propagate through before changing visibility

jquense:


jquense:
I think this makes more sense here then in the RCW itself

jquense:
dang tests

taion:
What's the specific issue here? Do the dropdowns not open any more, or do they not close any more?

jquense:
they don't open, because the `open` is pushed to the Menu before the click event triggering it bubbles up to the document. So the RCW sees the click and toggles it to closed immediately

test/DropdownButtonSpec.js
```diff
@@ -122,11 +129,13 @@ describe('<DropdownButton>', () => {
     const node = ReactDOM.findDOMNode(instance);
 
     const buttonNode = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'BUTTON');
-    ReactTestUtils.Simulate.click(buttonNode);
+    await simulateClick(buttonNode);
```
jquense: this feels like the best example of async/await being awesome :P

jquense:
does this seem reasonable? I want get a fixed release out before folks start complaining. If we need to bikeshed more I can just revert the other PR as well

taion:
Give me a sec.

jquense:
fixed lower down in react-overlays 👍

PR #2244, opened 2016-09-30
Remove unnecessary noWrap

taion:


taion:
Should be good to go now.

Issue #1587, opened 2016-01-05
Add support for checkbox/radio buttons


richardscarrott:
i.e. http://getbootstrap.com/javascript/#buttons-checkbox-radio

taion:
Why don't you try looking at our docs before asking this kind of question on the issue tracker...

taion:
Oh, I see - this is different. My bad. I don't believe we have this.

egeriis:
I would like this too. Open to PR?

taion:
Always.

konung:
@taion Ha - Looked through the issues but missed this one. thanks for merging. 

@egeriis @richardscarrott  - gentleman - have you had any luck with this?

egeriis:
@konung Yes, we've had some luck. We've build the component, but didn't get to do a PR yet. Will take a look at this moment :)

egeriis:
@taion I've looked into this now. These are my findings.

There are a lot of differences from a regular button:

1. Instead of being a `<button>` element, these checkboxes/radio buttons are `<label>`s with an `<input type="checkbox|radio">` as child
2. The `<label>` requires a parent element with a `data-toggle="buttons"` attribute
3. This parent element needs to be present, and can be either a _button group_ or just a regular `<div>`

These differences makes me consider that these special buttons should be their own element(s).

-----

Here's an outline of my suggested implementation. Naming is not thought through at all 🙂

**Changes to `<Button>`**
Allow `<Button>` to have `componentClass` property, which provides the ability to render it as `<label>`.

**Add new element `<CheckboxRadioButton>`**
This element would basically render:

```js
<Button componentClass="label" bsStyle={this.props.bsStyle} bsSize={this.props.bsSize} active={this.props.active}>
  <input autocomplete="off" type={this.props.type} value={this.props.value} onChange={this.props.onChange} checked={this.props.active}>
</Button>
```

**Consider adding `addToggleWrapper` property to `<CheckboxRadioButton>`**
`addToggleWrapper` will control whether the component will wrap the `<Button>` in a `<div data-toggle="buttons">` element. This will default to `true`.

**Handle `<CheckboxRadioButton>` when adding as child to `<ButtonGroup>`**
Set `addToggleWrapper` to `false` when `<CheckboxRadioButton>` is a child of a `<ButtonGroup>` and ensure that the `<ButtonGroup>` will have the `data-toggle="buttons"` attribute set.

[...]

PR #2252, opened 2016-10-03
Add toggle button components

jquense:
I’m tried of rewriting this component all the time :P

### DIFF
--- a/docs/examples/.eslintrc
+++ b/docs/examples/.eslintrc
@@ -62,6 +62,8 @@
     "TabPane": false,
     "Tabs": false,
     "Thumbnail": false,
+    "ToggleButton": false,
+    "ToggleButtonGroup": false,
     "Tooltip": false,
     "Well": false,
 
--- a/docs/examples/ButtonGroupToggleCheckbox.js
+++ b/docs/examples/ButtonGroupToggleCheckbox.js
@@ -1,0 +1,10 @@
+const toggleButtonGroup = (
+  <ToggleButtonGroup type="checkbox" defaultValue={[1, 3]}>
+    <ToggleButton value={1}>Checkbox 1 (pre-checked)</ToggleButton>
+    <ToggleButton value={2}>Checkbox 3</ToggleButton>
+
+    <ToggleButton value={3}>Checkbox 3 (pre-checked)</ToggleButton>
+  </ToggleButtonGroup>
+);
+
+ReactDOM.render(toggleButtonGroup, mountNode);
--- a/docs/examples/ButtonGroupToggleRadio.js
+++ b/docs/examples/ButtonGroupToggleRadio.js
@@ -1,0 +1,12 @@
+const buttonGroupInstance = (
+  <ToggleButtonGroup type="radio" name="options" defaultValue={1}>
+    <ToggleButton value={1}>
+      Radio 1 (pre-checked)
+    </ToggleButton>
+    <ToggleButton value={2}>Radio 3</ToggleButton>
+
+    <ToggleButton value={3}>Radio 3</ToggleButton>
+  </ToggleButtonGroup>
+);
+
+ReactDOM.render(buttonGroupInstance, mountNode);
--- a/docs/src/ReactPlayground.js
+++ b/docs/src/ReactPlayground.js
@@ -65,6 +65,8 @@ const Table = require('../../src/Table');
 const TabPane = require('../../src/TabPane');
 const Tabs = require('../../src/Tabs');
 const Thumbnail = require('../../src/Thumbnail');
+const ToggleButton = require('../../src/ToggleButton');
+const ToggleButtonGroup = require('../../src/ToggleButtonGroup');
 const Tooltip = require('../../src/Tooltip');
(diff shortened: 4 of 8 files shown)


taion:
ref https://github.com/react-bootstrap/react-bootstrap/issues/1587

taion:
Is it really better to have `<ToggleButton>` rather than some sort of styling props for `<Checkbox>` and `<Radio>`? The actual markup for the buttons seems almost identical to that for inline checkboxes and radios – just the CSS classes are different.

jquense:
I should not that the API here differs from upstream, in that it's more restrictive. here you must provide a `value` to the children which i also use like `eventKey`...I'm not at all sure why you would want radio/checkbox groups with no values but upstream does allow it.

jquense:
> Is it really better to have <ToggleButton> rather than some sort of styling props for <Checkbox> and <Radio>?

I don't know, do you think it's clear to add the button classes to those components/ remove the existing classes?

taion:
I dunno. The other bit is that we took out `<RadioGroup>`, so it's a bit odd to have group-style management here only in the context of toggle buttons.

We could add it back if it's a useful primitive even when using e.g. React-Formal and Redux Form, though, but it probably shouldn't be an "only-for-toggle-buttons" thing.

jquense:
The problem with the Group components is the dang styling that is sort of intractable. I agree though that it's odd to have one and not the other, _except_ that twbs doesn't have a RadioGroup component specifically.

I don't honestly really understand what the point of this component is in the context of React. Like is the checkbox/radio inputs bit more of an implementation detail that lets you do some native state tracking or is this styling for these form inputs? Personally I don't need the form wrapping bit, but I didn't want to jsut remove it lots of ppl seems to still use normal html forms with React :p

drsloth:
Hey all, where would you say the process is to get this implemented? This is exactly what I have need for

jquense:
> Hey all, where would you say the process is to get this implemented?

if by "this" you mean the toggle buttons, i've mostly implemented, it just needs tests

taion:
@jquense 

I think we can re-use the logic in `<Checkbox>` and `<Radio>`. At a glance, these render correctly:

```js
<Button componentClass={Checkbox} bsStyle="primary" inline active>Left</Button>
```

The other thing, I see that the actual button components in our context have to be controlled because `.active` has to get attached to the `label`.

However, given that, I think it'd make more sense to have the "group" component be generic, and have the `<ButtonGroup>` version just be a special case there.

Also, for checkboxes it should probably be an object rather than an array.

taion:
IOW, this look right:

```js
<ButtonGroup data-toggle="buttons">
  <Button componentClass={Checkbox} bsStyle="primary" inline active>Left</Button>
  <Button componentClass={Checkbox} bsStyle="primary" inline>Middle</Button>
  <Button componentClass={Checkbox} bsStyle="primary" inline>Right</Button>
</ButtonGroup>
```

but obviously doesn't work.

jquense:
The problem with that is you get a bunch of checkbox classes that shouldn't be added.

I think it makes sense to roll it into ButtonGroup.

> Also, for checkboxes it should probably be an object rather than an array.

not sure what you mean by that, like the value, what would be the key/value?

taion:
`<ButtonGroup>` feels weird because the core of the logic is not button-specific at all.

For checkboxes, I think it ought to be `{[eventKey]: bool}`... with the limitation that eventKeys are all strings, I guess.

jquense:
putting aside the wonkiness of this whole component for second, upstream has it has "part of th buttonGroup" so that makes sense in my head.

For checkbox values, I think it has to an array of the checkbox values, that's usually how like named checkboxes are traditionally serialized in forms (e.g. jquery will give you the array). It's like a multiselect sort of thing. Admittedly there is a tension between that and unnamed checkboxes, which wouldn't work here. However i'd suggest that this component makes no sense as a form input without named checkboxes.

jquense:
or rather if we treat it like a multiselect, vs a collection of checked things, its more useful in other contexts i think, like react-formal ;)

taion:
Well, okay, if other places do the array of values...

I guess is the idea that, in the general case, there's not much point to the non-presentational bits of the "group" logic here?

taion:
This LGTM net of tests. Should we add http://getbootstrap.com/javascript/#buttons-single-toggle as well?

jquense:
>  Should we add http://getbootstrap.com/javascript/#buttons-single-toggle as well?

Maybe the ToggleButton here covers that? Should I call this CheckboxButton and make ToggleButton the none input version? Might be confusing for folks i dunno

taion:
I dunno.

jquense:
I say leave off the single toggle for now since it's Essentially just a button with an active prop and aria-pressed. I'll add tests here also

taion:
Sounds good.

jquense:
updated and should be good. Who knows what codecov's deal is

taion:
@jquense Err what happened to the tests here?

jquense:
wtf?!?

taion:
They got lost?

PR #2258, opened 2016-10-06
Fix "Unknown props" warnings with Navbar Form, Text & Link

leth:
I see `Unknown props 'active', 'activeKey', 'activeHref' on <div> tag.` warnings when using Navbar.Form

I'm not certain this is the best way of fixing this, or whether this will have any fallout, I'd appreciate any advise here.


jquense:
I think you are seeing those warnings because you are using them inside of a `Nav` component, which is not valid

leth:
Ah I had no idea, Thanks!

PR #2268, opened 2016-10-09
Fixes #2256 by adding SR controls by default

### LINKED ISSUE #2256: Carousel component does not include all accessibility labels
I'm currently working on bringing a project using React Bootstrap up to ADA compliance. One of the issues we have is the Carousel component doesn't render all of the accessibility labels given in the example template in the Bootstrap documentation, specifically the "Next" and "Previous" labels.

Are these labels actually present but perhaps I'm missing some option to have them render? If not, I'd be happy to see if I can provide a pull request to bring this functionality in.

ghusse:
Adds SR controls by default, following the example on the bootstrap documentation:

https://getbootstrap.com/javascript/#carousel

I changed the API of the component, allowing to pass either a single node, or an array of nodes.

This way, it will not change the DOM tree in websites already using this component, but still allow to customize SR labels.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -36,14 +36,8 @@ const propTypes = {
   activeIndex: React.PropTypes.number,
   defaultActiveIndex: React.PropTypes.number,
   direction: React.PropTypes.oneOf(['prev', 'next']),
-  prevIcon: React.PropTypes.oneOfType([
-    React.PropTypes.node,
-    React.PropTypes.arrayOf(React.PropTypes.node),
-  ]),
-  nextIcon: React.PropTypes.oneOfType([
-    React.PropTypes.node,
-    React.PropTypes.arrayOf(React.PropTypes.node),
-  ]),
+  prevIcon: React.PropTypes.node,
+  nextIcon: React.PropTypes.node,
 };
 
 const defaultProps = {
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -36,8 +36,14 @@ const propTypes = {
   activeIndex: React.PropTypes.number,
   defaultActiveIndex: React.PropTypes.number,
   direction: React.PropTypes.oneOf(['prev', 'next']),
-  prevIcon: React.PropTypes.node,
-  nextIcon: React.PropTypes.node,
+  prevIcon: React.PropTypes.oneOfType([
+    React.PropTypes.node,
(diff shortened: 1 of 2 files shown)


src/Carousel.js
```diff
@@ -36,8 +36,14 @@ const propTypes = {
   activeIndex: React.PropTypes.number,
   defaultActiveIndex: React.PropTypes.number,
   direction: React.PropTypes.oneOf(['prev', 'next']),
-  prevIcon: React.PropTypes.node,
-  nextIcon: React.PropTypes.node,
+  prevIcon: React.PropTypes.oneOfType([
+    React.PropTypes.node,
+    React.PropTypes.arrayOf(React.PropTypes.node),
```
taion: you don't need this, node supports arrays out of the box

src/Carousel.js
```diff
@@ -47,8 +53,14 @@ const defaultProps = {
   wrap: true,
   indicators: true,
   controls: true,
-  prevIcon: <Glyphicon glyph="chevron-left" />,
-  nextIcon: <Glyphicon glyph="chevron-right" />,
+  prevIcon: [
+    <Glyphicon glyph="chevron-left" key="icon" />,
+    <span className="sr-only" key="sr">Previous</span>,
```
taion: I'd rather we just always render this I think

jquense: we should also expose the message as a prop so folks can localize it

ghusse:
That was my first idea, but sometimes the label is not necessary : if you
use an image and set the alt property, (I think) you don't need the sr
label.
This solution looks like the more easily customizable without adding any
property on the component.
And you can also localize it as you wish.

src/Carousel.js
ghusse: By setting the `prevIcon` property, they can customize and change the label

ghusse:
I removed the unnecessary proptype declaration.

Do you want me to add 2 more properties, `prevLabel` and `nextLabel`, or is it OK this way?

ghusse:
`closeLabel` on alerts
`toggleLabel` on split buttons
`aria-label` on Modal.Header

taion:
I think @jquense is right then. Maybe have `prevLabel` and `nextLabel`, and omit the label node if the corresponding label prop is `null`?

ghusse:
Closed in favor of #2275

PR #2269, opened 2016-10-09
Fixes #2209 by adding collapsed to className

### LINKED ISSUE #2209: Accordion Panel .collapsed class missing
Just updated to 0.30.3 but it seems fixes #2133 and #2137 are not effective.
Anchor toggles between true and false values for aria-expanded and aria-selected, but no collapsed class appears.

ghusse:
Adds `collapsed` in the className following the example in bootstrap docs.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -99,6 +99,7 @@ class Panel extends React.Component {
         role={role}
         href={id && `#${id}`}
         onClick={this.handleClickTitle}
+        className={ expanded ? '' : 'collapsed' }
         aria-controls={id}
         aria-expanded={expanded}
         aria-selected={expanded}
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -94,16 +94,21 @@ class Panel extends React.Component {
   }
 
   renderAnchor(header, id, role, expanded) {
+    const props = {
+      role,
+      href: id && `#${id}`,
+      onClick: this.handleClickTitle,
+      'aria-controls': id,
+      'aria-expanded': expanded,
+      'aria-selected': expanded,
+    };
+
+    if (!expanded) {
+      props.className = 'collapsed';
+    }
+
     return (
-      <a
-        role={role}
-        href={id && `#${id}`}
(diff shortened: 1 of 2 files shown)


src/Panel.js
```diff
@@ -99,6 +99,7 @@ class Panel extends React.Component {
         role={role}
         href={id && `#${id}`}
         onClick={this.handleClickTitle}
+        className={ expanded ? '' : 'collapsed' }
```
jquense: can we use `classNames(expanded && 'collapsed')`? it'll prevent having an empty `class` attribute.

ghusse: I just wanted to follow the docs.

Is it an issue to have an empty className? Why?

jquense: its just unnecessary if avoidable, cleaner markup.

ghusse:
I changed the way className is added, by only adding it when necessary.

src/Panel.js
```diff
@@ -94,15 +94,21 @@ class Panel extends React.Component {
...
+      role,
+      href: id && `#${id}`,
+      onClick: this.handleClickTitle,
+      'aria-controls': id,
+      'aria-expanded': expanded,
+      'aria-selected': expanded,
+    };
+
+    if (!expanded) {
+      props.className = 'collapsed';
```
taion: you don't need to do all this, just keep the previous code but do `className={expanded ? null : 'collapsed'}` if you really don't want to use `classNames` there.

test/PanelSpec.js
```diff
@@ -284,6 +284,7 @@ describe('<Panel>', () => {
       );
       const anchor = ReactDOM.findDOMNode(instance).querySelector('.panel-title a');
       assert.equal(anchor.getAttribute('aria-expanded'), 'true');
+      assert.equal(anchor.getAttribute('class'), null);
```
taion: just check `anchor.className` as elsewhere

ghusse:
Closed in favor of #2276

PR #2275, opened 2016-10-11
Adds prevLabel and nextLabel on carousel for screen readers

ghusse:
Replaces #2268

Creates 2 more properties on carousel: `prevLabel` and `nextLabel`, for screen readers.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -37,7 +37,19 @@ const propTypes = {
   defaultActiveIndex: React.PropTypes.number,
   direction: React.PropTypes.oneOf(['prev', 'next']),
   prevIcon: React.PropTypes.node,
+  /**
+   * Label shown to screen readers only, can be used to show the previous element
+   * in the carousel.
+   * Set to null to deactivate.
+   */
+  prevLabel: React.PropTypes.string,
   nextIcon: React.PropTypes.node,
+  /**
+   * Label shown to screen readers only, can be used to show the next element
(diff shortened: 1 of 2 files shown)


src/Carousel.js
```diff
@@ -269,6 +288,7 @@ class Carousel extends React.Component {
           onClick={this.handlePrev}
         >
           {prevIcon}
+          {prevLabelNode}
```
jquense: looks good, could you move the `prevLabel && <span className="sr-only">` inline here, ditto for the nextLabel as well

taion: should this be an explicit `null` check? what behavior do we want if the user sets `prevLabel=""`?

ghusse:
It seems that I forgot 2 strings with double quotes. I'll fix it.

src/Carousel.js
ghusse: I did it on purpose, because I cannot see an useful usage of empty labels for screen readers. They'll be unable to read them!

That's also why I added a test on this.

ghusse: @jquense : ok, I'll change this

ghusse: @jquense done

taion:
I'm mostly okay with this. @jquense do you think we want the explicit `null` check instead of the falsiness check?

jquense:
i think the falsey check is fine, LGTM. thanks!

PR #2276, opened 2016-10-11
Fixes #2209 by adding collapsed in anchor className

### LINKED ISSUE #2209: Accordion Panel .collapsed class missing

ghusse:
Replaces #2269

Adds `collapsed` in panel anchor's className, to follow bootstrap examples.

`className` is set to null if there is no need to add `collapsed`

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -102,6 +102,7 @@ class Panel extends React.Component {
         aria-controls={id}
         aria-expanded={expanded}
         aria-selected={expanded}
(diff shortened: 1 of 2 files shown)


jquense:
lgtm

jquense:
thanks!

PR #2280, opened 2016-10-12
Feature: Add closeOnSelect & onSelect to Navbar API with tests & docs

kevinzwhuang:
**Update 10/17:**
This PR has been improved from its original implementation. To see a running demo of the current iteration, you can run this Plunker here: https://plnkr.co/edit/43i4XA

The proposal has been simplified and improved with these two new props to the `<Navbar>` API:

   > **onSelect: React.PropTypes.func**
   > A callback fired when a `<NavItem>` grandchild is selected inside a child
   > `<Nav>`. Should be used to execute complex toggling or other miscellaneous
   > actions desired after selecting a `<NavItem>`. Does nothing if no `<Nav>`
   > & `<NavItem>` children exist. The callback is called with an eventKey,
   > which is a prop from the `<NavItem>`, and an event.
   >
   > ```js
   > function (
   > 	Any eventKey,
   > 	SyntheticEvent event?
   > )
   > ```
   >
   > For basic toggling after all `<NavItem>` select events, try using
   > toggleOnSelect.

-----
   > **toggleOnSelect: React.PropTypes.bool**
   >
   > Fires the onToggle callback whenever a `<NavItem>` inside the child `<Nav>` is
   > selected. Does nothing if no `<Nav>` & `<NavItem>` children exist.
   >
   > The onSelect callback should be used instead for more complex operations
   > desired for after `<NavItem>` select events.
   >

/Update

-----
Description:
----
Auto toggling as a behavior in react-bootstrap has been discussed at length in #1301 & #1692. @jquense and @taion have made good points against implementing it into the API due to complexities of different cases. There have also been valid arguments made for a simple version of this behavior to have available for users that does not need to catch every edge case. I think there is value to having a simple implementation of this expected behavior (normal click menus) for many users.

After reading over the discussion, I'd like to propose a simple addition to the Navbar API - `lazyAutoToggle`. I thought of naming it `lazy` to make it very apparent that it goes about autotoggling in a nondiscriminant & noncomplex way on `<Navbar.Collapse>` and its subcomponents. It does not attempt to be an omni-solution for all toggling needs.

Demo:
----
I've written a demo of this functionality in the plnkr. You can run the example at the link below:
https://plnkr.co/edit/vDejOB

There you can compare the functionality of a `<Navbar>` with & without `lazyAutoToggle` on.
It uses the built version of `react-bootstrap` from this PR.

How to use:
---
Simply add `lazyAutoToggle` to a `<Navbar>` with a child `<Navbar.Collapse>`.

Changes Proposed:
----
- [x] Add `lazyAutoToggle` prop to Navbar
- [x] Add `lazyAutoToggle` to `childContext` of Navbar
- [x] Check for `lazyAutoToggle` and `onToggle` from  `this.context.$bs_navbar` in `<Navbar.Collapse>` to fire `onToggle` when both exist

This solution is fairly straightforward. It involves adding the prop value of `lazyAutoToggle` to the `childContext` of `<Navbar>` and its children - particularly for use by `<Navbar.Collapse>`.

`<Navbar.Collapse>` will then fire `onToggle` from its parent `<Navbar>` if and only if `lazyAutoToggle` && `onToggle` exist in `this.context.$bs_navbar`.

To support this new prop, I've also added:

- [x] Test for `lazyAutoToggle` on `Navbar.Collapse` clicks
- [x] Test for `lazyAutoToggle` on `Navbar.Collapse` subcomponent click + onClick events
- [x] Documentation for `lazyAutoToggle` in `Navbar` propTypes.

Not included:
----
Since this is a dead simple solution, this will not attempt to cover any complex cases where individual items might need to prevent toggling. I can foresee a roadmap towards providing an array prop of components that should be ignored by lazyAutoToggle, but that is not a part of this PR.

As a single prop solution, `lazyAutoToggle` works extremely well with regular links, buttons, etc. (normal click-related menu items).

---
cc: @jquense and @taion

### DIFF
--- a/src/Nav.js
+++ b/src/Nav.js
@@ -93,6 +93,8 @@ const defaultProps = {
 const contextTypes = {
   $bs_navbar: React.PropTypes.shape({
     bsClass: React.PropTypes.string,
+    lazyAutoToggle: React.PropTypes.bool,
+    onToggle: React.PropTypes.func,
   }),
 
   $bs_tabContainer: React.PropTypes.shape({
@@ -319,6 +321,7 @@ class Nav extends React.Component {
           const childOnSelect = createChainedFunction(
             child.props.onSelect,
             onSelect,
+            navbar && navbar.lazyAutoToggle && navbar.onToggle || null,
             tabContainer && tabContainer.onSelect
           );
 
--- a/src/Nav.js
+++ b/src/Nav.js
@@ -93,8 +93,6 @@ const defaultProps = {
 const contextTypes = {
   $bs_navbar: React.PropTypes.shape({
     bsClass: React.PropTypes.string,
-    toggleOnSelect: React.PropTypes.bool,
-    onToggle: React.PropTypes.func,
     onSelect: React.PropTypes.func,
   }),
 
@@ -322,7 +320,7 @@ class Nav extends React.Component {
           const childOnSelect = createChainedFunction(
             child.props.onSelect,
             onSelect,
-            navbar && (navbar.onSelect || navbar.toggleOnSelect && navbar.onToggle) || null,
+            navbar && navbar.onSelect,
             tabContainer && tabContainer.onSelect
           );
 
--- a/src/Nav.js
+++ b/src/Nav.js
@@ -93,8 +93,9 @@ const defaultProps = {
 const contextTypes = {
   $bs_navbar: React.PropTypes.shape({
     bsClass: React.PropTypes.string,
-    lazyAutoToggle: React.PropTypes.bool,
+    toggleOnSelect: React.PropTypes.bool,
     onToggle: React.PropTypes.func,
+    onSelect: React.PropTypes.func,
   }),
 
   $bs_tabContainer: React.PropTypes.shape({
@@ -321,7 +322,7 @@ class Nav extends React.Component {
           const childOnSelect = createChainedFunction(
             child.props.onSelect,
             onSelect,
-            navbar && navbar.lazyAutoToggle && navbar.onToggle || null,
+            navbar && (navbar.onSelect || navbar.toggleOnSelect && navbar.onToggle) || null,
             tabContainer && tabContainer.onSelect
           );
 
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -57,21 +57,36 @@ const propTypes = {
    * @controllable navExpanded
    */
   onToggle: React.PropTypes.func,
-
+  /**
+   * A callback fired when a `<NavItem>` grandchild is selected inside a child
+   * `<Nav>`. Should be used to execute complex toggling or other miscellaneous
+   * actions desired after selecting a `<NavItem>`. The callback is called with
+   * an eventKey, which is a prop from the `<NavItem>`, and an event.
+   * ```js
+   * function (
+   * 	Any eventKey,
+   * 	SyntheticEvent event?
+   * )
+   * ```
+   * For basic toggling after all `<NavItem>` select events, try using
+   * toggleOnSelect. Note: When onSelect is set, toggleOnSelect will do nothing.
+   */
+  onSelect: React.PropTypes.func,
+  /**
+   * Fires the onToggle callback whenever a <NavItem> inside the child <Nav> is
+   * selected. Does nothing if an onSelect callback is provided to `<Navbar>` or
+   * if no `<Nav>` & `<NavItem>` children exist.
+   *
+   * The onSelect callback should be used instead for more complex operations
+   * desired for after `<NavItem>` select events.
+   */
+  toggleOnSelect: React.PropTypes.bool,
   /**
    * Explicitly set the visiblity of the navbar body
    *
    * @controllable onToggle
    */
   expanded: React.PropTypes.bool,
-
-  /**
-   * Sets 'lazy auto-toggling' that nondiscriminantely fires the onToggle callback
-   * after the onClick & onSelect callbacks for every item within a child <Nav>.
-   * Does nothing if no child <Nav> is present. lazyAutoToggle should not be used
-   * for complex nav menus.
-   */
-  lazyAutoToggle: React.PropTypes.bool,
 
   role: React.PropTypes.string,
 };
@@ -83,7 +98,7 @@ const defaultProps = {
   staticTop: false,
   inverse: false,
   fluid: false,
-  lazyAutoToggle: false,
+  toggleOnSelect: false,
 };
 
 const childContextTypes = {
@@ -91,7 +106,8 @@ const childContextTypes = {
     bsClass: PropTypes.string,
     expanded: PropTypes.bool,
     onToggle: PropTypes.func.isRequired,
-    lazyAutoToggle: PropTypes.bool,
+    onSelect: PropTypes.func,
+    toggleOnSelect: PropTypes.bool,
   })
 };
 
@@ -103,14 +119,15 @@ class Navbar extends React.Component {
   }
 
   getChildContext() {
-    const { bsClass, expanded, lazyAutoToggle } = this.props;
+    const { bsClass, expanded, onSelect, toggleOnSelect } = this.props;
 
     return {
       $bs_navbar: {
         bsClass,
         expanded,
         onToggle: this.handleToggle,
-        lazyAutoToggle,
+        onSelect,
+        toggleOnSelect,
       },
     };
   }
@@ -135,7 +152,7 @@ class Navbar extends React.Component {
     } = this.props;
 
     const [bsProps, elementProps] = splitBsPropsAndOmit(props, [
-      'expanded', 'onToggle', 'lazyAutoToggle'
+      'expanded', 'onToggle', 'onSelect', 'toggleOnSelect'
     ]);
 
     // will result in some false positives but that seems better
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -63,7 +63,9 @@ const propTypes = {
    * `<Nav>`. Should be used to execute complex toggling or other miscellaneous
    * actions desired after selecting a `<NavItem>`. Does nothing if no `<Nav>`
    * & `<NavItem>` children exist. The callback is called with an eventKey,
-   * which is a prop from the `<NavItem>`, and an event.
+   * which is a prop from the `<NavItem>`, and an event. If you would like
+   * onSelect to only fire when in a mobile viewport, add a condition for it to
+   * fire only when `window.innerWidth` is less than 768.
    *
    * ```js
    * function (
@@ -72,13 +74,14 @@ const propTypes = {
    * )
    * ```
    *
-   * For basic toggling after all `<NavItem>` select events, try using
-   * toggleOnSelect.
+   * For basic toggling after all `<NavItem>` select events in mobile viewports,
+   * try using toggleOnSelect.
    */
   onSelect: React.PropTypes.func,
   /**
    * Fires the onToggle callback whenever a `<NavItem>` inside the child `<Nav>` is
-   * selected. Does nothing if no `<Nav>` & `<NavItem>` children exist.
+   * selected. Does nothing if no `<Nav>` & `<NavItem>` children exist or if
+   * the window is not in a mobile-sized viewport.
    *
    * The onSelect callback should be used instead for more complex operations
    * desired for after `<NavItem>` select events.
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -128,7 +128,7 @@ class Navbar extends React.Component {
         bsClass,
         expanded,
         onToggle: this.handleToggle,
-        onSelect: createChainedFunction(onSelect, toggleOnSelect && this.handleToggle || null),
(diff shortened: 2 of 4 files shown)


taion:
I don't think this is the right approach. I think you should hook into `onSelect` in `<Nav>`.

kevinzwhuang:
Hooking into `onSelect` in `<Nav>` is a good idea, I'll look into that later today.

kevinzwhuang:
You were totally right @taion, the hook on `<Nav>` is a much better approach. Overall, a way simpler strategy compared to the `<Navbar.Collapse>` idea and needs only 3 lines of code in `Nav.js`

I also updated the docs & test for this approach in the latest commit.
Here is an updated working demo in Plunker for you to run in your browser: https://plnkr.co/edit/AYyb2G

src/Nav.js
```diff
@@ -93,6 +93,8 @@ const defaultProps = {
 const contextTypes = {
   $bs_navbar: React.PropTypes.shape({
     bsClass: React.PropTypes.string,
+    lazyAutoToggle: React.PropTypes.bool,
```
taion: i'd just pass down an `onSelect` here

src/Navbar.js
```diff
@@ -65,6 +65,14 @@ const propTypes = {
    */
   expanded: React.PropTypes.bool,
 
+  /**
+   * Sets 'lazy auto-toggling' that nondiscriminantely fires the onToggle callback
+   * after the onClick & onSelect callbacks for every item within a child <Nav>.
+   * Does nothing if no child <Nav> is present. lazyAutoToggle should not be used
+   * for complex nav menus.
+   */
+  lazyAutoToggle: React.PropTypes.bool,
```
taion: I'd call this `toggleOnSelect` (or maybe `autoToggle`)

kevinzwhuang:
`toggleOnSelect` sounds good with me. Could you clarify a bit more on the `onSelect`, it doesn't seem `$bs_navbar` has an `onSelect` set from `<NavBar>`'s [getChildContext()](https://github.com/react-bootstrap/react-bootstrap/pull/2280/files#diff-9d651e5c92548c4d9e89560e0df67ce9R105)

taion:
@kevinzwhuang I mean add one. @jquense wdyt?

kevinzwhuang:
@taion ah I see what you mean. 👍 

In that case we might not even need a `toggleOnSelect`/`autoToggle` flag, since the exposed `onSelect` hook into `<Nav>`s chained onSelect function provides even more flexibility and allows users to do the exact same `autoToggle` functionality themselves by passing in the same callback for `onToggle` for the `onSelect` context (or whatever modified toggle function they want). No need for prescribing behavior into the API and lets users handle their own edge cases!

Though I think maybe it should be named `onNavSelect` or `onNavSelectCompleted` to denote that it's a hook into `<Nav>`'s onSelect chain and not just an `onSelect` callback meant for `<NavBar>` itself (which is what I thought you meant at first)

taion:
Personally I think `onSelect` is fine. I think it's pretty clear in context that `onSelect` on a `<Navbar>` would refer to the `<NavItem>`s under it.

I feel like this "toggle on select" behavior you're describing is common enough that it'd make sense to have as an option. It may not account for all edge cases, but it's probably what most users want.

kevinzwhuang:
Sounds good @taion. I'll work on updating this PR with that approach sometime this weekend.

kevinzwhuang:
Updated to the latest changes as discussed with `toggleOnSelect` & `onSelect`. I made it so that when an `onSelect` callback is provided, `toggleOnSelect` will be ignored (makes sense if you view `onSelect` as a more complex option to `toggleOnSelect`).

Docs & tests have been updated to reflect this new API. 

Also I've created an updated Plunker with the latest build of this PR. Works well with `NavItems` & `NavDropdown`/`MenuItems` mixed in. (I believe NavDropdown's are ignored by toggleOnSelect & onSelect due to it not having a select event inherently). 

In this latest demo, I've added a particular example using the `onSelect` callback to skip toggling for eventKey 2 - the potential for this callback is promising given it gives users access to the `eventKey` + `event`. You can try it out here: https://plnkr.co/edit/cOc8ZI

Let me know what you think @taion @jquense

kevinzwhuang:
Folded the `navbar && (navbar.onSelect || navbar.toggleOnSelect && navbar.onToggle) || null` logic into `handleSelect` in `<Navbar>`. 

Tests did not need to be modified since it's the same behaviour, updated plunker with latest build continues to work, which you can run here: https://plnkr.co/edit/PZhOUx

src/Navbar.js
```diff
@@ -100,6 +126,7 @@ class Navbar extends React.Component {
         bsClass,
         expanded,
         onToggle: this.handleToggle,
+        onSelect: this.handleSelect,
```
taion: sorry, one last thing – i think it might be better to make this just

```js
createChainedFunction(onSelect, toggleOnSelect && this.handleToggle)
```

I think it's pretty unlikely that users will specify both `toggleOnSelect` and `onSelect`, and I think it'd be a bit cleaner this way.

Otherwise, this is looking really good. Thanks!

taion:
Thanks!

kevinzwhuang:
Thanks for your guidance on this @taion! Any more thoughts on this @jquense, @taion?

src/Navbar.js
```diff
@@ -93,13 +121,14 @@ class Navbar extends React.Component {
...
   getChildContext() {
-    const { bsClass, expanded } = this.props;
+    const { bsClass, expanded, onSelect, toggleOnSelect } = this.props;
 
     return {
       $bs_navbar: {
         bsClass,
         expanded,
         onToggle: this.handleToggle,
+        onSelect: createChainedFunction(onSelect, toggleOnSelect && this.handleToggle || null),
```
jquense: should this actually toggle or only close the navbar? Theoretically you can't select something if its collapsed, but double clicks may get through, or other weird races.

test/NavbarSpec.js
```diff
@@ -254,6 +256,64 @@ describe('<Navbar>', () => {
...
+
+    ReactTestUtils.Simulate.click(ReactDOM.findDOMNode(link));
+
+    expect(navItemSpy).to.be.calledOnce;
+    expect(toggleSpy).to.be.calledOnce;
+    expect(toggleSpy).to.be.calledWith(true);
+  });
+
+  it('Should fire onSelect with eventKey for nav children', () => {
+    const selectSpy = sinon.spy();
```
jquense: does this work correctly with a NavDropdown?

kevinzwhuang:
I added a condition for `toggleOnSelect` to fire `this.handleToggle` only if `window` is not undefined(i.e. within a node build/test env) && `window.innerWidth < 768`, since it doesn't make sense for it to toggle when in a viewport larger than a mobile screen. Hope that makes sense @taion @jquense 

Updated plunker with this build version: https://plnkr.co/edit/4sn165?p=preview

src/Navbar.js
```diff
@@ -93,13 +124,19 @@ class Navbar extends React.Component {
   }
 
   getChildContext() {
-    const { bsClass, expanded } = this.props;
+    const { bsClass, expanded, onSelect, toggleOnSelect } = this.props;
+    const mobileToggle = () => {
+      if (typeof(window) !== 'undefined' && window.innerWidth < 768) {
```
jquense: I don't think we need this check.  it's too brittle anyway as a user can change the navbar collapse breakpoint to whatever they want so hard coding a pixel width here isn't ideal. 

I do think that it illustrates  a more general problem with this feature tho,  you don't really want to be toggling expanded state while the expansion is not relevant, however there isn't really any way to know that from JS

taion:
@jquense I don't care about Navbar buttons. How do those even show up on mobile/collapsed anyway? IMO it's a non-problem.

src/Navbar.js
kevinzwhuang: Yeah it's brittle hard-coded as is.

One approach I can think of dealing with is is exposing a prop `toggleOnSelectMaxWidth`. It's of `React.PropTypes.number` and defaults to 768. So then when `toggleOnSelect == true` the window.innerWidth check would compare with `toggleOnSelectMaxWidth`.

Or `toggleOnSelect` could be the numeric value itself, and default to 0, so it's falsey when not set.

taion: Does anything bad happen if we unnecessarily toggle collapsed state? The styling is a no-op on desktop views, plus the user has to opt-in to this behavior anyway.

kevinzwhuang: The whole navbar looks like it sort of re-renders when it tries to collapse while in a non-mobile viewport.

Here's an older plunker that demonstrates this: https://plnkr.co/edit/PZhOUx?p=preview

If we can figure out a way to fix this behavior in the lifecycle of the navbar, then we could get away with not having to deal with a arbitrary `innerWidth` breakpoint.

jquense: I'm really against users needing to input there grid breakpoint in order for this to work well. IMO it defeats the value of having this "add on" functionality if is comes with config, or a gotcha.

The only way i can think of tho to handle this is some sort of computedStyle sniffing to detect if the nav is in the right media query...that's not much less brittle tho :/

kevinzwhuang: I agree - users shouldn't have to set a breakpoint.

I think I found a solution to this problem. It goes back to the point you mentioned about toggling vs closing the navbar! We were tackling it from the wrong angle with `toggleOnSelect`, which is causing the navbar to toggle between `expanded` states even when the screen size is not appropriate. In fact, screen size should not matter - only the value of `expanded` should really matter.

The right functionality should be something more like a `closeOnSelect` or `collapseOnSelect`, which will set `expanded` to false, only if it was true. If `expanded` is false, then it does nothing. I'll update this PR with a commit to demonstrate this new approach shortly.

taion:
That's a great idea. `closeOnSelect` sounds better, too. 😄

kevinzwhuang:
Latest commit adds `closeOnSelect`.Works as expected :)
Here's a demo of the new build: https://plnkr.co/edit/SNvGZyLkiBmWOnyT86yG

src/Navbar.js
```diff
@@ -57,7 +58,36 @@ const propTypes = {
    * @controllable navExpanded
    */
   onToggle: React.PropTypes.func,
-
+  /**
+   * A callback fired when a `<NavItem>` grandchild is selected inside a child
```
taion: "descendant" rather than "grandchild" (since it's actually Navbar > Navbar.Collapse > Nav > NavItem, and there might even be a LinkContainer or so in there)

src/Navbar.js
```diff
@@ -57,7 +58,36 @@ const propTypes = {
...
+   * viewports, try using closeOnSelect.
+   * Note: If you are manually closing the navbar using this `OnSelect` prop,
+   * ensure that you are setting `expanded` to false and not *toggling* between
+   * true and false.
+   */
+  onSelect: React.PropTypes.func,
+  /**
+   * Sets `expanded` to false after the onSelect event of a child of `<Nav>`
+   * (such as `<NavItem>` or `<MenuItem>`). Does nothing if no `<Nav>` or
+   * `<NavItem>`/`<MenuItem>` children exist.
```
taion: `<MenuItem>` only goes under dropdowns actually

src/Navbar.js
kevinzwhuang: What about "...when a descendant of a child `<Nav>` is selected..."? This might refer to NavItem, LinkContainer, etc., basically any descendants that can be *selected* without referring to any specific ones.

src/Navbar.js
kevinzwhuang: Working off of what I brainstormed in the comment above, how's this?
"
Sets `expanded` to `false` after the onSelect event of a descendant of a child `<Nav>`. Does nothing if no `<Nav>` or `<Nav>` descendants exist.
"

src/Navbar.js
taion: Sounds great.

src/Navbar.js
taion: :+1:

jquense:
looks good to me. Thanks for the greate work. Let me know if you've like to be part of the react-bootstrap organization*. Its really low commitment, contribute as you have time. 

 (baring unlikely disagreement from other @react-bootstrap/collaborators)

kevinzwhuang:
Just updated the PR with those comment improvements discussed.

@taion @jquense Thank you for the help on this. I'd be glad to be a part of the org and contribute. :smile:

GerardoMachadoGandaria:
Does not seem to work on mobile device?

kevinzwhuang:
@GerardoMachadoGandaria could you provide some more details? I use this in my own apps and just tested this again myself on Android and iPhone and there's no problem.

Are you using `react-bootstrap@0.30.6`? What device are you using? Are you using the `closeOnSelect` or `collapseOnSelect` prop? The API has since been renamed to `collapseOnSelect` in #2308 - there's also a note about this at the top of this PR, so using `closeOnSelect` will be the incorrect prop. If you have some code you could share that would be great.

GerardoMachadoGandaria:
Sorry my fault. Works fine. 
Something was probably buffering an old version of the app. 
Restart of the server did the trick. Also. Thank you for this change!

kevinzwhuang:
Glad it worked out :smile:

PR #2316, opened 2016-10-31
Refactor Panel to use shouldRenderFill in renderBody function

andela-aabdulwahab:
In the Panel Component implementation, there is a `shouldRenderFill` function which I think was meant to be used in the `renderBody` function, but was not use. For a cleaner code base, get rid of it or put it to use. I'm putting it to use because it's a good code abstraction as it enhances code readability.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -149,7 +149,7 @@ class Panel extends React.Component {
 
     // Convert to array so we can re-use keys.
     React.Children.toArray(rawChildren).forEach(child => {
-      if (React.isValidElement(child) && child.props.fill) {
+      if (this.shouldRenderFill(child)) {
         maybeAddBody();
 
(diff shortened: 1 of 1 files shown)


jquense:
this is fine, however the logic in `shouldRenderFill` is different than the check it's replacing. `fill != null`, it's not an a change that I think should matter, but it does slightly change the existing semantics. I think it's probably better to remove the helper method since it's only used once.

andela-aabdulwahab:
@jquense that makes sense. I will remove the function and make another push.

andela-aabdulwahab:
@jquense do you think the code should stay, happy to take this PR down and work on something else.

jquense:
looks good to me :thumbsup:

taion:
Thanks!

Issue #2206, opened 2016-09-09
Docs: Move "placement" property from "Tooltip" to "Overlay Trigger"


skitterm:
From the [docs](https://react-bootstrap.github.io/components.html#tooltips), I thought I needed to put the "placement" prop on the Tooltip component itself. Only by looking at someone's example did I realize I have to put it on the OverlayTrigger. Can the docs be updated so that it's clear that the "placement" property should be on the OverlayTrigger and not the Tooltip component?

taion:
Interesting. It actually is on the overlay and the tooltip. I think we need to add the prop types to `<Overlay>`.

ghusse:
Does it make sense to have this property in other cases the `Overlay` is used? Or is it specific to tooltips?

kevinzwhuang:
Yep, `placement` is used for custom Overlays too - https://react-bootstrap.github.io/components.html#custom-overlays-overlay

taion:
fixed by #2318?

kevinzwhuang:
I think the new version of the docs still need to be released to the site repo

taion:
Oh... yeah, I'll do that in a bit. Looks like the docs are properly fixed though. Thanks!

PR #2318, opened 2016-11-04
Add placement PropType & default prop to Overlay for docs

kevinzwhuang:
Addresses #2206 

-----

This should display for `OverlayTrigger` too in the Docs, since it inherits from these props.


taion:
LGTM

PR #2322, opened 2016-11-08
Allow empty validationState on FormGroup

Legogris:
Sometimes it is reasonable with no validation feedback, for example before the user has entered any input.


taion:
Set it to null then

Legogris:
But of course. Thanks.

Issue #2219, opened 2016-09-18
NavDropdown and onMouseLeave broken between 0.29.5 and 0.30.2


acomito:
I have a `NavDropdown` that I need to have open the dropdown on hover, _but also stay open when hovering on the children_ (e.g. `MenuItems`, or `LinkContainer`). 

I was having trouble with this at first because I was using **onMouseOut** which closed the dropdown if you left the top-level nav item (the NavDropdown)-- but I was able to use **onMouseLeave**, which would keep the dropdown open on the nav item _and all of its children_. Great, problem solved.

I just updated from 0.29.5 to 0.30.3 and noticed this no longer works. I tested both MenuItems and MenuItems wrapped in LinkContainer and it seems broken either way. I downgraded to 0.30.2 and it still is not working, so at some point between 0.29.5 to 0.30.2 there seems to have been a change that is breaking my code.

Has anyone else experienced this? Here is some pseudo code:

```
//top-level imports
import React from 'react';
import { LinkContainer } from 'react-router-bootstrap';
import { Nav, NavItem, Row, NavDropdown, MenuItem } from 'react-bootstrap';


// INTERNAL COMPONENTS
//------------------------------

 class MyNavDropdown extends React.Component {
  constructor(props) {
      super(props);
      this.state = {
        isHovering: false
      };
    }

    render() {

    const onMouseOver = () => {
      console.log('onMouseOver fired');
      this.setState({ isHovering: true });
    };

    const onMouseLeave = () => {
      console.log('onMouseLeave fired');
      this.setState({ isHovering: false });
    };

    const dropDownClasses = this.state.isHovering ? 'open nav-links' : 'nav-links';

    return (
      <NavDropdown 
        className={dropDownClasses}
        onMouseEnter={onMouseOver}
        onMouseLeave={onMouseLeave}
        ref="myDropdown"
        eventKey="4" 
        title={this.props.title} 
        id="nav-dropdown" 
        noCaret={true}>
          {this.props.links}
      </NavDropdown>
    );
  }
};


const OpportunitiesLinks = () => (
            <MenuItem className="dropdownMenuItem" eventKey="4.2">Sponsors</MenuItem>
);


// EXPORTED COMPONENT
//------------------------------

export const PublicNavigation = ({festivals}) => (
  <Nav pullRight>
      <MyNavDropdown links={<OpportunitiesLinks />} title="Opportunities" />
  </Nav>
);
```

taion:
[...]

PR #2335, opened 2016-11-15
Pass onMouseEnter and onMouseLeave to <Dropdown>

taion:
Fixes https://github.com/react-bootstrap/react-bootstrap/issues/2219


travisdahl:
if you use this to set the open state you will get the following error...
`Failed prop type: You have provided a `open` prop to `Dropdown` without an `onToggle` handler. This will render a read-only field. If the field should be mutable use `defaultOpen`. Otherwise, set `onToggle``

PR #2337, opened 2016-11-16
Add inputRef to FormControl component

### LINKED ISSUE #1887: FormControl support for inputRef
Moving from `Input` to `FormControl` seems to have broken support for specifying an input ref. It looks like `Props` and `Radio` support this via the `inputRef` property, but there is no equivalent for `FormControl`.

sorahn:
This looks like what should be here.  I was unable to get the tests to run on master, or on my branch.  Here is a gist with the console output trying to run the tests.

https://gist.github.com/sorahn/a9a4f9d4d1fd0af24faa438336ca389d

Fixes #1887

### DIFF
--- a/src/FormControl.js
+++ b/src/FormControl.js
@@ -17,6 +17,14 @@ const propTypes = {
    * Uses `controlId` from `<FormGroup>` if not explicitly specified.
    */
   id: React.PropTypes.string,
+  /**
+   * Attaches a ref to the `<input>` element. Only functions can be used here.
+   *
+   * ```js
+   * <FormControl inputRef={ref => { this.input = ref; }} />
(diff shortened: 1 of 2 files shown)


jquense:
this LGTM. thoughts from everyone else?

taion:
Question is `inputRef` v `elementRef` or `nodeRef` (`domNodeRef`?) to me... I guess `inputRef` is semantically safer, since even if the markup changes, it would still point to the `<input>`?

Code LGTM

jquense:
I'd say the name is safe since the whole point of this ref is to imperatively get access to the `value` (yuck) so even if the markup changes it should always point to the input, like Checkbox

PR #2342, opened 2016-11-21
On Accordion title hover change cursor to pointer

zatziky:
It is quite confusing if you point at an accordion title and the cursor is a caret. It should be a pointer instead:

image

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -98,6 +98,7 @@ class Panel extends React.Component {
         aria-controls={id}
         aria-expanded={expanded}
         aria-selected={expanded}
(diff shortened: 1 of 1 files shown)


jquense:
Thanks for the contribution, we always appreciate pull requests. While I tend to agree that panel headers should often have pointer cursors. React-bootstrap doesn't provide or make changes to the bootstrap css or styles, we only implement the javascript components. 

you can wrap your panel headers in links (as suggested in the documentation) or provide a css override in your own code using the `.panel-header` selector. thanks!

zatziky:
Ok, I see. Thank you for explanation.

PR #2344, opened 2016-11-21
bump teaspoon and react

jquense:
wanna see if anything fails...I think uncontrollable will (maybe)


jquense:
why were we doing that...

taion:
legacy

jquense:
gtg?

PR #2347, opened 2016-11-23
feat(tabs): add componentClass support

acateland:
This contribution is a non breaking change.
The scope of this contribution is to help user define with wich html tag the use want to have the tab list elements rendered.

Exemple: rendering tabs as `button` instead of `a`

### DIFF
--- a/src/Tab.js
+++ b/src/Tab.js
@@ -25,6 +25,7 @@ class Tab extends React.Component {
     delete props.title;
     delete props.disabled;
     delete props.tabClassName;
+    delete props.tabComponentClass;
 
     return <TabPane {...props} />;
   }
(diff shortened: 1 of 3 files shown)


taion:
I'm :-1: on this. We really want to minimize the API here – users in need of flexibility should use the lower-level components.

acateland:
@taion this api was available on other components, so its a matter of ease of use and homogenization.

jquense:
hey thanks, glad you got the tests working :) and thanks for the PR. I do tend to agree with @taion here. We generally want to try and limit the API surface as much as possible as we have a ton of it in this library. I definitely see your use-case here, but its a hard line to draw; why not also allow tabPaneComponentClass for instance. In this case we already have a really robust and (I think) easy to use lower level Tab components that cover this use case without to much work

acateland:
@jquense its what i ended up to do, but since we as organisation will use this extensively, i will probably build an abstraction over those low level component for our ease of use.

taion:
given agreement that we don't want this, i'm closing ofr now

PR #2359, opened 2016-11-28
Don’t set aria-describedby unless the referenced element is in the DOM

### LINKED ISSUE #1827: Popover id causes broken ARIA reference
Hi,

The documentation for the Popover component states that the id prop is required for accessibility. But I think it actually causes an error. The problem is that when given an id, the child element (e.g. the button that triggers the popover) is given an aria-describedby attribute with that id. However, the Popover div that is assigned this id does not actually exist until the OverlayTrigger is triggered (i.e. the button is clicked).

This causes the error "Broken ARIA reference" when scanned with WebAIM's WAVE toolbar extension. WebAIM is one of the leaders in accessibility and are referenced in react-bootstrap's own roadmap, so it seems like this is an error that should be considered.

From the perspective of someone using a screen reader, if there are a few Popovers on a site, then there will be a few aria-describedbys that refer to something that doesn't exist yet. This will likely cause confusion.

Unfortunately I haven't been able to come up with a solution, but I look forward to discussing the problem. Right now I just don't use ids in my Popovers.

andrew-cc:
Having `aria-describedby` contain a DOM ID which is not present is invalid behavior, and creates an error in e.g. Google Accessibility Audit (see [Rule AX_ARIA_04](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_04)).

This fixes #1827.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -261,10 +261,11 @@ class OverlayTrigger extends React.Component {
 
     const child = React.Children.only(children);
     const childProps = child.props;
+    const triggerProps = {};
 
-    const triggerProps = {
-      'aria-describedby': overlay.props.id
-    };
+    if (this.state.show) {
+      triggerProps['aria-describedby'] = overlay.props.id;
+    }
 
     // FIXME: The logic here for passing through handlers on this component is
     // inconsistent. We shouldn't be passing any of these props through.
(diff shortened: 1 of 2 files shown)


jquense:
LGTM

andrew-cc:
Thanks for the speedy merge! 👍

aintnorest:
v0.30.7 was released Nov 22, 2016 just 6 days too soon :(
Any idea when we'll see a new release?

Issue #1889, opened 2016-05-11
FormControl doesn't correctly support input-sm input-lg


jquense:
Did we forget about these or is there another reason we didn't cover this?

taion:
It's because the Bootstrap CSS sucks – `.form-control` with `.input-sm` doesn't work well with how we do `bsClass`. The size variants on `<FormGroup>` and `<InputGroup>` by contrast look sane.

gtwilliams03:
I ran into this as well - react-bootstrap adds `.form-control-lg` instead of `.input-lg` to a `<FormControl>` inside of a `<FormGroup bsSize="large">` element

joelseq:
I think this should be closed since #2382 was merged

taion:
Oops, thanks!

PR #2382, opened 2016-12-08
Add support for input-sm and input-lg in FormControl

joelseq:
This is a fix that I came up with for #1889 that allows FormControl to have a bsSize prop. I know the solution probably may not be the most scalable and elegant but since this seems to be a one off case I thought it might work here.

Let me know if you think this is an acceptable fix, if not, I'm willing to try to find an alternative solution.

### DIFF
--- a/src/FormControl.js
+++ b/src/FormControl.js
@@ -6,6 +6,8 @@ import warning from 'warning';
 import FormControlFeedback from './FormControlFeedback';
 import FormControlStatic from './FormControlStatic';
 import { bsClass, getClassSet, splitBsProps } from './utils/bootstrapUtils';
+import { SIZE_MAP } from './utils/StyleConfig';
+import { prefix } from './utils/bootstrapUtils';
 
 const propTypes = {
   componentClass: elementType,
@@ -46,6 +48,7 @@ class FormControl extends React.Component {
       id = controlId,
       inputRef,
       className,
+      bsSize,
       ...props
     } = this.props;
 
@@ -60,6 +63,13 @@ class FormControl extends React.Component {
     let classes;
     if (type !== 'file') {
       classes = getClassSet(bsProps);
+    }
+
+    // If user provides a size, make sure to append it to classes as input-
+    // e.g. if bsSize is small, it will append input-sm
+    if (bsSize) {
+      const size = SIZE_MAP[bsSize] || bsSize;
+      classes[prefix({ bsClass: 'input' }, size)] = true;
     }
 
     return (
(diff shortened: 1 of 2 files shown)


src/FormControl.js
```diff
@@ -62,6 +64,13 @@ class FormControl extends React.Component {
       classes = getClassSet(bsProps);
     }
 
+    // If user provides a size, make sure to append it to classes as input-
+    // e.g. if bsSize is small, it will append input-sm
+    if (bsSize) {
+      const size = SIZE_MAP[bsSize] || bsSize;
+      classes[`input-${size}`] = true;
```
taion: use the `prefix` helper from `bootstrapUtils`

src/FormControl.js
```diff
@@ -46,6 +47,7 @@ class FormControl extends React.Component {
       id = controlId,
       inputRef,
       className,
+      bsSize,
```
taion: @jquense do you think `bsSize` is correct here, given that it's not using the `bsClass`? should this instead be `inputSize`?

src/FormControl.js
joelseq: @taion initially I was planning to but then `prefix` uses the bsClass in prop and I thought it might just be quicker to add this one line instead. I'll change it anyway :)

src/FormControl.js
jquense: I think it's correct from a user's perspective, and wrong from ours :P 

the className mismatch is a bug/implementation detail on bootstraps parts not an intentional "different thing" like validation state sort of is. my evidence: it's fixed in v4 https://v4-alpha.getbootstrap.com/components/forms/#control-sizing

test/FormControlSpec.js
```diff
@@ -77,4 +77,12 @@ describe('<FormControl>', () => {
     expect(instance.input.tagName).to.equal('INPUT');
   });
 
+  it('should properly display size of FormControl', () => {
+    $(
+      <FormControl id="foo" type="text" bsSize="lg" className="bar" />
+    )
+      .render()
+      .single('input#foo.bar.form-control.input-lg');
```
jquense: I'd simplify this selector to just `input.input-lg` so it's a bit easier to see that the important assertion bit is

src/FormControl.js
taion: fair enough. i'm fine with this then

joelseq:
@taion @jquense made the changes that y'all requested :)

jquense:
LGTM

PR #2388, opened 2016-12-09
Add prop that prevents dropdown from closing on click

codebreach:
this is used for multiple select dropdowns
the prop added is `preventCloseOnItemSelect `

Please let me know if you have a better name for it

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -101,6 +101,11 @@ const propTypes = {
   rootCloseEvent: React.PropTypes.oneOf(['click', 'mousedown']),
 
   /**
+   * Don't close the menu when an item onSelect is fired
+   */
+  preventCloseOnItemSelect: React.PropTypes.bool,
+
+  /**
    * @private
    */
   onMouseEnter: React.PropTypes.func,
@@ -111,7 +116,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  componentClass: ButtonGroup,
+  componentClass: ButtonGroup
 };
 
 class Dropdown extends React.Component {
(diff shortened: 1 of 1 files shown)


codebreach:
Turns out this doesnt work because of `RootCloseWrapper`
Any ideas on how to make this work?

jquense:
thanks for the PR. I think this approach isn't quite ideal, generally we'd like behavior like this to be controlled via the dropdown `onToggle`/`open` props which let the user more flexibily control when the dropdown closes in response to it's events. The one missing bit there however, is the information about what triggers an onToggle. 

You can check out https://github.com/react-bootstrap/react-bootstrap/issues/1490 for some conversation about this.

codebreach:
Right. I was trying to use the same approach but realized that there is no way to tell if the onToggle is called by the component internally or externally.

Looking at #1490 it seems that the best way to do this would be to pass a source argument to onToggle and let the consumers use that. 

I can try doing that in this PR if you are willing to merge that?

codebreach:
Also, what do you expect source argument look like?
Like how do we tell its coming from the component and not form an external event?

jquense:
so generally i'd pass an object as a second argument to the onToggle callback something like:

```js
onToggle(open, { source: 'select' })
```

where source can be: ''select'`, '`keyboard'`, `'clickoutside'` or similar...we can bikeshed the names

taion:
Closing this PR given that we're not going to use this approach.

codebreach:
Is this functionality getting merged in another PR?

taion:
In a different form as https://github.com/react-bootstrap/react-bootstrap/pull/2422

PR #2394, opened 2016-12-13
Add sizes to FormControl docs

joelseq:
I noticed that after my previous PR the sizes did not get added to the docs so I added the line that @jquense told me about.


jquense:
thanks!

PR #2395, opened 2016-12-13
Correct tabindex on alert modal close button

### LINKED ISSUE #2380: Unable to tab to Alert close button
When attempting to tab to the close button for an alert, only the screen reader version of the close button is focusable, not the actual close button. This is how the alert looks like in chrome:
screen shot 2016-12-07 at 4 27 30 pm

This behavior was introduced in https://github.com/react-bootstrap/react-bootstrap/pull/1545

dcondrey:
Resolves #2380.  Corrects to #1545

### DIFF
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -22,7 +22,6 @@ class Alert extends React.Component {
         className="close"
         onClick={onDismiss}
         aria-hidden="true"
-        tabIndex="-1"
       >
         <span>&times;</span>
       </button>
@@ -35,6 +34,7 @@ class Alert extends React.Component {
         type="button"
         className="close sr-only"
         onClick={onDismiss}
+        tabIndex="-1"
       >
         {closeLabel}
       </button>
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -24,7 +24,7 @@ class Alert extends React.Component {
         tabIndex="-1"
       >
         <span aria-hidden="true">&times;</span>
-        <span class="sr-only">{closeLabel}</span>
+        <span className="sr-only">{closeLabel}</span>
       </button>
     );
   }
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -15,28 +15,16 @@ const defaultProps = {
 };
 
 class Alert extends React.Component {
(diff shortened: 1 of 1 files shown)


taion:
Surely an SR-only button that isn't tabbable is not right?

jquense:
why do we have two buttons again? seems like the better of the tradeoffs here is to have one button both for purposes yeah?

taion:
I dunno, the second button got added when the SR thing did. I think we should have one button if we care about this tab thing.

taion:
Fix the tests, obviously, but the code looks fine.

taion:
Structurally that's fine but fix the bugs in your code that lint is telling you about

taion:
On second thought, we should be consistent between this and the close button in `<Modal.Header>`.

dcondrey:
Working on conforming to <Modal.Header> but there's a comment about using `closeLabel`.  Not sure this is correct.

taion:
The prop name on `<Modal.Header>` is definitely wrong per the comments. I'm wondering more about the markup.

Looking at https://github.com/react-bootstrap/react-bootstrap/pull/1237, the old way of doing things caused issues with JAWS, so perhaps we can't just use `aria-label`, i.e. the way things are set up in `<Modal.Header>` is incorrect.

Note the comment in the docs added in #1237 that would now be incorrect with the changes in this PR.

dcondrey:
Should <Modal.Header> and <Alert> be combined?  They're very similar.

taion:
They can't be combined. They're different things.

dcondrey:
I'm not clear as to what the final decision was for this.  If you could please let me know I'll get it resolved.  Thanks.

jquense:
I think this approach is fine. upstream uses SR only text as well instead of aria-label. I'm pretty sure the aria label choice was mine, thinking it was practically the same thing... I guess however SRs may treat them differently. 

can you update the docs noted by @taion,

taion:
@dcondrey Have you had a chance to update the docs?

taion:
ping @dcondrey

taion:
ping

PR #2402, opened 2016-12-16
Update CONTRIBUTING.md

annie-l:
If someone is contributing visual changes, I don't think it's an undue burden to provide screenshots of the change, so I'm proposing dropping the "if at all feasible" wording. Screenshots/screencasts are vital for explaining visual changes and GitHub makes it especially easy to drag and drop images into comments.

Here's my PR as an example...

BEFORE:
screen shot 2016-12-16 at 12 43 59 pm

AFTER:
screen shot 2016-12-16 at 12 43 45 pm

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,7 +37,7 @@ linter.
 
 ## Visual Changes
 
-When making a visual change, if at all feasible please provide screenshots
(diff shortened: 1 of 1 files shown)


taion:
Thanks! Sounds reasonable to me as a guideline. LGTM.

PR #2419, opened 2017-01-04
Update copyright year in LICENSE

nicmitchell:


taion:
Thanks! Updated it a bit to use a date range.

PR #2422, opened 2017-01-08
Add eventType to onToggle and source for handleClose

### LINKED ISSUE #1490: possible to override Dropdown toggle onClick behavior

joelseq:
This PR attempts to resolve #1490. The main things that I changed are:
1) Added a parameter (eventType) to onToggle so that every time onToggle is fired you know exactly what event fired it ('click' if from handleClick, 'keydown' if from handleKeyDown, 'close' if from onClose, 'select' if from onSelect).
2) Added a parameter (source) to handleClose which is the source function that fired the handleClose.

I tried it out locally in my own app and I was able to override the default closing of the dropdown by creating my own state that I called `dropdownOpen` that I passed in as the `open` prop to the Dropdown component. I created a `handleToggle` function that I pass in as the `onToggle` prop which only closes when `open` is `false` and the event that closed it is `close`. Here's what my handleToggle looks like:
```
handleToggle(open, eventType) {
    // If open is false and type of event is 'close'
    if (!open && eventType === 'close') {
      this.setState({
        dropdownOpen: false,
      });
    }
    else {
      this.setState({
        dropdownOpen: true,
      });
    }
  } 
```

Here's what you get when you console.log the inputs to onToggle:
onToggle Screenshot

I didn't know how to write a test for this so if you could advise me on how I could then I'll add it to this PR, thanks!

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -71,10 +71,10 @@ const propTypes = {
 
   /**
    * A callback fired when the Dropdown wishes to change visibility. Called with the requested
-   * `open` value.
+   * `open` value, the DOM event, and the source that fired it: `'click'`,`'keydown'`,`'close'`, or `'select'`.
    *
    * ```js
-   * function(Boolean isOpen) {}
+   * function(Boolean isOpen, Object event, { String source }) {}
    * ```
    * @controllable open
    */
@@ -156,12 +156,12 @@ class Dropdown extends React.Component {
     }
   }
 
-  handleClick() {
+  handleClick(event) {
     if (this.props.disabled) {
       return;
     }
 
-    this.toggleOpen('click');
+    this.toggleOpen(event, 'click');
   }
 
   handleKeyDown(event) {
@@ -172,7 +172,7 @@ class Dropdown extends React.Component {
     switch (event.keyCode) {
       case keycode.codes.down:
         if (!this.props.open) {
-          this.toggleOpen('keydown');
+          this.toggleOpen(event, 'keydown');
         } else if (this.menu.focusNext) {
           this.menu.focusNext();
         }
@@ -180,30 +180,30 @@ class Dropdown extends React.Component {
         break;
       case keycode.codes.esc:
       case keycode.codes.tab:
-        this.handleClose(event);
+        this.handleClose(event, 'keydown');
         break;
       default:
     }
   }
 
-  toggleOpen(eventType) {
+  toggleOpen(event, source) {
     let open = !this.props.open;
 
     if (open) {
-      this.lastOpenEventType = eventType;
+      this.lastOpenEventType = source;
     }
 
     if (this.props.onToggle) {
-      this.props.onToggle(open);
+      this.props.onToggle(open, event, { source });
     }
   }
 
-  handleClose() {
+  handleClose(event, source) {
     if (!this.props.open) {
       return;
     }
 
-    this.toggleOpen(null);
+    this.toggleOpen(event, source);
   }
 
   focusNextOnOpen() {
@@ -274,10 +274,10 @@ class Dropdown extends React.Component {
       labelledBy: id,
       bsClass: prefix(props, 'menu'),
       onClose: createChainedFunction(
-        child.props.onClose, onClose, this.handleClose,
+        child.props.onClose, onClose, (event) => this.handleClose(event, 'close'),
       ),
       onSelect: createChainedFunction(
-        child.props.onSelect, onSelect, this.handleClose,
+        child.props.onSelect, onSelect, (key, event) => this.handleClose(event, 'select'),
       ),
       rootCloseEvent
     });
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -161,7 +161,7 @@ class Dropdown extends React.Component {
       return;
     }
 
-    this.toggleOpen(event, 'click');
+    this.toggleOpen(event, { source: 'click' });
   }
 
   handleKeyDown(event) {
@@ -172,7 +172,7 @@ class Dropdown extends React.Component {
     switch (event.keyCode) {
       case keycode.codes.down:
         if (!this.props.open) {
-          this.toggleOpen(event, 'keydown');
+          this.toggleOpen(event, { source: 'keydown' });
         } else if (this.menu.focusNext) {
           this.menu.focusNext();
         }
@@ -180,7 +180,7 @@ class Dropdown extends React.Component {
         break;
       case keycode.codes.esc:
       case keycode.codes.tab:
-        this.handleClose(event, 'keydown');
+        this.handleClose(event, { source: 'keydown' });
         break;
       default:
     }
@@ -194,7 +194,7 @@ class Dropdown extends React.Component {
     }
 
     if (this.props.onToggle) {
-      this.props.onToggle(open, event, { source });
+      this.props.onToggle(open, event, source);
     }
   }
 
@@ -274,10 +274,10 @@ class Dropdown extends React.Component {
       labelledBy: id,
       bsClass: prefix(props, 'menu'),
       onClose: createChainedFunction(
-        child.props.onClose, onClose, (event) => this.handleClose(event, 'close'),
+        child.props.onClose, onClose, this.handleClose,
       ),
       onSelect: createChainedFunction(
-        child.props.onSelect, onSelect, (key, event) => this.handleClose(event, 'select'),
+        child.props.onSelect, onSelect, (key, event) => this.handleClose(event, { source: 'select' }),
       ),
       rootCloseEvent
     });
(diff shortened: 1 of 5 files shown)


src/Dropdown.js
```diff
@@ -71,10 +71,10 @@ const propTypes = {
 
   /**
    * A callback fired when the Dropdown wishes to change visibility. Called with the requested
-   * `open` value.
+   * `open` value and the type of event that fired it: `'click'`,`'keydown'`,`'close'`, or `'select'`.
```
jquense: i don't see `keydown` below? did that go missing somewhere?

src/Dropdown.js
```diff
@@ -194,16 +194,16 @@ class Dropdown extends React.Component {
     }
 
     if (this.props.onToggle) {
-      this.props.onToggle(open);
+      this.props.onToggle(open, eventType);
```
jquense: i'd make this second argument an object like `{ eventType: 'click', }` that way its more future proof if we want to add mroe stuff there.

src/Dropdown.js
joelseq: It was already being passed from handleKeyDown

image

joelseq:
@jquense made the change that you requested

jquense:
this looks fine to me. We should probably also pass through the DOM event at somepoint as well like [I started doing here](https://github.com/jquense/react-widgets/blob/v4/packages/react-widgets/src/DropdownList.js#L414) but that would require https://github.com/react-bootstrap/react-overlays/pull/142 first

taion:
i think we should probably _just_ pass through the DOM event from https://github.com/react-bootstrap/react-overlays/pull/142, actually. that's enough, no?

jquense:
in this case yeah it's enough. tho events are always a bit weird given the pooling component. I'm good waiting for the event to be passed through and updating this. I'd still use an object here tho to reserve real estate for the future

taion:
I thought that, parallel with `onSelect`, it could just be `onToggle(open, event)`. If we need more details we can add a third arg, like with carousel.

joelseq:
So should I wait till https://github.com/react-bootstrap/react-overlays/pull/142 is merged and then send in the DOM event instead?

taion:
I think so, yes.

taion:
On second thought, I think the signature should be `(open, event, { source })` to make it easier

jquense:
yes :thumbsup: to that

codebreach:
is this going to be merged anytime soon?

jquense:
blocked by https://github.com/react-bootstrap/react-overlays/pull/142

taion:
That PR was released in react-overlays v0.6.11.

joelseq:
Made the requested changes but the AppVeyor build seems to be failing. All the tests pass locally for me but my hunch is that it's because the react-overlays lib hasn't been updated to reflect the latest PR and so for now when onClose is triggered from the rootClose there's undefined passed instead of the DOM event. I made sure to test it while playing around with it locally in a project by manually updating react-overlays and passing in the event. It seems to be working as expected.

taion:
Please bump the dep version in `package.json`

jquense:
what is the rational behind having three args vs just an object bag as the second one? Consistency with our handlers elsewhere?

taion:
That's what we do for carousels, no?

jquense:
yeah, just wanted to pause for a sec to note that. The Carousel code isn't generally our best foot forward :P

taion:
Sure, so compat with existing signatures is good – `eventKey, event, extraStuff`... means that it could just be `eventKey, event` for users who don't care.

joelseq:
Made the requested changes. Previous build failed because of a [test](https://github.com/react-bootstrap/react-bootstrap/pull/2422/commits/3b8dfb63ae4304cd64b31d7bc1683a2c515ea6f8) in DropdownMenuSpec that was checking to make sure that no args were being passed to onClose but after the update to RootCloseWrapper the DOM event is being passed in so I had to change that test.

Besides that,  I wrote unit tests to make sure that for each source (keydown, click, select, close) the appropriate params are being passed to onToggle. One thing that I did notice though, if you click the DropdownToggle button to close the Dropdown, the onToggle gets called twice, the first time with 'click' as the source and the second time with 'close' as the source. I just wanted to make sure whether this is the expected behavior.

taion:
You should set the event when calling `handleClose` on the key down handler as well... though I wonder, is `source` really needed at all?

I think the "clicking on the button" thing comes from both the root close and the click events firing.

jquense:
yeah I think we need to distinguish between the clickoutside vs the toggling via the button yeah?

taion:
You think looking at e.g. event target isn't enough? I guess maybe it's just `select` that we care about? I'm not sure about naming for the rest.

jquense:
well isn't the type `click` for most of the options here if you just went with the event type?

joelseq:
@taion whoops sorry about that, fixed now. With regards to passing the source, I agree that it's mostly `select` that's important but I think it might be nice to provide users with an easy way to figure out what sources triggered onToggle.

An interesting use case I thought about while playing around with it is to set up a GA event to fire inside the onToggle to log what events trigger onToggle, potentially for UX or usability research.

src/Dropdown.js
```diff
@@ -274,10 +274,10 @@ class Dropdown extends React.Component {
       labelledBy: id,
       bsClass: prefix(props, 'menu'),
       onClose: createChainedFunction(
-        child.props.onClose, onClose, this.handleClose,
+        child.props.onClose, onClose, (event) => this.handleClose(event, 'close'),
```
taion: `close` is a weird event type for a close action. `rootClose` might be nice, but it's only appropriate for the default implementation. i wonder if we should let the child deal with setting the event type here, something like

```
onRootClose = (e) => {
  this.props.onClose(e, { source: 'rootClose' });
}
```

in `<DropdownMenu>`.

jquense:
ok sorry. what is the action item for this? it'd be nice to get it in and released. are we just picking type names?

taion:
@jquense The last thing is the `source` to use for root close actions. I'm wondering if we should use some mechanism such that it can be `'rootClose'` instead of `'close'`.

src/Dropdown.js
jquense: Why don't we like `close` here? It's too generic? I do think that its weird for the Dropdown to fire a `rootClose` type since it can't actually know that...

Bah yeah, the whole point is to differentiate rootClose from any other close, so yeah probably belongs in DropdownMenu. We should think about moving RootClose up out of the menu maybe in the longer term, enough people seem to expect it to live in Dropdown anyway.

taion:
@jquense 

In the short term, I think the solution I have above works – essentially let `<DropdownMenu>` pass in `source` when calling `props.onClose`.

Or we could make this a breaking change and move `<RootCloseWrapper>` out of `<DropdownMenu>`, but it seems like if we pass down `onClose` at all, we should provide more detail as to the source of the invocation of that cb.

joelseq:
I am a little bit confused. Isn't onClose getting called from the DropdownMenu and handleClose just getting called via the chained onClose function created? So wouldn't just renaming the `close` event to `rootClose` suffice?

taion:
@joelseq 

From a correctness perspective, it's possible that a custom dropdown menu might call its `onClose` callback for some reason other than "root close", so it wouldn't be quite correct for us to specify `rootClose` here.

We'd either need `<DropdownMenu>` to set the source, or we could move `<RootCloseWrapper>` into `<Dropdown>` itself and just release this as a breaking change.

joelseq:
@taion okay that makes more sense now. Will send in a PR later tonight

joelseq:
@taion as we discussed on Discord, I pushed what I currently have but a unit test in DropdownMenuSpec is failing for some reason after I made the changes. Also, I'm hoping to get some advice on how to right a test for the rootClose event being fired.

kraizybone:
Is this going to be merged anytime soon? Thanks

taion:
@joelseq Can you enable me to push to your branch?

taion:
Oops, I messed up. I fixed this PR. Should be good to go now.

isaachinman:
Does this mean you guys are ready to release `v0.30.8`?

taion:
Needs review from another org member

PR #2423, opened 2017-01-08
Give `props.active` precedence over children active for `NavDropdown`

ml9951:
The logic for determining if a `NavDropdown` component is active first checks to see if `props.active` is true.  If not, then it gets set to active if any of its children are active [[1](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/NavDropdown.js#L24-L40)].  I think that setting `props.active = false` should take precendence over the children's activeness.  This will allow the programmer to explicitly set the active field to `false` regardless of the childrens' active state.

### DIFF
--- a/src/NavDropdown.js
+++ b/src/NavDropdown.js
@@ -22,7 +22,9 @@ const propTypes = {
 
 class NavDropdown extends React.Component {
   isActive({ props }, activeKey, activeHref) {
-    if (
+    if (props.active != null) {
+      return props.active;
+    } else if (
       props.active ||
       activeKey != null && props.eventKey === activeKey ||
       activeHref && props.href === activeHref


jquense:
seems reasonable to me, tho this needs to go into the `next` branch since it's a breaking change (albeit a small one).

taion:
We should do the same thing on `<Nav>` then no? Though this code should be cleaned up a bit.

ml9951:
I'm attempting to add this to the `next` branch, however, I'm getting a bunch of errors when running the testing script:

`Module not found: Error: Cannot resolve module 'react/lib/ReactTestUtils'`

[This](https://github.com/airbnb/enzyme/issues/684) issue suggests that version 15.4 moved `ReactTestUtils`.  React 15.3 is listed as a dependency, but the previously mentioned issue goes on to say that having `react-addons-test-utils` as a dependency can cause npm to auto-upgrade react.  We don't explicitly have `react-addons-test-utils` as a dependency, but `npm ls` shows:

```
├─┬ teaspoon@6.4.2
│ ├─┬ bill@3.2.6
│ │ ├── css-selector-parser@1.3.0
│ │ ├── escape-regexp@0.0.1
│ │ ├── fn-name@2.0.1
│ │ └── lodash@3.10.1
│ ├── lodash@3.10.1
│ ├── react-addons-test-utils@15.4.2
│ └── warning@2.1.0
```

Is anyone else having problems running the test script in the `next` branch?

jquense:
O yeah we probably need to refresh that branch, it hasn't been synced with. aster recently

taion:
I updated `next`

taion:
ping – we should be consistent on `<Nav>`. and have test coverage

ml9951:
Sorry, this fell off my radar.  I'll try and finish this up in the next day or two.

ml9951:
@taion, after looking into this further, I'm not sure I see your point about applying this to `<Nav>`.  The issue that I originally mentioned pops up when a `NavDropdown`'s child is active, we also set the `NavDropdown` itself to active with no way of explicitly overriding this.  What does it mean for a `Nav` to be active?  It's not even a PropType.  I would expect only `NavItem`s to be able to take on an active state.  Perhaps I'm missing something?

taion:
`<Nav>` uses essentially the same logic to determine whether a child is active based on `eventKey`.

ml9951:
Right, but my issue is not about children being active, it's about the parent being active.  Maybe a picture will make this more clear:

Imgur

On the left is a `NavDropdown`.  I'd like *one* of the children to be `active` based on the city that is currently selected, however, I **do not** want the parent ("Choose Location") to be active.  I'd like to override this by explicitly telling `NavDropdown` to not be active, but still allow it's children to be active.  

I'm failing to see the `Nav` analogy.  Regardless of what `Nav`'s children are active, the `Nav` itself is never active.

taion:
`isActive` gets called on all the children there as well: https://github.com/ml9951/react-bootstrap/blob/ba941cba550faed73453ba9e8b169414d81721cc/src/NavDropdown.js#L79

it's the same code regardless

jquense:
what is the state of this PR?

taion:
Closing for staleness. We'd welcome another PR that applies this consistently, though.

Issue #2429, opened 2017-01-13
When a MenuItem is wrapped, it won't close the DropdownButton when clicked


natevw:
With this sample code, clicking the second menu item leaves the list open:

```
import { DropdownButton, ButtonGroup, Button, MenuItem } from 'react-bootstrap'

let MyMenuItem = props => <MenuItem
  {...props}
  onSelect={(arg1,arg2) => console.log("wrapped MenuItem onSelect", arg1,arg2)}
/>

let Test = () => <div>
  <DropdownButton title="Test" id="test123"
    onSelect={(arg1,arg2) => console.log("DropdownButton onSelect", arg1,arg2)}>
      <MenuItem onSelect={(arg1,arg2) => console.log("MenuItem onSelect", arg1,arg2)}>Bare MenuItem</MenuItem>
      <MyMenuItem>Wrapped MenuItem</MyMenuItem>
    </DropdownButton>
</div>

render(<Test/>, document.querySelector('#test'))
```

The cause seems to be in https://github.com/react-bootstrap/react-bootstrap/blob/17fccf68b65cd58a80b9be84b87038548e50f622/src/DropdownMenu.js#L126 where the DropdownMenu attempts to chain its own onSelect to each child's — this works if the child is a MenuItem, but if it's not then the interloping MyMenuItem gets the chained handler, and when the click happens the MenuItem inside it still has its original handler.

Removing the `onSelect` property from the nested MenuItem also hides the issue, since the wrapper in this example then passes the chained version along via `{...props}`. If instead I'd simply wrapped it in a simpler fashion like:

    let MyMenuItem = ({name}) => <MenuItem onClick={someHandler}>{name}</MenuItem>

…then it would also break.

WORKAROUND: any wrapper for shared `MenuItem` logic needs to be aware of this internal implementation detail i.e. that the parent dropdown relies on a monkey-patched `onSelect` to close itself. It must either pass along `this.props.onSelect` verbatim, or if it wants to provide onSelect behavior itself it needs to be sure to call that.

jquense:
seems like you figured out. however, it's not an implementation detail, the MenuItem is injected with props, if you don't pass them through to the real MenuItem then yeah it doesn't work. This is a common pattern and generally how the prop system works. I can understand how it might feel annoying but it actually enables really powerful composition, since your MenuItem can control its behavior.

natevw:
[...]

PR #2432, opened 2017-01-14
Document caution required for MenuItem's onSelect

natevw:
As described in more detail in https://github.com/react-bootstrap/react-bootstrap/issues/2429, there is currently a "gotcha" if one wraps a MenuItem but overrides (or fails to pass along) the monkey-patched `onSelect` property that will be unexpectedly received via the DropdownMenu internals.

I.e. logic like this will break menu closing:

```
let MyMenuItem = ({obj}) => <MenuItem
  onSelect={ () => handleSelect(obj) }
>{obj.name}</MenuItem>

let objects = [{id:123, name:"My object}, …]
let Test = () => <DropdownButton title="Test" id="naiveLogic">
  {objects.map((obj) => <MyMenuItem key={obj.id} obj={obj}/>}
</DropdownButton>
```

Instead you currently need to implement the wrapper as something like this, aware of the DropdownMenu internal behavior:

```
let MyMenuItem = ({obj, onSelect:internalOnSelectOverride}) => <MenuItem
  onSelect={ () => { handleSelect(obj); internalOnSelectOverride(); } }
>{obj.name}</MenuItem>
```

Or as something like this, passing along all received properties (even if your own calling code doesn't ever provide any extra!) and being sure to avoid using onSelect in your wrapper itself:

```
let MyMenuItem = ({obj, ...props}) => <MenuItem
  {...props}
  onClick={ () => handleSelect(obj) }
>{obj.name}</MenuItem>
```

### DIFF
--- a/docs/src/sections/MenuItemSection.js
+++ b/docs/src/sections/MenuItemSection.js
@@ -22,7 +22,7 @@ export default function MenuItemSection() {
         <li><code>divider</code>: Adds an horizontal divider between sections</li>
         <li><code>disabled</code>: shows the item as disabled, and prevents the onclick</li>
         <li><code>eventKey</code>: passed to the callback</li>
-        <li><code>onSelect</code>: a callback that is called when the user clicks the item.</li>
+        <li><code>onSelect</code>: a callback that is called when the user clicks the item. <b>NOTE:</b> the containing DropdownMenu actually re-writes this property on its direct child elements. For simple (direct) uses of MenuItem, this will not affect you. However, if you wrap a MenuItem in your own component, be careful! You <em>must</em> handle the onSelect property accordingly, or the <a href="https://github.com/react-bootstrap/react-bootstrap/issues/2429">dropdown will not close</a> after a user selects a menu item!</li>
       </ul>
       <p>The callback is called with the following arguments: <code>event</code> and <code>eventKey</code></p>
       <ReactPlayground codeText={Samples.MenuItem} />


docs/src/sections/MenuItemSection.js
```diff
@@ -22,7 +22,7 @@ export default function MenuItemSection() {
         <li><code>divider</code>: Adds an horizontal divider between sections</li>
         <li><code>disabled</code>: shows the item as disabled, and prevents the onclick</li>
         <li><code>eventKey</code>: passed to the callback</li>
-        <li><code>onSelect</code>: a callback that is called when the user clicks the item.</li>
+        <li><code>onSelect</code>: a callback that is called when the user clicks the item. <b>NOTE:</b> the containing DropdownMenu actually re-writes this property on its direct child elements. For simple (direct) uses of MenuItem, this will not affect you. However, if you wrap a MenuItem in your own component, be careful! You <em>must</em> handle the onSelect property accordingly, or the <a href="https://github.com/react-bootstrap/react-bootstrap/issues/2429">dropdown will not close</a> after a user selects a menu item!</li>
```
jquense: this wording isn't accurate...

Probably something more like "Note: The Dropdown menu component passes props to the MenuItem. When providing a custom MenuItem component be sure to pass through the `onSelect` handler or call it explicitly in order to inform the Dropdown when an Item is selected and should close."

natevw: Your version is certainly less awkward, +1.

But to clarify, when you say "passes props" (generically) — does that mean that neither workaround above is actually valid? Does DropdownMenu reserve the right to pass along any/all other props as well? (I.e. even the second workaround above isn't correct, because how do I know it won't someday need careful `onClick` handling too? Do I have to defensively plan that *any* property that my wrapper wants to set on the actual MenuItem might also have content provided by DropdownMenu for it?)

jquense: You're looking at the wrong API here, what the dropdown does isn't really relevant to your custom MenuItem, only the API of `MenuItem` is important. Phrased another way, the DropdownMenu expects a component that implements the same API as a MenuItem. If you break that expectation then yes, it doesn't work as expected. So no you don't have have to plan on any prop, being provided, just the ones defined in the MenuItem propTypes (which is sort of like an interface). , 

 If you wrap MenuItem then you need to at a _minimum_ provide the same API to expect it to act the same. In this case, MenuItem requires an onSelect prop in order to communicate to it's parent when it's been selected. You are free to ignore or change that expectation by wrapping and not calling the onSelect handler, but then no parent component knows the MenuItem is selected as such doesn't do anything. That's all not an implementation detail is the explicitly defined API contract between the two components. You have the power to change that by implementing a MenuItem that ignores onSelect. Like i said this is a feature, it meanas users can really control how these component behave, whereas if we did it differently you'd have no ability to prevent selection of a menu item.

natevw: >  Phrased another way, the DropdownMenu expects a component that implements the same API as a MenuItem.

Okay, I see now where you're coming from. I'll be stubborn and still say it's suprising, and would be nice to handle either internally (perhaps configurable via an API on the menu) or by turning Select into an actual custom event that bubbles (not sure if React supports that yet) — but I agree this does make more sense when you put it in terms of essentially: a dropdown menu only really supports having MenuItems as its direct children. By wrapping them I ventured into somewhat uncharted territory, and so I had to be aware of some internal details.

Thanks for your patience :-)

taion:
Can we just add a `<CustomMenuItem>` to the custom dropdown example? Really this is about custom dropdowns, right?

taion:
@jquense What do you think about my idea of adding a `<CustomMenuItem>` component to the example to demonstrate what's going on, instead of inserting a long snippet in the docs?

joelseq:
Just out of curiosity, in what use cases would you consider wrapping a MenuItem?

natevw:
Say you have some objects in your app and you want a menu item for each one. This means common dispatch, labelling, etc. and a wrapper to take just the object and provide a menu item appropriately configured is convenient.On Feb 14, 2017 4:31 PM, Joel Sequeira <notifications@github.com> wrote:Just out of curiosity, in what use cases would you consider wrapping a MenuItem?

jquense:
Gonna close for now but if someone wants to Follow up this comment is the way to go: https://github.com/react-bootstrap/react-bootstrap/pull/2432#issuecomment-274185855

PR #2443, opened 2017-01-21
Actually fix <Pagination boundaryLinks>

### LINKED ISSUE #2431: Ellipsis between consecutive page numbers in pagination component
If `boundaryLinks=true` on the `Pagination` component, ellipsis are rendered between the boundary page link and the main page links, even when they are consecutive numbers.

## Expected Behaviour
screen shot 2017-01-13 at 14 15 04

_Note, I just directly modified the DOM, via Chrome DevTools to produce this screenshot._

## Actual Behaviour
From [sample page](https://react-bootstrap.github.io/components.html#pagination)
screen shot 2017-01-13 at 14 11 04

taion:
Fixes #2431

Still inclined to get rid of all this code. It's hairy and I don't see why it should be in React-Bootstrap.

### DIFF
--- a/src/Pagination.js
+++ b/src/Pagination.js
@@ -11,7 +11,8 @@ const propTypes = {
   maxButtons: React.PropTypes.number,
 
   /**
-   * When `true`, will display the first and the last button page
+   * When `true`, will display the first and the last button page when
+   * displaying ellipsis.
    */
   boundaryLinks: React.PropTypes.bool,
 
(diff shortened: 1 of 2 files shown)


jquense:
ltgm. I agree that we should remove this. RB feels like a weird place for it and it's jsut far and away more logic then upstream bootstrap includes. @react-bootstrap/collaborators thoughts?

Issue #2450, opened 2017-01-27
Static Modal does not work with closeButton={true}


vladimir-rovensky:
Hey everyone,

I came across an issue with the Modal component - when used as simple static html component (very first use case in the docs), and setting Modal.Header's closeButton=true, it crashes and doesn't render anything. Seems to be because the context is not initialized properly. I work around this by setting up the context in our wrapper component, but it should probably just work.

Example:
```
<Modal.Dialog>
      <Modal.Header closeButton={true}>
        <Modal.Title>Modal title</Modal.Title>
      </Modal.Header>

      <Modal.Body>
        One fine body...
      </Modal.Body>

      <Modal.Footer>
        <Button>Close</Button>
        <Button bsStyle="primary">Save changes</Button>
      </Modal.Footer>
</Modal.Dialog>
```

joelseq:
I'm pretty sure if you want to have the closeButton prop you need to wrap the entire thing inside a <Modal> otherwise it doesn't really make sense since the closeButton prop is supposed to fire the onHide of its parent Modal which in turn can be used to set the show prop of the Modal to false. The basic example is used to illustrate what the Dialog itself inside the Modal will look like and isn't actually a Modal.

vladimir-rovensky:
I have to disagree, the first use case is perfectly valid - I just want to render the Modal.Dialog and handle its showing/hiding myself. This makes the component much more flexible and usable - you have means to render just the modal as a siple React component and nothing else, no overlays, no portals. We have been using it this way for ages and it always worked.

To your second comment, the closeButton can just fire the onClose callback from the Header props, it doesn't have to crash if the context value is not there. This seems to be the intent in code as well (from the ModalHeader source):
```
  /**
   * A Callback fired when the close button is clicked. If used directly inside
   * a Modal component, the onHide will automatically be propagated up to the
   * parent Modal `onHide`.
   */
  onHide: React.PropTypes.func,
```
Nothing here says you HAVE to use it within Modal, there's simply no need to couple the two components like this.

joelseq:
@vladimir-rovensky I think that comment is a little ambiguous but what I understood from it is that you can directly put onHide in the parent Modal and not have to put it inside the Modal.Header as the `createChainedFunction` util will automatically take care of that. So basically it's like you can choose to put it either in Modal or Modal.Header. I see no mention anywhere in the docs saying that Modal.Header can be used independently of Modal. Here's a quote directly from the docs:
```
[...]

PR #2453, opened 2017-01-29
Add support for Modal.Header closeButton without containing Modal component

joelseq:
Implements what was discussed in #2450 . Also added ModalHeaderSpec since it wasn't previously there and wrote a test to make sure the component renders with a closeButton prop now.

### DIFF
--- a/src/ModalHeader.js
+++ b/src/ModalHeader.js
@@ -65,7 +65,7 @@ class ModalHeader extends React.Component {
             type="button"
             className="close"
             aria-label={label}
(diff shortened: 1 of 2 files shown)


jquense:
LGTM

taion:
Can we assert that clicking on `closeButton` triggers `onClose` here?

joelseq:
@taion added an assertion that makes sure that `onHide` is triggered when the close  button is clicked (I'm guessing that's what you wanted?).

taion:
Thanks!

PR #2466, opened 2017-02-08
Fix Carousel slide=false bug

isaachinman:
Do not set previousActiveIndex if slide animation is disabled


isaachinman:
@taion Sure. I didn't overlook that in the contributing docs, I was having trouble running the `test` script on my local machine. Looks like it only works with quite recent versions of node. Will write a test shortly.

PR #2469, opened 2017-02-10
Clarify naming expectation for glyph prop

sidewaysgravity:
It was not immediately obvious to remove the prefix since we specify using the icon name with a link to bootstrap which does use the prefix. It was only after looking at the sample and my written code that I noticed the difference.

While its nothing major I feel that this just adds a bit of clarity to the naming convention.

### DIFF
--- a/src/Glyphicon.js
+++ b/src/Glyphicon.js
@@ -6,7 +6,7 @@ import { bsClass, getClassSet, prefix, splitBsProps }
 
 const propTypes = {
   /**
-   * An icon name. See e.g. http://getbootstrap.com/components/#glyphicons
+   * An icon name without "glyphicon-" prefix. See e.g. http://getbootstrap.com/components/#glyphicons
    */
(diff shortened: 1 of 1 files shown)


jquense:
LGTM thanks!

sidewaysgravity:
@jquense When looking through the documentation, It looked as if the docs were built off the annotations which is what I have updated. I noticed that the docs at https://react-bootstrap.github.io/ still don't show the updated description. Are those docs built and updated separate from the pull requests or was there something else that I needed to update for the change to show?

jquense:
@sidewaysgravity the docs site isn't deployed automatically on a PR, we have to actually manually trigger a deploy for it. So no worries on your end, we jsut haven't pushed the update out yet.

sidewaysgravity:
@jquense Cool. No rush on it, I just wanted to make sure I didn't miss something.

PR #2475, opened 2017-02-14
Add closeOnSelect and closeOnTab options.

chexxor:
My attempt to fix https://github.com/react-bootstrap/react-bootstrap/issues/1490

I want to put some inputs in a dropdown. I don't want the dropdown to close when I do stuff with the inputs, and I want to be able to tab between the inputs.

What are your thoughts on these two new params to Dropdown?

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -90,6 +90,20 @@ const propTypes = {
   onSelect: React.PropTypes.func,
 
   /**
+   * Whether or not to close the Dropdown after selecting an item.
+   *
+   * @controllable onToggle
+   */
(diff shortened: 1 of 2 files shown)


taion:
We're planning on going with https://github.com/react-bootstrap/react-bootstrap/pull/2422 instead.

jquense:
Ya. thanks for the PR, btw. I think the #2422 offers better flexibility as a general solution and enables this behavior here to be built on top in your own tab

chexxor:
Ok, I'll take a look at using that one. Thanks!

PR #2488, opened 2017-02-17
Add support for styling pagination buttons (first, prev, next, & last)

kjschulz:
There is currently no non-hacky way to custom style the for first, prev, next, and last icon buttons.  The enterprise software development entity I contract with currently has a style requirement for showing and hiding these buttons based on the active page (ie if it's the first or last) and also custom styling.  All of this can be easily manged if there are some base classes on these buttons.  I admit it's a little bit of an edge case and so I'm already doing this on my fork but would like to propose it may benefit others to forward this option as well.

### DIFF
--- a/src/Pagination.js
+++ b/src/Pagination.js
@@ -205,6 +205,7 @@ class Pagination extends React.Component {
         {first && (
           <PaginationButton
             {...buttonProps}
+            className="btn-first"
             eventKey={1}
             disabled={activePage === 1}
           >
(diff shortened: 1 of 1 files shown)


taion:
This isn't really part of upstream Bootstrap. We'll eventually revise this API entirely in https://github.com/react-bootstrap/react-bootstrap/issues/1758, so I'd rather not merge this for now.

kjschulz:
Fair enough, thanks.

PR #2504, opened 2017-03-07
Wire up mountOnEnter

### LINKED ISSUE #2439: Tab unmountOnHide behavior
I want tabs which are hidden (all but initial visible tab) to mount lazily when shown.  But once they are mounted, I don't want them to unmount on hide.   Setting unmountOnHide to true accomplishes the first goal (although the property name doesn't suggest that it would) but not the second.   Is there any simple way to accomplish this?  

Another question: why are the tab's various onEnter on Leave methods only available when there's animation?  Accomplishing the above would be possible if I had those hooks but I don't want animation.  

Thanks!

taion:
Closes #2439

### DIFF
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -9,6 +9,11 @@ const propTypes = {
   in: React.PropTypes.bool,
 
   /**
+   * Wait until the first "enter" transition to mount the component (add it to the DOM)
+   */
+  mountOnEnter: React.PropTypes.bool,
+
+  /**
    * Unmount the component (remove it from the DOM) when it is faded out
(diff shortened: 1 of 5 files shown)


src/Fade.js
```diff
@@ -9,6 +9,15 @@ const propTypes = {
   in: React.PropTypes.bool,
 
   /**
+   * Wait until the first "enter" transition to mount the component (add it to the DOM)
+   */
```
jquense: double?

jquense:
LGTM modulo that one note

PR #2505, opened 2017-03-08
Export PaginationButton to people can build their own paginations

Tigraine:
A very small change but I noticed `PaginationButton` is not exported in `index.js` and I am currently working on a variation of the core Pagination that works a bit differently, and having the standard PaginationButton available would be nice..  Atm I have to re-implement the `PaginationButton` markup


taion:
LGTM for now, but note this API is likely to change, and that you shouldn't be importing from the main package index anyway

PR #2508, opened 2017-03-08
add unstyled SplitButton to docs #btn-dropdowns-split

sillyslux:
I was searching for an unstyled SplitButtonDropdown and since it already seems to be supported, i think it should be reflected in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,9 @@
+# data-toggle fork
+**This fork aims on adding `data-toggle` attributes in the right places.**
+
+You can still use the bundled js in `development`, but you may drop it in `production`
+and use e.g. [bootstrap.native](https://github.com/thednp/bootstrap.native/) there.
+
 # React-Bootstrap [![Travis][build-badge]][build] [![npm][npm-badge]][npm]
 
 [Bootstrap 3][bootstrap] components built with [React][react].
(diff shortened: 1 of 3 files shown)


taion:
probably should remove this from the other example actually. like technically it works but it doesn't look right. it's not in the upstream TWBS examples either: http://getbootstrap.com/components/#btn-dropdowns-single

sillyslux:
oh, really? it's useful though.
On the other side, users can override the styling with a custom class and custom css anyway.
So... whatever...

taion:
See other issues – if you're using `data-toggle`, you're doing something quite wrong

sillyslux:
Custom class and custom css has nothing to do with data-toggle.
if you want to discuss #2510, please do it there.

jquense:
for button styles we don't actually want to show the Link versions of Dropdowns b/c they aren't technically supported.

But you can add/use them in your own code without forking or editing check out https://react-bootstrap.github.io/components.html#custom-styles and just add in `link` as a "custom" style

sillyslux:
yea custom class/custom css, my words...  
but then, where do these unsupported link versions come from? will they be removed? that removal will break then, won't it?

sillyslux:
i'm using the link versions right now and thus i have to apply only little custom css.
_am i doing this one wrong too?_

PR #2515, opened 2017-03-11
Delegate the control over a collapsible panel to elements other than anchor

zefj:
Hello, this PR implements a change to collapsible panels with header - allowing nested elements or the entire heading to be toggles for collapsible panel, which is a must-have feature we needed at work. This is easily done with Bootstrap by simply reordering the elements or moving the control attributes up in the tree, but not so simple with react-bootstrap as far as I'm concerned. Up to this point, it was only possible for the text inside a header to be an anchor controlling the collapsible, and there was no elegant solution for making the entire header or nested elements toggles for collapsible panels. Here's how this works:

react-bootstrap-collapse-control

This change introduces a new prop for `PanelGroup` and `Panel` components: `collapseControl`, that takes one of three values: `heading`, `element` or `anchor`, with `anchor` being the default. Those should be self-explanatory. You can specify different `collapseControl` values for each `Panel` component, or have a "global" behaviour for a group of Panels by passing it to a `PanelGroup`. Note that this will always be prioritized, as in `PanelGroup` prop value will override any `Panel` prop values.

Notice how `heading` makes the entire header clickable, and removes the underline text decoration from the text. Same applies to `element` - before, passing a JSX component (eg. <Button>) as `header` resulted in the text inside the element being underlined on hover.

This is a non-breaking change, the default behaviour for this component should stay the same since the `collapseControl` prop is `anchor` by default. Please verify this statement though.

Docs and tests included, Travis build passes.

### DIFF
--- a/docs/examples/PanelCollapseControl.js
+++ b/docs/examples/PanelCollapseControl.js
@@ -1,0 +1,9 @@
+const collapseControlInstance = (
+  <PanelGroup accordion>
+    <Panel header="Clickable header" collapseControl="heading" eventKey="1">Panel 1 content</Panel>
+    <Panel header={<Button>Clickable button</Button>} collapseControl="element" eventKey="2">Panel 2 content</Panel>
+    <Panel header="Clickable anchor" eventKey="3">Panel 3 content</Panel>
+  </PanelGroup>
+);
+
+ReactDOM.render(collapseControlInstance, mountNode);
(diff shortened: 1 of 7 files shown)


taion:
This isn't the way we want to take these things – if you need control over things, use controlled components.

jquense:
note that we also have a PR up already that reworks the API to be alot more flexible.  thanks for taking the time to write this tho! you may want check out that PR

zefj:
So more or less like so, and then get rid of the underline with css?

```
<PanelGroup activeKey={this.state.activeKey} accordion>
  <Panel onClick={ ()=> this.setState({ activeKey: '1' })} header="Panel 1" eventKey="1">Panel 1 content</Panel>
  <Panel onClick={ ()=> this.setState({ activeKey: '2' })} header="Panel 2" eventKey="2">Panel 2 content</Panel>
</PanelGroup>
```

I can't say I'm a fan of rewriting the entire group collapse state logic just to get the header to be a toggle, especially in what used to be a nice stateless component. I admit I wanted to keep it stateless at all cost, and didn't know the lib enough at that point - hence the PR. 

If I may suggest one last thing - the Panels docs category could use a little note about this, let me be the example of how confusing the current state of this component is for fresh users :) To add to that, text anchors in collapsible panels with header are completely counter-intuitive as you probably understand, so I think the ability to customise this should be more clearly pointed out. Have a great evening

PR #2529, opened 2017-03-24
possibility to use only specified header controls for the "expand/col…

mrRozgo:
relates to #983 issue
now you can use only selected controls for expand/collapse operation. the whole header element will be intact by anchor-wrapping. 
All you need is to specify these controls by identifiers using the "headerControls" property: 
<Panel collapsible header={headerGenFunc()} headerControls={[id1', 'id2', 'id3']}>


jquense:
hey there, thanks for the PR. before you get to far though, we already have a PR up addressing the Panel API, and it substantially changes stuff. I'd check there to ensure it meets your needs bc additions to this component will com on top of that PR

PR #2533, opened 2017-03-28
Fix #2532

### LINKED ISSUE #2532: The Keyboard navigation doesn't work properly on dropdown menu
The keyboard navigation(up/down arrow) doesn't work properly. 
When down arrow is pressed it goes to the 3rd element than the 2nd element.

Anenth:
The issue right now is the use of `document.activeElement`. 

When the user press the *down button* the browser moves the *focus to the next element*(document.activeElement gets changed) and then our logic for focusing the next element is hit; which mean it will always focus the next to next element. 

The fix is to use the `event.target` instead of `document.activeElement`;

Fixes #2532

### DIFF
--- a/src/DropdownMenu.js
+++ b/src/DropdownMenu.js
@@ -38,13 +38,14 @@ class DropdownMenu extends React.Component {
   }
 
   handleKeyDown(event) {
+    const targetElement = event.target;
     switch (event.keyCode) {
       case keycode.codes.down:
-        this.focusNext();
+        this.focusNext(targetElement);
         event.preventDefault();
         break;
       case keycode.codes.up:
-        this.focusPrevious();
+        this.focusPrevious(targetElement);
         event.preventDefault();
         break;
       case keycode.codes.esc:
@@ -55,9 +56,9 @@ class DropdownMenu extends React.Component {
     }
   }
 
-  getItemsAndActiveIndex() {
+  getItemsAndActiveIndex(targetElement) {
     const items = this.getFocusableMenuItems();
-    const activeIndex = items.indexOf(document.activeElement);
+    const activeIndex = items.indexOf(targetElement);
 
     return { items, activeIndex };
   }
@@ -71,8 +72,8 @@ class DropdownMenu extends React.Component {
     return Array.from(node.querySelectorAll('[tabIndex="-1"]'));
   }
 
-  focusNext() {
-    const { items, activeIndex } = this.getItemsAndActiveIndex();
+  focusNext(targetElement) {
+    const { items, activeIndex } = this.getItemsAndActiveIndex(targetElement);
     if (items.length === 0) {
       return;
     }
@@ -81,8 +82,8 @@ class DropdownMenu extends React.Component {
     items[nextIndex].focus();
   }
 
-  focusPrevious() {
-    const { items, activeIndex } = this.getItemsAndActiveIndex();
+  focusPrevious(targetElement) {
+    const { items, activeIndex } = this.getItemsAndActiveIndex(targetElement);
     if (items.length === 0) {
       return;
     }


taion:
This works on our examples though? What's different about your use case?

Anenth:
@taion I am not sure how is it working on prod document, We were having this issue on our product [Recruiterbox](recruiterbox.com). When I tried this on prod document it worked fine but when tried on dev document(`npm run docs`) I faced the same issue. 

Not sure why it happens only dev, Might be because performance?

jquense:
maybe you have something hijacking focus in dev? I haven't seen this issue in any apps myself either...

Anenth:
@jquense I checked that, Its not just my machine it happens in most of the machines. But use `event.target` seems to fix the issue

jquense:
So my concern (besides the inability to replicate) with this solution is that `event.target` is only the same as "focusable items", incidentally. First it'll break if your menuItem is more complex , such as having a span instead it (the span will be the target). Second thing is that it won't work well with custom menu items, where the focusable element isn't the thing you click that receives the event handler.

jdanil:
I've been having the same problem using only react-bootstrap components on a page.
`<ReactBootstrap.Nav>
    <ReactBootstrap.NavDropdown id="rb-menu" title="Menu">
        <ReactBootstrap.MenuItem>Item 1</ReactBootstrap.MenuItem>
        <ReactBootstrap.MenuItem>Item 2</ReactBootstrap.MenuItem>
        <ReactBootstrap.MenuItem>Item 3</ReactBootstrap.MenuItem>
    </ReactBootstrap.NavDropdown>
</ReactBootstrap.Nav>`
The same issue seems to occur on react-bootstrap demos I've found (e.g. http://www.gulamshakir.com/apps/s3calc/).

jdanil:
Could we make a slightly less invasive change that will still protect us if the browser default occurs before it reaches handleKeyDown, as seems to be the case for @Anenth and I?
i.e.
```
focusNext(event) {
    const { items, activeIndex } = this.getItemsAndActiveIndex();
    if (items.length === 0) {
        return;
    }

    const targetIndex = event && items.indexOf(event.target);
    const currentIndex = targetIndex > -1 ? targetIndex : activeIndex;
    const nextIndex = currentIndex === items.length - 1 ? 0 : currentIndex + 1;

    items[nextIndex].focus();
}
```
This way if the browser default occurs first and we are already on the next element, it wont try to change focus to the next next element. And if the event target isn't in the collection of menu items it will fall back to using document.activeElement to determine the current/active index.

jquense:
I'm open to other solutions, but I'd really like to have a reproduction of the error first, I've never seen the bug described here

jdanil:
I reproduced it in [jsbin](https://jsbin.com/hotasevafe/edit?html,js,output), and figured out what was causing my problem. In my case it was jquery that was changing the focus. Removing the jquery import fixed the issue for me.

taion:
Okay, is this a non-issue on our end then?

jdanil:
Yeah its not a react-bootstrap issue from my end. Not sure about @Anenth.

Anenth:
I will try removing jQuery and test it out.

jdanil:
I've been thinking, jQuery is a pretty popular library. It would be reasonable for a project to use both react-bootstrap and jQuery. Should react-bootstrap protect against this behaviour if it can? Or will devs just have to disable jQuery keyboard nav on a case by case basis for each react-bootstrap dropdown menu?

taion:
React-Bootstrap already handles all the JS to make the dropdown work. You should not be mixing and matching R-B and TWBS JS on the same component.

jdanil:
But if jQuery is being used for something else on the same page as the react-bootstrap dropdown, simply importing jQuery on the page will still affect the dropdown's keyboard navigation.

Anenth:
@taion, We have backbone code in our code base, Its a huge task to completely move to react only to fix this issue. Right now we have forked it and made our changes and it works great. But its difficult when there is a new update.

taion:
Why does simply importing jQuery affect the keyboard navigation?

Anenth:
@taion I dont think its because of jquery. I am able to replicate, when I run the docs on my machine.

taion:
To be clear, you're seeing this on the React Bootstrap docs? And with what browser? I haven't seen this issue at all – and it really sounds like there's something else that's messing with your focus management here.

Anenth:
@taion It works fine when Docs(https://react-bootstrap.github.io/components.html). 
It doesn't work if I run Docs on my local machine

taion:
I can't reproduce this by running the docs page locally.

Barring a runnable repro and in light of the correctness concerns per https://github.com/react-bootstrap/react-bootstrap/pull/2533#issuecomment-289829304, I'm going to close this out for the time being.

Let us know if you can make a JSBin or something that demonstrates the problem.

Issue #2539, opened 2017-04-04
Tab eventKey cannot use zero as value


grovesNL:
In the [Tabs Controlled example](https://react-bootstrap.github.io/components.html#tabs-controlled), if `eventKey={0}` is used in place of the values, the tab and tab content do not update correctly _only when using the arrow keys to navigate between tabs_ (clicking the tab works as expected). I would guess there is a truthy check for `eventKey` somewhere in the key event handlers causing this issue.

Here is a full example using `eventKey={0|1|2}` instead of `eventKey={1|2|3}`:
```
const ControlledTabs = React.createClass({
  getInitialState() {
    return {
      key: 1
    };
  },

  handleSelect(key) {
    alert('selected ' + key);
    this.setState({key});
  },

  render() {
    return (
      <Tabs activeKey={this.state.key} onSelect={this.handleSelect} id="controlled-tab-example">
        <Tab eventKey={0} title="Tab 1">Tab 1 content</Tab>
        <Tab eventKey={1} title="Tab 2">Tab 2 content</Tab>
        <Tab eventKey={2} title="Tab 3" disabled>Tab 3 content</Tab>
      </Tabs>
    );
  }
});

ReactDOM.render(<ControlledTabs />, mountNode);
```

grovesNL:
I believe it's caused by [`child.props.eventKey` being falsy](https://github.com/react-bootstrap/react-bootstrap/blob/7d342c38861d61f3a93893da3d87e768eb621732/src/Nav.js#L164). Is this intentional?

taion:
so that's wrong then – feel free to fix eventKey falsiness checks to be nulliness checks

grovesNL:
@taion Thanks, I'll submit a PR when I have some time

PR #2541, opened 2017-04-04
#2539: change condition from truthy to not nully

grovesNL:
As described in #2539, certain falsy values such as `0` should be acceptable. The truthy condition has been changed to check whether the value is `null` or `undefined` instead.


taion:
huh, weird build failure.

code looks fine to me. wonder if we should have a test – the concern here would be regressions, not bugs in this code specificall

grovesNL:
@taion I changed another condition that I had missed initially, and added two tests for acceptable number/string event keys. Let me know if it's sufficient.

PR #2542, opened 2017-04-05
Add fullwidth support

trajano:


taion:
Not sure this makes sense here? This use isn't documented on http://getbootstrap.com/components/#navbar.

trajano:
There is no need for the container to be put in between as shown in https://github.com/BlackrockDigital/startbootstrap-sb-admin-2/blob/master/pages/index.html#L43 because of the forced container there was no way to make the navigation flush.

trajano:
Personally I would like to merge the two "fullwidth" and "fluid" into a single setting but that would involve a bit more change to the API and be really specific to the react implementation.

Another name we can use is `navbarContainer: false` or `navbarContainerWrapper: false`  that would prevent the container from being created.

jquense:
why doesn't `fluid` work here? My impression is that navbar should work fine without a container, but I don't know if it's officially supported

trajano:
@jquense because `fluid` still adds a bit of padding.  For the time being I created a workaround on my site to get around the issue https://github.com/trajano/reactjs-admin/blob/master/src/components/FullWidthNavbar.jsx

taion:
Why not just handle this in CSS?

jquense:
Ditto. I think this is probably a rare enough use case and the css reset is straightforward, adding a new api doesn't make sense. Shouldn't the css jsut be `.navbar > .container-fluid { padding: 0; }`

trajano:
@taion @jquense depends on train of thought, I prefer to avoid adding "CSS workarounds" if possible in addition it's not that bootstrap *required* a container in the `navbar` the implementation that was done by react-bootstrap explicitly added it.

trajano:
If it were to be "cleaner" then it should've been

    <Navbar>
        <Container>
                 ...

But that would add a breaking change to the users.

taion:
Again, all the TWBS examples use `.navbar > .container`. It's great that you're happy with what you have without using the `.container`, but that's an extremely weak sense of "not required".

jquense:
looping back around to this. thanks for the PR, but I think we are going to pass on adding a new api for this at the current time.

PR #2543, opened 2017-04-05
remove onClose prop from dropdown

parikhshiv:
Removed onClose prop from \<Dropdown\> per #1894.


jquense:
thanks, this needs to go into the `next` branch since it's a breaking change

PR #2579, opened 2017-04-22
Update react-overlays

jquense:
I think this is better as a major bump...safer. I'd like to not do that, but `prop-types` really isn't actually supported on < 15.3


taion:
looks fine to me, but it looks like something broke around root close wrapper in the upgrade?

csillag:
So, what's the next step with this?

csillag:
Great! Now, we just need a new release, and then we will have one less useless warning on the browser console...

PR #2584, opened 2017-04-24
Pr/2395 updated

jquense:
updated #2395

### DIFF
--- a/src/ModalHeader.js
+++ b/src/ModalHeader.js
@@ -42,7 +42,7 @@ const contextTypes = {
 class ModalHeader extends React.Component {
   render() {
     const {
-      'aria-label': label,
+      'aria-label': closeLabel,
       closeButton,
       onHide,
       className,
@@ -65,9 +65,9 @@ class ModalHeader extends React.Component {
           <button
             type="button"
             className="close"
-            aria-label={label}
             onClick={createChainedFunction(modal && modal.onHide, onHide)}
           >
+            <span className="sr-only">{closeLabel}</span>
             <span aria-hidden="true">
               &times;
             </span>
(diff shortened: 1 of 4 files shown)


taion:
@jquense Should `<Modal.Header>` use a `.sr-only` label as well, instead of `aria-label`?

jquense:
I didn't change it because the upstream docs use `aria-label` in that case. There may be more support for it perhaps? Or you just mean the close button in Modal.Header?

taion:
the upstream docs use an `aria-label` for the alert close button as well: http://getbootstrap.com/javascript/#alerts-usage

i thought there was something with the `.sr-only` label working better with more SRs, though

jquense:
The usage is slightly different tho, since in this case one is a Button and the other is jsut a div. I think the problem is due to JAWS handling of button text, not just it's reading of aria-label (which is does generally)

taion:
what do you mean? it's a `<button>` in both cases, no?

jquense:
what aria-label are we talking about? I was looking at the Modal.Title, do you jsut mean the close button there?

jquense:
wait...

jquense:
wtf

jquense:
nvm I'm confusing two things. yes we should make this consistent with the alert close btn

src/ModalHeader.js
```diff
@@ -65,9 +65,9 @@ class ModalHeader extends React.Component {
           <button
             type="button"
             className="close"
-            aria-label={label}
             onClick={createChainedFunction(modal && modal.onHide, onHide)}
           >
+            <span className="sr-only">{closeLabel}</span>
```
jquense: order shouldn't matter but i can change it. I do want to fix the naming, i forgot this was going into `next` 👍

taion: better to be consistent than inconsistent between two parallel use cases :p

jquense: better yet just make it a component

PR #2585, opened 2017-04-24
remove onClose from dropdown

jquense:
cleaned up #2543


taion:
not really sure why that showed conflicts...

taion:
oh dammit, we need to update `next`

taion:
oh or just this PR didn't reflect it :p

PR #2586, opened 2017-04-24
remove createClass in tests

jquense:


jquense:
gonna merge since it's a trivial doc only fix

PR #2587, opened 2017-04-24
remove pagination logic

jquense:


taion:
i'm fine with this, but do we want to just drop this as-is with no deprecation for the next point release? it might be nice to factor this out into a separate library to provide an upgrade path of some sort

jquense:
well i'm really sure what a deprecation would like, and i don't want to just sit on a point release for a while to let folks see warnings. I think we could put the old code in a separate library if you want? The goal is to not maintain stuff though.

taion:
yeah, i don't know. the difference w/the form case is that the logic here even in special cases is fairly non-trivial (as evidenced by the number of bugs we've hit around edge cases), so i think it's valuable _somewhere_. it just doesn't necessarily make a lot of sense to me here. i think pulling it out into a single-file, unmaintained library with a big "looking for maintainers" flag might be reasonable. something like... emitting sentinels for all the special values or whatever

jquense:
Ya that's reasonable I'll move the old code to a separate repo

jquense:
ugh. the original code is super wonky. I was pulling out, but like its all sort of wrong. It doesn't render links/spans correctly (everything is a link)...the whole buttonComponentClass is...weird

jquense:
Ok upgrade path provided. gtg?

taion:
@jquense What did you think of my comment on the pagination repo? Would a "smaller" API like what I suggested there make sense?

jquense:
I think either we should provide a drop-in replacement or just point people to one of the many libraries that already do what you'r suggesting, ie. return a bunch of sentinels. As in the experiment there `ultimate-pagniation` already does it

taion:
Okay, if there are plenty of libraries that already do that, would it make sense just to show integration with those? I feel like essentially re-implementing the current API, just in a separate library where people's issues will presumably get ignored (or still take time from us) is not necessarily a win.

jquense:
I do have some info on doing the implementation in the pagination repo. I think in general its ok to have a ropin-replacement for a version or so. lets close the issue-tracker and add a bit warning on it?

taion:
Well if we close the issue tracker then people won't be able to file an issue to claim maintainership :p

jquense:
I was thinking that given how easy it is to implement this component using something else that would be ok. The world doesn't need another bootstrap pagination component

taion:
ok, so just "do nothing" for now, then? and see if people complain?

jquense:
maybe? I'd like to not have anything new to maintain, while also not have angry ppl complaining about no drop-in upgrade path

jquense:
I attempted to do that the first pass but it was complicated to do without knowledge of the function of the buttons, or putting an event key on each one, which felt really verbose. Open to suggestions tho

taion:
Well, mostly the items will be programmatically generated, right? We could do something like supporting a:

```js
<Pagination.Item page={3} />
```

that unrolls into

```js
<Pagination.Item eventKey={3}>
  3
</Pagination.Item>
```

jquense:
you can't do that tho unless the top-level component reads `page` instead of `eventKey` which is ine but weird as it differs from other places. plus it doesn't really help with waht you do when someone clicks ellipsis or prev/next buttons

taion:
Can `<Pagination>` can just pass `onSelect` down to each of its children?

jquense:
what would it do tho? just be a callback with the evetKey? we usually do some controlled input thing in these cases, but we can't here without codifying a lot of logic

taion:
yeah so `<PaginationItem>` gets an `onSelect(eventKey, event)` callback, and then it can call it as appropriate, and figure out its own `eventKey`? is that too inconsistent with everything else we have?

jquense:
i'm just not sure if that's particularly useful considering you need to add your own onSelect logic on top of that if you use the other button types. I don't want do the bsRole thing here to distinguish children types either. To much annoyance around that with ppl wrapping them, etc.

taion:
I think the way I'd have that work would be to have e.g. `<Pagination.Prev>` default to using an implicit `eventKey` of `'prev'`, maybe?

It just seems like it ought to be possible to put the `onSelect` on `<Pagination>`. On the other hand, adding it back would be a feature, so perhaps it's not blocking here? What do you think?

jquense:
I think there isn't really any way of having an onSelect without having the component handle all the logic, unless onSelect doesn't do anything aside from handle back it's event key. you can't control it or have it manage something like activeItem. In that case the page buttons are the thing that can be active while the other buttons only switch the active page. in any case tho you end up adding in a bunch of logic to handle pagination in which case its sort of just the same as what we already have.

jquense:
which is to say I don't think the "simple" on select case gains you much but limits generic use, and makes it weirder to incorporate in the complex case

taion:
I don't think `onSelect` should do anything other than just call itself with `eventKey`. I just feel like the natural API is something like:

```js
<Pagination
  onSelect={(eventKey) => { goToPage(eventKey); }
>
  <Pagination.Page page={1} />
  <Pagination.Page page={2} />
  <Pagination.Page page={3} />
</Pagination>
```

Rather than the user passing `onSelect` to each item.

taion:
Fair enough. I guess we can always add this later as a non-breaking change, if it becomes necessary?

jquense:
Yes very true easier to add then take away again

jquense:
gtg for now then?

src/PaginationItem.js
```diff
@@ -0,0 +1,77 @@
...
+    }
+  };
+}
+
+export const First = createButton('First', '\u00ab');
+export const Prev = createButton('Prev', '\u2039');
+export const Ellipsis = createButton('Ellipsis', '\u2026', 'More');
+export const Next = createButton('Next', '\u203a');
+export const Last = createButton('Last', '\u00bb');
+
```
taion: delete extra newline

src/index.js
```diff
@@ -46,7 +46,6 @@ export PageHeader from './PageHeader';
 export PageItem from './PageItem';
 export Pager from './Pager';
 export Pagination from './Pagination';
-export PaginationButton from './PaginationButton';
```
taion: should we explicitly expose `PaginationItem`?

jquense:
they will _always_ complain

alasdairhurst:
So, what's the upgrade path for this?  Changelog isn't overly helpful on what we need to modify on breaking changes.

NFNChris:
Clearly people haven't complained enough about this change.  I'm late to the party, and not familiar with whatever issues prompted this change, but the goal of these components should always be to *reduce boilerplate code*.  This change has significantly *increased boilerplate code*.  That we have to write a for loop to enumerate page numbers is pretty ridiculous and unnecessary.

Was the impetus for this change discussed in more detail somewhere else?  There's plenty of examples of logic built into react-bootstrap, so I'm very curious why you've suddenly decided that this is "bad".

taion:
As noted in the docs, the logic for generating the pagination entries is available in https://www.npmjs.com/package/@react-bootstrap/pagination. Ultimately it was just not a great API to confuse the presentation with the real logic.

NFNChris:
So prior to the change, my pagination code looked like this:

```
<Pagination 
  prev next first last ellipsis boundaryLinks 
  bsSize="large" 
  items={search.pages} 
  maxButtons={5} 
  activePage={search.page} 
  onSelect={search.setPage} />
```


Now it looks like this:

```
let items = [];
const minPage = Math.max(search.page - 2, 1);
const maxPage = Math.min(search.page + 2, search.pages);

for (let page = minPage; page <= maxPage; page++) {
  items.push(
    <Pagination.Item 
      active={page === search.page}
      onClick={() => search.setPage(page)}>
        {number}
    </Pagination.Item>
  );
}

<Pagination>
  <Pagination.First />
  <Pagination.Prev />
  {search.page - 2 > 1 &&
    <Pagination.Item>{1}</Pagination.Item>
    <Pagination.Ellipsis />
  }
  {items}
  {search.page + 2 < search.pages &&
    <Pagination.Ellipsis />
    <Pagination.Item>{search.pages}</Pagination.Item>
  }
  <Pagination.Next />
  <Pagination.Last />
</Pagination>
```

The only real logic implemented in this case is reducing boilerplate -- generating the page numbers, ellipsis, noting the active page, and attaching a callback to each item.  That sounds like pretty reasonable presentation logic. 

> Ultimately it was just not a great API to confuse the presentation with the real logic.

I guess I'm a bit confused by what you consider 'real logic'.  Generating markup for page numbers?

jquense:
You can go ahead and read the old code, there is way more logic than generating markup. I'm glad you were happy with old code, continuing to use it, that's why we extracted into a new package. While also making the core components more in line with what upstream bootstrap provides. Thanks

PR #2590, opened 2017-04-25
Upgraded `react-overlays` version

jankowiakdawid:
# Why
`react-overlays@0.6.12` was not using `prop-types` and that was producing a worning after upgrading to React 15.5.4.

This PR should fix our problem.


jquense:
Already done in the next branch, thanks

PR #2601, opened 2017-04-28
Some minor improvements

SebastiaanNijland:
* Set `type="button"` with a `defaultProp` instead of logic
* Use Button components instead of plain buttons
* Added a simple test for the Accordion component
* Added a small test case for a Label component with multiple children
* Removed some already implemented TODO’s


src/CloseButton.js
```diff
@@ -10,14 +12,13 @@ class CloseButton extends React.Component {
   render() {
     const { label, onClick } = this.props;
     return (
-      <button
-        type="button"
+      <Button
```
jquense: This won't work. Close buttons aren't implemented in terms of Buttons they are a separate thing. It's odd I know but that's how the styling works

src/NavbarToggle.js
```diff
@@ -39,19 +39,19 @@ class NavbarToggle extends React.Component {
 
     if (children) {
       return (
-        <button {...buttonProps}>
+        <Button {...buttonProps}>
           {children}
-        </button>
+        </Button>
```
jquense: Same as close button, this isn't a bootstrap btn

src/NavbarToggle.js
```diff
@@ -39,19 +39,19 @@ class NavbarToggle extends React.Component {
...
 
     return (
-      <button {...buttonProps}>
+      <Button {...buttonProps}>
         <span className="sr-only">Toggle navigation</span>
         <span className="icon-bar" />
         <span className="icon-bar" />
         <span className="icon-bar" />
-      </button>
+      </Button>
```
jquense: Ditto

test/AccordionSpec.js
```diff
@@ -0,0 +1,20 @@
...
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Accordion>
+        Accordion content
+      </Accordion>
+    );
+
+    const accordion = ReactDOM.findDOMNode(instance);
+
+    assert.equal(accordion.nodeName, 'DIV');
+    assert.equal(accordion.getAttribute('role'), 'tablist');
```
jquense: Can this go in the PanelGroup tests?

SebastiaanNijland: Of course!

src/CloseButton.js
SebastiaanNijland: Ahh, too bad. But makes sense.

src/Button.js
```diff
@@ -47,7 +48,7 @@ class Button extends React.Component {
     return (
       <Component
         {...elementProps}
-        type={elementProps.type || 'button'}
```
taion: slightly different semantically... do we ever pass in falsy (but defined) `type` here?

SebastiaanNijland: That shouldn't be possible though, since type is defined as `oneOf` the truthy button types above?

taion: ah, okay, great

taion:
What's up?

SebastiaanNijland:
The PR is quite out of date :) I'll hopefully be able to get it up to date soon.

PR #2630, opened 2017-05-23
Fix position bug when changing placement for OverlayTrigger

andreisoare:
There is a bug when a component changes the placement for an OverlayTrigger.

If for example the original placement is to the right, and the overlay doesn't fit fully into the window, then OverlayTrigger squeezes it (increases height) so that the entire contents are still visible.

Then when the placement is changed to "top", `<Position>` uses the stretched overlay's width and height to compute the new `top` and `left` styles for the overlay.

My PR forces the recreation of the overlay so that it has its original width and height.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -129,6 +129,10 @@ class OverlayTrigger extends React.Component {
   }
 
   componentDidUpdate() {
+    if (this.state.show) {
+      ReactDOM.unmountComponentAtNode(this._mountNode);
+      this._mountNode = document.createElement('div');
+    }
     this.renderOverlay();
   }
 


taion:
This case shouldn't be handled here – all the utilities for managing the measurement and positioning live in https://github.com/react-bootstrap/react-overlays.

andreisoare:
The problem originates here though. It's because the `_mountNode` from previous renders is being reused.

andreisoare:
Are you thinking that `<Position>` should make a copy of the overlay without height/width constraints when computing `top` and `left`?

taion:
Well, sizing is all handled by `<Position>`. Either way, this sort of DOM thrashing should not be the default behavior, anyway.

Issue #2560, opened 2017-04-12
pound sign href="#" changes my url unexpected


rfvyhn:
I am using a SPA

route is using `/#/route`

now I have tabs build with react-bootstrap

```
        <Tabs id='stati-tabs'
          bsStyle='tabs'
          activeKey={this.state.tabActive}
          onSelect={this.showTab}>
          <Tab eventKey={1} title={this.t('tab.counts')}></Tab>
          <Tab eventKey={2} title={this.t('tab.users')}></Tab>
          <Tab eventKey={3} title={this.t('tab.hours')}></Tab>
        </Tabs>
```

evenything is good.

but click the Tab will changes my url to `/#`

I tried

`<Tab onClick={e => e.preventDefault()} title={this.t('stati.tab.view_counts')}></Tab>`

`<Tab href={undefined} title={this.t('stati.tab.view_counts')}></Tab>`

`<Tab href={''} title={this.t('stati.tab.view_counts')}></Tab>`

any ways not working.

I have no idea now.

taion:
your router is probably hijacking clicks on `<a>` elements everywhere in your document to do navigation

use a better router

zourtney:
Requesting this be reopened. "Use a better router" is not an acceptable solution for me (slowly adding React to an old Angular 1.x / UI-Router project).

I've created a pull request that passes along a `Nav`'s `href` prop down to the `SafeAnchor`. When combined with the old `href="javascript:void(0)` hack, this component is at least usable.

Example:

```jsx
<Tabs id="test" bsStyle="pills" defaultActiveKey={1} animation={false}>
  <Tab title="Tab 1" eventKey={1} href="javascript:void(0)">Tab 1 content</Tab>
</Tabs>
```

zourtney:
_Revising my previous comment (pull request denied on the grounds of maintaining a simple API --
 totally understandable). So, for posterity's sake:_

If your hash-router is prematurely navigating, [recreate the component tree](https://react-bootstrap.github.io/components.html#tabs-with-dropdown), then create a `NavItem` with a custom `href` prop. The value doesn't seem to matter, so long as it's not falsy.

For example:

```jsx
<Tab.Container id='MyTabContainer' defaultActiveKey={0}>
  <Row className='clearfix'>
    <Col sm={12}>
      <Nav bsStyle='tabs'>
        <NavItem eventKey={0} href='javascript:void(0)'>
          {m.name}
        </NavItem>
      </Nav>
[...]

Issue #2476, opened 2017-02-14
Link on Tab


Shadowman4205:
So how add link to tab?
i tried this on Tab.Container
```
<NavItem  href="https://www.google.com" eventKey="Google">Google</NavItem>
```
But nothing happened.

jquense:
hi there,  please look at the code for the examples on the docs site.

Shadowman4205:
I did i used this 

```
const tabsInstance = (
  <Tab.Container id="tabs-with-dropdown" defaultActiveKey="first">
    <Row className="clearfix">
      <Col sm={12}>
        <Nav bsStyle="tabs">
          <NavItem eventKey="first">
            Tab 1
          </NavItem>
          <NavItem eventKey="second">
            Tab 2
          </NavItem>
          <NavItem href="https://www.google.com" target="_blank" eventKey="1">
                                    Google
         </NavItem>
        </Nav>
      </Col>
      <Col sm={12}>
        <Tab.Content animation>
          <Tab.Pane eventKey="first">
            Tab 1 content
          </Tab.Pane>
          <Tab.Pane eventKey="second">
            Tab 2 content
          </Tab.Pane>
        </Tab.Content>
      </Col>
    </Row>
  </Tab.Container>
);

ReactDOM.render(tabsInstance, mountNode);
```
Are you refer on different example?

huyvlam:
I experience the same issue. Tab.Container is rendering the <a href> inside <li class="tab"> but stop the click event from propagating. I used onSelect event to redirect to the href value, but this solution is rather undesirable. What's the rationale for this? The component design here doesn't quite make sense.

PR #2632, opened 2017-05-23
Tab: pass `href` prop down to inner <a> element

zourtney:
Allow `href` to be passed to `<Tab>`s. This can be used for normal navigation, or combined with the old `javascript:void(0)` hack to circumvent link-jacking by overzealous hash routers.

```jsx
<Tab title=“Tab 1” href=‘javascript:void(0)’>…</Tab>
```

See #2560, #2476.

### DIFF
--- a/src/Tab.js
+++ b/src/Tab.js
@@ -10,6 +10,8 @@ const propTypes = {
 
   disabled: PropTypes.bool,
 
+  href: PropTypes.string,
+
   title: PropTypes.node,
 
   /**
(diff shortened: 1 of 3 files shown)


zourtney:
**Note:** I realize I can get the same functionality by manually constructing the `Tab.Container`, `Nav`, `NavItem` tree. This is cumbersome, so this pull request is meant to be a shortcut around that.

taion:
Thanks for your PR!

We're deliberately keeping the `<Tabs>` and `<Tab>` API minimal; in our experience, trying to build more than the minimum flexibility into these types of API lead relatively quickly to thinks become unmanageable.

It is our intention that these use cases be handled with the lower-level, more flexible API.

zourtney:
Understood. Thanks for the quick response!

Sounds like the voice of experience. And hey, if nothing else, I got a quick peek into the internals of the library. Everything was easy to follow, so good job 👍

Issue #2634, opened 2017-05-24
Checkbox applies title prop to checkbox rather than label


Billy-:
Title says it all. This means the tooltip only appears if you hover over the checkbox itself and not for the text. Working on a PR right now.

taion:
fixed in #2635

PR #2635, opened 2017-05-24
Update Checkbox.js to apply title to label

Billy-:
RE: #2634

### DIFF
--- a/src/Checkbox.js
+++ b/src/Checkbox.js
@@ -9,6 +9,7 @@ import { bsClass, getClassSet, prefix, splitBsProps }
 const propTypes = {
   inline: PropTypes.bool,
   disabled: PropTypes.bool,
+  title: PropTypes.string,
   /**
    * Only valid if `inline` is not set.
    */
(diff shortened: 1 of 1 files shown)


jquense:
looks ok to me, can you fix the lint issue tho please thanks

Billy-:
Sorry, I edited on GitHub so didn't see that. Have just pushed another commit which should fix it.

taion:
This should apply to `<Radio>` as well, at least.

Probably also all the other cases where we pass `className` and `style` to the parent, but other props to the child?

Billy-:
Probably, I'm afraid I don't have time to do all of them, at least not any time soon. We're very pushed for time on our current project :confounded:. Sorry!

Billy-:
Any chance this can be merged and a separate issue filed for the other inputs? If no one has time to do it there's no point holding this back...

hostr:
> We'd like to fix Radio at least, they are essentially the same input so the fix should be quick

Done -> https://github.com/react-bootstrap/react-bootstrap/pull/2647

PR #2638, opened 2017-05-26
upgrade to Webpack 2

jochenberger:
This doesn't seem to have a huge impact on the bundle size, so maybe something's still missing.


GoAlek:
When this change is expected be released and available in npm repository?

jquense:
@GoAlek not sure, but it shouldn't affect you anyway this a change to build tooling, at best it might make the bundled size of the browser build smaller

webpack/base.config.js
```diff
@@ -20,8 +23,8 @@ const baseConfig = {
   externals: undefined,
 
   module: {
-    loaders: [
-      { test: /\.js/, loader: jsLoader, exclude: /node_modules/ }
+    rules: [
+      { test: /\.js/, use: [jsLoader], exclude: /node_modules/ }
```
taion: @jochenberger One quick q – why use an array here?

jochenberger: Oh, I assumed that an array was required there. I guess I saw it being used like that in an example. Want me to change it?

jochenberger:
As I said, the bundle change does not change very much.

webpack/base.config.js
taion: i'd just use a non-array here, but it's your call

taion:
I rebased to try to fix CI

jochenberger:
Yes, there are issues with HMR. I couldn't get it to work yet.

taion:
I think the old RHL just doesn't work with webpack 2. This is still a good step forward though – we just need to figure out how to get HMR working again, because the components page is too big to fully rerender.

jochenberger:
getting closer, now I get XHR failures because of CORS.

jochenberger:
Huh, I'm getting those on master too.

docs/dev-run.js
```diff
@@ -64,7 +64,7 @@ portfinder.getPorts(2, {}, (portFinderErr, [docsPort, webpackPort]) => {
     process.exit(1);
   }
 
-  runCmd('webpack-dev-server', `nodemon --watch webpack --watch webpack.docs.js --exec webpack-dev-server -- --config webpack.docs.js --color --port ${webpackPort} --debug --hot --host ${ip.address()}`);
+  runCmd('webpack-dev-server', `nodemon --watch webpack --watch webpack.docs.js --exec webpack-dev-server -- --config webpack.docs.js --color --port ${webpackPort} --debug --hot --inline --host ${ip.address()}`);
```
jquense: can we leave host out of this? I'm not sure why it's needed to begin with at this point

taion: I think @mtscout6 added it to avoid CORS issues or something.

mtscout6: It was for cross browser testing from another machine: https://github.com/react-bootstrap/react-bootstrap/commit/9536fa6a13e4414b823720b498177405026c0d75

jochenberger: Thanks @mtscout6. So, do you want it removed?

mtscout6: I'm not in this code much anymore so I'll leave that decision to you guys. My use case for this was to launch the docs on my mac, then load the docs site on my Windows VM in Internet Explorer. Since we all know how awesome Internet Explorer can be it may be beneficial to leave this in here so you guys can debug in that browser without hosting your dev environment on Windows. It's entirely up to you guys.

jquense: i'm ok with removing it for now. I think something like the SauceLabs tunnel can handle these cases for manual tests if we need them :)

taion:
Is this irrelevant now given #2818?

jochenberger:
Probably not.

taion:
Oh, shoot, I guess either way we should be using webpack 2 for bundling?

jochenberger:
https://github.com/react-bootstrap/react-bootstrap/pull/2818/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R85 even moves to Webpack 3.

jochenberger:
We might consider using rollup, but we can check that out after #2818, there's no use comparing it with Webpack 1.

PR #2645, opened 2017-05-30
Fix where we get ReactTestUtils

taion:


jquense:
lgtm

taion:
Rebuilding on Travis without cache

PR #2648, opened 2017-06-01
Removed all propTypes references from react package

manubamba:


Tiagodsn:
👍

jononu:
Any idea when this will ship in a new version?  We'd like to have it before upgrading our React version.  Thanks for the awesome library!

Issue #2338, opened 2016-11-17
Warning popup when using "Tabs with Dropdown" generated `id` and `aria-controls` attributes for the sake of proper component accessibility


mrmons:
I follow the sample provider by here 

https://react-bootstrap.github.io/components.html#tabs
I am using the "Tab with Dropdown"

And I get this error
browser.js:49
Warning: In the context of a `<TabContainer>`, `<NavItem>`s are given generated `id` and `aria-controls` attributes for the sake of proper component accessibility. Any provided ones will be ignored. To control these attributes directly, provide a `generateChildId` prop to the parent `<TabContainer>`.

Issue been file before but it seen like been fixed? But maybe not "Tab with Dropdown"?
https://github.com/react-bootstrap/react-bootstrap/issues/1854

priand:
I'm seeing something similar with 0.31.2:
Warning: Failed prop type: The prop `id` is required to make `Tabs` accessible for users of assistive technologies such as screen readers.

Issue #2623, opened 2017-05-15
Export CloseButton for public use?


natevw:
There is a CloseButton component used internally. It ± follows Bootstrap's own documented suggestion for such: <https://v4-alpha.getbootstrap.com/utilities/close-icon/>

Any particular reason this can't be exposed for reuse outside of react-bootstrap, either alongside the normal button component or at least under utilities?

taion:
sure go for it

skwon-mdsol:
this can be closed, correct?

taion:
thanks!

PR #2652, opened 2017-06-04
Export CloseButton for public use & Fix example docs error

BrennerSpear:
Comment #2623 

Includes tests

### DIFF
--- a/test/CloseButton.js
+++ b/test/CloseButton.js
@@ -7,11 +7,9 @@ import CloseButton from '../src/CloseButton';
 describe('<CloseButton>', () => {
   it('Should output a button', () => {
     let noOp = () => {};
-    let label = 'Close';
     let instance = ReactTestUtils.renderIntoDocument(
       <CloseButton
         onClick={noOp}
-        label={label}
       />
     );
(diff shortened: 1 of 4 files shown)


BrennerSpear:
What are y'alls thoughts on defaulting the label to "Close" or "Close Item"?

jquense:
I fee good about "Close" as a default yeah

BrennerSpear:
Will add that. Should it be on the same PR or a new one?

BrennerSpear:
Newest commit fixes issue #2338.

Nothing wrong with the codebase - just the example. and ```id``` was getting passed into ```NavDropdown``` when it shouldn't have been.

I can put these into two different PRs if you'd like me to, but I think this commit is small enough to just tack on.

test/CloseButton.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import ReactTestUtils from 'react-addons-test-utils';
```
taion: should be `react-dom/test-utils`

BrennerSpear: Oops. Fixed

PR #2672, opened 2017-06-22
added description for carousel interval prop

rallgood:
Bootstrap's documentation for carousel says that `interval` should be set to `false` to prevent automatic cycling, but this is a failing prop in react-bootstrap. Clarifying that `interval={0}` should be used to prevent automatic cycling in react-bootstrap could be helpful.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -17,6 +17,10 @@ import ValidComponentChildren from './utils/ValidComponentChildren';
 const propTypes = {
   slide: PropTypes.bool,
   indicators: PropTypes.bool,
+  /**
+   * The amount of time to delay between automatically cycling an item.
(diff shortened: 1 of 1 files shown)


jquense:
maybe we should add an explicit false to the propType and docs to be clear?

taion:
Or `null`.

skwon-mdsol:
@rallgood has this been addressed? I'm in support for null due to a mixing of types. It's a slight hack but I'd rather respect the prop types here. I'm all for consistency with the standard bootstrap docs

taion:
I think `null` makes the most sense semantically with how we want typing to work.

We want to preserve the same functionality, but the API doesn't have to be the exact same down to every detail when we have reason to not do so.

PR #2674, opened 2017-06-26
Fixes #2365: CollapseOnSelect in Responsive Navbars prevents external links from working

### LINKED ISSUE #2365: CollapseOnSelect in Responsive Navbars prevents external links from working

emilyemorehouse:
See the issue for more information.

### DIFF
--- a/src/NavItem.js
+++ b/src/NavItem.js
@@ -28,7 +28,7 @@ class NavItem extends React.Component {
   }
 
   handleClick(e) {
-    if (this.props.onSelect) {
+    if (this.props.onSelect && this.props.onSelect.name !== 'bound handleCollapse') {
       e.preventDefault();
 
       if (!this.props.disabled) {


src/NavItem.js
```diff
@@ -28,7 +28,7 @@ class NavItem extends React.Component {
   }
 
   handleClick(e) {
-    if (this.props.onSelect) {
+    if (this.props.onSelect && this.props.onSelect.name !== 'bound handleCollapse') {
```
jquense: my guess is that this will be too fragile across browsers and platforms, but it does illustrate the issue nicely

emilyemorehouse: agreed. my other thought was to pass down collapseOnSelect as a prop and check against that.

jquense: TBH i'm not sure why we even preventDefault here at all... @taion and idea? seems like bad practice to mess with any link

emilyemorehouse: when using onSelect, it's stopping the default browser behavior (if any) and runs only the custom function.

emilyemorehouse: but yes, i'm not sure why onSelect would be expected to preventDefault. it would make more sense to allow default behavior to happen and put it on the user to ensure that they use (or don't use) anything that would trigger a behavior even when using onSelect.

taion:
Closing this out, since we don't want to fix things this way.

Issue #2687, opened 2017-07-05
'onError' was not passed down in Thumbnail


Yuyz0112:
It looks like we are not passing down 'onError' in Thumbnail component. Users may need that event handler to show a placeholder when image broken.

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Thumbnail.js#L27

skwon-mdsol:
http://andrewhfarmer.com/detect-image-load/ Or you could detect whether the image loads correctly or not and implement accordingly via a parent component.

skwon-mdsol:
Actually bump for this, makes sense that we should pass down onError and onLoad props. I can take a stab at this?

jquense:
go for it @skwon-mdsol 👍  thanks

taion:
fixed in #2691

PR #2691, opened 2017-07-06
#2687 Pass onError and onLoad props to its child image

skwon-mdsol:
# Description

Addressing https://github.com/react-bootstrap/react-bootstrap/issues/2687

We want to be able to pass the onLoad and onError props from the parent Thumbnail prop to the child img tag. This pull request implements that functionality, as well as adding a few minimal descriptions for each prop in the thumbnail component

* [x] Added tests for the new props
* [x] Check to see if the docs display the new prop types and their descriptions

### Also
This is my first pr for this repo so I don't entirely understand the specific pr process for this repo. Please request changes if needed.

### DIFF
--- a/test/ThumbnailSpec.js
+++ b/test/ThumbnailSpec.js
@@ -11,7 +11,7 @@ describe('<Thumbnail>', () => {
     alt: 'test',
     onError: () => {},
     onLoad: () => {}
-  }
+  };
   it('Should have a thumbnail class and be an anchor', () => {
     let instance = ReactTestUtils.renderIntoDocument(
       <Thumbnail {...defaultProps} />
@@ -28,9 +28,8 @@ describe('<Thumbnail>', () => {
   });
 
   it('Should have a thumbnail class and be a div', () => {
-    const {href, ...otherProps} = defaultProps;
     let instance = ReactTestUtils.renderIntoDocument(
-      <Thumbnail {...otherProps} />
+      <Thumbnail src="#" alt="test" />
     );
     assert.ok(ReactDOM.findDOMNode(instance).className.match(/\bthumbnail\b/));
     assert.equal(ReactDOM.findDOMNode(instance).nodeName, 'DIV');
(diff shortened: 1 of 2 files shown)


test/ThumbnailSpec.js
```diff
@@ -5,17 +5,24 @@ import ReactTestUtils from 'react-dom/test-utils';
 import Thumbnail from '../src/Thumbnail';
 
 describe('<Thumbnail>', () => {
+  let defaultProps = {
```
taion: probably it was fine without this `defaultProps`? worth verifying that things still work if you don't specify all of these, like we were implicitly doing before

test/ThumbnailSpec.js
```diff
@@ -58,4 +65,25 @@ describe('<Thumbnail>', () => {
...
   });
+
+  it('Should have an img with an onError callback', () => {
+    const onErrorSpy = sinon.spy();
+    let instance = ReactTestUtils.renderIntoDocument(
+      <Thumbnail src="#" href="#" onLoad={() => {}} onError={onErrorSpy} />
+    );
+    let img = ReactTestUtils.findRenderedDOMComponentWithTag(instance, 'img');
+    ReactTestUtils.Simulate.error(img);
+    assert.ok(onErrorSpy.calledOnce);
```
taion: let's do something like

```js
expect(onErrorSpy).to.have.been.calledOnce
```

test/ThumbnailSpec.js
```diff
@@ -58,4 +65,25 @@ describe('<Thumbnail>', () => {
     );
     assert.ok(ReactDOM.findDOMNode(instance).lastChild.className.match(/\bcaption\b/));
   });
+
+  it('Should have an img with an onError callback', () => {
+    const onErrorSpy = sinon.spy();
+    let instance = ReactTestUtils.renderIntoDocument(
```
taion: use `const` for these

test/ThumbnailSpec.js
skwon-mdsol: ok, from here on, should i rely on `expect`? I've seen assert.ok in a lot of places

test/ThumbnailSpec.js
skwon-mdsol: hmm, makes sense. will address in a commit soon.

test/ThumbnailSpec.js
taion: I think in general `expect` is nicer. It wouldn't be worth the effort to replace all `assert` instances, but it'd be good to use `expect` for new code.

skwon-mdsol: nice, I agree 👍

Issue #2449, opened 2017-01-26
Remove h3 from popover title


cjtafoya:
Since a node can be passed in (which is great), doesn't make sense to nest the prop within an h3 tag  (the .popover-title class overrides all the relevant mark up anyway)

drsloth:
bump, I also would like to replace the h3

taion:
@jquense I wonder if we should have `<Popover.Title>` and `<Popover.Content>`... ahaha

bpas247:
So would this just be a re-design of the `Popover` component to make it more modular @taion?

taion:
yup, i think that's how this would look. instead of taking a `title` prop, have a `<Popover.Title>` sub-component as in everything else

PR #2694, opened 2017-07-07
#2449 removed h3 from Popover

jshaker:


skwon-mdsol:
Not sure if this is the solution to the popover markup, nor was it ever decided by the owners to do such a thing. This could potentially be  abreaking change.

taion:
Yeah, this isn't safe – this changes what gets rendered.

skwon-mdsol:
I think this one should be closed.

taion:
Okay, closing this for now. I think the right way to go about things here is to have a separate `<Popover.Title>` component per https://github.com/react-bootstrap/react-bootstrap/issues/2449#issuecomment-277114258.

PR #2695, opened 2017-07-07
Fixes to minor typos

cyanliu:
I also adjusted the length of lines for some of the paragraph blocks to allow for more readable code (this doesn't affect how the text is rendered on the page)

### DIFF
--- a/docs/src/sections/NavbarSection.js
+++ b/docs/src/sections/NavbarSection.js
@@ -42,8 +42,8 @@ export default function NavbarSection() {
       <h3><Anchor id="navbars-mobile-friendly">Responsive Navbars</Anchor></h3>
       <p>
         To have a mobile-friendly Navbar, add a <code>Navbar.Toggle</code> to your Header and wrap your
-        Navs in a <code>Navbar.Collapse</code> component. The <code>Navbar</code> will automatically wire
-        the toggle and collapse together!
+        Navs in a <code>Navbar.Collapse</code> component. The <code>Navbar</code> will automatically combine
+        the toggle and collapse.
       </p>
       <p>
         Set the <code>defaultExpanded</code> prop to make the Navbar start expanded. Set <code>collapseOnSelect</code>
--- a/docs/src/sections/NavbarSection.js
+++ b/docs/src/sections/NavbarSection.js
@@ -41,12 +41,14 @@ export default function NavbarSection() {
 
       <h3><Anchor id="navbars-mobile-friendly">Responsive Navbars</Anchor></h3>
       <p>
-        To have a mobile friendly Navbar, Add a <code>Navbar.Toggle</code> to your Header and wrap your
+        To have a mobile-friendly Navbar, add a <code>Navbar.Toggle</code> to your Header and wrap your
         Navs in a <code>Navbar.Collapse</code> component. The <code>Navbar</code> will automatically wire
         the toggle and collapse together!
       </p>
       <p>
-        Set the <code>defaultExpanded</code> prop to make the Navbar start expanded. Set <code>collapseOnSelect</code> to make the Navbar collapse automatically when the user selects an item. You can also finely control the collapsing behavior by using the <code>expanded</code> and <code>onToggle</code> props.
+        Set the <code>defaultExpanded</code> prop to make the Navbar start expanded. Set <code>collapseOnSelect</code>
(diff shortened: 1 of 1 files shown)


docs/src/sections/NavbarSection.js
```diff
@@ -41,12 +41,14 @@ export default function NavbarSection() {
 
       <h3><Anchor id="navbars-mobile-friendly">Responsive Navbars</Anchor></h3>
       <p>
-        To have a mobile friendly Navbar, Add a <code>Navbar.Toggle</code> to your Header and wrap your
-        Navs in a <code>Navbar.Collapse</code> component. The <code>Navbar</code> will automatically wire
-        the toggle and collapse together!
+        To have a mobile-friendly Navbar, add a <code>Navbar.Toggle</code> to your Header and wrap your
+        Navs in a <code>Navbar.Collapse</code> component. The <code>Navbar</code> will automatically combine
```
taion: This is not exactly right – the idea here is that the toggle is made to automatically control the collapse, not that they're combined in some way.

docs/src/sections/NavbarSection.js
```diff
@@ -41,12 +41,14 @@ export default function NavbarSection() {
...
+        the toggle and collapse.
       </p>
       <p>
-        Set the <code>defaultExpanded</code> prop to make the Navbar start expanded. Set <code>collapseOnSelect</code> to make the Navbar collapse automatically when the user selects an item. You can also finely control the collapsing behavior by using the <code>expanded</code> and <code>onToggle</code> props.
+        Set the <code>defaultExpanded</code> prop to make the Navbar start expanded. Set <code>collapseOnSelect</code>
```
taion: Usually editors can be set to wrap lines. I've come around to the idea that it's better to not manually wrap lines – otherwise for line length limits, you end up potentially having to touch a lot of lines of docs source just to add or remove a new word or something.

docs/src/sections/NavbarSection.js
skwon-mdsol: yeah, toggle and collapse are actually two separate functionalities, at least semanticaly. was the `together` there necessary?

taion: Their functionality is getting wired together – i.e. the toggle will properly control the collapse.

skwon-mdsol: In that case, how about `The Navbar will grant control of the collapse to the toggle` ? Something in that vein?

docs/src/sections/NavbarSection.js
jquense: You soft wrap in js files? :P I like the shorter lines, easier to read but no need to hard limit it yeah?

docs/src/sections/NavbarSection.js
jquense: It's s more "when used inside a navbar the Toggle will automatically collapse or expand the navbar without any additional setup"

jquense:
Thanks for the the PR, addressing typos and confusing wording is always appreciated.

docs/src/sections/NavbarSection.js
skwon-mdsol: Nice, i like that wording better. I vote for that explicit definition. @taion @jquense you all right with that?

jquense: No need to vote, just explaining the original intent. Lets not all pile on in the comments please

docs/src/sections/NavbarSection.js
cyanliu: err sorry I'm a little new to this, I was just trying to keep it consistent and since there wasn't a lot to this file, and most of the other paragraphs were wrapped, I figured this would be as well. I couldn't find a style guide or anything so this was just a leap of faith that I had taken. There's no need to merge if it's unnecessary!

docs/src/sections/NavbarSection.js
cyanliu: Sorry if the wording had become just as ambiguous as it was before, I had hoped to make it a bit clearer in the rewrite but I wasn't super familiar with the exact nuances of how this worked. I'm glad we were able to talk and clear it up though 😅

taion:
Thanks! I clarified the collapse/toggle interaction in https://github.com/react-bootstrap/react-bootstrap/pull/2701. Let's figure out the docs line wrapping separately.

Issue #2613, opened 2017-05-08
defaultOpen not recognized


aaron-m-kent:
I am trying to add the `defaultOpen` property to a `DropdownButton`, but it is not opening the button and is generating the below warning. 

```Warning: Unknown prop `defaultOpen` on <button> tag. Remove this prop from the element. For details, see https://fb.me/react-unknown-prop
    in button (created by Button)
    in Button (created by DropdownToggle)
    in DropdownToggle (created by Dropdown)
    in div (created by ButtonGroup)
    in ButtonGroup (created by Dropdown)
    in Dropdown (created by Uncontrolled(Dropdown))
    in Uncontrolled(Dropdown) (created by DropdownButton)
    in DropdownButton (created by Unknown)
    ...```

**Comment by taion** [comment]
need to add it to `propTypes`. care to PR?

**Comment by aaron-m-kent** [comment]
Thank you. Even if not in `propTypes`, shouldn't behavior be observed? There is no observable change in behavior. Will adding to `propTypes` produce desired behavior?

**Comment by jquense** [comment]
something else is not right. uncontrollable won't pass through `default` props I suspect the problem is the prob is on the wrong element. can you post your code

**Comment by aaron-m-kent** [comment]
Thanks. Here is my usage:
```
<DropdownButton
  id={id}
  title={title}
  disabled={isDisabled()}
  defaultOpen={true}
>
  {content}
</DropdownButton>
```

taion:
@jquense think the issue is that https://github.com/react-bootstrap/react-bootstrap/blob/v0.31.0/src/DropdownButton.js#L27-L28 won't include `defaultOpen`

skwon-mdsol:
should this issue be addressed? I can open a pr for this if necessary?

taion:
Yeah it should be a simple fix. Thanks!

skwon-mdsol:
nice, will open a pr later.

PR #2696, opened 2017-07-07
#2613 added defaultOpen as a prop for DropdownButton

jshaker:


taion:
You don't need to do this – just add it to propTypes for Dropdown.

jshaker:
#2613 mentions adding it to DropdownButton though

Also, it seems Dropdown is already aware of the defaultOpen prop? 
Line 484:
https://github.com/react-bootstrap/react-bootstrap/blob/master/test/DropdownSpec.js

taion:
The `splitComponentProps` looks explicitly at `propTypes` and will do the right thing. `<Dropdown>` knows about `defaultOpen`, yes, but it doesn't have it listed under `propTypes`.

taion:
I added `defaultOpen` in #2702. Thanks!

Issue #2681, opened 2017-06-30
a11y - NavItem with onClick should also have onKeyDown


andrewdushane:
A NavItem component that gets an `onClick` prop will generate this warning from react-a11y:

```
SafeAnchor You have `role="button"` but did not define an `onKeyDown` handler. Add it, and have the "Space" key do the same thing as an `onClick` handler. 
```

`onClick` is an explicitly specified prop for `NavItem`, so I believe my usage here is correct. It's basically
```javascript
<NavItem onClick={callbackFunc}>
  Do stuff
</NavItem>
```

I don't see this specific issue addressed in any currently open issues or PRs. I am happy to work on this, but wanted to make sure:
a. this isn't user error
b. nobody's already working on it

skwon-mdsol:
Any reason you're not just using the onSelect prop for the nav component that is presumably wrapping your navitem?

taion:
This does seem like a potential oversight on `<SafeAnchor>`. Is there an explicit reason we're doing `role="button"` but not responding to "space" keys @jquense?

andy-j-d:
@skwon-mdsol I definitely could use `onSelect`, but I don't need `selected` state, so it adds a bit of unneeded complexity.

skwon-mdsol:
@andy-j-d hmmm, fair enough

jquense:
I think it's reasonable to add a keydown for space in SafeAnchor, if we can confirm that that is the spec'd expected behavior 👍

taion:
https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-4

looks like it... yet again upstream TWBS does the wrong thing for a11y, i guess

not fixed in TWBSv4 either

andy-j-d:
So according to that link, the keydown behavior should match `onClick` for "enter" and "space". I should be able to open a PR for that next week. Thanks for the feedback @skwon-mdsol, @taion and @jquense.

taion:
"Enter" is already handled on `<a>`, so we just need to handle "space". I think it should be safe to do this? Thoughts anyone?

jquense:
Should be OK to do yeah

PR #2697, opened 2017-07-10
SafeAnchor a11y - Provide keydown handler for "space"

andy-j-d:
~If `onKeyDown` prop is supplied, prefer that.~
If SafeAnchor renders an `a` tag, supply onKeyDown handler for "space" that calls the `onClick` handler.
Test that space bar calls click handler.

Addresses https://github.com/react-bootstrap/react-bootstrap/issues/2681

### DIFF
--- a/src/SafeAnchor.js
+++ b/src/SafeAnchor.js
@@ -1,10 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import elementType from 'prop-types-extra/lib/elementType';
+import keycode from 'keycode';
 
 const propTypes = {
   href: PropTypes.string,
   onClick: PropTypes.func,
+  onKeyDown: PropTypes.func,
   disabled: PropTypes.bool,
   role: PropTypes.string,
   tabIndex: PropTypes.oneOfType([
@@ -36,6 +38,7 @@ class SafeAnchor extends React.Component {
     super(props, context);
 
     this.handleClick = this.handleClick.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
   }
 
   handleClick(event) {
@@ -55,8 +58,14 @@ class SafeAnchor extends React.Component {
     }
   }
 
+  handleKeyDown(event) {
+    if (keycode(event) === 'space') {
+      this.handleClick(event);
+    }
+  }
+
   render() {
-    const { componentClass: Component, disabled, ...props } = this.props;
+    const { componentClass: Component, disabled, onKeyDown, ...props } = this.props;
 
     if (isTrivialHref(props.href)) {
       props.role = props.role || 'button';
@@ -74,6 +83,7 @@ class SafeAnchor extends React.Component {
       <Component
         {...props}
         onClick={this.handleClick}
+        onKeyDown={onKeyDown || this.handleKeyDown}
       />
     );
   }
--- a/src/SafeAnchor.js
+++ b/src/SafeAnchor.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import elementType from 'prop-types-extra/lib/elementType';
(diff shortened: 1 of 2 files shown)


src/SafeAnchor.js
```diff
@@ -55,8 +58,14 @@ class SafeAnchor extends React.Component {
     }
   }
 
+  handleKeyDown(event) {
+    if (keycode(event) === 'space') {
```
taion: just do

```js
event.key === ' '
```

src/SafeAnchor.js
```diff
@@ -74,6 +83,7 @@ class SafeAnchor extends React.Component {
       <Component
         {...props}
         onClick={this.handleClick}
+        onKeyDown={onKeyDown || this.handleKeyDown}
```
taion: should probably chain them

src/SafeAnchor.js
```diff
@@ -55,8 +58,14 @@ class SafeAnchor extends React.Component {
     }
   }
 
+  handleKeyDown(event) {
+    if (keycode(event) === 'space') {
+      this.handleClick(event);
```
taion: we should probably `preventDefault` here regardless to prevent scrolling?

src/SafeAnchor.js
taion: we don't need to do this if the element is actually `<button>`

andy-j-d:
👍 @taion

src/SafeAnchor.js
```diff
@@ -55,8 +58,18 @@ class SafeAnchor extends React.Component {
...
+    }
+  }
+
   render() {
-    const { componentClass: Component, disabled, ...props } = this.props;
+    const { componentClass: Component, disabled, onKeyDown, ...props } = this.props;
+
+    const handleKeyDown = Component === 'a' ?
+      createChainedFunction(this.handleKeyDown, onKeyDown) : onKeyDown;
```
andy-j-d: It would be more performant to do this in the constructor, but if the `onKeyDown` prop were changed after the component mounted, the handler wouldn't update. Don't see another way to do this without breaking backwards compatibility.

src/SafeAnchor.js
```diff
@@ -55,8 +58,18 @@ class SafeAnchor extends React.Component {
     }
   }
 
+  handleKeyDown(event) {
+    event.preventDefault();
```
taion: only if key is space

src/SafeAnchor.js
taion: oops, sorry, i take that back – i forgot that using `<button>` was handled upstream in `<Button>`. you don't need to check for `a` here

test/SafeAnchorSpec.js
```diff
@@ -39,6 +39,17 @@ describe('SafeAnchor', () => {
...
 
+  it('provides onClick handler as onKeyDown handler for "space"', () => {
+    const handleClick = sinon.spy();
+
+    tsp(<SafeAnchor onClick={handleClick} />)
+      .shallowRender()
+      .find('a')
+      .trigger('keyDown', { key: ' ', preventDefault() {} });
+
+    handleClick.should.have.been.calledOnce;
```
skwon-mdsol: `expect(handleClick).to.have.been.calledOnce` ?

andy-j-d: They're equivalent, I was following the pattern of the "it forwards onclick" test https://github.com/andy-j-d/react-bootstrap/blob/b8800dbba51e86ed4bdd34c046f820a67194029c/test/SafeAnchorSpec.js#L39

taion: yeah this is fine as long as we're not using assert – i prefer `expect` but consistency has some value

taion:
@jquense thoughts?

PR #2704, opened 2017-07-13
export ToggleButton, ToggleButtonGroup for public use

AnthonyGreate:


joeellis:
Thanks for this - I was wondering why I kept getting errors when trying to export these two things

meltedice:
I'm looking forward to see v0.31.2 including this PR.

Following code doesn't work with v0.31.1 via yarn/npm.
```
import { ToggleButtonGroup, ToggleButton } from 'react-bootstrap'
// => 'react-bootstrap' does not contain an export named 'ToggleButtonGroup'
```

Need to do like this:
```
import ToggleButtonGroup from 'react-bootstrap/lib/ToggleButtonGroup'
import ToggleButton from 'react-bootstrap/lib/ToggleButton'
```

sm1th:
Thanks @meltedice. Yes I hope it is fixed soon so I can just do `import { ToggleButtonGroup } from 'react-bootstrap'`

taion:
Again, note that it's not ideal to import from top-level. You _should_ import from `react-bootstrap/lib` per https://react-bootstrap.github.io/getting-started.html#commonjs, or use a Babel plugin to rewrite your imports.

webpack 2 tree-shaking is not good enough to appropriately minimize your bundle sizes if you import from the top-level export.

sm1th:
@taion you're right. Thanks

PR #2706, opened 2017-07-13
Allow custom component in close button

d4rky-pl:
Example use case: close button with custom image.


taion:
This doesn't do what you think it does. That label is marked sr-only. The visible one is the `&times;`, and the correct typing would be `PropTypes.node` anyway.

d4rky-pl:
@taion I have no idea how could I've screwed up this pull request so bad, guess I was tired... Since there's no way of replacing &times; anyway, I'm closing this in favour of just adding the button inside `<Modal.Header/>`.

Apologies for extra noise.

PR #2708, opened 2017-07-13
Allow optional 'target' prop on OverlayTrigger

### LINKED ISSUE #1694: Support 'target' prop on OverlayTrigger for explicit positioning
`OverlayTrigger` ignores the `target` prop allowed by `Position` in `react-overlays`. I can't see any reason why you wouldn't want to allow that to be specific with an OverlayTrigger. It would default, naturally, to the current behavior and use the `OverlayTrigger` as target.

```
<OverlayTrigger
  overlay={filterMenu}
  placement='bottom'
  target={() => ReactDOM.findDOMNode(this.refs.trigger)}>
  <span>I want all of this to be part of the trigger, and it's a *lot* of content.
    But, I want the overlay to be positioned directly under 
    <span ref="trigger">this</span> word.
  </span>
</OverlayTrigger>
```

If I get a :+1:, I'll submit a MR, it should be really easy.

timothykang:
Resolves #1694

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -84,10 +84,6 @@ const propTypes = {
   /**
    * @private
    */
-  target: PropTypes.oneOf([null]),
-   /**
-   * @private
-   */
   onHide: PropTypes.oneOf([null]),
   /**
    * @private
@@ -229,7 +225,7 @@ class OverlayTrigger extends React.Component {
         {...props}
         show={this.state.show}
         onHide={this.handleHide}
-        target={this}
+        target={this.props.target || this}
       >
         {overlay}
       </Overlay>


timothykang:
Correct, `target` is ultimately defined on the `Position` component of `react-overlays` and is not marked as required:

```js
  target: PropTypes.oneOfType([
    componentOrElement, PropTypes.func
  ]),
```

([source](https://github.com/react-bootstrap/react-overlays/blob/v0.7.0/src/Position.js#L134))

taion:
@jquense Do we want to add this? It seems simple enough in the code.

bpas247:
What's currently stopping this PR from being merged? I'd like to help get this finalized

mxschmitt:
Closed due the current solution in v1 is completely different. But thanks for your work anyway!

PR #2711, opened 2017-07-15
fixes #2365 click on href with collapseOnSelect

### LINKED ISSUE #2365: CollapseOnSelect in Responsive Navbars prevents external links from working

sevgibson:
Note: My first contribution to an open source project. Please be kind :)

From the discussion on #2365 it sounds like there is no reason for ```preventDefault()``` unless the item is disabled. All other tests still pass after removing this. Taken from dmison suggestion [here](https://github.com/react-bootstrap/react-bootstrap/issues/2365#issuecomment-290325516)

Added 2 specs:
- NavBar with collapseOnSelect that failed prior to the change, but passes now.
- NavBar without collapseOnSelect that passes (this showed the expectation should work).

The specs may not be the correct way to test this, so feel free to correct me.


taion:
@jquense Do you recall offhand why we do `preventDefault` there? I guess because it would be weird to actually follow a link if the user specifies `onSelect` (and presumably expects the result of that callback to be relevant).

On the other hand, if we're going to do this, we should be consistent and drop the `preventDefault` bit from all the `onSelect` handlers – but I'm not sure we want to do this.

jquense:
I have no idea why we preventDefault onSelect ever. My instinct is that that behavior is _wrong_ upstream doesn't do that. We really should defer that logic to the routing libs right? LinkContainer should do that for the client side routes and otherwise it should just go. I guess the issue is that for real links you'll see the menu collapse as it loads away.

taion:
I think I'd be okay with that. It would be really odd to try to implement your own routing system via `onSelect` callbacks, or in any case to expect to be able to do so without calling `preventDefault()` yourself.

src/NavItem.js
```diff
@@ -28,10 +28,10 @@ class NavItem extends React.Component {
   }
 
   handleClick(e) {
-    if (this.props.onSelect) {
+    if (this.props.disabled) {
```
taion: This also changes the logic to unconditionally `preventDefault` on `disabled`, even with no `onSelect`?

sevgibson: You're right -- I'm not sure what should happen with ```disabled```? Or is ```preventDefault``` simply not needed at all? Feel free to correct the logic.

Tang8330:
@taion What is the state of this PR? How can we get this merged in so that this solves https://github.com/react-bootstrap/react-bootstrap/issues/2365?

taion:
Still trying to figure out what the right thing to do is.

We'd need to either drop `preventDefault` in `onSelect` in a blanket way, or use a different fix here.

taion:
I think dropping `preventDefault` everywhere is the right thing to do. Can you update this to apply to all cases of `onSelect`?

LorienOlive:
Hello.  Thanks to everyone who is working on this issue.  I'm having problems with internal links when collapseOnSelect is turned on.  I've tried incorporating the new handleClick code, but it doesn't seem to be fixing the problem.  I've basically copied and pasted the body of the new handleClick function into the NavItem.js file in my node_modules.  Is there another/better way to incorporate this fix into my project?

zhanwenchen:
This issue has been around for 9 months. Someone please do something.

PsiXdev:
>  I guess because it would be weird to actually follow a link if the user specifies onSelect (and presumably expects the result of that callback to be relevant).

First, it is strange that the user pass both the `href` and `onSelect`.
Secondly, if he really does this intentionally, then he most likely wants to process the referencing within the `onSelect` function. He can prevent it or not.
I think, that href ceases to influence anything - a more strange behavior than a simultaneous `onSelect` call and a link jump.
Why not just remove this preventDefault from NavItem.prototype.handleClick? Once again I repeat my opinion, this is a user problem, not a library.

jackpetraitis:
I am also experiencing pain points from this issue. Can this be merged?

jquense:
@jackpetraitis nope, if you want to help it along feel free to implement the requested chagnes on top of this and submit a PR 👍

sevgibson:
Note that I'm not sure the fix is the right one, and further, I'm no longer dependent on this being fixed, for the moment. I have opted to use react-router with an onClick handler that performs the navigation by pushing the url on the history of the react-router.

There is definitely a problem in href with collapseOnSelect, but I am not qualified to say what is the right solution.

I am perfectly happy if someone more knowledgable submits a better solution -- and closes this PR.

src/NavItem.js
jackpetraitis: I mean, what's wrong with preventing a user from doing a default action on a disabled element?

jackpetraitis:
Another member of my team implemented a workaround for our project. He wrote an onSelect handler for the NavItems that takes in an EventKey that was also placed on each NavItem. Then it routes the user to the proper location as specified by EventKey, but it seems an unnecessary overhead. I like what sevgibson had provided so far, but there is still issue with the fix provided not breaking the item's previously correct handling of 'disabled'. I think the two contributors, taion and jquense, are also debating whether or not to drop the preventDefault behavior across all onSelect behaviors. Not sure what the status is on that decision between them, but I will take a stab with a fork tonight and see if we can fix this.

taion:
I think #2790 is the active one we're working on now.

taion:
Thanks for getting this process started.

sevgibson:
Glad it's moving forward.. hope this was actually a contribution :)

PR #2719, opened 2017-07-19
Pass className from ToggleButtonGroup to ButtonGroup

ujovlado:
Since `ButtonGroup` accepts `className` I think it's a good idea to allow `className` also in `ToggleButtonGroup`.

This will allow additional class to append after `btn-group`.

If it's ok, I can add also changes to docs.

Thanks.

### DIFF
--- a/src/ToggleButtonGroup.js
+++ b/src/ToggleButtonGroup.js
@@ -69,7 +69,7 @@ class ToggleButtonGroup extends React.Component {
   }
 
   render() {
-    const { children, type, name } = this.props;
+    const { children, type, name, className } = this.props;
 
     const values = this.getValues();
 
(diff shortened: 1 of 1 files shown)


taion:
We should probably generically spread props through to `<ButtonGroup>` here, unless there's a specific reason to not do so. @jquense?

jquense:
Yeah I think so, I have no idea what is supported upstream there, but that's not our business i don't think. @ujovlado can you switch this to a `{...props}`?

ujovlado:
@jquense ok, I'll update this PR

jquense:
Sorry this is actually handled by #2731 which was a missing commit we had opriginally. It covers this case :) thanks!

ujovlado:
Oh, I see. Thank you.

PR #2723, opened 2017-07-21
Add backdropClassName to Modal props to allow extra class names on backdrop

dancrew32:
Similar to how <Modal/> allows `dialogClassName` to add class names on top of `modal`, it would be convenient to be able to add class names to the backdrop. Currently the only option is to modify the bsClass.


taion:
v0.31.2

PR #2738, opened 2017-07-27
use onMouseDown instead of onClick for Modal close

jgrisafe:
This prevents the modal from closing when the user clicks inside the modal, drags the cursor into the backdrop and releases. This is discussed in issue #1785

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -247,7 +247,7 @@ class Modal extends React.Component {
           {...dialogProps}
           style={{ ...this.state.style, ...style }}
           className={classNames(className, inClassName)}
-          onClick={backdrop === true ? this.handleDialogClick : null}
+          onMouseDown={backdrop === true ? this.handleDialogClick : null}
         >
           {children}
         </Dialog>


taion:
Per discussion on https://github.com/react-bootstrap/react-bootstrap/issues/1785, switching the event here is not the correct fix.

jgrisafe:
@taion would you mind explaining why this fix does not work? It does not seem like any other solution would be as simple. The linked thread doesn't seem to discuss that.

jquense:
The issue is fleshed out a bit more in the actual bootstrap repo, I'm nit sure anymore what the fox was but there were problems with switching to onMousedown

rikas:
This works for me. What are the issues?

taion:
https://github.com/twbs/bootstrap/issues/14263

TomiS:
Experiencing this problem too. My initial thought was to require both mouseDown and mouseUp to happen outside dialog to cause modal closing. That would solve the issue I think.

PR #2741, opened 2017-07-30
Added target property

BaReinhard:
Added target property to allow for img href to open in new window


taion:
Maybe we should switch `<Thumbnail>` to use the pattern from https://github.com/react-bootstrap/react-bootstrap/blob/502fe0044e5edef530f944953154a85e9daaebeb/src/BreadcrumbItem.js#L35.

BaReinhard:
Misunderstanding target is already allowed to be passed and is correctly placed on the safe anchor. I somehow made the mistake of thinking the target should have been placed on the img element.

PR #2742, opened 2017-08-01
Fix `import * as PropTypes` to `import PropTypes`

pastak:
Why it is `import * as PropTypes` ?

`import PropTypes` is fine I think.


taion:
Thanks!

PR #2775, opened 2017-08-30
[FEATURE] Added ability for multiple panels to expand

gfunk1230:
Configured accordion to be able to open multiple panels on click. Please see gif for example.

_Credits: @rovolution_ & Myself.

react-bootstrap-multipanel


jquense:
hey there! Thanks for the PR! I think we may not need to add an additional prop here for a few reasons.  First the panel API is changing fairly dramatically in the next version so this may not translate, second It's a bit confusing to call something an accordian but then have it not act like one (one open at a time). Finally tho you can actually already do this, without any changes!

```
<PanelGroup defaultActiveKey="2">
    <Panel collapsible header="Panel 1" eventKey="1">Panel 1 content</Panel>
    <Panel collapsible header="Panel 2" eventKey="2">Panel 2 content</Panel>
  </PanelGroup>
```

taion:
What @jquense said.

Although you don't need the `defaultActiveKey` prop I think?

gfunk1230:
Thanks, You guys were right but maybe we could update your documentation with an example of this use case. I didn't realize I had to use a PanelGroup.

jquense:
If you want to support a docs PR adding an example we'll be happy to merge it :)

taion:
BTW, I don't think you technically _need_ `<PanelGroup>`. In this context, `<PanelGroup>` is just a `<div className="panel-group">`. You just need to not set `accordion` on the `<PanelGroup>`.

jquense:
i want to say the panel-group adds some margins but i don't actually know that

PR #2784, opened 2017-09-08
Upgraded codecov because execSync subdependency fails to install

chillyistkult:
With "newer" node versions (>= 11) you are not able to run `npm install` on react-bootstrap because of execSync became a native node extension and fails (https://github.com/mgutz/execSync/issues/36) with

`Attempting to compile native extensions.`

Upgrading codecov to 2.X.X solves the problem because they dropped dependency to execSync.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-bootstrap",
-  "version": "0.31.2",
+  "version": "0.31.3",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


taion:
Which Node version, BTW? This works fine for me with the latest Node on Homebrew.

I do see the error message, but it doesn't fail the install – it's just logging.

chillyistkult:
I am working on a mac with a fresh node v8.1.3 and npm 5.4.1.

PR #2789, opened 2017-09-12
fix(Navbar): Allow external links to fire

### LINKED ISSUE #2365: CollapseOnSelect in Responsive Navbars prevents external links from working

joelseq:
- Remove preventDefault for NavItem onSelect which was preventing the external link from opening

Fixes #2365. This PR implements the solution suggested by @taion [here](https://github.com/react-bootstrap/react-bootstrap/pull/2711#issuecomment-325151744) in the discussion for #2711.

Tested it out locally and it works perfectly. Let me know if this is not the expected solution for this problem.

### DIFF
--- a/test/NavbarSpec.js
+++ b/test/NavbarSpec.js
@@ -353,4 +353,34 @@ describe('<Navbar>', () => {
 
     ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'test');
   });
+
+  it('Should open external href link in collapseOnSelect', () => {
+    const selectSpy = sinon.spy();
+    const navItemOnClick = sinon.stub();
+    const instance = ReactTestUtils.renderIntoDocument(
+      <Navbar onSelect={selectSpy}>
+        <Navbar.Header>
+          <Navbar.Toggle />
+        </Navbar.Header>
+        <Navbar.Collapse>
+          <Nav>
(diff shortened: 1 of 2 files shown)


test/NavbarSpec.js
```diff
@@ -353,4 +353,34 @@ describe('<Navbar>', () => {
...
+
+    ReactTestUtils.Simulate.click(ReactDOM.findDOMNode(link));
+
+    const event = navItemOnClick.getCall(0).args[0];
+    const preventDefaultSpy = sinon.spy(event.preventDefault);
+
+    expect(selectSpy).to.be.calledOnce;
+    expect(navItemOnClick).to.be.calledOnce;
+    expect(event.target.getAttribute('href')).to.be.equal('https://www.google.com');
+    expect(preventDefaultSpy).to.not.be.called;
```
joelseq: I didn't have a good way to test if the browser location actually changes because Karma throws an error if the browser page reloads. So instead I asserted that `preventDefault` doesn't get called and it failed prior to my change.

taion:
To clarify, in https://github.com/react-bootstrap/react-bootstrap/pull/2711#issuecomment-325151744, I mean drop `preventDefault` in all `onSelect` logic everywhere in this library, not just here.

joelseq:
@taion sorry I misunderstood. I'm going to close this since this issue is being handled in #2790

taion:
Thanks anyway!

PR #2790, opened 2017-09-12
preventDefault specified to work only for disabled && onSelect; it is…

jackpetraitis:
… a fix for issue #2711


taion:
As elsewhere, see https://github.com/react-bootstrap/react-bootstrap/pull/2711#issuecomment-325151744.

jackpetraitis:
Oh, @taion I've been meaning to ask you why. Why would we want to drop it across all onSelect?
Just wondering. But also, I found the related behavior in these items:

- NavItem
- Panel
- PanelGroup

I will update my PR with a fix for those as well. I saw some mention of this.props.onSelect in Carousel too, but I don't think it applies here.

jquense:
thanks! the reason why is that it ends up being more harm than help to prevent the link behavior. There are a few places where it still makes sense but I'm not sure which places.

ALSO please make the PR against the `next` branch not master as this is a breaking change

taion:
You need to rebase this branch on `next` too.

We want to change it everywhere because we want all components here to offer similar APIs. It would be confusing if `onSelect` on different components behaved differently.

jackpetraitis:
Thanks. I'm currently rebasing this branch onto next in my fork. Then I will re-open a PR with a request to merge my next into react-bootstrap's next.

taion:
You need to run `git rebase --interactive next` on your branch, and restrict the branch to only your own commits.

jackpetraitis:
So I had done an interactive rebase before on this. I think that's what messed it up.

I had done my changes off master. And then I rebased master onto next. Then I force pushed my next branch up to my forked repo. And now here we are. I'm wondering why all these other people's changes are still showing up as part of this PR. Maybe you could direct me as to how I can restrict the next branch to only my own commits. Perhaps you meant I should be creating a feature branch for just my own commits and then doing a PR to next with that feature branch?

Let me know if you can decipher what I may have done wrong from above.
I really only changed:

- NavItem.js
- NavBarSpec.js

Panel, PanelGroup, and their respective spec files already got refactored with the latest from next, so I think their preventDefault issue is no longer an issue. Although their changes are still reflected in this PR.

So let me fix the PR and sort this out properly. I will try this out in a few hours.

jackpetraitis:
Found the behavior in more items:

- DropdownButtonCustomMenu
- NavDropdown
- SubNav
- Dropdown
- DropdownMenu
- MenuItem
- Nav
- PagerItem
- SafeAnchor
- SafeAnchorSpec

I'm going to investigate each of these and drop the preventDefault if possible.
And testing each of the components rendered in a browser.

taion:
I rebased and force pushed to your branch. You'll want to run something like:

```
$ git fetch
$ git reset --hard origin/master
```

assuming you don't have any other work you want to preserve on your current `master` branch.

It turns out that due to https://github.com/react-bootstrap/react-bootstrap/pull/1769, we don't need this for `<Panel>` and related components any more.

The only one left is actually just `<PagerItem>`, as far as I can tell.

src/NavItem.js
```diff
@@ -28,12 +28,8 @@ class NavItem extends React.Component {
   }
 
   handleClick(e) {
-    if (this.props.onSelect) {
-      e.preventDefault();
```
taion: we probably still want to do `e.preventDefault()` if `disabled` is set?

test/NavbarSpec.js
```diff
@@ -256,6 +256,61 @@ describe('<Navbar>', () => {
...
+        </Navbar.Collapse>
+      </Navbar>
+    );
+
+    const link = ReactTestUtils.findRenderedDOMComponentWithClass(instance, 'link-text');
+
+    ReactTestUtils.Simulate.click(ReactDOM.findDOMNode(link));
+
+    expect(navItemSpy.getCall(0).args[0].isDefaultPrevented()).to.be.false;
+  })
```
taion: missing semi here is causing lint failure

src/NavItem.js
jquense: Yes

test/NavbarSpec.js
jackpetraitis: thanks

src/NavItem.js
jackpetraitis: ok

jackpetraitis:
My preventDefault's snooping gave me the same result you found @taion 

- DropdownButtonCustomMenu  <- don’t need as this was an example
- NavDropdown <- also an example
- SubNav <- in the docs
- Dropdown <- seems to be an important part of the handleKeyDown logic
- DropdownMenu <- same, has a role to play in handleKeyDown
- MenuItem <- this can stay here
- Nav <- part of handleTabKeyDown logic
**- PagerItem <- changed to preventDefault only if PagerItem is disabled && onSelect**
- SafeAnchor <- this can also stay the same
- SafeAnchorSpec <- don’t need to change

src/NavItem.js
```diff
@@ -28,12 +28,10 @@ class NavItem extends React.Component {
   }
 
   handleClick(e) {
-    if (this.props.onSelect) {
+    if (this.props.onSelect && !this.props.disabled) {
+      this.props.onSelect(this.props.eventKey, e);
+    } else if (this.props.disabled) {
```
jquense: you can rewrite this a bit simpler as 

```
if (this.props.disabled) {
  e.preventDefault();
  return;
}

if (this.props.onSelect)
   this.props.onSelect(this.props.eventKey, e);
```

src/PagerItem.js
```diff
@@ -30,7 +30,7 @@ class PagerItem extends React.Component {
   handleSelect(e) {
     const { disabled, onSelect, eventKey } = this.props;
 
-    if (onSelect || disabled) {
+    if (onSelect && disabled) {
```
jquense: again, check for disabled and return early

jackpetraitis: oh okay

src/NavItem.js
jackpetraitis: I am for sure a fan of simpler.

jackpetraitis:
I think I fixed the linting errors. I ended up removing a Navbar test because it was a duplicate of another test. 

I had issues writing tests for the NavItem + PagerItem code branches involving e.preventDefault. Essentially, once you add a disabled property on a tag you can no longer Simulate a click on it, and that would throw off all my sinon spy and stub logic. If you can figure out a way to do it I'd be all ears.

jackpetraitis:
I also don't know why appveyor build is failing. It seems that build fix is outside of my control. The log mentioned something about trying to run an npm install command as administrator.

taion:
AppVeyor is just flaky and janky.

@jquense What's up with the `tsp.fn.log` thing?

taion:
ping @jquense

jquense:
LGTM, the log this was just a convenience for logging why chaining assertions. It gets removed in the enzyme switch anyway

taion:
I resolved the merge commit. Good to go on my end now, just need signoff

PR #2791, opened 2017-09-12
Next

jackpetraitis:
Rebased master over next. Added changes to remove preventDefault from NavItem components. I had added it as well to Panel and PanelGroup, but those items have since been refactored.


taion:
This isn't needed; we'll merge `master` into `next` when we're ready for release.

PR #2816, opened 2017-10-02
Issue Fix #2554 - Docs should have individual page <title>s

### LINKED ISSUE #2554: Docs should have individual page <title>s (for browser history, etc.)
This is how it looks like right now:

reactbs_docs_no_title

This is poor UX, and improving it would make it up to par with vanilla Bootstrap's docs, as well as the docs of the other UI kits for React.

Working on this issue seems like a good "starter contribution", and it's something I'd like to do myself once I get some confirmation/feedback.

fumblehool:
Added page titles for individual Doc pages
Fixes #2554


fumblehool:
I'll try make it work for server-side rendered pages.

taion:
This still isn't going to work for SSR. You need to add the support to https://github.com/react-bootstrap/react-bootstrap/blob/61be58cfdda5e428d8fb11d55bf743661bb3f0b1/docs/build.js and https://github.com/react-bootstrap/react-bootstrap/blob/61be58cfdda5e428d8fb11d55bf743661bb3f0b1/docs/server.js to handle the "server-side" bits.

taion:
Err, maybe it'd be better to go back to what you had, but use `DocumentTitle.rewind()` per https://github.com/gaearon/react-document-title#server-usage?

jquense:
Maybe lets not spend a ton of time on this since #2818 is a thing that's happening. We still need this functionality but better figure it out in that PR perhaps (using react-helmet) and we'll get the SSR functionality for free

fumblehool:
@taion Should I revert back to old implementation or you should I leave it here?

taion:
Okay, sorry, it looks like #2818 will replace this.

PR #2819, opened 2017-10-04
Lint

jquense:
I’m ignoring docs for now since we will probably replace them


.eslintrc
```diff
@@ -9,32 +9,75 @@
...
+    "no-mixed-operators": ["error", {
+      "groups": [
+        ["&", "|", "^", "~", "<<", ">>", ">>>"],
+        ["==", "!=", "===", "!==", ">", ">=", "<", "<="],
+        ["&&", "||"],
+        ["in", "instanceof"]
+      ],
+      "allowSamePrecedence": false
+    }],
+    "class-methods-use-this": "off",
```
taion: i think the 4catalyzer config already handles many of these? it'd save some extra overrides on your end maybe

jquense: true, i didn't think about using that...

PR #2821, opened 2017-10-05
update babel-config

jquense:


.babelrc
```diff
@@ -1,14 +1,19 @@
 {
   "presets": [
-    ["es2015", { "loose": true }],
-    "stage-1",
+    ["env", {
+      "targets": {
+        "ie": 9,
+        "uglify": true
```
taion: what's this do?

.babelrc
```diff
@@ -1,14 +1,19 @@
...
+      }
+    }],
     "react"
   ],
   "plugins": [
+    "transform-class-properties",
+    "transform-object-rest-spread",
+    "transform-export-extensions",
```
taion: maybe just use stage 1?

package.json
```diff
@@ -45,16 +45,16 @@
     "babel-cli": "^6.11.4",
     "babel-core": "^6.13.2",
     "babel-eslint": "^8.0.1",
-    "babel-loader": "^6.2.5",
+    "babel-loader": "6.0.0",
```
taion: what?

jquense: YARN

.babelrc
jquense: stage 1 is less maintainable, plus adds everything 1 -4 negating most of the env preset

.babelrc
jquense: ensures it compiles code down to at least what uglify can parse, e.g es5

taion: isn't that what no-options `env` does? just target ES5, like with `latest`? i would just do plain `env` then, with `loose`.

jquense: i'm copying the CRA config here, they generally have good reasons for the settings

taion: fair enough. let's use `loose` though?

jquense: good call

.babelrc
```diff
@@ -1,14 +1,20 @@
 {
   "presets": [
-    ["es2015", { "loose": true }],
-    "stage-1",
+    ["env", {
+      "targets": {
+        "ie": 9,
+        "uglify": true,
+        "loose": true
```
taion: not a target

PR #2822, opened 2017-10-05
switch to enzyme

jquense:
will fix tests in #2821  as well


webpack/base.config.js
```diff
@@ -22,9 +22,12 @@ const baseConfig = {
   module: {
     loaders: [
       { test: /\.js/, loader: jsLoader, exclude: /node_modules/ },
+      { test: /\.json$/, loader: 'json-loader' },
```
taion: why is this part of this PR?

jquense: Cherrio via enzyme loads json files apparently

PR #2837, opened 2017-10-13
Fixes #2831 added missing devDependencies

### LINKED ISSUE #2831: build fails due to missing babel packages
If I do the following:

```
cd /Source
git clone https://github.com/react-bootstrap/react-bootstrap.git
cd react-bootstrap
yarn install
yarn build
```

I get the following error:

```
ReferenceError: Unknown plugin "transform-es3-member-expression-literals" specified in "base" at 2, attempted to resolve relative to "/Source/react-bootstrap/src"
    at /Source/react-bootstrap/node_modules/babel-core/lib/transformation/file/options/option-manager.js:180:17
    at Array.map (<anonymous>)
    at Function.normalisePlugins (/Source/react-bootstrap/node_modules/babel-core/lib/transformation/file/options/option-manager.js:158:20)
    at OptionManager.mergeOptions (/Source/react-bootstrap/node_modules/babel-core/lib/transformation/file/options/option-manager.js:234:36)
    at OptionManager.init (/Source/react-bootstrap/node_modules/babel-core/lib/transformation/file/options/option-manager.js:368:12)
    at File.initOptions (/Source/react-bootstrap/node_modules/babel-core/lib/transformation/file/index.js:212:65)
    at new File (/Source/react-bootstrap/node_modules/babel-core/lib/transformation/file/index.js:135:24)
    at Pipeline.transform (/Source/react-bootstrap/node_modules/babel-core/lib/transformation/pipeline.js:46:16)
    at buildContent (/Source/react-bootstrap/tools/buildBabel.js:8:18)
    at buildFile (/Source/react-bootstrap/tools/buildBabel.js:18:5)
```

The same error arises when installing with npm instead of yarn as well.

It seems that some of the build scripts include babel presets/plugins that are not included in package.json, like this file: https://github.com/react-bootstrap/react-bootstrap/blob/0e93f934e426685c6570ac63aacd8fae6cacafda/tools/es/build.js

I was able to fix the errors by adding the following to package.json under "devDependencies":

```
"babel-plugin-transform-es3-member-expression-literals": "^6.22.0",
"babel-plugin-transform-es3-property-literals": "^6.22.0",
"babel-preset-es2015": "^6.24.1",
"babel-preset-stage-1": "^6.24.1",
```

Was it just an oversight that these packages were not included?

fumblehool:
* Fixes #2831

### DIFF
--- a/tools/es/build.js
+++ b/tools/es/build.js
@@ -17,10 +17,10 @@ export default function BuildES() {
           modules: false,
           targets: {
             ie: 9,
-            uglify: true
-          }
+            uglify: true,
+          },
         }],
-        'react'
+        'react',
       ],
       plugins: [
         'transform-class-properties',
@@ -28,7 +28,7 @@ export default function BuildES() {
         'transform-export-extensions',
         'dev-expression',
         'transform-runtime',
-        'add-module-exports'
+        'add-module-exports',
       ],
     }))
     .then(() => console.log('Built: '.cyan + 'es module'.green));
--- a/tools/es/build.js
+++ b/tools/es/build.js
@@ -12,7 +12,7 @@ export default function BuildES() {
     .then(() => buildBabel(srcRoot, esRoot, {
       babelrc: false,
       presets: [
-        [env, {
+        ['env', {
           loose: true,
           modules: false,
           targets: {
--- a/tools/es/build.js
+++ b/tools/es/build.js
@@ -12,16 +12,28 @@ export default function BuildES() {
     .then(() => buildBabel(srcRoot, esRoot, {
       babelrc: false,
       presets: [
-        ['es2015', { loose: true, modules: false }],
-        'stage-1',
-        'react',
+        ['env', {
+          'loose': true,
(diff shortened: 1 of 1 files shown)


jquense:
Hey there, thanks for the PR. The issue here is that we recently changed the babel config to not need these plugins but missed a config in the change. Rather than add these we want to change the build script that uses them to match the `.babelrc` (which doesn't need these)

tools/es/build.js
```diff
@@ -12,16 +12,28 @@ export default function BuildES() {
     .then(() => buildBabel(srcRoot, esRoot, {
       babelrc: false,
       presets: [
-        ['es2015', { loose: true, modules: false }],
-        'stage-1',
-        'react',
+        ['env', {
+          'loose': true,
+          'targets': {
```
jquense: the diifference here is in this one we wnat `"modules": false"` so they aren't compiled

taion:
I fixed this; going to merge so I can publish the release.

PR #2838, opened 2017-10-13
DropdownButton supports block

hitchcockwill:
Related to issue #1680 

The DropdownButton component now supports the block prop. This will add a `.btn-block` class to the button element in the Dropdown.Trigger but will _not_ do anything to the `btn-group` itself. In order to make the `btn-group` display as a block level element, users will need to add additional styling.

I chose to make the `block` prop on DropdownButton a private prop to avoid any confusion with having to add additional styling.

### DIFF
--- a/src/DropdownButton.js
+++ b/src/DropdownButton.js
@@ -13,6 +13,13 @@ const propTypes = {
   title: PropTypes.node.isRequired,
   noCaret: PropTypes.bool,
 
+  // The block prop does not work 100% as you'd expect out of the box
+  // Requires extra styling to make it display correctly, but WILL add .btn-block to button element
+  /**
+   * @private
+   */
+  block: PropTypes.bool,
+
   // Override generated docs from <Dropdown>.
   /**
    * @private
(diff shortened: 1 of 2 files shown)


taion:
See https://github.com/react-bootstrap/react-bootstrap/issues/1680#issuecomment-336305886.

The issue is that, per stock TWBS styling, `.btn-block` doesn't do anything to `.btn-group` elements. The `display: block` property gets overridden by `display: inline-block` on `.btn-group`, so you only end up with `width: 100%`.

My inclination remains to drop the `block` prop on `<ButtonGroup>` entirely. It shouldn't be there at all. If users want a 100%-wide `<ButtonGroup>`, they can add such styling directly, instead of relying on a janky interaction with `.btn-group`.

taion:
cc @jquense thoughts?

hitchcockwill:
@taion I don't disagree with your line of thinking there. With or without the changes in this PR users will still have to add their own styling.

I think that this may be the best option if we actually want to take action on #1680, but a `wontfix` seems more reasonable TBH.

jquense:
yeah i think adding this for the sake of "sort of works accidentally" isn't ideal.

taion:
I mean not just a WONTFIX but also "drop the `block` prop from `<ButtonGroup>`".

@hitchcockwill Any interest in submitting a PR to deprecate that prop?

hitchcockwill:
@taion submitted a new PR for this: #2850

PR #2844, opened 2017-10-17
make setModalRef a bound function so that it doesn't get called with null

### LINKED ISSUE #2842: Resizing window with Modal showing throws error in Safari 11 with React 16
I'm seeing the following error when resizing the browser window with Safari 11:

```
TypeError: null is not an object (evaluating 'this._modal.getDialogElement')
at updateStyle(/node_modules/react-bootstrap/es/Modal.js:199:0)
at handleWindowResize(/node_modules/react-bootstrap/es/Modal.js:183:0)
```

This is with `react-bootstrap 0.31.3` `react-overlays 0.7.3` and `react 16.0.0`.

It works fine in Chrome

jharris4:
Fixes #2842

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -162,6 +162,10 @@ class Modal extends React.Component {
     this.handleExited();
   }
 
+  setModalRef(ref) {
+    this._modal = ref;
+  }
+
   handleDialogClick(e) {
     if (e.target !== e.currentTarget) {
       return;
@@ -208,10 +212,6 @@ class Modal extends React.Component {
           getScrollbarSize() : undefined,
       },
     });
-  }
-
-  setModalRef(ref) {
-    this._modal = ref;
   }
 
   render() {
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -157,6 +157,10 @@ class Modal extends React.Component {
     };
   }
 
+  componentWillMount() {
+    this._pendingUpdateStyle = false;
+  }
+
   componentDidUpdate() {
     if (this._pendingUpdateStyle) {
       this.updateStyle();
@@ -166,7 +170,6 @@ class Modal extends React.Component {
   componentWillUnmount() {
     // Clean up the listener if we need to.
     this.handleExited();
-    this._pendingUpdateStyle = false;
   }
 
   handleDialogClick(e) {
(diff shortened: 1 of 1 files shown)


jharris4:
@taion @jquense I need this fix to prevent modals from blowing up on me whenever the window is resized on Mac/iOS Safari.

I tried to build a minimal repro but it was too difficult without exposing proprietary code. Basically the app I'm using react-bootstrap with is rendering some SVG for which I need to manually measure text of using `svg:text.getComputedTextLength()` in `componentDidUpdate`. It seems like this is triggering Safari to "flush" the window resize events before the whole React tree has updated, which is what exposed this bug. Yay for the new Async nature of Fiber! lol

I hope I've sufficiently documented what's happening in #2842 and justified why this fix is necessary with React 16.

I'm basically just using a flag to defer updating the modal style when the reference to the portal is null because the ref hasn't been set yet.

That said, I'm all ears if you guys have suggestions on a better way to fix this issue.

src/Modal.js
```diff
@@ -156,9 +157,16 @@ class Modal extends React.Component {
...
+  componentDidUpdate() {
+    if (this._pendingUpdateStyle) {
+      this.updateStyle();
+    }
+  }
+
   componentWillUnmount() {
     // Clean up the listener if we need to.
     this.handleExited();
+    this._pendingUpdateStyle = false;
```
taion: eh? can `cDU` get called after `cWU`?

src/Modal.js
```diff
@@ -189,24 +197,29 @@ class Modal extends React.Component {
...
-
-    this.setState({
-      style: {
-        paddingRight: bodyIsOverflowing && !modalIsOverflowing ?
-          getScrollbarSize() : undefined,
-        paddingLeft: !bodyIsOverflowing && modalIsOverflowing ?
-          getScrollbarSize() : undefined,
-      },
-    });
+    if (!this._modal) {
```
taion: i'd write this as

```js
if (!this._modal) {
  this._pendingUpdateStyle = true;
  return;
}
```

jharris4:
@taion I addressed your comments. I moved clearing the pendingUpdateStyle flag to cWM from cWU (I was just avoiding adding the extra life cycle method...)

jharris4:
FYI: I'm not sure if the ref being null is expected behaviour with React 16, but I filed an issue just in case (or to get clarification): https://github.com/facebook/react/issues/11258

taion:
It could well be something race-y with the use of the new portal API plus mounting the portal in `cDM`.

jharris4:
Ahhh! TIL that ref callbacks can cause issues if they're declared inline...

See Dan Abromov's reply to the issue I filed: https://github.com/facebook/react/issues/11258#issuecomment-337379350

I'll take a look tomorrow at whether moving the ref callback function to a bound method of the class fixes the issue without needing the pending update style flag

jharris4:
@taion @jquense changing the ref callback to a bound function totally resolved all of the problems without needing the pending flag(/hack), so I've updated the PR code changes accordingly.

As per https://reactjs.org/docs/refs-and-the-dom.html#caveats:

```
Caveats

If the ref callback is defined as an inline function, it will get called twice during updates, first with null and then again with the DOM element. This is because a new instance of the function is created with each render, so React needs to clear the old ref and set up the new one. You can avoid this by defining the ref callback as a bound method on the class, but note that it shouldn’t matter in most cases.
```

jochenberger:
lgtm

jharris4:
Any chance of getting a new release with this fix? (Please and thank you ;-) )

taion:
Soon – need to get a couple of more RO fixes merged.

PR #2850, opened 2017-10-19
Deprecate 'block' prop on ButtonGroup

hitchcockwill:
* Added a deprecation warning when the block prop is used on the ButtonGroup component
* Made the block prop private on ButtonGroup to hide it from the docs
* Updated docs to include example for full width ButtonGroup using `style` prop
* Added test to make sure ButtonGroup will render with inline style

Related to #1680, #2838


docs/examples/ButtonGroupBlock.js
```diff
@@ -1,5 +1,5 @@
 const buttonGroupInstance = (
-  <ButtonGroup vertical block>
+  <ButtonGroup vertical style={{ display: 'block' }}>
```
taion: maybe `style={{ width: '100%' }}`? or does doing `display: block` actually work for making the group be full-width?

src/ButtonGroup.js
```diff
@@ -14,14 +14,19 @@ const propTypes = {
...
   block: all(
     PropTypes.bool,
-    ({ block, vertical }) => (
-      block && !vertical ?
-        new Error('`block` requires `vertical` to be set to have any effect') :
-        null
-    ),
+    ({ block, vertical }) => {
+      if (block) {
+        console.warn('`block` is deprecated on ButtonGroup and will be removed in v0.31.5. See https://react-bootstrap.github.io/components.html#btn-groups for more information.');
```
taion: use `deprecated` from `prop-types-extra`; we don't need to specify an explicit version, it'll just be the next minor release

docs/examples/ButtonGroupBlock.js
hitchcockwill: `.btn-group-vertical > .btn` already has a default style for being displayed block and 100% width.

`display: block` on the ButtonGroup accomplishes 100% width in a more clear and predictable way than relying on the inherited `display: inline-block` and adding `width: 100%` IMO.

src/ButtonGroup.js
hitchcockwill: Will do.

docs/examples/ButtonGroupBlock.js
taion: ah, cool

src/ButtonGroup.js
hitchcockwill: @taion I may be confused on the implementation of `deprecated` and `warning` but it seems like they throw a fatal error when that prop is present. For a deprecation warning we'd want something that just warns, but does not raise an exception right?

This is what I'm trying right now in the ButtonGroup propTypes:

```
block: all(
    PropTypes.bool,
    deprecated(({ block, vertical }) => {
      if (block && !vertical) {
        return new Error('`block` requires `vertical` to be set to have any effect');
      }
      return true;
    }, 'The `block` prop has been deprecated on ButtonGroup. See https://react-bootstrap.github.io/components.html#btn-groups for more information.'),
  ),
```

With that in there the "Should add block variation" test is failing with:

```
Error: Warning: Failed prop type: Warning: The prop `block` of `ButtonGroup` is deprecated. The `block` prop has been deprecated on ButtonGroup. See https://react-bootstrap.github.io/components.html#btn-groups for more information..
        in ButtonGroup
        at console.<anonymous> (webpack:///test/index.js:41:0 <- test/index.js:102:12)
        at Object.invoke (node_modules/sinon/pkg/sinon.js:427:32)
        at console.functionStub (node_modules/sinon/pkg/sinon.js:2893:47)
        at Function.invoke (node_modules/sinon/pkg/sinon.js:2441:47)
        at console.proxy [as error] (node_modules/sinon/pkg/sinon.js:2337:22)
        at printWarning (webpack:///~/fbjs/lib/warning.js:33:0 <- test/index.js:40770:16)
        at warning (webpack:///~/fbjs/lib/warning.js:57:0 <- test/index.js:40794:21)
        at checkReactTypeSpec (webpack:///~/react/lib/checkReactTypeSpec.js:78:0 <- test/index.js:42892:18)
        at validatePropTypes (webpack:///~/react/lib/ReactElementValidator.js:160:0 <- test/index.js:42336:6)
        at Object.createElement (webpack:///~/react/lib/ReactElementValidator.js:214:0 <- test/index.js:42390:6)
```

taion: PropType errors are never fatal. They're just console log spam. They log to `console.error` but does not otherwise cause problems.

taion:
Oh gosh, those build failures. I guess we need to re-sync `next` to `master`? The code looks fine, though.

hitchcockwill:
Ya some of the warnings in there right now are pretty rough. Anything else I can do to help right now?

taion:
Not sure. A bit busy the next few days but we'll get this sorted out.

PR #2851, opened 2017-10-19
Update Yarn lock

taion:
Will merge once CI goes green given the trivial change.


taion:
Oh, looks like we somehow had both a `yarn.lock` and a `package-lock.json`. I nuked the latter. Going to hold off on merging though.

Issue #428, opened 2015-03-03
Put a link to the github source for each component in the docs


jeremydmiller:
Just a suggestion. The quickest way to understand what options a component takes is to just look at the code for that component. A link in the docs to the jsx code in GitHub might be nice.

mtscout6:
Agreed! For consistency sake we should have the url be to the latest tagged release. Since we generate the static docs pages during the now automated release process after the version bump that should work fine. We will just need some mechanics to generate that link based on that version. Shouldn't be difficult though.

morlay:
Hi guys, I start a project to auto generate a doc for react,  <https://github.com/morlay/react-doc>.
not good enough now, and anyone could join me to continue it or any suggestion.

Generated doc json: https://github.com/morlay/react-doc/blob/master/example/docs/result.json
DocUI preview: http://morlay.github.io/react-doc/example/docs/

Thanks.

mtscout6:
@morlay I REALLY like where this is going! Let's discuss this in #431.

AlexKVal:
It will add too many maintanance burden.

The latest [FormControls](https://github.com/react-bootstrap/react-bootstrap/tree/0c61f4634b404171abdea856c17b0e88a9e63a3a/src/FormControls) components refactoring made it plain https://github.com/react-bootstrap/react-bootstrap/issues/825 that problems will arize.
Just imagine how much it has to be done then, with every component split or moving into new folder / namespace. 

(the probability of realizing this issue in reality is 'wontfix')
I'm closing it

mathieumg:
I hadn't seen this issue since I was really busy with school at the time. I believe this could be automated in a way that makes it maintainable.

AlexKVal:
Then you can prove it by starting from a small (but having more priority now) step
fix https://github.com/react-bootstrap/react-bootstrap/issues/825 :stuck_out_tongue_winking_eye: 
Then you'll realize that this `docs + links` has the same 'burden' and is not so simple to automate
as it seems.

jquense:
well we now have some infrastructure for automating some documentation. It wouldn't be two hard to have the `generate-metadata` do it without much trouble or increased maintenance burden.

I think if something makes the docs better, it should be seriously considered, sure it increases maintenance burden, but that, to me, is just part of running a good project.

AlexKVal:
Then:
Pull requests are welcome :wink:

Grahack:
[...]

PR #2818, opened 2017-10-03
[WIP] gatsby docs

jquense:
Should clean up a ton of old tooling and just be better :P

PR #2852, opened 2017-10-23
Add link to each component's PropTable to GitHub source

isaachinman:
Unless I'm missing something, completing #428 should be as simple as this. Not sure exactly where the link is best served, so I just put it in a footer of the PropTable component itself.

### DIFF
--- a/docs/src/PropTable.js
+++ b/docs/src/PropTable.js
@@ -7,6 +7,14 @@ import Label from '../../src/Label';
 import Table from '../../src/Table';
 import capitalize from '../../src/utils/capitalize';
 
+import packageJSON from '../../package.json';
+
+let version = packageJSON.version;
+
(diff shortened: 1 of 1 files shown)


taion:
Maybe put a `[source]` link next to the prop table headers? See https://github.com/react-bootstrap/react-bootstrap/pull/2818 though. We'll be merging that soon.

isaachinman:
Seems like #2818 maintains the PropTables, but not in the same way. I think props are being set [here](https://github.com/react-bootstrap/react-bootstrap/blob/gatsby-test/www/bsPropUtils.js)?

Anyways, if that's going to be merged soon, I'd probably prefer to close this and open a new PR once it's merged.

isaachinman:
Rewrote approach in #2980.

PR #2862, opened 2017-10-26
temp use fork of uncontrollable, update tests (add support to React v…

vadimka123:
…16 prod build)


taion:
Per https://github.com/jquense/uncontrollable/pull/26 this is not necessary.

PR #2864, opened 2017-10-27
remove twitter association

gijsbotje:
According to the latest released documentation http://getbootstrap.com/docs/4.0/about/brand/#name
Bootstrap should not be associated with Twitter


taion:
Thanks!

PR #2877, opened 2017-11-07
add note re: allowed # of children for Collapse + Fade

dangerismycat:
Wrapping either component around multiple children yields `Error: React.Children.only expected to receive a single React element child`.

Took me a while to debug this error, as it was swallowed and not available in the console. Would like to save others / FutureMe the trouble. 😊

### DIFF
--- a/docs/src/sections/TransitionSection.js
+++ b/docs/src/sections/TransitionSection.js
@@ -18,7 +18,7 @@ export default function TransitionSection() {
         <Anchor id="transitions-collapse">Collapse</Anchor>
       </h3>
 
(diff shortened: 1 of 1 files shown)


taion:
Hmm, I thought we had an issue somewhere to document the expected child types for various components. This is sort of like a special case of that. I wonder if it'd be better just to have the prop table say "a single React element" for `children`.

bpas247:
Is this PR still relevant to the current version?

taion:
looks like this is before the docs revamp; i think the current explanation is clear enough. feel free to re-open though

PR #2878, opened 2017-11-08
Remove fill from docs

jamesallison:
It doesn't look like the fill property is necessary anymore, so it can be removed from the docs.


taion:
It's consumed by `<Panel>`: https://github.com/react-bootstrap/react-bootstrap/blob/0e93f934e426685c6570ac63aacd8fae6cacafda/src/Panel.js#L105-L112

This janky API did get rewritten on `next`, though.

PR #2899, opened 2017-12-08
Updated Modal example to use ES6 class since latest react since 15 do…

tuxrace:
…es not support createClass method.


taion:
This is cool. Could you switch the function binding to the constructor, or else at instantiation time, though, rather than doing it in render? It's a non-preferred style in general.

mxschmitt:
Closing due outdated and now correctly implemented.

PR #2901, opened 2017-12-10
FormControl should accept `xs` as bsSize

Liooo:
right?

### DIFF
--- a/src/FormControl.js
+++ b/src/FormControl.js
@@ -92,5 +92,5 @@ FormControl.Feedback = FormControlFeedback;
 FormControl.Static = FormControlStatic;
 
 export default bsClass('form-control',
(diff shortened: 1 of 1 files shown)


taion:
Well, we pass it down to the input, and there is no `.input-xs`.

Liooo:
@taion Absolutely. 

Sorry, forgot `.input-xs` was defined manually and was used in my app. Sorry for bothering.

taion:
I forget... can we update the prop types to add new supported sizes?

Liooo:
@taion sorry I didn't understand, what's new supported sizes?

taion:
@Liooo Use `bsSizes` from `bootstrapUtils` to add `xs` as a supported `bsSize`.

Liooo:
But `.input-xs` isn't supported officially by bootstrap as you have pointed out

PR #2903, opened 2017-12-12
Add documentation note about rootCloseEvent being unavailable to custom dropdowns

### LINKED ISSUE #2891: rootCloseEvent doesn't work on custom dropdown
Exactly what the title says,

I found this issue: https://github.com/react-bootstrap/react-bootstrap/issues/2762
Pointing out that the logic is implemented in DropdownMenu.

It is at least confusing that the documentation shows rootCloseEvent as a property in the Dropdown Component. See:
image

Perhaps a mention of this would be appropriate?

kevinzwhuang:
Summary
-----------
The documentation about `rootCloseEvent` for Dropdown does not indicate that it is not usable for `CustomDropdown`.

Changes proposed
-------------------
- [x] Add note about `rootCloseEvent` not working for custom dropdowns due it being implemented internally by `<Dropdown.Menu>`.

-------------------
Fixes #2891

### DIFF
--- a/docs/src/sections/DropdownSection.js
+++ b/docs/src/sections/DropdownSection.js
@@ -57,6 +57,15 @@ export default function DropdownSection() {
       <p>
         For those that want to customize everything, you can forgo the included Toggle and Menu components, and create your own. In order to tell the Dropdown component what role your custom components play, add a special prop <code>bsRole</code> to your menu or toggle components. The Dropdown expects at least one component with <code>bsRole="toggle"</code> and exactly one with <code>bsRole="menu"</code>. Custom toggle and menu components must be able to accept refs.
       </p>
+      <div className="bs-callout bs-callout-warning">
+        <h4>Using <code>rootCloseEvent</code> with custom dropdown components</h4>
+        <p>
+          If you want to use the <code>rootCloseEvent</code> prop with your custom dropdown components, you will have to pass the <code>rootCloseEvent</code> to <code>{ '<RootCloseWrapper>' }</code> in your custom dropdown menu component <a href="https://github.com/react-bootstrap/react-bootstrap/blob/e41e041709045db657ac02c46fa4ff2df4e339a3/src/DropdownMenu.js#L115-L119">similarly to how it is implemented in <code>{ '<Dropdown.Menu>' }</code></a>.
+        </p>
+        <p>
+          You will have to add <code>react-overlays</code> as a dependency and import <code>{ '<RootCloseWrapper>' }</code> from <code>react-overlays</code> yourself like: <code>import RootCloseWrapper from 'react-overlays/lib/RootCloseWrapper'</code>.
+        </p>
+      </div>
       <ReactPlayground codeText={Samples.DropdownButtonCustomMenu} />
 
       <h3><Anchor id="btn-dropdowns-props">Props</Anchor></h3>
--- a/docs/src/sections/DropdownSection.js
+++ b/docs/src/sections/DropdownSection.js
@@ -60,7 +60,7 @@ export default function DropdownSection() {
       <div className="bs-callout bs-callout-warning">
         <h4>Using <code>rootCloseEvent</code> with custom dropdown components</h4>
         <p>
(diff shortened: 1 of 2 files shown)


taion:
Could you just say that users need to explicitly pass this down to the `<RootCloseWrapper>`?

kevinzwhuang:
@taion So telling users to import `<RootCloseWrapper>` from react-overlay themselves in their custom dropdowns and passing down the event there?

taion:
I think so! Is it in the formal docs section on custom menus?

kevinzwhuang:
Doesn't look like it - there hasn't been a public mention of `<RootCloseWrapper>` in the docs thus far (did a global search in the docs and the codebase, only appears in the `src/DropdownMenu.js` source, not the docs)

We could also add a mention there in the `## Custom Dropdown Components` section recommending the usage of `<RootCloseWrapper>` for this

kevinzwhuang:
@taion This seems like it might be a little difficult for most users to reimplement directly using `<RootCloseWrapper>`...but here's a try at some notes about it here:

----------------

Note: To simulate the `rootCloseEvent` prop for custom dropdown menu components, you will need to import `<RootCloseWrapper>` from [react-overlays](https://react-bootstrap.github.io/react-overlays/) and render it as the top node of your custom menu component. Then pass the `rootCloseEvent` to `<RootCloseWrapper>` as the event prop it expects ([similar to how `<Dropdown.Menu>` is implemented](https://github.com/react-bootstrap/react-bootstrap/blob/e41e041709045db657ac02c46fa4ff2df4e339a3/src/DropdownMenu.js#L115-L119)).

How it looks in that section: (I also started writing out some example code, but I feel like that might make it more complicated, maybe they should just look at DropdownMenu and reimplement it)
image

taion:
It might be good to show this in the actual example, since we already have that. Though maybe the import from react-overlays won't work.

Maybe just mention it briefly for now, and we can clean it up once we merge the Gatsby docs.

kevinzwhuang:
For sure @taion - I've added callout to this in the custom dropdown component section and updated the note in the props section to hint at this.

Custom Dropdown Components section:
image

Dropdown Props:
image

docs/src/sections/DropdownSection.js
```diff
@@ -57,6 +57,15 @@ export default function DropdownSection() {
       <p>
         For those that want to customize everything, you can forgo the included Toggle and Menu components, and create your own. In order to tell the Dropdown component what role your custom components play, add a special prop <code>bsRole</code> to your menu or toggle components. The Dropdown expects at least one component with <code>bsRole="toggle"</code> and exactly one with <code>bsRole="menu"</code>. Custom toggle and menu components must be able to accept refs.
       </p>
+      <div className="bs-callout bs-callout-warning">
+        <h4>Using <code>rootCloseEvent</code> with custom dropdown components</h4>
+        <p>
+          If you want to use the <code>rootCloseEvent</code> prop with your custom dropdown components, you will have to pass the <code>rootCloseEvent</code> to <code>{ '<RootCloseWrapper>' }</code> in your custom dropdown menu component <a href="https://github.com/react-bootstrap/react-bootstrap/blob/e41e041709045db657ac02c46fa4ff2df4e339a3/src/DropdownMenu.js#L115-L119">similarly to how it is implemented in <code>{ '<Dropdown.Menu>' }</code></a>.
```
taion: Use a tag instead of the commit here, like https://github.com/react-bootstrap/react-bootstrap/blob/v0.31.5/src/DropdownMenu.js#L115-L119

src/Dropdown.js
```diff
@@ -95,6 +95,10 @@ const propTypes = {
 
   /**
    * Which event when fired outside the component will cause it to be closed
+   *
+   * *Note: For custom dropdown components, you will have to pass the
+   * `rootCloseEvent` to `<RootCloseWrapper>` in your custom dropdown menu
+   * component ([similarly to how it is implemented in `<Dropdown.Menu>`](https://github.com/react-bootstrap/react-bootstrap/blob/e41e041709045db657ac02c46fa4ff2df4e339a3/src/DropdownMenu.js#L115-L119)).*
```
taion: and here

kevinzwhuang:
Good call @taion, merging after it passes TravisCI

PR #2908, opened 2017-12-20
Use different properties for mouseover and mouseout in handleMouseOverOut

jcestibariz:
The current fallback for `relatedTarget` in `handleMouseOverOut` is always `toElement`, but this is not correct for the `mouseover` event,which should use `fromElement` instead (see [the documentation for relatedTarget](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/relatedTarget)).

This bug causes the overlay to not be triggered when tested under chromedriver since `relatedTarget` can be null in some cases.


taion:
Sounds good to me!

PR #2921, opened 2018-01-02
Prettier

jquense:
I intentionally didn’t format anything yet, since it’ll make big merges with the docs and next branch much harder

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build-docs": "npm run bootstrap-docs && npm -C www run build",
     "ci-lint": "eslint --rule 'prettier/prettier: 2'",
     "format": "npm run ci-lint -- --fix",
-    "lint": "npm run lint .",
+    "lint": "npm run ci-lint .",
     "precommit": "lint-staged",
     "release": "release",
     "tdd": "karma start",
--- a/package.json
+++ b/package.json
@@ -14,8 +14,9 @@
     "bootstrap-docs": "npm run build lib && cd www && yarn install && cd ..",
     "build": "node tools/build.js",
     "build-docs": "npm run bootstrap-docs && npm -C www run build",
-    "format": "npm run lint -- --fix",
-    "lint": "eslint .",
+    "ci-lint": "eslint --rule 'prettier/prettier: 2'",
+    "format": "npm run ci-lint -- --fix",
+    "lint": "npm run lint .",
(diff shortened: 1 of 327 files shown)


package.json
```diff
@@ -7,22 +7,34 @@
...
-  "jsnext:main": "es/index.js",
-  "module": "es/index.js",
+  "lint-staged": {
+    "*.js": [
+      "eslint --fix",
+      "git add"
+    ]
+  },
+  "prettier": {
+    "printWidth": 80,
```
taion: this is the default, no?

jquense: huh, TIL

taion: yeah i just set it to 79 in many places because, uhh, never mind https://prettier.io/docs/en/options.html#print-width

package.json
```diff
@@ -7,20 +7,32 @@
...
+  "jsnext:main": "es/index.js",
+  "module": "es/index.js",
   "scripts": {
     "bootstrap-docs": "npm run build lib && cd www && yarn install && cd ..",
     "build": "node tools/build.js",
     "build-docs": "npm run bootstrap-docs && npm -C www run build",
-    "lint": "eslint www/src src test tools *.js",
+    "ci-lint": "eslint --rule 'prettier/prettier: 2'",
+    "format": "npm run ci-lint -- --fix",
+    "lint": "npm run eslint .",
```
jquense: I think this is a better pattern. I really don't like having the in-editor prettier lint errors, it's very nosy and something that is fixed for me, so its not helpful. This way it's only run in a CI or precommit setting.

package.json
```diff
@@ -7,20 +7,32 @@
...
   "homepage": "https://react-bootstrap.github.io/",
+  "main": "lib/index.js",
+  "jsnext:main": "es/index.js",
+  "module": "es/index.js",
   "scripts": {
     "bootstrap-docs": "npm run build lib && cd www && yarn install && cd ..",
     "build": "node tools/build.js",
     "build-docs": "npm run bootstrap-docs && npm -C www run build",
-    "lint": "eslint www/src src test tools *.js",
+    "ci-lint": "eslint --rule 'prettier/prettier: 2'",
```
taion: i guess this is ok

jquense: yeah idk, maybe we find something a bit less awkward but it seems to work well enough for now anyway

PR #2923, opened 2018-01-02
Release prep

jquense:


package.json
```diff
@@ -12,7 +12,7 @@
     "build": "node tools/build.js",
     "build-docs": "npm run bootstrap-docs && npm -C www run build",
     "lint": "eslint www/src src test tools *.js",
-    "release": "release",
+    "release": "npm run build-docs && release",
```
taion: isn't this already part of `build`? i thought we already did all this

jquense: I think I removed it along the way bc it was weird to build do a with the library? Idk if this is nicer

CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+## [v0.32.0]
+> 2018-01-02
```
taion: i guess update the date

Issue #2926, opened 2018-01-04
Accordions allow more than one panel open as of 0.32.0


leviwheatcroft:
Previously, the behaviour with a PanelGroup accordion was that when you expanded one panel, the previous panel would collapse. This no longer occurs. You can see this behaviour [in the docs](https://react-bootstrap.github.io/components/panel/#panels-accordion).

I note that previously the example in the docs used `<Accordion>` where now the docs are using `<PanelGroup accordion>`, but I'm not sure if that's relevant.

mreishus:
Running into the same problem.  If I copy and paste the `PanelGroupControlled.js` example ( https://github.com/react-bootstrap/react-bootstrap/blob/5ff1962588facf4a51d59ac9aea558330133cc3c/docs/examples/PanelGroupControlled.js ) I can open multiple panels.

jquense:
anyone want to take a stab at a fix?

taion:
This was fixed in #2940

PR #2940, opened 2018-01-13
Fix PanelGroup Accordion allow more than one panel open #2926

rluiten:
I ran into this issue myself last week https://github.com/react-bootstrap/react-bootstrap/issues/2926.
Created a test case and a fix.

### DIFF
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -71,7 +71,7 @@ class Panel extends React.Component {
     let ids;
 
     if (idKey !== null) {
-      const panelGroup = this.context.$bs_panelGroup;
+      const panelGroup = this.context.$bs_panelGroup || {};
       const getId = panelGroup.getId || defaultGetId;
 
       ids = {
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -95,7 +95,7 @@ class Panel extends React.Component {
     const { eventKey } = this.props;
     const { activeKey } = this.context.$bs_panelGroup || {};
     return activeKey === undefined
-      ? !!this.props.expanded 
+      ? !!this.props.expanded
       : activeKey === eventKey;
   }
 
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 import uncontrollable from 'uncontrollable';
+import warning from 'warning';
 
 import {
   bsStyles,
@@ -16,6 +17,8 @@ import Title from './PanelTitle';
 import Footer from './PanelFooter';
 import Toggle from './PanelToggle';
 import Collapse from './PanelCollapse';
+
+const has = Object.prototype.hasOwnProperty;
 
 const defaultGetId = (id, type) => (id ? `${id}--${type}` : null);
 
@@ -61,20 +64,16 @@ const childContextTypes = {
 };
 
 class Panel extends React.Component {
-  constructor(...args) {
-    super(...args);
-    this.handleToggle = this.handleToggle.bind(this);
-  }
-
   getChildContext() {
     const { eventKey, id } = this.props;
-    let { getId } = this.context.$bs_panelGroup || {};
+    const idKey = eventKey == null ? id : eventKey;
 
     let ids;
-    let idKey = eventKey == null ? id : eventKey;
 
     if (idKey !== null) {
-      getId = getId || defaultGetId;
+      const panelGroup = this.context.$bs_panelGroup;
+      const getId = panelGroup.getId || defaultGetId;
+
       ids = {
         headingId: getId(idKey, 'heading'),
         bodyId: getId(idKey, 'body')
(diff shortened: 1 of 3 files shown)


rluiten:
No sure what I can do about the test "sets focus on next menu item when the key "down" is pressed" failing, should I have not created the PR against master maybe ?

src/Panel.js
```diff
@@ -94,9 +94,8 @@ class Panel extends React.Component {
   getExpanded() {
     const { eventKey } = this.props;
     const { activeKey } = this.context.$bs_panelGroup || {};
-
-    return this.props.expanded != null || activeKey === undefined
```
taion: Shouldn't `this.props.expanded` take precedence over `activeKey` though? I think the bug is more in `<PanelGroup>`.

rluiten: If `<PanelGroup>` is not an `accordion` then `activeKey` is not on the `$bs_panelGroup` context  and none of the children controls like `<Panel>` even have `activeKey` available.

It took me a good hour of instrumenting code with log statements to figure out the first point above, the code looks simple but I took me a while to get the pattern in my head then think about it.

This then lead me to `<Panel>` really does not care about its expanded status at all if `activeKey` is present. This works out that if `activeKey` changes via `<PanelGroup>.handleSelect()` then all children need to decide if they are then open or closed purely based on `activeKey`.

Please feel free to point out holes in my logic.

taion: I see, so this is the Uncontrollable thing. @jquense is there a way to distinguish between an `expanded` coming down from Uncontrollable v. an explicit one?

jquense: nope :/

rluiten: @taion is there specific condition you are thinking of that you have in mind that requires knowing about source of expanded being Uncontrollabe or explicit for some behaviour ?

taion: Could it come up that someone will do something like:

```js
<PanelGroup accordion>
  <Panel expanded>{...}</Panel>
  <Panel>{...}</Panel>
  <Panel>{...}</Panel>
</PanelGroup>
```

Or what should we do in that case? Warn, I guess?

rluiten: Do you have a development vs production model warning similar to React ? 
If so that would seem easy enough to warn about ?

taion: Right and we do something like that for tabs, so it would make sense to do the same thing here.

rluiten: The basic check would be in constructor of Panel throw Error if expanded is set true or false (not undefined) if there is a `$bs_panelGroup.activeTab` present on context.

This type of parameter checking Error I see raised in a dozen or so places in the code base all seem to not check for production or development so they will still be active in production builds unlike many of React's notifications.

eg. https://github.com/react-bootstrap/react-bootstrap/blob/master/src/TabContainer.js#L21-L27

rluiten: While checking behaviours, i just found that a `<Panel>` with `expanded` but no `onToggle` will get an error generated by uncontrollable as follows.

```
ERROR: 'Warning: Failed prop type: You have provided a `expanded` prop to `Panel` 
without an `onToggle` handler. This will render a read-only field. If the 
field should be mutable use `defaultExpanded`. Otherwise, set `onToggle`
in Uncontrolled(Panel)'
```
This error is great in general but in our case is not help this case  as providing `expanded` (without `onToggle`) to a `Panel` inside a `PanelGroup` of type `accordion` would generate this error, but the developer is just as likely to add `onToggle` to address the error as remove `expanded`, now given the current bug fix this won't stop the Panel from behaving as a member of and accordion panel group I believe, but its dead code that currently does nothing and may cause harm in the future if implementation of controls changes. If the developer is then relying on something in the panels `onToggle` he may get hard to understand behaviour.

After an initial attempt at detecting if a Panel inside an Accordion is being passed `expanded` or `onToggle` to generate an Error. I believe I can't do this without knowing if uncontrolled is managing the `expanded` and `onToggle` properties or not as @taion raised earlier.

Because `uncontrollable` provides these properties if the user does not then I can't just see if the props `expanded` or `onToggle` are provided to a `Panel` with a accordion panel group parent.

taion: Yeah. That's fine. Just emit a `warning` if `expanded` is not nully but `activeKey` is present?

rluiten: I am not sure if this is feasible.

Could I map over the children provided to PanelGroup with accordion enabled and emit an Error from that side if there is expanded onToggle or defaultExpanded explicitly set ?

taion: It's just a warning. No harm emitting a second warning.

rluiten: So far in my initial attempts the second one does not get emitted.
I will look at his again.

taion:
I pushed an update.

src/Panel.js
```diff
@@ -92,23 +91,32 @@ class Panel extends React.Component {
...
 
   getExpanded() {
-    const { eventKey } = this.props;
-    const { activeKey } = this.context.$bs_panelGroup || {};
+    const panelGroup = this.context.$bs_panelGroup;
 
-    return this.props.expanded != null || activeKey === undefined
-      ? this.props.expanded
-      : activeKey === eventKey;
+    if (panelGroup && has.call(panelGroup, 'activeKey')) {
```
taion: This is annoying but `activeKey` starts out as `undefined` if the panel group has no default value... not sure what we can do there

src/PanelGroup.js
```diff
@@ -96,11 +90,13 @@ class PanelGroup extends React.Component {
     };
   }
 
-  handleSelect(key, expanded, e) {
+  handleSelect = (key, expanded, e) => {
     if (expanded) {
       this.props.onSelect(key, e);
+    } else if (this.props.activeKey === key) {
```
taion: You should be able to click on a panel in an accordion again to close it.

I'm not sure setting `activeKey` to `null` is the best way to do this, though. It'd be weird to use `undefined`.

src/Panel.js
rluiten: I had got to the same point of checking hasOwnProperty() myself.

src/Panel.js
```diff
@@ -61,20 +64,16 @@ const childContextTypes = {
...
     const { eventKey, id } = this.props;
-    let { getId } = this.context.$bs_panelGroup || {};
+    const idKey = eventKey == null ? id : eventKey;
 
     let ids;
-    let idKey = eventKey == null ? id : eventKey;
 
     if (idKey !== null) {
-      getId = getId || defaultGetId;
+      const panelGroup = this.context.$bs_panelGroup;
```
rluiten: If not in PanelGroup this causes failure on next line at `panelGroup.getId`.
So maybe use.
`const panelGroup = this.context.$bs_panelGroup | {};`

rluiten:
Your changes look good to me, cover more bases than mine.
I added a comment on one line which is causing the test failures for a trivial reason.

rluiten:
Sorry finally figured out that your updated applied to my original branch, I havn't encountered this before so took me a bit of reading and snooping to figure out it was my own branch updated. Cheers.
Fixed the problem causing test failures.

taion:
Oops, forgot to push my fix. Could you change that last bit like:

```diff
diff --git a/src/Panel.js b/src/Panel.js
index a6f3c273..7a73f506 100644
--- a/src/Panel.js
+++ b/src/Panel.js
@@ -72,7 +72,7 @@ class Panel extends React.Component {

     if (idKey !== null) {
       const panelGroup = this.context.$bs_panelGroup;
-      const getId = panelGroup.getId || defaultGetId;
+      const getId = (panelGroup && panelGroup.getId) || defaultGetId;

       ids = {
         headingId: getId(idKey, 'heading'),
```

Instead of doing `this.context.$bs_panelGroup || {}`? I'd like to avoid the extra object here.

Issue #2941, opened 2018-01-14
Bug in exemplary code on web in section forms.


JanKaifer:
There is custom function which doesn't work properly:
```
function FieldGroup({ id, label, help, ...props }) {
	return (
		<FormGroup controlId={id}>
			<ControlLabel>{label}</ControlLabel>
			<FormControl {...props} />
			{help && <HelpBlock>{help}</HelpBlock>}
		</FormGroup>
	);
}
```
It throws this error:
```
TypeError: Cannot read property 'line' of undefined
```
However, when I remove this one by using props as an argument:
```
function FieldGroup(props) {
	return (
		<FormGroup controlId={props.id}>
			<ControlLabel>{props.label}</ControlLabel>
			<FormControl {...props} />
			{props.help && <HelpBlock>{props.help}</HelpBlock>}
		</FormGroup>
	);
}
```
It works.

So if someone could tell why react-live doesn't support this function({arg1, arg2, arg3, ...args}) {...} feature.

Link to playground: https://react-bootstrap.github.io/components/forms/?radioGroup=on#forms-controls

taion:
dupe https://github.com/react-bootstrap/react-bootstrap/issues/2933

PR #2943, opened 2018-01-14
website - Quick fix for code in example in forms section

JanKaifer:
This is just quick and ugly fix fro issue #2941.


taion:
That doesn't do what you think it does. You can just manually spread out the fields there if you want.

taion:
We fixed this properly.

Issue #2906, opened 2017-12-17
Don't always inject active-related props to Nav children


taion:
What if we gate this on `bsRole` like for `<Dropdown>`?

taion:
resolved in v1

PR #2955, opened 2018-01-21
fix(Nav): set activeKey and activeHref to lowerCase

trevorwhealy:
The 	`activeKey` and `activeHref` property on the `Nav` component are throwing a React warning in my console. I'm not sure what those attributes are used for, but if I do as the warning asks and make the properties lowercase instead of camel-case, the errors go away.

Cool library by the way. Thanks to the contributors.

### DIFF
--- a/src/Nav.js
+++ b/src/Nav.js
@@ -342,8 +342,8 @@ class Nav extends React.Component {
               childOnSelect
             ),
             active,
-            activekey,
-            activehref,
+            activekey: activeKey,
+            activehref: activeHref,
             onSelect: childOnSelect
           });
         })}
--- a/src/Nav.js
+++ b/src/Nav.js
@@ -342,8 +342,8 @@ class Nav extends React.Component {
               childOnSelect
             ),
             active,
-            activeKey,
-            activeHref,
+            activekey,
+            activehref,
             onSelect: childOnSelect
           });
         })}


trevorwhealy:
@taion + @jquense can I get a review here, please?

taion:
These are semantic props for NavItem. Look at the second part of the error. 

@jquense wdyt of switching Nav to use context? Might be the ideal fix for avoiding issues like this going forward.

jquense:
i think context is gonna cause more issues then it solves especially with the upcoming changes, since it makes mapping props harder for custom navitems. maybe it's worth inspecting the type tho and bailing for host components (type === string).

I'm gonna close this since it's not the right fix @trevorwhealy in general tho we probably aren't going to accept PR's when CI is failing, especially when tests are failing if you're interested in continuing to contribute :)

trevorwhealy:
@jquense I was mostly hoping to put this on the radar, because I was a bit confused by the React code in this project, and I'm not sure where the actual fix would go. So is there not going to be a fix for these warnings any time in the near future ?

jquense:
the issue you are running into is that you are putting a plain `li` element inside of a Nav instead of a NavItem, it's not technically a bug in RB but more of a gotcha with how prop passing in react works

taion:
Ref https://github.com/react-bootstrap/react-bootstrap/issues/2906

Issue #2933, opened 2018-01-06
Forms supported controls example 


VonTran:
![error](https://user-images.githubusercontent.com/15680299/34638941-1701dd48-f2a4-11e7-8f4c-cb28787de801.png)

mengwong:
farther up the page, https://react-bootstrap.github.io/components/forms/?#forms gives this error in the console:

```
Uncaught TypeError: this.setState is not a function
    at handleChange (eval at s (evalCode.js:18), <anonymous>:30:10)
    at Object.executeOnChange (LinkedValueUtils.js:130)
```

taion:
Seems like this is happening because React Live uses Buble v0.15, which doesn't support object rest properties.

@jquense Can we get them to upgrade to a more recent Buble version? It looks like Buble added support here in v0.17: https://github.com/Rich-Harris/buble/blob/master/CHANGELOG.md#0170.

taion:
PRed https://github.com/FormidableLabs/react-live/pull/51. Let's see how that goes.

taion:
We fixed the error up the page in https://github.com/react-bootstrap/react-bootstrap/pull/2932.

justbill2020:
this is still an issue... and #2932 still hasn't fixed the isssue on the page

taion:
I pinged the React Live maintainers again.

PR #2957, opened 2018-01-22
Fix and update examples

taion:
Fixes https://github.com/react-bootstrap/react-bootstrap/issues/2933

### DIFF
--- a/www/src/examples/.eslintrc
+++ b/www/src/examples/.eslintrc
@@ -2,7 +2,7 @@
   "rules": {
     "no-alert": "off",
     "no-unused-expressions": "off",
-    "semi": "off",
+    "no-unused-vars": "off",
     "react/jsx-no-undef": "off",
     "react/no-multi-comp": "off"
   },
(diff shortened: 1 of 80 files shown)


www/src/examples/.eslintrc
```diff
@@ -2,7 +2,7 @@
   "rules": {
     "no-alert": "off",
     "no-unused-expressions": "off",
-    "semi": "off",
```
taion: This was pointless because Prettier added back the semis anyway.

www/src/examples/Badge.js
```diff
@@ -1,7 +1,3 @@
-const badgeInstance = (
-  <p>
-    Badges <Badge>42</Badge>
-  </p>
-);
-
-render(badgeInstance);
```
taion: I figure we might as well use the shorthand syntax everywhere. We're already using it in many places, anyway.

jquense: 👍 ya definitely i wanted to do that but didn't have time

PR #2960, opened 2018-01-23
Change docs footer version reference

tomscott1:
I noticed that the documentation version was 1.0.0 which was looking at the `www/package.json` version number so I directed it to look at `/package.json`.

image

Alternatively we could change `www/package.json` version number.


taion:
I'm not sure this quite works with our build process. Don't we build the docs before we bump the npm version? Which would mean that the docs would show the old version rather than the new verison.

taion:
We might need to move the docs building back into the main build.

PR #2965, opened 2018-01-26
Changed ProgressBar check to work with proxied components

### LINKED ISSUE #2963: React Hot Loader v4 causes ProgressBar prop type warning
I am trying to upgrade my project from React Hot Loader v3 to v4, and it's causing me to have a prop type warning with the ProgressBar component. It looks like the ProgressBar checks that all children are also ProgressBars, but I guess RHLv4 basically wraps every component, so this check fails even though the child is actually a (proxied) ProgressBar.

Is there a way make this check more lenient and work with proxied components? I haven't found a way to disable a specific warning in React and I don't want to turn off all warnings.

Kimahriman:
Fixes #2963

This is the recommended way of doing Component type checking in react hot loader v3. The updated version in v4 using `areComponentsEqual` doesn't seem to work correctly, but this method still seems to work with v4. 

I don't really know how to add any new tests to show this work with RHL, but all the current tests still work.

### DIFF
--- a/src/ProgressBar.js
+++ b/src/ProgressBar.js
@@ -30,7 +30,12 @@ function onlyProgressBar(props, propName, componentName) {
       return;
     }
 
-    // eslint-disable-next-line no-use-before-define
+    /**
+     * Compare types in a way that works with libraries that patch and proxy
+     * components like react-hot-loader.
+     *
+     * see https://github.com/gaearon/react-hot-loader#checking-element-types
+     */
     const element = <ProgressBar />;
     if (child.type === element.type) return;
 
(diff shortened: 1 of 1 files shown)


src/ProgressBar.js
```diff
@@ -31,7 +31,8 @@ function onlyProgressBar(props, propName, componentName) {
     }
 
     // eslint-disable-next-line no-use-before-define
-    if (child.type === ProgressBar) return;
+    const element = <ProgressBar />;
+    if (child.type === element.type) return;
```
taion: Add a comment explaining why you do this

PR #2967, opened 2018-01-27
Fix ordering of arguments in Dropdown onSelect

rtelicak:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Dropdown.js#L88


taion:
It's failing in the Prettier check. Looks good otherwise.

PR #2968, opened 2018-01-28
include props for modal dialog

### LINKED ISSUE #2961: Missing prop in modal component docs
Search the modal docs https://react-bootstrap.github.io/components/modal/
for the prop dialogClassName

There is an example, but no description of the prop in the table below
dialogClassName

tomscott1:
fixes #2961


taion:
Merging as docs-only fix

Issue #2978, opened 2018-02-04
Fix Bootstrap CSS version in CDN link in documentation


rimatla:
Navigation broke the UI out of the blue without me touching any of the files with the navigation code using react-bootstrap "^0.32.1"

In one page refresh the site went from:
![screen shot 2018-02-04 at 2 39 42 pm](https://user-images.githubusercontent.com/14306250/35782145-5c726074-09b9-11e8-9139-ccac6f8a1040.png)

to:
![screen shot 2018-02-04 at 2 41 46 pm](https://user-images.githubusercontent.com/14306250/35782159-b0fb0560-09b9-11e8-8bf4-b3ac83a6d010.png)

There are no errors showing in the console.

**EDIT:**
Issue occurring on Chrome, Firefox and Safari, However still works on Opera (for now).

VonTran:
same issue a long with misalignment of columns 
![firefox](https://user-images.githubusercontent.com/15680299/35782361-b1b922ec-09c4-11e8-9533-0129d26ec367.png)
![jan24](https://user-images.githubusercontent.com/15680299/35782363-b3c5290a-09c4-11e8-81dd-3d0b450173c0.png)

rimatla:
How's possible to have all broken after I had the library version locked into my package.json?

VonTran:
It probably doesn't have much to do with the actual package. You use a required cdn in your head to pull down the styles it's something wrong with that.

rimatla:
You're right, updating the CDN worked. But... That's a vulnerability, my question to the developers here is how can I prevent such issue after my app is shipped to prod?

taion:
Seems unrelated to this package then. In general if these things matter, either use hashes or don’t use third party CDNs.

Planeshifter:
While the issue was not caused by the react-bootstrap package, it can be reproduced by loading the stylesheets from the CDN suggested in the documentation:

```
!-- Latest compiled and minified CSS -->
<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap.min.css">

<!-- Optional theme -->
<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap-theme.min.css">
 ```

I suggest you change these as otherwise many more users will run into this issue. Thanks!

rimatla:
My case exactly. I used that cdn per documentation's instructions.

taion:
[...]

PR #2979, opened 2018-02-05
Lock bootstrap version in Getting Started guide

drugoi:
Locked bootstrap version from MaxCDN for prevent loading of new versions of Bootstrap.

Fixed this issue - https://github.com/react-bootstrap/react-bootstrap/issues/2978


jquense:
thanks!

PR #2987, opened 2018-02-06
Fix ModalDialog docs

taion:


taion:
Gonna merge this when CI goes green as a trivial docs fix.

PR #3022, opened 2018-03-04
Added a prop to enable adding classes to the modal-content dom node

jmpridgen:
I needed a way to apply a css class to the modal-content dom node. Making my own custom modalDialogComponent broke the drop in modal transition. This enables a custom class on the modal-content that is prepended to the other classes applied.

### DIFF
--- a/src/ModalDialog.js
+++ b/src/ModalDialog.js
@@ -15,7 +15,11 @@ const propTypes = {
   /**
    * A css class to apply to the Modal dialog DOM node.
    */
-  dialogClassName: PropTypes.string
+  dialogClassName: PropTypes.string,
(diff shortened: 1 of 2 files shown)


jmpridgen:
Did I not follow the correct contribution guidelines correctly? I would love to get this merged so I can switch back from my repository.

mxschmitt:
@jquense @taion what do you think? In my mind we could add something like this, but this PR is quite old and so I would create a new one for that.
(ignore this last commit^^)

bpas247:
@mxschmitt Maybe we can open a new issue to keep track of this? That way we can just close this PR for now and open another one sometime in the future.

PR #3035, opened 2018-03-05
Docs attribution

jquense:
We should really have had this in here...


taion:
should be "CC-BY" instead of "CC BY" though

taion:
err, no it's not. nvm sorry

PR #3044, opened 2018-03-05
update to babel 7

jquense:
prepares for a gatsby v2 upgrade as well


taion:
Can you double-check the output bundle sizes? I mean that they don't get larger

jquense:
at some point, what is the worry there tho? should be same. i don't care that much about the dist builds otherwise we'd be using rollup :P

taion:
With webpack 4 scope hoisting they should actually be fine :p

But I saw them getting like 25% bigger, which is nontrivial

jquense:
master:
image

babel 7 webpack-atoms@5:
image

rehevkor5:
Babel 7 is a pre-release version (still in beta), but these changes were published as a minor version change from 0.32.1 to 0.32.2 (according to the "master" branch: the default branch in github, "v4-support", does not include this change). Not sure if this was intentional?

PR #3046, opened 2018-03-05
Style context

jquense:
w/d/y/t ?


src/Button.js
```diff
@@ -3,100 +3,105 @@ import React from 'react';
...
+      className,
+      block,
+      type,
+      componentClass,
+      ...props
+    } = this.props;
 
-    if (elementProps.href) {
-      return this.renderAnchor(elementProps, fullClassName);
+    delete props.bsRole;
```
taion: can we do something like `bsRole: _0` in the spread above, and add a lint rule to not warn on unused vars starting with `_`? i think that's what we're largely doing now, and i think it's good

src/Button.js
```diff
@@ -3,100 +3,105 @@ import React from 'react';
...
+
+    const classes = classNames(
+      className,
+      bsPrefix,
+      active && 'active',
+      `${bsPrefix}-${variant}`,
```
taion: do we want to support the null `bsPrefix` case?

src/ThemeProvider.js
```diff
@@ -0,0 +1,51 @@
...
+      </ThemeContext.Provider>
+    );
+  }
+}
+
+function createBootstrapComponent(Component, prefix) {
+  const name = Component.displayName || Component.name;
+  // eslint-disable-next-line
+  return class extends React.Component {
+    static displayName = `Boostrap(${name})`;
```
taion: "Bootstrap"

src/Button.js
jquense: I think no. I couldn't think of any cases where it'd make sense. I guess if you want a completely unstyled component

PR #3050, opened 2018-03-05
Breadcrumbs

kendaganio:
First attempt implement breadcrumbs. Please let me know if I have missed anything.


www/src/pages/components/breadcrumb.js
```diff
@@ -15,8 +15,9 @@ export default function BreadcrumbSection({ data }) {
       </h2>
 
       <p>
-        Breadcrumbs are used to indicate the current page's location. Add{' '}
-        <code>active</code> attribute to active <code>Breadcrumb.Item</code>.
+        Indicate the current page’s location within a navigational hierarchy
+        that automatically adds separators via CSS. Add <code>active</code>{' '}
+        attribute to active <code>Breadcrumb.Item</code>.
```
jquense: lets use `prop` instead of `attribute` since it's a bit more accurate :)

src/BreadcrumbItem.js
```diff
@@ -39,7 +41,10 @@ class BreadcrumbItem extends React.Component {
     const linkProps = { href, title, target };
 
     return (
-      <li className={classNames(className, classes, { active })}>
+      <li
+        className={classNames(className, classes, { active })}
+        {...(active ? { 'aria-current': 'page' } : {})}
```
jquense: I believe that props would get sent to the DOM if they are undefined so you could simplify this like:

```jsx
aria-current={active ? 'page' : undefined}
```

src/BreadcrumbItem.js
```diff
@@ -7,7 +7,9 @@ import { bsClass, getClassSet, splitBsProps } from './utils/bootstrapUtils';
 
 const propTypes = {
   /**
-   * If set to true, renders `span` instead of `a`
+   * If set to true, renders `span` instead of `a` and
+   *
+   * adds `aria-current="page"` attribute
```
jquense: The original wording here was a bit wonky maybe: "Add's a visual "active" state to a Breadcrumb Item and disabling the link.

src/Breadcrumb.js
```diff
@@ -12,12 +12,13 @@ class Breadcrumb extends React.Component {
     const classes = getClassSet(bsProps);
 
     return (
-      <ol
-        {...elementProps}
-        role="navigation"
-        aria-label="breadcrumbs"
-        className={classNames(className, classes)}
-      />
+      <nav aria-label="breadcrumbs">
```
jquense: I think the aria-label should probably be configurable. I'd make it an explicit prop but set the default to "breadcrumbs" that way a user can change the language or wording as needed

kendaganio:
Hi @jquense! I'll fix the conflicts later, it's a bit late in my part of the world. Thank you for the opportunity to contribute. ❤️

src/Breadcrumb.js
```diff
@@ -1,27 +1,43 @@
...
 import React from 'react';
+import PropTypes from 'prop-types';
 
 import BreadcrumbItem from './BreadcrumbItem';
 import { bsClass, getClassSet, splitBsProps } from './utils/bootstrapUtils';
 
+const propTypes = {
+  /**
+   * ARIA label for the nav, if no label is provided
+   * it will default to "breadcrumbs".
```
jquense: no need to list the default the docs generation will pick up the defaultProps for it.

src/Breadcrumb.js
```diff
@@ -1,27 +1,43 @@
...
-      <ol
-        {...elementProps}
-        role="navigation"
-        aria-label="breadcrumbs"
-        className={classNames(className, classes)}
-      />
+      <nav aria-label={label}>
+        <ol
+          role="navigation"
+          {...elementProps}
```
jquense: @taion do we have an opinion on how to spread through props? Always the root element? Maybe we should also add a `listProps` prop as well?

taion: we never did sort this out \:

shouldn't the styling props go to the root element, at least, though?

jquense: i think so yeah.

kendaganio: Just let me know how you want me to proceed. :D

alacroix:
@kendaganio watch out, it seems you have commited `yarn-error.log` by mistake.

src/Breadcrumb.js
jquense: lets spread `elementProps` into the `nav` and add another prop `listProps` that is spread into the list element 👍 and in general going forward we'll try and be consistent on that pattern

src/Breadcrumb.js
```diff
@@ -1,27 +1,61 @@
...
 import { bsClass, getClassSet, splitBsProps } from './utils/bootstrapUtils';
 
+const propTypes = {
+  /**
+   * ARIA label for the nav element
+   * https://www.w3.org/TR/wai-aria-practices/#breadcrumb
+   */
+  label: PropTypes.string,
+  /**
+   * Optional prop that hold properties that will be spread to
```
jquense: "Additional props passed as-is to the underlying `<ul>` element"

src/Breadcrumb.js
```diff
@@ -1,27 +1,61 @@
...
+        aria-label={label}
         className={classNames(className, classes)}
-      />
+        {...elementProps}
+      >
+        <ol
+          role="navigation"
+          className={bsClass}
+          children={children}
+          {...listProps}
```
jquense: I think it's ok for `children` to not be overridden by `listProps` since a user can control them directly already

kendaganio: Umm, I needed to add that because the `children` are the `BreadCrumb.Item`s

jquense: sorry I mean lets be a bit more intentional about how `listProps` is spread in. right now if a user provides it it will override `className` and `children` on `ul` which isn't what we want. I would move `children` down and express it normally e.g. `<ul>{children}</ul` and use `classNames(bsClass, listProps.className)`

kendaganio: Yup, agreed. Didn't thought about that thanks for the clarificaiton.

src/Breadcrumb.js
```diff
@@ -1,27 +1,61 @@
...
+      <nav
+        aria-label={label}
         className={classNames(className, classes)}
-      />
+        {...elementProps}
+      >
+        <ol
+          role="navigation"
+          className={classNames(bsClass, listProps.className)}
+          {...listProps}
```
jquense: switch the order here to:

```
role="navigation"
{...listProps}
className={classNames(bsClass, listProps.className)}
```

otherwise it'll still override the className 👍

kendaganio:
Hi @jquense is this okay now? I want to take up some more items if this is passable.

src/Breadcrumb.js
```diff
@@ -1,27 +1,61 @@
...
+      bsClass: bsClassName,
+      listProps,
+      children,
+      className,
+      label,
+      ...props
+    } = this.props;
     const [bsProps, elementProps] = splitBsProps(props);
 
     const classes = getClassSet(bsProps);
```
jquense: lets switch to the newly-merged pattern for handling styles with `createBootstrapComponent` and bsPrefix, Check out the Button.js component for a reference. We want to not using anything from bootstrapUtils ideally

jquense:
friendly ping :)

PR #3057, opened 2018-03-06
Issue #3051 - Update Media Object

### LINKED ISSUE #3051: v4: Media Object
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/layout/media-object/)

I think this one is mostly deleting things, like MediaList and updating the markup

sliw:
Fixes #3051

1. Modified the code in Media.js

2. Removed src documents and test documents

of the following:

- MediaList
- MediaListItem
- MediaLeft
- MediaRight
- MediaHeading


jquense:
looks great 👍 can we also update the documentation to:

- reflect the components that exist
- show how to align media objects using the utility classes

The bootstrap docs are a good template/starting point there

jquense:
Also you need to submit this against the `v4-support` branch

sliw:
@jquense 

I was going to ask whether I have to change the documentation as well. Thanks for bringing this up.

I have changed the branch to `v4-support` and modified documentation.

For documentation, I have added examples about `MediaNesting` and  `MediaOrder`. But I am not sure whether I have done it correctly or not, particularly the classnames (e.g. `class="align-self-end"` to `align="self-end"`). Is there anyway I can check the documentation page locally (I tried to use `npm` but did not work)?

On top of this, may I ask is there any style guide on commenting because I feel a bit messy without commenting in the documentation.

Thanks 😄

www/src/examples/MediaAlignment.js
```diff
@@ -19,12 +23,17 @@
       </p>
     </Media.Body>
   </Media>
+
   <Media>
-    <Media.Left align="middle">
-      <img width={64} height={64} src="/thumbnail.png" alt="thumbnail" />
-    </Media.Left>
+    <img
+      align="self-center"
```
jquense: this should be a className not an attribute

www/src/examples/MediaOrder.js
```diff
@@ -0,0 +1,14 @@
+<div>
```
jquense: don't need the wrapping div here

src/Media.js
```diff
@@ -3,11 +3,7 @@ import React from 'react';
 import elementType from 'prop-types-extra/lib/elementType';
 
 import MediaBody from './MediaBody';
-import MediaHeading from './MediaHeading';
-import MediaLeft from './MediaLeft';
-import MediaList from './MediaList';
-import MediaListItem from './MediaListItem';
-import MediaRight from './MediaRight';
+
 import { bsClass, getClassSet, splitBsProps } from './utils/bootstrapUtils';
```
jquense: We recently added a different way to handle the styling bits, lets upgrade the Media component while we're here to use `createBootstrapComponent`, check out Button.js for a reference, we wnat to completely get rid of bootstrapUtils in each component 👍

jquense:
friendly ping :)

jquense:
merged in https://github.com/react-bootstrap/react-bootstrap/pull/3108

Thanks for starting this off @sliw much appreciated!

PR #3061, opened 2018-03-06
fix readme, react-16-enzyme doc running

jquense:


jquense:
ok netlify is at least deploying, and ci running, expectedly tests are failing tho.

PR #3062, opened 2018-03-06
Navs

jquense:
sits on top of #3046 

still a WIP, but the functionality is there, simpler and works!


src/Nav.js
```diff
@@ -1,359 +1,228 @@
...
     }
+    if (!nextActiveChild) return;
 
     event.preventDefault();
-
-    if (onSelect && nextActiveChild && nextActiveChild.props.eventKey != null) {
-      onSelect(nextActiveChild.props.eventKey);
-    }
-
+    onSelect(nextActiveChild.dataset.rbEventKey);
```
jquense: this is some OG DOM work here

src/Nav.js
```diff
@@ -1,359 +1,233 @@
...
+    /**
+     * The visual variant of the nav items.
+     *
+     * @type {('tabs'|'pills')}
+     */
+    variant: PropTypes.string,
+
+    /**
+     * Marks the NavItem with a matching `eventKey` as active. Has a
+     * higher precedence over `activeHref`.
```
taion: `activeHref` looks like it's gone now

src/Nav.js
```diff
@@ -1,359 +1,233 @@
...
+    this.state = { navContext: null };
+  }
 
-      return tabContainer;
-    }
+  componentDidUpdate() {
+    if (!this._needsRefocus || !this.listNode) return;
 
-    return this.props;
+    let activeChild = qsa(this.listNode, `.nav-link.active`).pop();
```
taion: this seems like an awkward way to just do `querySelector` :p

taion: maybe just `.active`? hard-coding `.nav-link` seems wrong if `bsPrefix` is customizable

src/Nav.js
```diff
@@ -1,359 +1,233 @@
...
-    const numValidChildren = validChildren.length;
-
-    if (nextIndex >= numValidChildren) {
-      nextIndex = 0;
-    } else if (nextIndex < 0) {
-      nextIndex = numValidChildren - 1;
-    }
-
-    return validChildren[nextIndex];
+    return items[items.indexOf(activeChild) + offset];
```
taion: suppose there's no active child. then this will give the item with index "offset - 1". is that right?

src/Nav.js
taion: not bad

src/Nav.js
```diff
@@ -1,359 +1,233 @@
...
+    delete props.onSelect;
+    delete props.getControlledId;
+    delete props.getControllerId;
 
     const classes = {
-      ...getClassSet(bsProps),
-      [prefix(bsProps, 'stacked')]: stacked,
-      [prefix(bsProps, 'justified')]: justified
+      [bsPrefix]: !navbar,
+      [`navbar-nav`]: navbar,
```
taion: ouch... can we read "navbar" from the context provider?

src/NavLink.js
```diff
@@ -0,0 +1,114 @@
...
+            ...props
+          } = this.props;
+
+          delete props.onSelect;
+
+          const navItemKey = String(eventKey || href);
+
+          const isActive =
+            active == null
+              ? String(navContext.activeKey) === navItemKey
```
taion: why stringify this from navcontext?

src/NavLink.js
```diff
@@ -0,0 +1,114 @@
...
+
+          let role = propsRole;
+          if (navContext.role === 'tablist') {
+            role = 'tab';
+            props['data-rb-event-key'] = navItemKey;
+            props['aria-selected'] = isActive;
+            props.tabIndex = isActive ? props.tabIndex : -1;
+          }
+
+          this.navContext = navContext;
```
taion: lol react

src/TabContainer.js
```diff
@@ -1,101 +1,132 @@
...
       }
     };
   }
 
-  render() {
-    const { children, ...props } = this.props;
+  getKey(key, type) {
+    const { generateChildId, id } = this.props;
+    if (generateChildId) return generateChildId(key, type);
+    return id ? `${id}-${type}-${key}` : null;
```
taion: can we use a default prop here?

src/TabContext.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const TabContext = React.createContext(
+  null /* {
```
taion: ?

src/Nav.js
jquense: ah true, 👍

src/Nav.js
jquense: yes upcoming in the navbar PR

src/NavLink.js
jquense: do i stringify it further up? i was doing it b/c setting it to data-wahtever stringifies it

src/NavLink.js
jquense: 😞 i know

src/TabContext.js
jquense: i was signaling to myself waht this type will be, but leaving it null to test presence. Good example of where types would be nice.

src/TabContainer.js
jquense: no /bc we need the id in the one case but not the other

src/NavLink.js
jquense: i have this more optimized in the Dropdowns PR, since that one finally covers all the "selectabe" components

src/NavLink.js
jquense: this is apparently not safe for async mode /shrug, fixed in the Dropdowns

jquense:
gonna get it in, since it's easier to keep refining as we go and my pr's are starting to balloon off thse branches :P

src/Nav.js
```diff
@@ -1,359 +1,231 @@
...
-    let nextIndex = activeChildIndex + offset;
-    const numValidChildren = validChildren.length;
+    if (!this.listNode) return null;
 
-    if (nextIndex >= numValidChildren) {
-      nextIndex = 0;
-    } else if (nextIndex < 0) {
-      nextIndex = numValidChildren - 1;
-    }
+    let items = qsa(this.listNode, '[data-rb-event-key]:not(.disabled)');
```
taion: this is really great

src/Nav.js
```diff
@@ -1,359 +1,231 @@
...
 
-    const tabContainer = this.context.$bs_tabContainer;
-    const role = propsRole || (tabContainer ? 'tablist' : null);
-
-    const { activeKey, activeHref } = this.getActiveProps();
-    delete props.activeKey; // Accessed via this.getActiveProps().
-    delete props.activeHref; // Accessed via this.getActiveProps().
-
-    const [bsProps, elementProps] = splitBsProps(props);
+    delete props.activeKey;
```
taion: could use the `: _1` pattern

src/NavLink.js
```diff
@@ -0,0 +1,113 @@
...
+            eventKey,
+            href,
+            role: propsRole,
+            onClick: _,
+            ...props
+          } = this.props;
+
+          delete props.onSelect;
+
+          const navItemKey = String(eventKey || href);
```
taion: can we pull this out? or do we really need this? might be more consistent to just always require event keys

PR #3063, opened 2018-03-07
V4 ProgressBar

grady-lad:
According to [migration guide](https://getbootstrap.com/docs/4.0/migration/) the new changes for the `ProgressBar` component are the following:

- `.active` class now becomes `progress-bar-animated`
  - Changed `active` prop to `animated`

- `.progress-bar-[bgStyle]` now becomes `.bg-[bgStyle]`
  - Props remain the same
  - `.progress-bar-[bgStyle]` is removed
 -  new class added to component `.bg-[bg-style]`

### DIFF
--- a/src/ProgressBar.js
+++ b/src/ProgressBar.js
@@ -58,7 +58,7 @@ const propTypes = {
   label: PropTypes.node,
   srOnly: PropTypes.bool,
   striped: PropTypes.bool,
-  active: PropTypes.bool,
+  animated: PropTypes.bool,
   children: onlyProgressBar,
 
   /**
@@ -70,7 +70,7 @@ const propTypes = {
 const defaultProps = {
   min: 0,
   max: 100,
-  active: false,
+  animated: false,
   isChild: false,
   srOnly: false,
   striped: false
@@ -89,19 +89,19 @@ class ProgressBar extends React.Component {
     label,
     srOnly,
     striped,
-    active,
+    animated,
     className,
     style,
+    bsStyle,
     ...props
   }) {
     const [bsProps, elementProps] = splitBsProps(props);
-
     const classes = {
       ...getClassSet(bsProps),
-      active,
-      [prefix(bsProps, 'striped')]: active || striped
+      [prefix({ bsClass: 'bg' }, bsStyle)]: bsStyle !== undefined,
+      [prefix(bsProps, 'animated')]: animated,
+      [prefix(bsProps, 'striped')]: animated || striped
     };
-
     return (
       <div
         {...elementProps}
@@ -131,7 +131,7 @@ class ProgressBar extends React.Component {
       label,
       srOnly,
       striped,
-      active,
+      animated,
       bsClass,
       bsStyle,
       className,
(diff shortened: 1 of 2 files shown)


src/ProgressBar.js
```diff
@@ -89,19 +89,19 @@ class ProgressBar extends React.Component {
...
   }) {
     const [bsProps, elementProps] = splitBsProps(props);
-
     const classes = {
       ...getClassSet(bsProps),
-      active,
-      [prefix(bsProps, 'striped')]: active || striped
+      [prefix({ bsClass: 'bg' }, bsStyle)]: bsStyle !== undefined,
+      [prefix(bsProps, 'animated')]: animated,
+      [prefix(bsProps, 'striped')]: animated || striped
```
jquense: Let's not use the bootstrapUtils here, we are getting rid of a lot of them in #3046 and we can preempt that change by handling the bits manually here like:

```
classes = {
  [bsClass]: true,
  [`bg-${bsStyle}`]: bsStyle,
  [`${bsClass}-animated`]: animated
}
```

We want to not rely on the `splitBsProps` helper as well so we can just spread out the bs props like we do   `animated` or `striped`

test/ProgressBarSpec.js
```diff
@@ -185,13 +180,13 @@ describe('<ProgressBar>', () => {
...
     const instance = ReactTestUtils.renderIntoDocument(
-      <ProgressBar min={1} max={11} now={6} active />
+      <ProgressBar min={1} max={11} now={6} animated />
     );
 
     const barClassName = ReactDOM.findDOMNode(instance).firstChild.className;
 
     assert.ok(barClassName.match(/\bprogress-bar-striped\b/));
-    assert.ok(barClassName.match(/\bactive\b/));
+    assert.ok(barClassName.match(/\banimated\b/));
```
jquense: `progress-bar-animated`

test/ProgressBarSpec.js
```diff
@@ -226,12 +221,12 @@ describe('<ProgressBar>', () => {
...
 
     assert.ok(bar1.className.match(/\bprogress-bar\b/));
-    assert.ok(bar1.className.match(/\bactive\b/));
+    assert.ok(bar1.className.match(/\banimated\b/));
     assert.ok(bar1.className.match(/\bprogress-bar-striped\b/));
 
     assert.ok(bar2.className.match(/\bprogress-bar\b/));
     assert.ok(bar2.className.match(/\bprogress-bar-striped\b/));
-    assert.notOk(bar2.className.match(/\bactive\b/));
+    assert.notOk(bar2.className.match(/\banimated\b/));
```
jquense: same for these two

www/src/examples/ProgressBarAnimated.js
```diff
@@ -1 +1 @@
-<ProgressBar active now={45} />;
+<ProgressBar animated now={45} />;
```
jquense: lets also move the examples to `examples/ProgressBar` like we've started to do with the other components 👍

grady-lad:
@jquense I just realised I my branch steams from `master` rather than `v4-support`  🤦‍♂️ .

Can I close this pr and create a new one steaming from the correct branch ?

Also thanks for the comments on the pr! 😄

jquense:
yeah go for it :)

PR #3064, opened 2018-03-08
Add v4 support for Jumbotron

### LINKED ISSUE #3036: v4: Jumbotron
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/components/jumbotron/)

stephanieinez:
This is the v4 support for the `Jumbotron` component (fixes #3036) as per the [v4 migration docs](https://getbootstrap.com/docs/4.0/components/jumbotron/).

I've included a new prop called `fluid` which adds a `jumbotron-fluid` modifier class.

@jquense -> I have included both  `Jumbotron` examples to `examples/Jumbotron.js`, is this correct?


src/Jumbotron.js
```diff
@@ -1,31 +1,37 @@
...
-      <Component {...elementProps} className={classNames(className, classes)} />
-    );
+    const {
+      componentClass: Component,
+      className,
+      fluid,
+      ...props
+    } = this.props;
+    const classes = {
+      jumbotron: true,
```
jquense: instead of a literal `'jumbotron'` here, we can now use `props.bsPrefix` which is passed in by the `createBootstrapComponent` so you can write the classes like: the below (after spreading out bsPrefix in the props block above)
```js
{
  [bsPrefix]: true,
  [`${bsPrefix}-fluid`]: fluid
}
```

jquense: we also want to add a propType in for it with a comment similar to: https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/src/ButtonGroup.js#L13

www/src/examples/Jumbotron/Basic.js
```diff
@@ -5,6 +5,6 @@
     extra attention to featured content or information.
   </p>
   <p>
-    <Button variant="primary">Learn more</Button>
+    <Button bsStyle="primary">Learn more</Button>
```
jquense: this should be variant, the prop name changed 👍

iamnapo:
Shouldn't a `<div class="container">` be added when `fluid: true`?

src/Jumbotron.js
iamnapo: the `[]` are unnecessary here. No?

jquense:
I think we can leave out the container. I think the idea is that fluid is appropriate for when you wnat a container inside a Jumbotron, but isn't required. Plus it'd conflict with what `fluid` means in the navbar case which is "no container"

iamnapo:
so, good to merge?

PR #3065, opened 2018-03-08
V4 progress bars

grady-lad:
Ok round 2 😄 

According to [migration guide](https://getbootstrap.com/docs/4.0/migration/) the new changes for the `ProgressBar` component are the following:

- `.active` class now becomes `progress-bar-animated`
  - Changed `active` prop to `animated`

- `.progress-bar-[bgStyle]` now becomes `.bg-[bgStyle]`
  - Props remain the same
  - `.progress-bar-[bgStyle]` is removed
 -  new class added to component `.bg-[bg-style]`

ProgressBar docs moved to `example/ProgressBar`


grady-lad:
@jquense I have updated `ProgressBar` to use the new `createBootstrapComponent` function, But this results in 3 tests failing, which is all about rendering `ProgressBar` children.

The reason is because there is a constraint that says that `ProgressBar` only allows children of `ProgressBar`  see [here](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ProgressBar.js#L40).

 But with the introduction of `createBootstrapComponent` what we are exporting is `Bootstrap(ProgressBar)`. 

I am not sure what to should check against on this [line](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ProgressBar.js#L40)

src/ProgressBar.js
```diff
@@ -2,20 +2,14 @@ import classNames from 'classnames';
...
-import { State } from './utils/StyleConfig';
+import { createBootstrapComponent } from './ThemeProvider';
+
 import ValidComponentChildren from './utils/ValidComponentChildren';
 
 const ROUND_PRECISION = 1000;
 
 /**
- * Validate that children, if any, are instances of `<ProgressBar>`.
+ * Validate that children, if any, are instances of `Bootstrap(<ProgressBar>)`.
```
jquense: We can leave this as `ProgressBar` as user won't have a concept of what the wrapper component is here since it's an implementation detail.

src/ProgressBar.js
```diff
@@ -89,22 +83,24 @@ class ProgressBar extends React.Component {
...
     ...props
   }) {
-    const [bsProps, elementProps] = splitBsProps(props);
 
     const classes = {
-      ...getClassSet(bsProps),
-      active,
-      [prefix(bsProps, 'striped')]: active || striped
+      [bsPrefix]: true,
+      [`bg-${bsStyle}`]: bsStyle,
```
jquense: let's change `bsStyle` to `variant` to be consistent with the new styling names :)

src/ProgressBar.js
```diff
@@ -164,7 +160,5 @@ class ProgressBar extends React.Component {
 ProgressBar.propTypes = propTypes;
 ProgressBar.defaultProps = defaultProps;
 
-export default setBsClass(
-  'progress-bar',
-  bsStyles(Object.values(State), ProgressBar)
-);
+export default createBootstrapComponent (ProgressBar, 'progress-bar');
```
jquense: in order to check the right thing above lets first put this in a local identifier before exporting like: 
```
const DecoratedProgressBar = createBootstrapComponent (ProgressBar, 'progress-bar');
export default DecoratedProgressBar
```


Then at [line](https://github.com/react-bootstrap/react-bootstrap/pull/3065/files#diff-bf9af2152e797180921eba199402e1c3L39) we can check `<DecoratedProgressNav>` and the tests should work again, you might need to make a local exception for eslint if it complains about use-before-define like: https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

src/ProgressBar.js
jquense: Lets also explicitly add `bsPrefix` and `variant` to the prop-types with docs comments.  For the bsPrefix if can be very simple like: https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/src/ButtonGroup.js#L13

src/ProgressBar.js
```diff
@@ -58,7 +52,14 @@ const propTypes = {
...
   striped: PropTypes.bool,
-  active: PropTypes.bool,
+  animated: PropTypes.bool,
+  bsPrefix: PropTypes.string,
+  /**
+   * Sets the background class of the progress bar.
+   *
+   * @type ('success'|'danger'|'warning'|'info')
+   */
+  variant: PropTypes.string,
```
royling: what about making this prop type definition exact as the comment?
```js
PropTypes.oneOf(['success','danger','warning','info'])
```

src/ProgressBar.js
```diff
@@ -58,7 +52,14 @@ const propTypes = {
...
   striped: PropTypes.bool,
-  active: PropTypes.bool,
+  animated: PropTypes.bool,
+  bsPrefix: PropTypes.string,
+  /**
+   * Sets the background class of the progress bar.
+   *
+   * @type ('success'|'danger'|'warning'|'info')
+   */
+  variant: PropTypes.oneOf(['success','danger','warning','info']),
```
jquense: We want to keep the propType here `string` so folks can extend with there own variants, the doclet `@type` above tho will show the "default" variants in the docs which is good.

src/ProgressBar.js
```diff
@@ -58,7 +52,14 @@ const propTypes = {
   label: PropTypes.node,
   srOnly: PropTypes.bool,
   striped: PropTypes.bool,
-  active: PropTypes.bool,
+  animated: PropTypes.bool,
```
jquense: I know a bunch of comments are missing, but we want to try and add/backfill docs for any of the components we're working on. Lets add descriptive comments 

(you don't need the the type doclet for everyone btw, docs will infer it from the propType, we use it as an override like in variant below)

www/src/pages/components/progress.js
```diff
@@ -5,13 +5,13 @@ import LinkToSource from '../../components/LinkToSource';
...
 import ReactPlayground from '../../components/ReactPlayground';
 
-import ProgressBarBasic from '../../examples/ProgressBarBasic';
-import ProgressBarWithLabel from '../../examples/ProgressBarWithLabel';
-import ProgressBarScreenreaderLabel from '../../examples/ProgressBarScreenreaderLabel';
-import ProgressBarContextual from '../../examples/ProgressBarContextual';
-import ProgressBarStriped from '../../examples/ProgressBarStriped';
-import ProgressBarAnimated from '../../examples/ProgressBarAnimated';
-import ProgressBarStacked from '../../examples/ProgressBarStacked';
+import ProgressBarBasic from '../../examples/ProgressBar/ProgressBarBasic';
```
jquense: lets remove the `ProgressBar` from the beginning of the file names since they are now namespaced in a directory 👍

src/ProgressBar.js
jquense: we want to allow folks to extend with variants of their own with minimal effort so the string is a bit more fexible

jquense:
thanks!

Issue #3055, opened 2018-03-06
v4: Tables


jquense:
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/content/tables/)

Need to support `responsive` breakpoints!

astronomersiva:
I would like to take this up!

jquense:
go for it!

astronomersiva:
Hi @jquense , I have pushed some changes [here](https://github.com/astronomersiva/react-bootstrap/commit/e9b9604de908d10d65c4074519e2727a4b1328c0).

Just wanted to make sure I am going on the right track especially with responsive-breakpoints.

jquense:
Hey awesome looking good. feel free to push a PR even if it's not finished,it's easier to comment on that way

PR #3070, opened 2018-03-09
Add v4 support for tables

astronomersiva:
PR for implementing #3055

- Changed `table-condensed` to `table-sm`.
- Changed `table-inverse` to `table-dark`. There was a confusion caused by the [migration guide](https://getbootstrap.com/docs/4.0/migration/#tables) but I guess I will go fix it there too.
- Brought back the wrapping div for responsive tables. There is again a conflict between the migration guide and the [documentation](https://getbootstrap.com/docs/4.0/content/tables/#responsive-tables).

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -2,63 +2,88 @@ import classNames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import {
-  bsClass,
-  getClassSet,
-  prefix,
-  splitBsProps
-} from './utils/bootstrapUtils';
-
-const propTypes = {
-  inverse: PropTypes.bool,
-  striped: PropTypes.bool,
-  bordered: PropTypes.bool,
-  condensed: PropTypes.bool,
-  hover: PropTypes.bool,
-  responsive: PropTypes.bool
-};
-
-const defaultProps = {
-  inverse: false,
-  bordered: false,
-  condensed: false,
-  hover: false,
-  responsive: false,
-  striped: false
-};
+import { createBootstrapComponent } from './ThemeProvider';
 
 class Table extends React.Component {
+  static propTypes = {
+    /**
+     * @default 'table'
+     */
+    bsPrefix: PropTypes.string,
+
+    /**
+     * Adds zebra-striping to any table row within the `<tbody>`.
+     */
+    striped: PropTypes.bool,
+
+    /**
+     * Adds borders on all sides of the table and cells.
+     */
+    bordered: PropTypes.bool,
+
+    /**
+     * Enable a hover state on table rows within a `<tbody>`.
+     */
+    hover: PropTypes.bool,
+
+    /**
+     * Make tables more compact by cutting cell padding in half by setting
+     * size as `sm`.
+     */
+    size: PropTypes.string,
+
+    /**
+     * Invert the colors of the table — with light text on dark backgrounds
+     * by setting variant as `dark`.
+     */
+    variant: PropTypes.string,
+
+    /**
+     * Responsive tables allow tables to be scrolled horizontally with ease.
+     * Across every breakpoint, use `responsive` for horizontally
+     * scrolling tables. Responsive tables are wrapped automatically in a `div`.
+     * Use `responsive="sm"`, `responsive="md"`, `responsive="lg"`, or
+     * `responsive="xl"` as needed to create responsive tables up to
+     * a particular breakpoint. From that breakpoint and up, the table will
+     * behave normally and not scroll horizontally.
+     */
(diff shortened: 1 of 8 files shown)


src/Table.js
```diff
@@ -44,17 +47,26 @@ class Table extends React.Component {
 
     const classes = {
       ...getClassSet(bsProps),
-      [prefix(bsProps, 'responsive')]: responsive,
-      [prefix(bsProps, 'inverse')]: inverse,
+      [prefix(bsProps, 'dark')]: dark,
       [prefix(bsProps, 'striped')]: striped,
       [prefix(bsProps, 'bordered')]: bordered,
-      [prefix(bsProps, 'condensed')]: condensed,
+      [prefix(bsProps, 'sm')]: sm,
```
jquense: this feels like a weird choice by the bootstrap team since there is only one size. I'm guessing they were leaving space for other sizes? in which case this might make more sense as a `size` prop, like in Button.

src/Table.js
```diff
@@ -44,17 +47,26 @@ class Table extends React.Component {
...
 
-    return (
+    const table = (
       <table {...elementProps} className={classNames(className, classes)} />
     );
+
+    if (responsive) {
+      let responsiveClass = 'responsive';
+      if (breakpoint) {
+        responsiveClass = `${responsiveClass}-${breakpoint}`;
```
jquense: What about making `responsive` a prop that can be `true` or a breakpoint? like: `responsive='sm'` or just `responsive`, @taion thoughts?

src/Table.js
```diff
@@ -10,32 +10,35 @@ import {
 } from './utils/bootstrapUtils';
```
jquense: We just merged some changes to how we do styling, can you take a look a the Button component as an example and use `createBootstrap` instead of any utils from bootstrapUtils? thanks!

src/Table.js
jquense: I think similarly with `dark` that should become `variant='dark'` like Button maybe

www/src/pages/components/table.js
```diff
@@ -16,22 +17,39 @@ export default function TableSection({ data }) {
...
-        wide, you will not see any difference in these tables.
+        Responsive tables allow tables to be scrolled horizontally with ease.
+      </p>
+
+      <h3>
+        <Anchor id="table-responsive">Always Responsive</Anchor>
+      </h3>
+      <p>
+        Across every breakpoint, use <code>responsive</code> for horizontally
+        scrolling tables.
```
jquense: lets maybe add a Callout as well here that responsive tables are wrapping in another div

src/Table.js
astronomersiva: Ah! That would be a better choice in fact.

src/Table.js
astronomersiva: In that case, striped and bordered should also be variants right? 

Also, apologies for the delay.

astronomersiva:
@jquense I have pushed the suggested changes. Please take a look :)

src/Table.js
```diff
@@ -2,63 +2,88 @@ import classNames from 'classnames';
...
-      [prefix(bsProps, 'condensed')]: condensed,
-      [prefix(bsProps, 'hover')]: hover
-    };
+    const table = <table {...props} className={classes} />;
 
-    return (
-      <table {...elementProps} className={classNames(className, classes)} />
-    );
+    if (responsive) {
+      let reponsiveClass = 'table-responsive';
```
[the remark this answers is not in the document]
astronomersiva: Good pick!

.gitignore
```diff
@@ -2,6 +2,7 @@
 .DS_Store
 npm-debug.log*
 node_modules
+.yarn
```
taion: ?

src/Table.js
```diff
@@ -2,63 +2,88 @@ import classNames from 'classnames';
...
-      [prefix(bsProps, 'condensed')]: condensed,
-      [prefix(bsProps, 'hover')]: hover
-    };
+    const table = <table {...props} className={classes} />;
 
-    return (
-      <table {...elementProps} className={classNames(className, classes)} />
-    );
+    if (responsive) {
+      let reponsiveClass = `${bsPrefix}-responsive`;
```
taion: "responsiveClass"

.gitignore
astronomersiva: This is yarn cache. Felt like adding it here since there's going to be a good number of people using Yarn.

src/Table.js
astronomersiva: Good catch, thanks! :) I have changed and pushed.

jquense:
🎉  Thanks a ton!

Issue #3023, opened 2018-03-04
v4: Alerts


jquense:
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/components/alerts/)

- [ ] new variants
- [ ] Heading component
- [ ] Link component

abdullahtariq1171:
I would love to take this.

jquense:
all yours 👍

PR #3072, opened 2018-03-09
V4 alerts

abdullahtariq1171:
Add new variants
Add support for 'Heading component'

Solves https://github.com/react-bootstrap/react-bootstrap/issues/3023

### DIFF
--- a/src/Alert.js
+++ b/src/Alert.js
@@ -14,16 +14,27 @@ import CloseButton from './CloseButton';
 
 const propTypes = {
   onDismiss: PropTypes.func,
-  closeLabel: PropTypes.string
+  closeLabel: PropTypes.string,
+  headingComponent: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6']),
+  headingText: PropTypes.string
 };
 
 const defaultProps = {
-  closeLabel: 'Close alert'
+  closeLabel: 'Close alert',
+  headingComponent: 'h4'
 };
 
 class Alert extends React.Component {
   render() {
-    const { onDismiss, closeLabel, className, children, ...props } = this.props;
+    const {
+      onDismiss,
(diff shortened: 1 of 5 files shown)


src/Alert.js
```diff
@@ -39,6 +65,11 @@ class Alert extends React.Component {
         className={classNames(className, classes)}
       >
         {dismissable && <CloseButton onClick={onDismiss} label={closeLabel} />}
+        {!!headingText && (
+          <HeadingComponent className="alert-heading">
+            {headingText}
+          </HeadingComponent>
```
jquense: lets split this out into another component `AlertHeading`, like we do for Modal, ModalHeader, etc.

src/utils/StyleConfig.js
```diff
@@ -22,6 +22,10 @@ export const State = {
   WARNING: 'warning',
   DANGER: 'danger',
   INFO: 'info',
+  PRIMARY: 'primary',
+  SECONDARY: 'secondary',
+  DARK: 'dark',
+  LINK: 'light',
```
jquense: we aren't using this for styles going forward :)

src/Alert.js
```diff
@@ -13,17 +13,43 @@ import { State } from './utils/StyleConfig';
...
+    const {
+      onDismiss,
+      closeLabel,
+      headingComponent: HeadingComponent,
+      headingText,
+      className,
+      children,
+      ...props
+    } = this.props;
     const [bsProps, elementProps] = splitBsProps(props);
```
jquense: We've updated how we are doing styling for components, Take a look at Button.js, which uses `createBootstrapComponent` and renames bsStyle to `variant` we want to do the same thing here, and not use anything from `bootstrapUtils`

jquense:
friendly ping :)

PR #3078, opened 2018-03-16
V4 support - Modal

jeffreyroshan:
# Changes

- Fixed close button alignment
- Updated example doc page
- Added **vertically centered modal** in example
- Added **centered** prop (bool) in Modal
- Added **modal with grid layout** in example

# Screenshot

modal-static
modal-basic
modal-basic-close
modal-container
modal-small
modal-large
modal-custom-size

## Vertically centered modal
modal-centered

## Using grid in modal
modal-grid

## New prop in Modal: centered (bool)
modal-prop-centered

### Need help with

- How do I run specific test files?
- Do we have a checklist for the Modal migration?


iamnapo:
To run a specific test file change [karma.conf.js#L15](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/karma.conf.js#L15) and [karma.conf.js#L18](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/karma.conf.js#L18) and run `yarn test`.

src/ModalDialog.js
```diff
@@ -36,7 +42,8 @@ class ModalDialog extends React.Component {
     const dialogClasses = {
       ...getClassSet(bsProps),
       [bsClassName]: false,
-      [prefix(bsProps, 'dialog')]: true
+      [prefix(bsProps, 'dialog')]: true,
+      [prefix(bsProps, 'dialog-centered')]: centered
```
jquense: lets switch to use `createBootstrapComponent` in all the Modal components instead of the bootstrapUtils, you can check out Button.js for an example on how to do it 👍

src/ModalHeader.js
```diff
@@ -59,14 +59,14 @@ class ModalHeader extends React.Component {
 
     return (
       <div {...elementProps} className={classNames(className, classes)}>
+        {children}
+
         {closeButton && (
           <CloseButton
             label={closeLabel}
             onClick={createChainedFunction(modal && modal.onHide, onHide)}
```
jquense: We want to update the `context.$bs_modal` usage to use the new `React.createContext` API outlined here: https://github.com/acdlite/rfcs/blob/new-version-of-context/text/0000-new-version-of-context.md

src/ModalTitle.js
```diff
@@ -9,7 +9,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  componentClass: 'h4'
+  componentClass: 'h5'
```
jquense: do the styles still require that this be a `h` tag? if not i think `div` is a better default, since it doesn't make an assumption about the document outline

www/src/examples/ModalStatic.js
```diff
@@ -1,14 +1,18 @@
-<div className="static-modal">
-  <Modal.Dialog>
-    <Modal.Header>
-      <Modal.Title>Modal title</Modal.Title>
-    </Modal.Header>
+const modalInstance = (
```
jquense: this looks like a bad merge, the old code is more correct (unnecessary render())

www/src/pages/components/modal.js
```diff
@@ -10,6 +10,8 @@ import Modal from '../../examples/Modal';
 import ModalContained from '../../examples/ModalContained';
 import ModalDefaultSizing from '../../examples/ModalDefaultSizing';
 import ModalCustomSizing from '../../examples/ModalCustomSizing';
+import ModalVerticallyCentered from '../../examples/ModalVerticallyCentered';
+import ModalGrid from '../../examples/ModalGrid';
```
jquense: Lets move all the example files under `examples/Modal/` (and drop the Modal prefix on the file name), like we've been doing with other components

jeffreyroshan:
@jquense after recent merge, I am getting the following error on `yarn run develop`

```
$ yarn run develop
yarn run v1.5.1
$ gatsby develop
success delete html and css files from previous builds — 0.020 s
error We encountered an error while trying to load your site's gatsby-config.js. Please fix the error and try again.


  Error: /Work/Github/react-bootstrap/www/  gatsby-config.js:12
    ...require('../.babelrc.js'),
    ^^^
  SyntaxError: Unexpected token ...
```

jquense:
@jeffreyroshan you need to use a newer version of Node, >= v8.3.0

src/ModalDialog.js
jeffreyroshan: To start with, I have updated the ModalFooter with createBootstrapComponent. Could you check if that's the recommended way?

src/ModalTitle.js
jeffreyroshan: Agreed. Made `div` the default.

www/src/examples/ModalStatic.js
jeffreyroshan: Fixed

www/src/pages/components/modal.js
jeffreyroshan: Done

src/ModalFooter.js
```diff
@@ -2,30 +2,24 @@ import classNames from 'classnames';
...
+    componentClass: 'div'
+  };
 
-class ModalFooter extends React.Component {
   render() {
     const { componentClass: Component, className, ...props } = this.props;
-    const [bsProps, elementProps] = splitBsProps(props);
 
-    const classes = getClassSet(bsProps);
+    const classes = classNames('modal-footer', className);
```
jquense: So `createBootstrapComponent` will inject the string you specified in as a `bsPrefix` prop, in this case it'll be `modal-footer`. That let's you not hardcode the `modal-footer` bit here, and instead for `className(className, bsPrefix)` to add the class. For cases where there are variants (like dialog), you can use both the plain bsPrefix, and then intropolated string to create all the classes like, ```classNames(bsPrefix, `${bsPrefix}-content`)```

src/ModalFooter.js
```diff
@@ -1,31 +1,32 @@
...
+      ...props
+    } = this.props;
 
-    const classes = getClassSet(bsProps);
+    const classes = classNames(className, bsPrefix);
 
-    return (
-      <Component {...elementProps} className={classNames(className, classes)} />
-    );
+    return <Component {...props} className={classNames(classes)} />;
```
jquense: this all looks great generally. i'd remove the extra `classNames` call here and right above, you can combine that all into `className={classNames(className, bsPrefix)}`

src/ModalDialog.js
jquense: yup looks great

src/ModalHeader.js
jeffreyroshan: Is there an example I can look at in another component?

isaacsanders:
@jquense @jeffreyroshan Thanks for your work on this, y'all!

jquense:
merged in #3187 thanks for all the initial work!

PR #3080, opened 2018-03-18
Forward ref

jquense:
Maybe we should go a step further and forward it through to the component so it can be attached directly to the DOM node?


taion:
Hmm... fair enough. I wonder if this might make tests harder. Maybe it doesn't matter.

PR #3087, opened 2018-03-22
WIP: V4 carousel

zshnr:
Hey @jquense,

I'm guessing the change from slide to animate might be breaking change. Is that acceptable for the v4-support branch or should it go to another and not as a part of the migration?

I guess a couple of things I'm having difficulty getting my head around are:

1. How are bootstrap classes resolved in our components? Because in V4 the class name structure for carousel has changed and I couldn't figure out how to make that change.
2. For the Page Visibility API, do we want to use it directly in the component, or as a HOC, or a render props type approach? Also, do we want to use an existing solution or (to keep dependencies to a minimum) build one ourselves for this project?
3. What would be the best way to visually test my changes to ensure that everything works as intended?

Thanks! :)

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -14,12 +14,10 @@ import {
 } from './utils/bootstrapUtils';
 import ValidComponentChildren from './utils/ValidComponentChildren';
 
-// TODO: `slide` should be `animate`.
-
 // TODO: Use uncontrollable.
 
 const propTypes = {
-  slide: PropTypes.bool,
+  animate: PropTypes.bool,
   indicators: PropTypes.bool,
   /**
    * The amount of time to delay between automatically cycling an item.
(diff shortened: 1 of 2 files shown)


jquense:
> How are bootstrap classes resolved in our components? Because in V4 the class name structure for carousel has changed and I couldn't figure out how to make that change.

Going forward they are resolved manually. we have a `createBootstrapComponent` HOC that injects a prop called `bsPrefix` (old `bsClass`) then you use that to build up the classes in the component. Check out how Button.js works in the v4-support branch

> For the Page Visibility API, do we want to use it directly in the component,

I'd use it directly in the component, the code should be fairly minimal, and a few direct checks on `document` are fine. I _would_ however see if most of the transition logic could be ripped out and replaced with Transition or TransitionGroup from `react-transition-group`, That might be worth doing tho in a follow up PR if it's to involved

>What would be the best way to visually test my changes

The best way is to use the documentation site, run `yarn start` to run it in dev mode.

PR #3089, opened 2018-03-25
Add v4 support for layout utility

chenesan:
So there is a `<Layout />` component wrap all the [layout utility](https://getbootstrap.com/docs/4.0/layout/utilities-for-layout/) in v4. All possible usage should be included in the tests.

And @jquense, where should I create the example page in `/www/src/pages` for this new component(layout or utilities or components?)?

### DIFF
--- a/src/Layout.js
+++ b/src/Layout.js
@@ -1,0 +1,277 @@
+import classNames from 'classnames';
+import React from 'react';
+import PropTypes from 'prop-types';
+import elementType from 'prop-types-extra/lib/elementType';
+
+const camelCaseToHyphen = str =>
+  str.replace(/[A-Z]/g, match => `-${match.toLowerCase()}`);
+
+const flexAlignPropTypes = PropTypes.oneOf([
+  'start',
+  'end',
+  'center',
+  'baseline',
+  'stretch'
+]);
+
+const displayPropTypes = PropTypes.oneOf([
+  'none',
+  'inline',
+  'inline-block',
+  'block',
+  'table',
+  'table-cell',
+  'table-row',
+  'flex',
+  'inline-flex'
+]);
+
+const flexDirectionPropTypes = PropTypes.oneOf([
+  'row',
+  'row-reverse',
+  'column',
+  'column-reverse'
+]);
+
+const flexWrapPropTypes = PropTypes.oneOf(['nowrap', 'wrap', 'wrap-reverse']);
+
+const justifyContentPropTypes = PropTypes.oneOf([
+  'start',
+  'end',
+  'center',
+  'between',
+  'around'
+]);
+
+const spaceUtilSizePropTypes = PropTypes.oneOfType([
+  PropTypes.number,
+  PropTypes.oneOf(['auto'])
+]);
+
+const spaceUtilPropTypes = PropTypes.oneOfType([
+  spaceUtilSizePropTypes,
+  PropTypes.shape({
+    /**
+     *
+     * the key is the side of spacing utility,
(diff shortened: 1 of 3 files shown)


jquense:
🎉 awesome work, thanks for starting this off!

I'd like to talk a bit a bout the api considerations and get some feedback back and forth before we do to much more coding (so as not to waste any effort). It seems like the challenge here is that for most props there is a "plain" and breakpoint equivalent yeah? The tough bit there is trying to get the "nesting" right so that the component is easy to user and easy to _read_, while also maintaining the flexibility of it.

Looking at this:
```
<Layout sm={{ m: { x: 1, l: 'auto' } }} xs={{ m: { x: 0, l: 1 } }} />
``` 
I think illustrates that concern well, it's very expressive but really hard to parse quickly. Some of this is coming from the class name mapping, bootstrap picked really terse names, presumably to save bytes and typing, but it sacrifices clarity in the process. I don't think that tradeoff is necessarily a bad one but its definitely out of place in react-land where the philosophy always leans on clarity and  explicitness over terseness (e.g. `componentWillReceiveProps()` vs `update()`), I think we probably want to go that direction as well, especially since we don't have the byte size concern to worry about (since they all boil down to the terse classes anyway).

The other thing to keep in mind is optimizing for the most common case, which will probably (a hunch) not be complex responsive combinations, but probably just a quick "use this margin", etc.

In terms of the prop api what about something like:
```jsx
<Layout marginX={1} marginLeft={2} />

<Layout marginX={{ sm: 1, xl: 0 }} marginLeft={{ sm: 'auto', xl: 1 }} />
```
It flattens it a bit more and non-breakpoint case is pretty clear? thoughts?

chenesan:
Hi, thanks for your comment!

(1) Originally I tried to preserve the bootstrap usage as much as possible because I think maybe it's easier to understand for some users who are familiar with bootstrap. But I agree that `{marginLeft: 1}` is better than `{m: { l: 1}}` generally. So I think it's ok to change all the sides abbreviation into full name like `margin`, `marginTop` or `marginX`

(2) I'd agree that `<Layout marginX={1} />` is easier to use than `<Layout xs={{ marginX: 1}} />` in such quick use case. However when I was implementing I'm not sure if it's easier to integrate with existed `<Col />` or `<Row />` component in the future with the breakpoint props form so I keep the `sm` `xl` props.

Maybe we can keep the both different forms of props? So we can write it like `<Layout sm={{marginLeft: 3}}` or `<Layout marginLeft={{sm: 3}} />`. But I'm not sure if this will make the api too fat or harder to understand. If we only can select one form of props I think it's indeed clearer in the latter non-breakpoint prop case.

jquense:
>  I'm not sure if it's easier to integrate with existed <Col /> or <Row />

I think the double usage of the breakpoint names would make that harder not easier actually.

> Maybe we can keep the both different forms of props? 

I'd say keep it simple and just the one way for now, we can always iterate as we hit additional use cases and applications.

chenesan:
> I'd say keep it simple and just the one way for now, we can always iterate as we hit additional use cases and applications.

Ok I see. I'll try your way to make utilities as props and make breakpoint settings nested in props. Just something like `<Layout display="inline-block" margin={2} paddingY={{ xs: 6, md: 2  }}  />`. May be done in this week.

bpas247:
What's the current status of this PR?

chenesan:
Sorry that this has been a long time. Since now I'm not working on this I'd close this. Feel free to open PR if you wanna go for it :)

PR #3096, opened 2018-03-28
Add engines to package.json

astronomersiva:


jquense:
@iamnapo you gotta hold up a bit on these. Please don't auto merge stuff in. 

I know i suggested this, but i'm not sure it's actually a good idea. This will prevent any one using bootstrap via yarn from installing it even tho the node version doesn't matter in that case

iamnapo:
@jquense I so sorry for that, it was my mistake.

For future reference, could you explain why node version wouldn't matter for yarn users?

jquense:
> For future reference, could you explain why node version wouldn't matter for yarn users?

yarn enforces the "engines" key very strictly by default it will error if you are trying to install react-bootstrap on a version of node lower than 8.3.0, and break. That is the behavior we want when someone is doing development on RB, but if they are just installing it to bundle in there application the version of node they are using doesn't matter because they aren't interacting without our build tooling.

Issue #3025, opened 2018-03-04
v4: Cards


jquense:
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/components/cards/)

manishsaraan:
I would like to take this.

jquense:
awesome thanks!

kendaganio:
Hi @manishsaraan just checking in if you're still interested in this issue. I can pick this up if you're a bit busy. Thanks!

kendaganio:
Hi @manishsaraan following up if you're still interested! 😺

jquense:
@kendaganio I think it's safe for you to work on this if you'r interested.

kendaganio:
Thanks @jquense, i'll take this one.

manishsaraan:
Sorry for the late reply guys. I am little busy these days so @kendaganio you can work on it. :)

jquense:
No problem.feel free to come one back if you get some time, plenty of other stuff to do

manishsaraan:
Sure will do.

PR #3099, opened 2018-03-30
WIP: v4-cards

kendaganio:
Issue #3025

TODO
- [x] Card
- [x] CardBody
- [x] CardImg
- [x] CardImgOverlay
- [x] CardTitle
- [x] CardSubtitle
- [x] CardText
- [x] CardLink
- [x] CardHeader
- [x] CardFooter
- [x] CardHeaderTabs (?)
- [x] CardHeaderPills (?)
- [x] CardGroup
- [x] CardDeck
- [x] CardColumns
- [x] CardStyles


kendaganio:
Hi @jquense, just want to know how you want to handle the card-img-{top, bottom, overlay} components. Do you think it's a good idea to add a variant for that? Top and bottom works the same, but the overlay one I think needs a separate component. Thanks!

src/CardLink.js
```diff
@@ -0,0 +1,34 @@
...
+      className,
+      componentClass: Component,
+      ...props
+    } = this.props;
+
+    return <Component className={classNames(bsPrefix, className)} {...props} />;
+  }
+}
+
+export default createBootstrapComponent(CardLink, 'card-link');
```
jquense: for these simple components I think there is a util for creating them dynamically, createWithBsPrefix i think? in /utils, might be easier for the logic-less ones

src/CardSubtitle.js
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * @default 'card-subtitle'
+     */
+    bsPrefix: PropTypes.string,
+
+    componentClass: elementType
+  };
+
+  static defaultProps = {
+    componentClass: 'h6'
```
jquense: generally we don't want to default anything to a `h*` tag because the one you use is completely context specific, e.g how deep in the app it is etc. What we might want to do is default to a `div className='h6'/>` instead that a user can override as via the componentClass

src/CardLink.js
kendaganio: I was thinking the same thing! I should have looked into the code more, thanks for the tip!

src/CardSubtitle.js
kendaganio: This makes sense, will change the Title/Subtitle to that.

jquense:
I think for the header tabs and pills we maybe want to do the same thing we do with navbar-nav's which is use context to auto add a `card` prop to the nav to add the additional classes

src/CardGroup.js
```diff
@@ -0,0 +1,3 @@
+import createWithBsPrefix from './utils/createWithBsPrefix';
+
+export default createWithBsPrefix('card-group');
```
kendaganio: @jquense do you reckon this is okay to do?

jquense: Sure, though we will need to figure out how to generate docs for these cases but that's something we need to do anyway.

www/src/pages/components/cards.js
```diff
@@ -102,6 +103,15 @@ export default function CardSection({ data }) {
       <p />
       <ReactPlayground codeText={CardImgOverlay} />
 
+      <h3>
+        <Anchor id="card-example-layout">Card layout</Anchor>
+      </h3>
```
jquense: As you're going here, might be worth rebaseing with v4-support and using the new `Heading` and `ComponentApi` components in the docs instead of Anchor and PropTable

jquense: It will still be open. you can do whatever you want to the branch and the pr will update, and force pushing to your own branch is fine

kendaganio:
@jquense almost done here! Please check the context implementation, not sure if that's how it works. I'm still learning the new 16.3 APIs. Also, what do you think about card styling? Should I add a `bg` and `text` prop or just let the user add a `className="bg-primary"`?

src/CardSubtitle.js
```diff
@@ -0,0 +1,38 @@
...
+
+  render() {
+    const {
+      bsPrefix,
+      className,
+      componentClass: Component,
+      ...props
+    } = this.props;
+
+    const isDefaultComponent = Component === 'div';
```
jquense: maybe we can define this one like: https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/src/Alert.js#L119-L127 and save some bytes?

src/CardTitle.js
```diff
@@ -0,0 +1,38 @@
...
+    const {
+      bsPrefix,
+      className,
+      componentClass: Component,
+      ...props
+    } = this.props;
+
+    const isDefaultComponent = Component === 'div';
+
+    const classes = classNames(bsPrefix, className, isDefaultComponent && 'h5');
```
jquense: ditto here.

src/CardDeck.js
```diff
@@ -0,0 +1,3 @@
+import createWithBsPrefix from './utils/createWithBsPrefix';
+
+export default createWithBsPrefix('card-deck');
```
jquense: Lets move any component that's simply a `createWithBsPrefix` one into the Parent component (e.g the Card here). There isn't much of an advantage to splitting for cherry-picking and there isn't an logic in these to makethe Card component confusing to read through

src/CardBody.js
```diff
@@ -0,0 +1,34 @@
...
+      className,
+      componentClass: Component,
+      ...props
+    } = this.props;
+
+    return <Component className={classNames(bsPrefix, className)} {...props} />;
+  }
+}
+
+export default createBootstrapComponent(CardBody, 'card-body');
```
jquense: `createWithBsPrefix`?

www/src/pages/components/cards.js
```diff
@@ -0,0 +1,161 @@
...
+      <Heading h="1" id="cards">
+        Cards
+      </Heading>
+
+      <p className="lead">
+        Bootstrap’s cards provide a flexible and extensible content container
+        with multiple variants and options.
+      </p>
+
+      <Heading h="3" id="card-example-basic">
```
jquense: h="2"

src/CardDeck.js
kendaganio: So the `Card{Deck, Group, Columns}` wrap a bunch of cards together like how a `ButtonGroup` would for buttons, so I think it makes sense for it to be outside of the `Card` component.

kendaganio:
@jquense please take a look if you're free. 🎉

www/src/examples/Card/BodyOnly.js
```diff
@@ -0,0 +1,3 @@
+<Card>
```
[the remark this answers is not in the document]
kendaganio: yup, sounds good to me

www/src/examples/Card/WithHeaderAndQuote.js
```diff
@@ -0,0 +1,15 @@
+<Card>
+  <Card.Header>Quote</Card.Header>
+  <Card.Body>
+    <blockquote className="blockquote mb-0">
+      <p>
+        {' '}
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer posuere
+        erat a ante.{' '}
+      </p>
+      <footer className="blockquote-footer">
```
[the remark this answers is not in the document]
kendaganio: shall we create a new issue in v4-support for this?

www/src/examples/Card/Basic.js
```diff
@@ -0,0 +1,11 @@
+<Card style={{ width: '18rem' }}>
+  <Card.Img variant="top" src="http://placehold.it/268x180" />
```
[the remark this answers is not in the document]
kendaganio: Might be worth looking into [this](https://github.com/imsky/holder) since bootstrap uses this as well?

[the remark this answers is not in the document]
jquense: do they? I'm happy to use whatever, jsut want to get them working tho :)

src/Card.js
```diff
@@ -0,0 +1,126 @@
...
+      bg && `bg-${bg}`,
+      text && `text-${text}`,
+      border && `border-${border}`
+    );
+
+    if (body) {
+      const { children, ...rest } = props;
+      return (
+        <Component className={classes} {...rest}>
+          <div className="card-body">{children}</div>
```
jquense: you're missing the Provider here, plus i'd use Card.Body directly, even if you have to create above first.

I'd combine this into one branch as well:
```jsx
<CardContext.Provider value={this.state.cardContext}>
  <Component className={classes} {...props}>
    {body ? <CardBody>{children}</CardBody> : children}
  </Component>
</CardContext.Provider>
```

jquense:
aweeeeesooooommmmeeeee

PR #3100, opened 2018-03-30
Update to webpack4 and simplify build tooling

jquense:
slowly trimming it all down to nothing


tools/build.js
```diff
@@ -1,31 +1,120 @@
...
+};
+
 const has = t => !targets.length || targets.includes(t);
 
+/**
+ * Run babel over the src directory and output
+ * compiled common js files to ./lib.
+ */
+const buildLib = step('commonjs modules', libRoot, () =>
+  execa.shell(`npx babel ${srcRoot} --out-dir ${libRoot} --env-name "lib"`, {
```
taion: why npx instead of just using a dev dep?

jquense: it is a dev dep, but this saves me from having to do `node_modules/.bin/node`

taion: fair enough i guess

PR #3109, opened 2018-04-04
remove Panel component

jquense:


jquense:
we need to add back some sort of generic CollapseParent component for accordions

PR #3113, opened 2018-04-06
use `as` instead of componentClass

jquense:


jquense:
probably, should be real quick to do. We can backport it to master as well

Issue #3041, opened 2018-03-05
Documentation design!


jquense:
Our documentation could use a facelift! We've recently updated it to a Gatsby js site, and could definitely benefit from some design work.

- [ ] Clean up all the forked bootstrap doc css, removing unused stuff
- [ ] Update from LESS to Sass

maaz93:
I'd like to take this up. I'm experimenting with Gatsby, so this would be a fun experience I hope. Could you elaborate a bit on the first task?

jquense:
Awesome! thanks for helping out.

>  Could you elaborate a bit on the first task

Yes, a lot of the styling you'll see in `www/src/css/` is a complete copy/paste of the v3 bootstrap docs styles and then heavily modified. We don't use a lot of it and the current docs site structure isn't even similar to the v3 docs. I'd love if we could essentially "start over" with the style files and build something up with a bit more modern of a style architecture.

Does that help? In terms of design I think we are ll pretty open to whatever, though maybe not a complete copy of the bootstrap docs :P

maaz93:
When you say _modern style architecture_, you mean the way the styles are organized or written?

Also honestly, when I first read the issue description, I got the impression that it's more of a clean-up (as in remove unused things, format the CSS to a way it is more readable etc.) task than a re-design the website task. Sadly, I won't have the time to do a new website from scratch 😞 But, I can definitely help with testing the new site, whenever somebody starts working on it. Hope that's ok!

jquense:
> as in remove unused things, format the CSS to a way it is more readable etc.

That would be great! We probably do need a bunch of redesign eventually, but a great first step would be to clean up the css that's there and update the existing styles to not be broken with the new bootstrap v4 css. If you're up for that please go for it 👍

maaz93:
Awesome! I'll start it off and come back if I have any questions.

maaz93:
@jquense I've started looking at this. Is the docs site broken in the v4-support branch? I get this error after doing `yarn run bootstrap` and then `yarn start` from the project root.

> Uncaught (in promise) ReferenceError: React is not defined
    at Basic.js:10
    at Array.map (<anonymous>)
    at Object.C:\Users\i317212\oss\react-bootstrap\www\node_modules\babel-loader\lib\index.js??ref--0-0!C:\Users\i317212\oss\react-bootstrap\www\src\examples\Alert\Basic.js (Basic.js:1)
[...]

PR #3115, opened 2018-04-07
Cleanup docs css

maazadeeb:
Related to #3041.

Using the code coverage feature from Chrome devtools, I've figured out the CSS which is never used in the new docs site. Along the way, I also realized that some CSS is outdated due to some DOM structure changes. I've kept note of the CSS which is not used based on DOM structure change and will do that next.

Let me know if you notice something deleted which might be required.

Also, I fixed a small CSS problem with the twitter followers number

TODO:

 - [x] Figure out the navbar CSS related to branding and hover
 - [ ] Things related to `bs-example` class are kind of broken because of change in DOM structure (Because of using react-live I believe). Fix it
 - [x] The RB icon has a bunch of CSS of the actual Bootstrap icon CSS, and it's overridden almost completely. Get rid of the extra styles
 - [x] The prop tables seem to broken on mobile, with extra horizontal scroll coming up for the whole page. This is not there in master. Need to investigate
 - [ ] Fix horizontal scrolling on mobile for
    - [ ] Buttons
    - [ ] Images
    - [ ] Navbar
    - [ ] Overlays
    - [ ] Pagination

Changes that require feedback:
 - Have fallen back to Bootstrap defaults for the colors of the nav links. No additional changes done there as the defaults look pretty good actually. The current pinkish bottom border might require a bit more hacking, if it's actually needed.


maaz93:
I just noticed that I've deleted some stuff in the navbar which was actually a change in DOM structure. Will introduce it back

maaz93:
As I'm looking to modify the examples CSS, I have this thought: Should there be any external CSS applied to the displayed examples which is not immediately obvious to a consumer, when they look the JSX and the rendered output?

As it stands currently, there is some extra CSS applied to the rendered examples which the consumer of the library will never know unless they see the source. Shouldn't we make it explicit somehow that "Here's the rendered button from react-bootstrap. But we've applied some extra CSS so that all the rendered buttons don't look cluttered on the screen"?

Thoughts?

maaz93:
Hmm. That makes sense.

maaz93:
Apologies for the lack of updates here. I haven't been able to gather the time to finish it off. I can complete the second TODO soon, I think I only have one or two more components.

Can we create a separate issue for the horizontal scrolling issues that I've mentioned in the TODO list?

www/src/components/PageFooter.js
```diff
@@ -35,7 +35,7 @@ class PageFooter extends React.Component {
               <li>
                 <iframe
                   src="https://platform.twitter.com/widgets/follow_button.html?screen_name=react_bootstrap&show_screen_name=true"
-                  width={230}
+                  width={235}
```
taion: ooo

www/src/pages/components/modal.js
```diff
@@ -24,7 +24,10 @@ export default withLayout(function ModalSection({ data }) {
         Static Markup
       </Heading>
       <p>A modal dialog component</p>
-      <ReactPlayground codeText={ModalStatic} />
+      <ReactPlayground
+        exampleClassName="bs-example-modals-static"
```
taion: i wonder if we should actually just have the modal dialog as a component... we had people confused by this before

jquense: Funny thing, that is what we are doing in the example now. This bonus class isn't doing anything, just noticed today

maaz93:
I was having problems merging yesterday. Some error about socket io client or something like that. @jquense were you able to get the docs running locally after the merge?

Issue #3027, opened 2018-03-04
v4: List Group


jquense:
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/components/list-group/)

- [ ] New sub component API

alacroix:
Hi @jquense ! I would like to take this one :)

jquense:
👍 awesome all yours.

jquense:
This one will involve some API changes, because the API we have now is not great, we want something more along the lines of the Panel api

jquense:
hey @alacroix just a friendly ping to see if you're still up for this. No problem either way, i want to know if someone else could work on it or if you're still planning on it 👍

alacroix:
Hey @jquense, sorry for the delay. I just changed job so I won't have free time to work on this, sorry 😕 

@kendaganio it's all yours.

kendaganio:
@jquense taking this on next!

PR #3124, opened 2018-04-19
WIP: v4 List Groups

kendaganio:
Issue #3027

TODO
- [x] fix docs
- [x] createBootstrapComponent
- [ ] fade (?)
- [x] variants
- [x] context classes 
- [x] clickable ListGroupItems


jquense:
hey @kendaganio sorry for the delay, do you need input here? No rush just want to make sure you aren't waiting on us

kendaganio:
Hello @jquense, I haven't completed the entire thing yet. Have been a bit busy at work lately. Will get back to it next week!

jquense:
@kendaganio I've rebased and taken a bit of a shot at finishing this off. Thanks a ton for your work so far!

kendaganio:
terribly sorry @jquense for leaving this hanging for quite some time, was a bit busy at work but will try my best to continue helping out whenever i'm free!

jquense:
no worries! we all get busy, i've also been super short on time

src/AbstractNav.js
```diff
@@ -0,0 +1,151 @@
...
+import PropTypes from 'prop-types';
+import elementType from 'prop-types-extra/lib/elementType';
+import mapContextToProps from 'react-context-toolbox/lib/mapContextToProps';
+import SelectableContext, { makeEventKey } from './SelectableContext';
+import NavContext from './NavContext';
+import TabContext from './TabContext';
+
+const noop = () => {};
+
+class AbstractNav extends React.Component {
```
jquense: cheeeck this out @taion i quite like the sort of composability of the new contexts

jquense:
The ul/li mismatch is even more annoying now that bootstrap upstream doesn't care about what element you use (nave as well). I think that for simple cases a ul/li syrucutre is semantically and a11y the most correct but we cant enforce they are paired together so I think having completely incorrect markup easily is a worse outcome. I think moving away from ul/li as default elements will probably cause less bugs than they create

taion:
i'm totally cool with that – seems fine to ask users to use the correct `as`

Issue #2825, opened 2017-10-10
[Docs] Accessibility : Should add aria-expanded and aria-controls for Collapse toggle


changLiuUNSW:
For accessibility we should add aria-expanded and aria-controls for Collapse toggle:
https://react-bootstrap.github.io/components.html#transitions-collapse
```
class Example extends React.Component {
  constructor(...args) {
    super(...args);

    this.state = {};
  }

  render() {
    return (
      <div>
        <Button onClick={ ()=> this.setState({ open: !this.state.open })}>
          click
        </Button>
        <Collapse in={this.state.open}>
          <div>
            <Well>
              Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid.
              Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident.
            </Well>
          </div>
        </Collapse>
      </div>
    );
  }
}

ReactDOM.render(<Example/>, mountNode);

```

The accessible one:

```
class Example extends React.Component {
  constructor(...args) {
    super(...args);

    this.state = {
        open: false
    };
  }

  render() {
    return (
      <div>
        <Button onClick={ ()=> this.setState({ open: !this.state.open })} aria-expanded={this.state.open} aria-controls="collapseExample">
          click
        </Button>
        <Collapse in={this.state.open}>
          <div id="collapseExample">
            <Well>
              Anim pariatur cliche reprehenderit, enim eiusmod high life accusamus terry richardson ad squid.
              Nihil anim keffiyeh helvetica, craft beer labore wes anderson cred nesciunt sapiente ea proident.
            </Well>
          </div>
        </Collapse>
      </div>
    );
  }
}

ReactDOM.render(<Example/>, mountNode);

```

coreybrown89:
I'd love to have accessibility support for `<Collapse>` as well!

taion:
We're definitely open to PRs here.

she1991:
Are you still taking PRs here, would love to make it happen.

jquense:
yup! please go for it 👍

bpas247:
Is this issue currently up for grabs? Trying to determine the state of v4 support atm

taion:
sort of.

- we automatically set `aria-expanded` on `<Collapse>` when `role` is set.
[...]

PR #3131, opened 2018-04-25
Added aria-expanded and aria-controls for Collapse toggle issue #2825

she1991:
WAIT. This PR does NOT contain tests. I have a question.

The role attribute of the Collapse component implicitly sets the values of the aria-expanded attribute. With an explicitly defined aria-expanded attribute, how will this conflict be handled?

I think this needs a bit of discussion - I am fairly new to react so please feel free to state the obvious etc.

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -119,7 +119,20 @@ const propTypes = {
   /**
    * ARIA role of collapsible element
    */
-  role: PropTypes.string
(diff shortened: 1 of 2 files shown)


mxschmitt:
Please create an issue when you have a question. Aria attributes etc. can be added manually to a control and should not be contained in the prop-types.

PR #3187, opened 2018-06-29
Pr/3078

jquense:


src/Modal.js
```diff
@@ -149,147 +148,151 @@ class Modal extends React.Component {
...
+  handleDialogClick = e => {
+    if (e.target !== e.currentTarget) return;
 
     this.props.onHide();
-  }
+  };
+
+  handleEnter = (node, ...args) => {
+    if (node) {
+      node.style.display = 'block';
```
taion: so twiddling this is a new requirement with TWBS v4?

www/gatsby-node.js
```diff
@@ -8,6 +8,7 @@ exports.onCreateWebpackConfig = function onCreateWebpackConfig({
   getConfig
 }) {
   actions.setWebpackConfig({
+    devtool: 'cheap-inline-module-source-map',
```
taion: remove?

src/ModalDialog.js
```diff
@@ -2,61 +2,56 @@ import classNames from 'classnames';
...
+    dialogClassName: PropTypes.string,
+
+    /**
+     * Specifies a large or small modal.
+     *
+     * @type ('sm'|'lg')
+     */
+    size: PropTypes.string,
+
+    /**
```
taion: sync up this comment with the one on `<Modal>`?

src/Modal.js
jquense: This more correct for v3 as well but we don't allow modals to be in the DOM prior to showing so it's sort of irrelevant but I figured making it more correct can't hurt

src/ModalTitle.js
```diff
@@ -1,31 +1,28 @@
...
+import { createBootstrapComponent } from './ThemeProvider';
 
-const propTypes = {
-  as: elementType
-};
-
-const defaultProps = {
-  as: 'h4'
-};
+const DivStyledAsH4 = divWithHeadingClass('h4');
```
taion: so the idea is that this looks more or less per the examples, but doesn't insert a weird semantic header component? cool!

src/utils/divWithHeadingClass.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import classNames from 'classnames';
+
+export default headingClass =>
```
taion: arguably this should just be `divWithClassName` :p

src/Modal.js
kt3k: @jquense @taion 
I think this change caused the issue https://github.com/react-bootstrap/react-bootstrap/issues/3399

`handleEnter` is called at onEnter of Transition object (in react-overlays/modal). When the user pass `animation={false}`, then the animation object is not created and therefore this method is not called.

PR #3199, opened 2018-07-20
Default all NavItems in Tabs to have tabindex="0"

jkrehm:
The `active` logic will take care of making the inactive ones un-focusable, leaving only the active NavItem as a focusable element.

Currently, the tabs are skipped over when using a keyboard to tab through the on-screen elements.

### DIFF
--- a/src/Tabs.js
+++ b/src/Tabs.js
@@ -82,7 +82,12 @@ class Tabs extends React.Component {
     }
 
     return (
-      <NavItem eventKey={eventKey} disabled={disabled} className={tabClassName}>
+      <NavItem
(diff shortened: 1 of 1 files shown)


jquense:
Hey, thanks for PR, I'm not sure what problem this is solving, is there an open issue or could you walk me through the specific problem?

jkrehm:
Well dang, it must just be a bug in Vivaldi (the Chromium-based browser that I use). I should have checked in Chrome itself. Apparently Vivaldi is alone in not tabbing through anchors unless they have a tabindex specified.

Sorry for the noise. I'll open an issue with Vivaldi instead.

PR #3201, opened 2018-07-21
Update Carousel

jquense:
this is such a meh component…

still doesn’t have keyboard events or support for the other animations…


jquense:
this is done i think

src/Carousel.js
```diff
@@ -63,233 +100,278 @@ const propTypes = {
...
-        previousActiveIndex: activeIndex,
-        direction:
-          nextProps.direction != null
-            ? nextProps.direction
-            : this.getDirection(activeIndex, nextProps.activeIndex),
-      });
-    }
+      let direction;
+      if (
+        (nextIndex === 0 && previousActiveIndex >= lastPossibleIndex) ||
```
taion: when does this trigger? if previously there were 3 items, but now there's only 1? why should that trigger "next", v going from 3 -> 2 trigger "prev"?

src/Carousel.js
```diff
@@ -63,233 +100,278 @@ const propTypes = {
...
     }
+
+    this.select(index, event);
   }
 
-  waitForNext() {
-    const { slide, interval, activeIndex: activeIndexProp } = this.props;
+  select(index, event, direction) {
+    clearTimeout(this.selectThrottle);
+    if (event && event.persist) event.persist();
```
taion: we can probably kill this logic of decorating the event

src/Carousel.js
```diff
@@ -63,233 +100,278 @@ const propTypes = {
...
+    const pendingIndex = this._pendingIndex;
+    this._isSliding = false;
+    this._pendingIndex = null;
+
+    if (pendingIndex != null) this.to(pendingIndex);
+    else this.cycle();
+  };
+
+  handleMouseOut = () => {
+    this.cycle();
```
taion: only if we're doing "pause on hover"? otherwise why should mousing out reset the timer?

src/Carousel.js
```diff
@@ -345,48 +432,48 @@ class Carousel extends React.Component {
       ...props
     } = this.props;
 
-    const { previousActiveIndex, direction } = this.state;
+    const {
+      activeIndex,
+      previousActiveIndex,
+      prevClasses,
```
taion: maybe `prevClassName` and `currentClassName` would be better

src/utils/triggerBrowserReflow.js
```diff
@@ -0,0 +1,5 @@
+// reading a dimension prop will cause the browser to recalculate,
```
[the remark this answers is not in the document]
jquense: Yes

src/Carousel.js
jquense: No, when the slides wrap you want to continue in the same direction you we're going. this isn't a perfect check but I think this component is dumb and don't want to spend lots of time on it :P

taion: hmm so that's if `previousActiveIndex == lastPossibleIndex`... except what if the user clicks on the indicator? or if you're cycling back from 0 to final

Issue #3203, opened 2018-07-24
V4-support: Update local setup instructions


abdullahtariq1171:
Local setup instructions aren't updated with the setup for V4-support.

Need a green flag from you guys and I will send a PR. 
@jquense @taion 

https://github.com/react-bootstrap/react-bootstrap/tree/v4-support#local-setup

taion:
Sounds good, thanks!

PR #3204, opened 2018-07-24
Update local docs script name in readme

abdullahtariq1171:
Solves https://github.com/react-bootstrap/react-bootstrap/issues/3203

Also, need input on should we update `yarn run bootstrap` to `yarn run setup`? Consdering the script installs dependencies of both app, the word `setup` sounds better for new developers.

Update readme demo here: https://github.com/abdullahtariq1171/react-bootstrap/tree/v4-support-update-readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # React-Bootstrap [![Travis][build-badge]][build] [![npm][npm-badge]][npm]
 
-[Bootstrap 3][bootstrap] components built with [React][react].
+[Bootstrap 4][bootstrap] components built with [React][react].
 
 [![Codecov][codecov-badge]][codecov]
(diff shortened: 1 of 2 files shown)


taion:
"Bootstrap" is the name of the command in Lerna, though, so there is some traction there

taion:
This looks like it's the same as the current `start` command. I think it'd be sufficient to just update the docs to ran "run `yarn start`".

abdullahtariq1171:
Actually, I just realized that there wasn't any inconsistency, I was actually on a different branch (`master`) when I checked code and had browser opened (`v4-support`).

abdullahtariq1171:
Closing this now :D.

PR #3208, opened 2018-07-27
write dropdown in terms of react-overlays

jquense:


jquense:
GREEEEEEENNNNNNNNNNN

I removed a bunch of tests because they were duplicates or irrelevant at this point

PR #3213, opened 2018-08-03
Fix OverlayTrigger's wrong state when unstable_AsyncMode enabled

KochiyaOcean:
In React 16, when unstable_AsyncMode is enabled, `handleDelayedShow` and `handleDelayedHide` of OverlayTrigger  could be fired before value `this.state.hide` changes, and it will make OverlayTrigger can't show or hide overlay correctly.
This pr fixed the bug by removing state check of `handleDelayedHide` and `handleDelayedHide` .

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -143,7 +143,7 @@ class OverlayTrigger extends React.Component {
       return;
     }
 
-    if (!this.state.show || this._hoverHideDelay != null) {
+    if (this._hoverHideDelay != null) {
       return;
     }
 
@@ -168,7 +168,7 @@ class OverlayTrigger extends React.Component {
       return;
     }
 
-    if (this.state.show || this._hoverShowDelay != null) {
+    if (this._hoverShowDelay != null) {
       return;
     }
 


taion:
Is this safe in all cases, though? If the overlay is shown, then couldn't there be an issue if the overlay gets hidden while this newly added callback is pending?

Perhaps we need a local field to track intended shown state?

KochiyaOcean:
@taion for example, when mouse move over the overlaytrigger quickly, it will fire both `handleDelayedShow` and `handleDelayedHide`, but `handleDelayedHide` is returned because `this.state.show` is not set to `true` yet. So the overlay will stay on the page until mouse move over it again (which fires `handleDelayedHide` again).

taion:
right, i understand why you'd need this, but i'm not sure this doesn't cause other problems

KochiyaOcean:
@taion The patch works fine for me, but maybe needs more test.

taion:
I'm not sure. I wonder if the state machine here should work a little differently. On the other hand, as noted in https://github.com/react-bootstrap/react-bootstrap/pull/3215#discussion_r208426004, this doesn't really match the upstream logic, so... I dunno.

bpas247:
What's the status of this PR? Looks pretty stale atm.

mxschmitt:
closed due no response

PR #3215, opened 2018-08-04
overlays and doc improvements

jquense:
cleans up the overlays to use popper and new react-overlays, but the bulk of the code changes here are DOCS.

written mostly at night from a Hospital after having a baby :P


jquense:
this PR is sort of ridiculous, i'll probably just merge it and issue a bunch of follow up PR's, but @taion do check the overlay's API adjustments, i wanna make sure we cover all the same use-cases well

src/Overlay.js
```diff
@@ -65,34 +66,50 @@ const propTypes = {
...
 
-class Overlay extends React.Component {
-  render() {
-    const { animation, children, ...props } = this.props;
+function wrapRefs(popper) {
+  const {
+    ref,
+    arrowProps: { ref: aRef },
+  } = popper;
+  popper.ref = ref.__wrapped || (ref.__wrapped = r => ref(findDOMNode(r)));
```
taion: could use `Symbol` here... we have `runtime`

src/OverlayTrigger.js
```diff
@@ -1,92 +1,64 @@
...
+    }),
+  ]),
 
-  // FIXME: This should be `defaultShow`.
   /**
    * The initial visibility state of the Overlay. For more nuanced visibility
    * control, consider using the Overlay component directly.
    */
-  defaultOverlayShown: PropTypes.bool,
+  defaultShow: PropTypes.bool,
```
taion: do we want to just uncontrollable this?

src/OverlayTrigger.js
```diff
@@ -203,118 +151,78 @@ class OverlayTrigger extends React.Component {
...
-  handleToggle() {
-    if (this.state.show) {
-      this.hide();
-    } else {
-      this.show();
-    }
-  }
+  handleFocus = e => {
+    const { onFocus } = this.getChildProps();
+    this.handleHide(e);
```
taion: show?

src/OverlayTrigger.js
```diff
@@ -203,118 +151,78 @@ class OverlayTrigger extends React.Component {
...
 
-    this._overlay = this.makeOverlay(overlay, props);
-
-    return cloneElement(child, triggerProps);
+    return (
+      <>
+        <RefHolder ref={this.trigger}>
+          {cloneElement(child, triggerProps)}
+        </RefHolder>
+        <Overlay
```
taion: sweeeeeet

src/OverlayTrigger.js
```diff
@@ -98,97 +70,73 @@ const defaultProps = {
...
-      return;
-    }
+  show() {
+    this.setState({ show: true });
+  }
 
-    if (!this.state.show || this._hoverHideDelay != null) {
-      return;
-    }
+  handleShow = () => {
```
taion: this doesn't quite match the old logic...

in the old logic, if there already was a show timeout pending, then we wouldn't reset the timeout. we'd just... do nothing.

this makes it so that if a "handleShow" triggers while we have a pending show, we reset the timer.

suppose the triggers were "hover" and "click". if you clicked after hovering, the new logic would reset the timeout, which i do not think is what we want

src/OverlayTrigger.js
```diff
@@ -203,118 +151,78 @@ class OverlayTrigger extends React.Component {
...
-        show={this.state.show}
-        onHide={this.handleHide}
-        target={this}
-      >
-        {overlay}
-      </Overlay>
-    );
-  }
+  handleClick = e => {
+    const { onClick } = this.getChildProps();
```
taion: we don't want to use the function chain thing any more?

src/OverlayTrigger.js
```diff
@@ -203,118 +151,78 @@ class OverlayTrigger extends React.Component {
...
-    delete props.delayHide;
-    delete props.defaultOverlayShown;
+    delete props.defaultShow;
 
     const child = React.Children.only(children);
-    const childProps = child.props;
+
     const triggerProps = {};
 
+    let triggers = trigger == null ? [] : [].concat(trigger);
```
taion: why `[].concat(trigger)`? why not `let triggers = trigger || []`?

src/Popover.js
```diff
@@ -22,27 +21,44 @@ const propTypes = {
...
+    style: PropTypes.object,
+  }),
+
+  /** @private */
+  innerRef: PropTypes.any,
+
+  /** @private */
+  scheduleUpdate: PropTypes.func,
+  /** @private */
+  outOfBoundaries: PropTypes.func,
```
taion: just to delete these?

src/Tooltip.js
```diff
@@ -22,80 +21,77 @@ const propTypes = {
...
-  }
+function Tooltip({
+  bsPrefix,
+  innerRef,
+  placement,
+  className,
+  style,
+  children,
+  arrowProps,
+  scheduleUpdate: _,
```
taion: why no `outOfBoundaries` here?

www/resolveHocComponents.js
```diff
@@ -0,0 +1,87 @@
+const { camelCase, upperFirst } = require('lodash');
```
taion: higher-order-component components?

www/src/examples/Modal/Basic.1.js
```diff
@@ -0,0 +1,123 @@
+class Example extends React.Component {
```
taion: basic.1.js?

www/resolveHocComponents.js
jquense: Lol ouch, yes, that is silly

src/OverlayTrigger.js
jquense: That's true, I didn't notice the logic change. Hmm I think I mostly copied this from what upstream does, so I agree our old behavior is seems more correct but it'd be divergent, unless I copied it wrong

src/Overlay.js
jquense: runtime in v7 no longer polyfills by default, there is like a whole separate runtime now for corejs polyfills

Salinn:
Would this fix the issue with overlays on ios mobile where clicking outside won't close when rootClose is set to true?

jquense:
It should fix the iOS issue yes,

jquense:
gonna go for it and maybe get a beta released

Salinn:
@jquense any chance you can comment on this with the beta version? I can give it a shot and let you know if it solves the ios rootClose

PR #3223, opened 2018-08-09
Honor mountOnEnter and mountOnExit when animation=false

vincentjames501:
This PR is more for discussion than anything else. Right now, `mountOnEnter` and `mountOnExit` are misleading as they only work if `animation=true`. If when `animation=false` we simply do a no-op transition, we gain the functionality back. I also have a separate PR #3222 to allow using a custom animation which could be used as a workaround to this.

### DIFF
--- a/src/TabPane.js
+++ b/src/TabPane.js
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import elementType from 'prop-types-extra/lib/elementType';
 import warning from 'warning';
+import Transition from 'react-transition-group/Transition';
 
 import {
   bsClass,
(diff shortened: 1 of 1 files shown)


taion:
I think a more direct approach would be to just render `null` when no transition is specified, but when `unmountOnExit` is set.

jquense:
That is more direct, I don't hate the symmetry of this approach tho. you even get all the callbacks which people are always asking about. Maybe it's worth having like a null transition?

vincentjames501:
Effectively this is a "no-op" transition so that all the callbacks work fine. There may be another approach, but most would involve reimplementing a lot of the functionality of the `Transition` component. I am personally using a custom "no-op" transition in my project and it's been working great and shouldn't be a breaking change for anyone.

taion:
hmm, it would nicely unify the branches... i'd be +0 on a null transition i guess

mxschmitt:
@taion @vincentjames501 whats the state of the PR? looks very stale.

taion:
updated base branch

the v1 betas have the correct behavior, and it's implemented there per https://github.com/react-bootstrap/react-bootstrap/pull/3223#issuecomment-411788273. i think that, if we want this change in the bs3 dev line, we should follow that impl

mxschmitt:
Closed because it's stale

PR #3242, opened 2018-08-21
Try mdx

jquense:
this is amazing, I’m so impressed
 by the whole thing here. also it just works with remark plugins!


dozoisch:
Wow, mdx is really awesome

Issue #2969, opened 2018-01-28
Tooltip remains visible after disabling element


captain-igloo:
See #1432 which has been closed but the problem still exists.  When you click on the button it becomes disabled and the tooltip remains visible.

I am using React 16.2.0 and react-bootstrap 0.32.0.

```javascript
import React, { Component } from 'react';
import { OverlayTrigger, Tooltip, Button } from 'react-bootstrap';

export default class LinkWithTooltip extends Component {

    constructor() {
        super();
        this.onClick = this.onClick.bind(this);
        this.state = {
            disabled: false,
        };
    }

    onClick() {
        this.setState({
            disabled: true,
        });
    }

    render() {
        const tooltip = <Tooltip id="tooltip">tooltip</Tooltip>;

        return (
            <OverlayTrigger overlay={tooltip}>
                <Button
                    onClick={this.onClick}
                    disabled={this.state.disabled}
                >
                    click me
                </Button>
            </OverlayTrigger>
        );

    }

}
```

taion:
Is this another side-effect of our disabling pointer-effects?

jquense:
i don't think so the `pointer-events` thing merely mimics the behavior of a native `button` element, and only for anchor elements which the example above isn't rendering right?

murphyke:
I'm seeing this also, react-bootstrap 0.32.1, React 16.4.0.

keul:
I'm working on this, see https://github.com/react-bootstrap/react-bootstrap/pull/3251

The solution if the PR is not the best probably, as at every click on the OverlayTrigger container the tooltip will be closed. This is probably acceptable for buttons, but what about form elements?

However: another solution is not that easy: I think this is happening because browsers stop triggering _onmouse-something-like_ events when an element became disabled, so there's no easy way to track this kind of change.

Any suggestion about a different approach is welcome.

keul:
Some other hint about this is mostly a browser behavior: https://developer.mozilla.org/en-US/docs/Web/Events/mouseleave#Browser_compatibility

* Issue is not hitting Firefox 61
* Issue can be reproduced on Chrome 68 (see below)

![react-bootstrap-2969](https://user-images.githubusercontent.com/284924/44838080-93f2a200-ac3b-11e8-98e8-a30a158bd032.gif)

keul:
[...]

PR #3251, opened 2018-08-29
Force hide tooltip when clicking on OverlayTrigger

keul:
This prevent tooltip to remain open if hover area became disabled.

See discussion at #2969


taion:
I don't think this quite makes sense to unconditionally hide the tooltip on any click.

keul:
@taion do you have any alternative suggestion? If my analysis at #2969 is right is not simple to solve this as we are hitting a browser behavior.

mxschmitt:
closed due inactivity.

jakeNiemiec:
@mxschmitt #2969 Can also be closed then.

PR #3256, opened 2018-09-03
Update babel to 7.0.0 - fixes #3255

### LINKED ISSUE #3255: Update to Babel 7.0.0
When building a project that uses babel 7 and react-bootstrap, 

We get a bunch of these errors : 

```
ERROR in ./node_modules/react-bootstrap/es/CarouselCaption.js
Module not found: Error: Can't resolve '@babel/runtime/helpers/es6/objectWithoutProperties' in '/Users/....../node_modules/react-bootstrap/es'
 @ ./node_modules/react-bootstrap/es/CarouselCaption.js 2:0-90 30:16-40
 @ ./node_modules/react-bootstrap/es/Carousel.js
 @ ./node_modules/react-bootstrap/es/index.js
 @ ./src/templates/Offre/index.js
 @ ./.cache/sync-requires.js
 @ ./.cache/app.js
 @ multi ./node_modules/react-hot-loader/patch.js (webpack)-hot-middleware/client.js?path=http://localhost:8000/__webpack_hmr&reload=true&overlay=false ./.cache/app
```

This is caused by babel who changed their helpers path from @babel/runtime/helpers/es6/ to @babel/runtime/helpers/esm/ :(

I'm proposing to change it to use babel 7.0.0

flegall:
I'm proposing to use Babel 7 final in order to fix the @babel/runtime helpers path issues.

I'm not so sure if releasing to master is the correct branch. I'd like to release this in a 0.32.4 version, if that's possible.

Feel free to comment, recommend changes, I've just updated package.json & yarn.lock.

Yarn lint does not pass as it currently does not work on master, I'm having a lot of import/no-unresolved.

However : 
- yarn test-browser, yarn test-node & yarn build pass 
- the produced code includes correct path to @babel/runtime helpers


taion:
Let's use caret ranges now that it's final. We should also be using `@babel/runtime-corejs2` instead of `@babel/runtime` I believe.

taion:
(With appropriate configuration of the transform.)

jquense:
I think ultimately we want to not include all the core-js stuff tho right? Let folks polyfill as appropriate?

taion:
hmm do we? idk i like having libraries that i can just install. but then again not including dupe polyfills is nice. your call

taion:
though if we dropped the polyfills on this branch, that would really be breaking, so we shouldn't do that on a non-breaking release

flegall:
Ok, it seems like I have to update gatsby to the latest v2 rc as well.
I'll try doing it.

About the @babel/runtime-corejs2  I'm not sure I fully understand the reason why it is needed on master, as it only includes @babel/runtime - https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/package.json#L120

Did @babel/runtime used to bring those polyfills earlier in beta versions ?

jquense:
> Did @babel/runtime used to bring those polyfills earlier in beta versions

yes, `@babel/runtime` is now doesn't polyfill it's own code, and `@babel/runtime-corejs2` is essentially the same as the old runtime

flegall:
I'm discovering the webpack + babel setup you made for integrating gatsby and keeping the gatsby site in sync with the lib code, it's awesome ! :)

I'll try to migrate that to gatsby v2 final apis !

jquense:
@flegall FYI i'm working to upgrade the v4-support branch to the latest gatsby versions as well, and there is a gatsby bug that prevents it from working right. I'll post the branch i have so you can see my workarounds

flegall:
Ah yes, I'd definitely have an eye, I'll see if I can help.

I've actually found the @babel/runtime helpers path errors while migrating a site (that uses react-bootstrap) to gatsby v2, I've successfully migrated to v2 rc1, but since rc2, the gatsby team have moved to babel 7.0.0 final...

jquense:
https://github.com/react-bootstrap/react-bootstrap/blob/ff2e719641cd8638ead47395f3bf6c16edb8d222/www/gatsby-node.js#L39-L47 here is the key lines, but check out the rest of the commit (it's a bit busy)

flegall:
Ok, this is a lot more than just bumping to the latest babel :)

I think I will let you finish the work on v4 as the v4 branch is already on a more advanced gatsby version than master.

PR #3277, opened 2018-09-17
Trim fat

jquense:


jquense:
no proptypes!

jquense:
no prop-types!

src/AbstractNav.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import qsa from 'dom-helpers/query/querySelectorAll';
 import PropTypes from 'prop-types';
-import elementType from 'prop-types-extra/lib/elementType';
+import { elementType } from 'prop-types-extra';
```
jquense: switched to this style in order to have the import removed entirely for prod

RKushnir:
Could you please release a new version with this change, as we're getting a warning because of the type mismatch for `outOfBoundaries`?

PR #3301, opened 2018-10-03
Specify React 16.3.0 as a peer dependency

davegw:
v1 uses the React context API, which was introduced in React v16.3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,8 +54,8 @@
   "author": "Stephen J. Collings <stevoland@gmail.com>",
   "license": "MIT",
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -54,7 +54,7 @@
   "author": "Stephen J. Collings <stevoland@gmail.com>",
   "license": "MIT",
   "peerDependencies": {
-    "react": "^0.14.9 || >=15.3.0",
+    "react": ">=16.3.0",
     "react-dom": "^0.14.9 || >=15.3.0"
```
taion: need the same change here i think

davegw: Good call! Updated.

PR #3338, opened 2018-10-18
Update DropdownMenu.js

jquense:


src/DropdownMenu.js
```diff
@@ -110,7 +110,7 @@ class DropdownMenu extends React.Component {
                     className,
                     bsPrefix,
                     show && 'show',
-                    alignRight && `${bsPrefix}-right`,
+                    alignEnd && `${bsPrefix}-right`,
```
taion: no extra intelligence needed here for left/right positioning?

jquense: There is via popper, but this is an additional class, presumably to reset styles appropriately for that or the navbar where popper isn't used.

PR #3339, opened 2018-10-19
feat: change default Nav components to non ul/li

jquense:
BREAKING CHANGE: components no longer default to a list

### DIFF
--- a/src/Nav.js
+++ b/src/Nav.js
@@ -94,7 +94,7 @@ class Nav extends React.Component {
   static defaultProps = {
     justify: false,
     fill: false,
-    as: 'ul',
+    as: 'nav',
   };
 
   render() {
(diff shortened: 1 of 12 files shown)


taion:
btw, with nesting `nav` in `nav` with `<Navbar>` and `<Nav>`, will anything bad happen?

jquense:
o shoot does navbar render a nav? this should probably be a plain `div` then...

taion:
by default yeah https://github.com/react-bootstrap/react-bootstrap/blob/84ef393827ba79c4b7ffba38ad733cdd5e661fea/src/Navbar.js#L120

jquense:
we need to figure out why FF keeps failing in tests...

jquense:
also fixes #3360

PR #3341, opened 2018-10-21
Nit: Fix grammar (it's vs its)

mikeyhew:
On the home page, replace "it's", the contraction of "it is", with "its", the possesive form.

Unfortunately the syntax highlighting on GitHub is still messed up — maybe changing the file extension to `.jsx` would fix that.


jquense:
Thanks!

Issue #3328, opened 2018-10-13
OverlayTrigger with Popover in overlay throw Warning: Failed prop type: Invalid prop `outOfBoundaries` ...


Daarien:
Warning: Failed prop type: Invalid prop `outOfBoundaries` of type `boolean` supplied to `Popover`, expected `function`.
    in Popover (created by Context.Consumer)
    in Transition (created by Fade)
    in Fade (created by InnerPopper)
    in InnerPopper (created by Context.Consumer)
    in Popper (created by Overlay)
    in WaitForContainer (created by Portal)
    in Portal (created by Overlay)
    in Overlay (created by WaitForContainer)
    in WaitForContainer (created by ForwardRef(withContainer(Overlay)))
    in Overlay (created by OverlayTrigger)

smallstepstoday:
I created a [repo](https://github.com/smallstepstoday/popover.git) that corroborates the error.

SailingSteve:
I just dropped the example code onto my page and reproduced the problem.  
https://react-bootstrap.netlify.com/components/overlays/#overlay-trigger    

Another minor issue about that demo code is that it uses `placement="right-start"` which I don't think is correct anymore.

PR #3342, opened 2018-10-21
fix Popover outOfBoundaries proptype

tylerreichle:
Popover's `outOfBoundaries` proptype should be boolean. (Proptype currently bool in master)

Reference:
 - [React Popper doc](https://github.com/FezVrasta/react-popper)
 - [react-popper/Popper.js](https://github.com/FezVrasta/react-popper/blob/4e35c6f83242aed0397e5d66c7f8ca17475f1977/src/Popper.js)

fixes issue: #3328

### DIFF
--- a/src/Popover.js
+++ b/src/Popover.js
@@ -58,7 +58,7 @@ const propTypes = {
   /** @private */
   scheduleUpdate: PropTypes.func,
   /** @private */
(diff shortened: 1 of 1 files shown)


imsergiobernal:
Hello I was looking for an easy issue. Why tabs testing is failing? I don't see any relationship between them

Issue #3047, opened 2018-03-05
v4: Custom form inputs (checkbox, select, radio, file)


jquense:
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/components/forms/#custom-forms)

- [x] Checkbox
- [x] Radio
- [x] File input
- [x] Select
- [x] Range

AnderssonChristian:
I can take this.

AnderssonChristian:
@jquense Looking at the official bs4 docs, all examples have the input-label association done with the `for` and `id` attributes. The other method of wrapping each input inside a label isn't documented and doesn't seem to work. Unless I am missing something, this is really a shame since it would be easier to implement that approach with React. 

Now we would need some creative way of generating a unique id. Ideas?

jquense:
Do you have any specific examples that you mean? Checkboxes and Radios are semantically valid inside of a label but i don't know if the styles work.

taion:
I think we should either have a component that just corresponds to e.g. `.form-check` and handles rendering both the `.form-check-input` and the `.form-check-label`, or else have the component corresponding to `.form-check` work like the current `<FormGroup>` as a context provider that takes a `controlId` (or perhaps better `inputId`) prop.

jquense:
I think in this case we mean the custom inputs, which need a different markup than the normal `form-check` and input case :/ https://getbootstrap.com/docs/4.0/components/forms/#checkboxes

I'd say in this cases we should just accept an id, and thread it to both the label and input

jquense:
ugh i ddin't even realize the normal inputs do this as well, sorry @taion we're on the same page :P

taion:
i mean this is good, this means that checkboxes are more like normal form groups... just need a slightly different API is all... `<FormCheck>`, `<FormCheck.Input>`, and `<FormCheck.Label>` perhaps?

jquense:
well sort of? the label is still in a different place. I've actually already started the normal form inputs @AnderssonChristian (not the custom style ones noted here tho!)

maybe I sohuld push that branch so folks can see what i was thinking?

taion:
We could do something cute with functional components, e.g.

```js
const normalFormCheck = (
[...]

PR #3343, opened 2018-10-22
Add support for custom checkboxes and radios

eladlevy:
Implemented custom checkboxes and radios mode as discussed in [3047](https://github.com/react-bootstrap/react-bootstrap/issues/3047).
Includes tests and documentation

image

### DIFF
--- a/src/FormCheck.js
+++ b/src/FormCheck.js
@@ -45,6 +45,9 @@ class FormCheck extends React.Component {
     title: PropTypes.string,
     label: PropTypes.node,
 
+    /** Use Bootstrap's custom form elements to replace the browser defaults */
+    custom: PropTypes.bool,
+
     /** The type of checkable. */
     type: PropTypes.oneOf(['radio', 'checkbox']).isRequired,
 
@@ -83,6 +86,7 @@ class FormCheck extends React.Component {
       type,
       label,
       children,
+      custom,
       ...props
     } = this.props;
 
@@ -102,14 +106,18 @@ class FormCheck extends React.Component {
 
     return (
       <FormContext.Transform
-        mapToValue={({ controlId }) => ({ controlId: id || controlId })}
+        mapToValue={({ controlId }) => ({
+          controlId: id || controlId,
+          custom,
+        })}
       >
         <div
           style={style}
           className={classNames(
             className,
-            bsPrefix,
-            inline && `${bsPrefix}-inline`,
+            !custom && bsPrefix,
+            custom && `custom-control custom-${type}`,
+            inline && `${custom ? 'custom-control' : bsPrefix}-inline`,
           )}
         >
           {children || (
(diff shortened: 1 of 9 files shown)


src/FormCheck.js
```diff
@@ -93,6 +97,7 @@ class FormCheck extends React.Component {
         {...props}
         type={type}
         ref={inputRef}
+        custom={custom}
```
jquense: I think may make sense to put `custom` on the Context object below with controlId, that way when using the individual components you don't need to explicitly pass `custom` to the input and label

www/src/pages/components/forms.js
```diff
@@ -235,6 +237,56 @@ export default withLayout(function FormControlsSection({ data }) {
...
+        completely custom form elements to replace the browser defaults. They’re
+        built on top of semantic and accessible markup, so they’re solid
+        replacements for any default form control.
+      </p>
+      <LinkedHeading h="3" id="forms-custom-checkboxes-and-radios">
+        Checkboxes and radios
+      </LinkedHeading>
+      <p>
+        Each checkbox and radio is wrapped in a <code>{'<div>'}</code> with a
+        sibling <code>FormCheckInput</code> to create our custom control and a{' '}
```
jquense: I would shorten this all to something like:

> Custom checkbox and radio styles are achieved with a resourceful use of the `:checked` selector and `::after` psuedo elements, but are Structurally similar to the default `FormCheck`.  By default the checked and indeterminate icons use embedded svg icons from Open Iconic...

src/FormCheck.js
eladlevy: Great feedback! I wasn't aware of the Context option. This helps to reduce code and makes the customize render option a lot easier to use.

taion:
It seems like this is slightly easier than adding e.g. `<CustomFormCheck.Input>`. That said this seems _almost_ like something `bsPrefix` can handle, though.

taion:
gah i clicked the wrong button

i meant _if_ we had support for `bsPrefix` from `FormCheck` cascading down into `FormCheck.Input`, then had some special-ish (?) logic to also add `custom-checkbox` or whatever if the bsPrefix is `custom-control`.

is that too magical? probably? (though i like having that sort of `bsPrefix` semantics)

in which case i think a `custom` prop is good

eladlevy:
@taion hmm I can try overriding bsPrefix in the render method. I thought about something like:
```javascript
  const activeBsPrefix = custom ? 'custom-control' : bsPrefix;
   <div
          style={style}
          className={classNames(
            className,
            activeBsPrefix,
            custom && `custom-${type}`,
            inline && `${bsPrefix}-inline`,
          )}
    >
```

Is there a better way to override `bsPrefix` ?
It does saves a bit of logic but if a user passes a custom `bsPrefix` it will override his prop and may cause unexpected behaviour.
WDTH?

taion:
no, i was sort of speculating that a "neat" API would be to do something like `bsPrefix="custom-control"`. but i think that's a little too cute.

i'm okay with this API

eladlevy:
So... @jquense  Do you think this PR can be merged?

eladlevy:
Sure. I've rebased latest master on top of my branch so no dirty merge commits will appear.

PR #3345, opened 2018-10-23
Update index.js / Fix grammar (stratch -> scratch)

jonisar:
Fixed to "built from scratch" instead of "stratch", which is a grammar mistake in this context.

Cheers :)


jquense:
Thanks!

Issue #27, opened 2014-02-07
Use Tether in place of Bootstrap overlays?


syllog1sm:
Overlays are difficult to get right, and Bootstrap doesn't do a perfect job. Their overlays mostly work "okay enough", but there are various weirdnesses.

For instance, you have to add special attributes to buttons that carry popovers and tooltips if they are within button groups: http://getbootstrap.com/javascript/#popovers

I also encounter bugs from the fact that hidden popovers are opacity: 0 in Bootstrap, instead of display: none. This means an inactive popover can "occlude" interface items.

This library claims to do overlays right: http://tether.io , and I think their approach looks very solid.

There's a real value for this project keeping all-and-only-Bootstrap. But there's a cost, too, to locking ourselves in to parts which could be better. I think overlays will be consistently painful, especially because in React you really want to make the overlay a DOM sibling to the trigger, which causes further problems.

How have you guys found Bootstrap's overlay library? Have you also had problems with it?

PR #3346, opened 2018-10-23
fix typo in README.md at #27 from

Nelson-Chinedu:
Yarn is the our package manager of choice
-to-
Yarn is our package manager of choice


jquense:
thanks!

PR #3350, opened 2018-10-25
Nav: Validate children in keyboard navigation handler

vojty:
This fixes situation when Nav children are added conditionally


vojty:
could you release patch version please?

vojty:
bump

PR #3368, opened 2018-11-11
Pass Context.Consumer to context toolbox

nathanstitt:
As outlined in https://github.com/4Catalyzer/react-context-toolbox/issues/9, context toolbox needs to only use `Context.Consumer` since it can't navigate to the `Consumer` property reliably.


nathanstitt:
Right you are @taion - thanks for noticing and setting me straight.  That's what I get for coding on weekend evening.  I'll update PR shortly

PR #3377, opened 2018-11-20
Update Getting Started Installation Steps

### LINKED ISSUE #3367: Getting Started missing important step for create-react-app
I use create-react-app to set up a new React app. When enabling react-bootstrap for my app, I need to do two steps. 
1. `yarn add react-bootstrap@next bootstrap`
2. In the `src/index.js` file, add `import 'bootstrap/dist/css/bootstrap.min.css';` as the first line

Step 1 is covered by the Getting Started page, but step 2 is missing. It may be obvious for seasoned web devs, but it will help for less seasoned ones if it is also covered.

bpas247:
Fixes #3367

### DIFF
--- a/www/src/pages/getting-started/introduction.mdx
+++ b/www/src/pages/getting-started/introduction.mdx
@@ -26,6 +26,13 @@ to use a CDN for the stylesheet, it may helpful to install
 npm install react-bootstrap@next bootstrap
 ```
 
+In order for react-bootstrap components to be styled correctly, bootstrap's css must be imported 
+somewhere in your project (usually in the entry point of your program).
+
+```js
+import 'bootstrap/dist/css/bootstrap.min.css';
+```
+
 ### Importing
 
 You should import individual components from


jquense:
Hey thanks for the PR, we do already have this info in the docs https://react-bootstrap.netlify.com/getting-started/introduction#stylesheets 

I'm not sure we want to tell people to definitely require the bootstrap.css in this way because: 

1. Often that's not what you want to do, it's better to construct a more focused style build with Sass.
2. If you do want to include all of bootstrap, requiring it in the app root is sub-optimal. you should link to a CDN version in that case to increase cache hits for the site viewer

It may be worth stating the above in the docs tho we try not to overwhelm in the getting started section. In all that's why we've chosen to link to bootstraps already excellent docs on the subject.

bpas247:
So the best course of action might be to leave the docs as is. I'm gonna close this PR for now

jquense:
Ya, i mean folks are clearly getting confused so something isn't ideal! i'm not sure tho what the correct advice is for CRA consumers specifically, since i'm not sure how you can even provide a link in the html without ejecting

PR #3380, opened 2018-11-21
Remove Navigation Role from Breadcrumb

### LINKED ISSUE #3371: Breadcrumbs should not have role of navigation
The Breadcrumb component is currently adding a role of "navigation" to the ordered list. This is not necessary and actually hurts accessibility since it takes away the default behavior of the ol. 

Deque suggests removing it from the ol: "role navigation is not allowed for given element"

The [WAI-ARIA breadcrumb](https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html) example also does not include a role.

bpas247:
Fixes #3371

### DIFF
--- a/src/Breadcrumb.js
+++ b/src/Breadcrumb.js
@@ -45,7 +45,6 @@ class Breadcrumb extends React.Component {
     return (
       <Component aria-label={label} className={className} {...props}>
         <ol
-          role="navigation"
           {...listProps}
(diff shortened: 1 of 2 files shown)


bpas247:
Should anything else be done with this PR? Any docs that need updating?

taion:
oh oops @jquense 	didn't see that you didn't approve this

PR #3388, opened 2018-12-03
Added typescript v0.32 types with some initial updates for v1

rathga:
Starting point for updating the typescript types

### DIFF
--- a/types/LICENSE
+++ b/types/LICENSE
@@ -1,0 +1,21 @@
+    MIT License
+
+    Copyright (c) Microsoft Corporation. All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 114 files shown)


jquense:
awesome thanks 👍 

I'm ok merging for now. we can iterate

types/LICENSE
```diff
@@ -0,0 +1,21 @@
+    MIT License
+
+    Copyright (c) Microsoft Corporation. All rights reserved.
```
taion: copyright microsoft?

taion: well, it does have to be MIT: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/LICENSE

but it's not (c) microsoft

Gillariel:
Hi,
How can we update the types to reflect thoses changes ? 
I'm currently using v1 and the last types package but import Container in a ts file is still not recognize :/
Should be a minor thing but i'm stuck ...

rathga:
> Hi,
> How can we update the types to reflect thoses changes ?
> I'm currently using v1 and the last types package but import Container in a ts file is still not recognize :/
> Should be a minor thing but i'm stuck ...

Are you using the v0.32 types from DT or this new types folder?  Container is new in v1 so you need the new types (in v.032 it was Grid).

Gillariel:
> Are you using the v0.32 types from DT or this new types folder? Container is new in v1 so you need the new types (in v.032 it was Grid).

Well that's the question, I'm using "@types/react-bootstrap": "^0.32.15" as type pâckages and "react-bootstrap": "^1.0.0-beta.3" so how does I import this folder? Does I need to manually reference it and import it as a part of my project itself? (surely a npm newbie question, sorry)

jquense:
We haven't released these types yet, but once we do you can drop the definitely Typed package and the types will automatically be used from here

rathga:
@Gillariel Until then, the simplest way to use the new types right now would be to download this repo and then manually paste the /types folder into your /node_modules/@types/react-bootstrap.  Do not install the old @types/react-boostrap package from npm

Please note there are only a handful of fixes vs v0.32 so far though!  So your mileage may vary.  

@jquense To support what you say above, once we get a little further on I think we need to move the index.d.ts into the root directory so the new types get automatically picked up.  I was not sure if you wanted this yet given the limited progress so far.  Therefore the index.d.ts remains in the /types folder for now and people will need to manually reference the new types if they want to work with them

jquense:
yup there is more discussion in #3392 we may try and just migrate stuff to typescript straight up as well

Gillariel:
@rathga @jquense 
Thank you for your answer :)

PR #3392, opened 2018-12-11
WIP types, ongoing work

jquense:


types/components/ThemeProvider.d.ts
```diff
@@ -0,0 +1,42 @@
...
+interface InferrableComponentProps<TRenderedType extends ElementType> {
+  as: TRenderedType;
+}
+
+export interface BootstrapProps<TRenderedType extends ElementType>
+  extends InferrableComponentProps<TRenderedType> {
+  bsPrefix?: string;
+}
+
+// declare class BootstrapComponent<
```
taion: ?

jquense: Trying to figure out d.ts limitations

taion: btw, i'm a little curious how good incremental typing support in TS is. given the extensiveness of the type defs, should we just use TS here full stop?

taion: worth a try i think... esp if it works component-by-component

glenjamin:
FYI I'm aiming to start a new side project where I wanted to use typescript and react-bootstrap, so I decided I'd start with some massive yak shaving and update the type definitions for 1.0.

I'll post up a PR when I have something properly worth looking at, but progress for now is at https://github.com/glenjamin/react-bootstrap/tree/types

I initially started trying to use really simple types to make sure the project is usable in typescript again, then I went down a similar rabbithole to @jquense when I understood what he was trying to do to get `<Button as>` to typecheck fully correctly. I think I'm going to go back to the original plan of trying to get a decent first pass of just the `react-bootstrap` props, and continue to allow any HTML props to be passed through, regardless of the underlying component.

jquense:
Awesome, happy for some help. I too think the simple first pass is probably the right way to go.

glenjamin:
I've now added #3411 with a load of types in it

PR #3397, opened 2018-12-13
fix(form-control): make inputRef backward-compatible with React creat…

OlegLustenko:
Hi, I faced an issue with the validation of props for FormControl especially with `inputRef`, if I suddenly passing down a `React.createRef` interface to it there are a warning in console.

There a small fix, hope you could find time to merge it :)

That PR should go to 0.32.4 tag and possibly published as 0.32.5. Thanks for your assistance

### DIFF
--- a/src/FormControl.js
+++ b/src/FormControl.js
@@ -32,7 +32,7 @@ const propTypes = {
    * <FormControl inputRef={ref => { this.input = ref; }} />
    * ```
    */
-  inputRef: PropTypes.func
+  inputRef: PropTypes.any,
 };
 
 const defaultProps = {


src/FormControl.js
```diff
@@ -32,7 +32,10 @@ const propTypes = {
    * <FormControl inputRef={ref => { this.input = ref; }} />
    * ```
    */
-  inputRef: PropTypes.func
+  inputRef: PropTypes.oneOfType([
```
jquense: let's just go with `any` maybe

taion: no exposed type for those ref things, eh?

jquense:
this should be handled in v1 automatically!

Issue #3386, opened 2018-11-30
How to achieve Panel Group accordion functionality using 1.0.0-beta.3


jainvizz:
I am using react-bootstrap 1.0.0-beta.3, which is build for supporting bootstrap 4 update.

Before this I was using react-bootstrap 0.32.1 and created Accordion using Panels and Panel group.

But after bootstrap upgrade it was suggested to Card component. I tried to achieve the same behavior like this:

`<CardGroup>
 <Card eventKey={this.state.eventKey} className="border-0">
              <Card.Header>
                <div className="row">
                  <div className="col-xs-9 col-sm-9 col-md-9 col-lg-9">
                    <Card.Title>
                      This is test
                    </Card.Title>
                  </div>
                  <div className="col-xs-3 col-sm-3 col-md-3 col-lg-3">
                    Test Text 123
                  </div>
                </div>
              </Card.Header>
              <Card.Body>
                Test Text 456
              </Card.Body>
           </Card>
</CardGroup>`

Please help me out with this implementation.

jquense:
At the moment there is no accordian behaviour out of the box with cards. We just haven't built it yet.

You can add it manually with the Collapse component just fine but there isn't a shortcut component right now

jainvizz:
Thanks @jquense for the quick reply.

I have checked collapse component but that only work for single card, can you please tell me how to achieve accordion behavior with the same.

Like when I open one card it should close others.

jquense:
You have to wrap each card in a collapse and control the open state for each of them yourself to achieve the effect. That's how we'd implement it internally at least

bpas247:
> At the moment there is no accordian behaviour out of the box with cards. We just haven't built it yet.

Is this something we should add support for? I can probably start work on a PR for this, if it's something we would like to add.

jquense:
We should add it, but I'm not sure how best to implement it. Not very happy with the way panel group works (or it's complexity) in previous versions so didn't want to just copy it

bpas247:
Looking at the previous implementation of PanelGroup right now, I see what you mean by being unhappy wit its complexity.

> You have to wrap each card in a collapse and control the open state for each of them yourself to achieve the effect. That's how we'd implement it internally at least

[...]

PR #3409, opened 2019-01-10
Add Accordion

bpas247:
Relevant to #3386, this PR is meant to add an Accordion component to the library (implementation of [this](https://getbootstrap.com/docs/4.1/components/collapse/#accordion-example) bootstrap example)

TODOs:
~~Remove hard-coded value for `getDimensionValue`.~~
~~Write Storybook stories to ensure the Accordion works visually~~
- [x] Write more functional tests to ensure it works as expected
- [x] Ensure that Card components that are passed in as children still have the full functionality of being a Card component
- [x] Add docs for using Accordion
- [ ] Typescript defintions

Any suggestions and/or criticisms are welcome, I'm currently at a block regarding how to fully implement this component.


src/Accordion.js
```diff
@@ -0,0 +1,79 @@
...
+    children: PropTypes.instanceOf(Object),
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      openedId: 0,
+    };
+    this.onClick = this.onClick.bind(this);
```
taion:
```suggestion
```

the class prop syntax below already binds it

src/Accordion.js
```diff
@@ -0,0 +1,79 @@
...
+    this.setState({
+      openedId: id,
+    });
+  };
+
+  render() {
+    const { children } = this.props;
+    return (
+      <div>
+        {React.Children.map(children, (CurCard, i) => (
```
taion: I'm not sure about this approach. Perhaps using some sort of AccordionContext might be better?

bpas247: This is where I ultimately got stuck. Any suggestion for how I should implement `AccordionContext`? I might take a look at how other `...Context` components were implemented

taion: i think the idea is to check for the context inside the card header, and interact with it if present

we use similar patterns around navs and tabs

jquense: @bpas247 yeah I think the API might be something like 

```jsx
<Accordion>
  <Card>
	<Card.Header>
	  <Accordion.Toggle eventKey="1">
		<Button >Click me</Button>
	  </Accordion.Toggle>
 	</Card.Header>
	<Collapse eventKey="1">
		{/* body etc blah blah */}
	</Collapse>
  </Card>
</Accordion>
```

It's verbose but I think probably the right starting point as a raw API

bpas247: I'll go with that approach @jquense :+1: I'm hoping I can get that working by this weekend

bpas247:
Current update: There is a working version of the Accordion component available in the docs on this PR. Please feel free to try it out and let me know what y'all think :smiley:

src/AccordionCollapse.js
```diff
@@ -0,0 +1,37 @@
...
+    eventKey: PropTypes.string.isRequired,
+  };
+
+  render() {
+    const { bsPrefix, children, className, eventKey, ...props } = this.props;
+
+    return (
+      <AccordionContext.Consumer>
+        {context => (
+          <Collapse in={context.activeEventKey === eventKey} {...props}>
```
taion: this kind of pattern has historically been problematic for us in the past. e.g. we're not passing `style` down to the wrapped child

src/AccordionToggle.js
```diff
@@ -0,0 +1,72 @@
...
+     */
+    onClick: PropTypes.func,
+  };
+
+  static defaultProps = {
+    as: 'button',
+  };
+
+  handleClick = () => {
+    const handleClick = this.accordionContext;
```
taion: maybe call this `accordionContextOnClick` or something. or just bind `onClick` inside `render`?

taion:
seems broadly fine. @jquense is this consistent with implementation elsewhere?

src/AccordionCollapse.js
jquense: In this case Context is just passing props through to the `div` and rendering nothing. Agree it's not great tho. It's more of a convenience to not have to split out transition specific props, but we should have better pattern for that.

src/AccordionToggle.js
```diff
@@ -0,0 +1,72 @@
...
+      eventKey,
+      ...props
+    } = this.props;
+
+    if (Component === 'button') props.type = 'button';
+
+    return (
+      <AccordionContext.Consumer>
+        {context => {
+          this.accordionContext = context.onClick;
```
jquense: this isn't safe to do apparently? I'm not sure why but somethin, something, concurrent mode. I think it's fine here for now. We'll refactor to hooks later :P

bpas247: I was planning on migrating to use hooks when dependencies get updated anyways, so I'll keep that in mind for future reference :ok_hand:

src/AccordionCollapse.js
bpas247: is there a better pattern I should look into? I'm not sure as to what other approach I should go with

jquense: I think this ok. It's the underlying Collapse component that's the problem

src/AccordionToggle.js
```diff
@@ -0,0 +1,71 @@
...
+    const {
+      as: Component,
+      bsPrefix,
+      children,
+      className,
+      eventKey,
+      ...props
+    } = this.props;
+
+    if (Component === 'button') props.type = 'button';
```
jquense: I think instead of introspecting `Component` it's better to set the default prop to something like `props => <button type="button {...props}/>` That said I'm not sure Toggle should render anything itself...I think maybe it should just inject props into `children` or have a render function

bpas247: Meaning the Toggle component should just pass onClick and other props to its children, such as this example?
```jsx
<Accordion.Toggle>
  <button>click me</button> 
</Accordion.Toggle>
```
or Should I make my own `<Accordion.Button>` that grabs onClick from the `Accordion` context?

jquense: I wouldn't add an AccordionButton, the above tho would work. Maybe in the in near future we also provide a `useAccordionToggle` hook to be cool

bpas247: to clarify, should I pass the `onClick` function to its children via context? I'm not sure how I would determine which child gets the function from context though...

src/Accordion.js
```diff
@@ -0,0 +1,53 @@
...
+
+const defaultProps = {
+  as: 'div',
+};
+
+const Accordion = React.forwardRef(
+  ({ as: Component, bsPrefix, children, className, ...props }, ref) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'accordion');
+
+    const [openedId, setOpenedId] = useState('0');
```
taion: should we use `uncontrollable` here and have something like `defaultActiveKey`? we should have some capacity for the accordion to start in a "fully collapsed" state

src/Accordion.js
```diff
@@ -0,0 +1,53 @@
...
+import { elementType } from 'prop-types-extra';
+import { useBootstrapPrefix } from './ThemeProvider';
+import AccordionContext from './AccordionContext';
+import AccordionToggle from './AccordionToggle';
+import SelectableContext from './SelectableContext';
+import AccordionCollapse from './AccordionCollapse';
+
+const propTypes = {
+  /** Set a custom element for this component. */
+  as: elementType,
```
taion: can we use elementType from `prop-types` proper?

src/AccordionCollapse.js
```diff
@@ -0,0 +1,36 @@
...
+  bsPrefix: PropTypes.string,
+  /**
+   * A key that corresponds to the toggler that triggers this collapse's expand or collapse.
+   */
+  eventKey: PropTypes.string.isRequired,
+};
+
+const AccordionCollapse = React.forwardRef(
+  ({ bsPrefix, children, className, eventKey, ...props }, ref) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'accordion-collapse');
```
taion: i don't think this is actually a class in bootstrap. perhaps this only needs to handle collapsing, per https://getbootstrap.com/docs/4.3/components/collapse/#accordion-example ?

src/AccordionToggle.js
```diff
@@ -0,0 +1,69 @@
...
+        {children &&
+          React.Children.map(children, child => {
+            if (child.props && child.props.type === 'button') {
+              // grab the old click
+              let givenClick = child.props.onClick;
+              let newOnClick = () => {
+                context.onClick(eventKey);
+                if (givenClick) givenClick(eventKey);
+              };
+              return React.cloneElement(child, { onClick: newOnClick });
```
taion: why do we both inject `onClick` into the child _and_ render our own component (the default `div`) here? probably just one or the other is enough?

src/AccordionToggle.js
```diff
@@ -0,0 +1,69 @@
...
+const defaultProps = {
+  as: 'div',
+};
+
+const AccordionToggle = React.forwardRef(
+  (
+    { as: Component, bsPrefix, children, className, eventKey, ...props },
+    ref,
+  ) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'accordion-toggler');
```
taion: this isn't a bootstrap CSS class i don't think

bpas247: hmm, what bootstrap CSS class should I apply to this component? I wasn't sure which one would make the most sense for what it's trying to do.

src/AccordionCollapse.js
bpas247: So don't apply any new classes? or would we still want to apply `collapse` as one of the classes?

src/AccordionToggle.js
bpas247: Seems like just injecting `onClick` should be efficient enough, yeah. Should I just wrap `children` in a fragment inside `render()` then?

src/AccordionToggle.js
taion: i'm not sure there is one – do you see one in the docs?

src/AccordionToggle.js
taion: or i'd say just render the wrapping component maybe?

src/AccordionToggle.js
bpas247: Looks like there's only `accordion` that's specific to this component, judging by the accordion example on the bootstrap docs. I don't think I should apply any bootstrap CSS class for this component.

taion: yup – so this can just render the `<a>` or whatever

gurkerl83:
What is the status of this component? The fact that this merge request touches no other parts of react-bootstrap means the merge should be uncritical. Also noticed, there is no typescript definition for accordion available.

src/AccordionCollapse.js
jquense: As a rule we don't creat new classes that don't exist in bootstrap since we don't alter the case. If the upstream component adds collapse then we should too otherwise no

src/Accordion.js
bpas247: I ended up using `useUncontrolled` for this, and updated the docs to include an example of how to set Accordion to a "fully collapsed" state

src/Accordion.js
```diff
@@ -0,0 +1,62 @@
...
+
+const propTypes = {
+  /** Set a custom element for this component. */
+  as: PropTypes.elementType,
+
+  /** @default 'accordion' */
+  bsPrefix: PropTypes.string,
+
+  /** Set a default active key for this component to be opened on start */
+  activeKey: PropTypes.string,
```
taion: might need `defaultActiveKey` as well?

src/AccordionToggle.js
```diff
@@ -0,0 +1,42 @@
...
+const propTypes = {
+  /**
+   * A key that corresponds to the collapse component that gets triggered
+   * when this has been clicked.
+   */
+  eventKey: PropTypes.string.isRequired,
+};
+
+const defaultProps = {
+  as: 'div',
```
taion: prop type?

src/AccordionToggle.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const AccordionToggle = ({ children, eventKey }) => {
```
taion:
```suggestion
const AccordionToggle = ({ as: Component, children, eventKey }) => {
```

src/AccordionToggle.js
```diff
@@ -0,0 +1,42 @@
...
+  const context = useContext(AccordionContext);
+
+  return (
+    <>
```
taion: then here do e.g.

```js
<Component onClick={...} (include other props) />
```

perhaps?

taion: so in the examples below then, you'd do `<Accordion.Toggle as="button" ...`

src/AccordionCollapse.js
```diff
@@ -0,0 +1,28 @@
...
+  eventKey: PropTypes.string.isRequired,
+};
+
+const AccordionCollapse = React.forwardRef(
+  ({ children, eventKey, ...props }, ref) => {
+    const context = useContext(AccordionContext);
+
+    return (
+      <Collapse in={context.activeEventKey === eventKey} {...props}>
+        <div ref={ref}>{children}</div>
```
taion: does this need to render this extra `<div>`? the `<Card.Body>` or whatever is going to render a `<div>` itself, no?

bpas247: The extra `<div>` seems to be there due to an issue with `Collapse`, if you don't render `children` in a wrapped element. For example, If I were to change the implementation to:
```jsx
<Collapse in={context.activeEventKey === eventKey} {...props}>
  {children}
</Collapse>
```
Then this error is thrown:
```
    Error: Uncaught TypeError: Cannot read property 'props' of undefined (webpack:///src/Collapse.js:213 <- test/index.js:136762)
        at Object.invokeGuardedCallbackDev (webpack:///node_modules/react-dom/cjs/react-dom.development.js:199 <- test/index.js:63588:16)
        at invokeGuardedCallback (webpack:///node_modules/react-dom/cjs/react-dom.development.js:256 <- test/index.js:63645:31)
        at replayUnitOfWork (webpack:///node_modules/react-dom/cjs/react-dom.development.js:18532 <- test/index.js:81921:5)
        at renderRoot (webpack:///node_modules/react-dom/cjs/react-dom.development.js:19422 <- test/index.js:82811:13)
        at performWorkOnRoot (webpack:///node_modules/react-dom/cjs/react-dom.development.js:20296 <- test/index.js:83685:7)
        at performWork (webpack:///node_modules/react-dom/cjs/react-dom.development.js:20208 <- test/index.js:83597:7)
        at performSyncWork (webpack:///node_modules/react-dom/cjs/react-dom.development.js:20182 <- test/index.js:83571:3)
        at requestWork (webpack:///node_modules/react-dom/cjs/react-dom.development.js:20051 <- test/index.js:83440:5)
        at scheduleWork (webpack:///node_modules/react-dom/cjs/react-dom.development.js:19865 <- test/index.js:83254:5)
```
Which seems to be somehow correlated to this code block in `Collapse`:
```jsx
className: classNames(
  className,
  children.props.className, // <-- This line may be causing the issue
  collapseStyles[state],
  this.getDimension() === 'width' && 'width',
),
```

src/AccordionToggle.js
bpas247: So if I have `as` to have a default val of `button`, then the doc example would be updated to something like this?
```jsx
  <Card>
    <Card.Header>
      <Accordion.Toggle
        onClick={key => console.log(`I've been clicked by : ${key}`)}
        eventKey="0"
      />
    </Card.Header>
      <Card.Body>Hello! I'm the body</Card.Body>
    </Accordion.Collapse>
  </Card>
</Accordion>
```

taion: I think so, right? I'm looking at e.g. how `DropdownToggle` works. Then if you wanted to make the entire header clickable, you could do:

```js
<Accordion.Toggle
  as={Card.Header}
  // ...
>
```

src/AccordionCollapse.js
taion: Well, `children` needs to be exactly one React element. Perhaps type `children` as `React.Element`, and perhaps do `React.Children.only(children)`?

src/Accordion.js
```diff
@@ -0,0 +1,65 @@
...
+    ...controlledProps
+  } = useUncontrolled(props, {
+    activeKey: 'onClick',
+  });
+
+  bsPrefix = useBootstrapPrefix(bsPrefix, 'accordion');
+
+  return (
+    <AccordionContext.Provider value={{ onClick, activeEventKey: activeKey }}>
+      <SelectableContext.Provider value={null}>
```
taion: i don't think this is needed; the reasons per https://github.com/react-bootstrap/react-bootstrap/blob/89693c7967594d4713149a17aca0993b45373cc0/src/TabPane.js#L167-L171 don't apply here

unless you want to use `SelectableContext` to pass down `onSelect`... might be more DRY than passing down `onClick` in `AccordionContext`

bpas247: I think I'm gonna go with just using `SelectableContext`, should be good enough for what I'm using it for

bpas247:
I am currently on the fence for adding typescript definitions for Accordion and its relevant components in this PR, as it might make the diffs a bit larger than it currently is. 

Thoughts on this? Maybe just open another PR after this one is merged?

src/Accordion.js
```diff
@@ -42,17 +41,15 @@ const Accordion = React.forwardRef((props, ref) => {
...
-        <Component
-          ref={ref}
-          {...controlledProps}
-          className={classNames(className, bsPrefix)}
-        >
-          {children}
-        </Component>
-      </SelectableContext.Provider>
-    </AccordionContext.Provider>
+    <SelectableContext.Provider value={{ onClick, activeEventKey: activeKey }}>
```
taion: if you look at the other use cases, we only use `<SelectableContext.Provider>` to pass down `onSelect`, while using a different context to pass down details of the active item. i guess because the latter varies from place to place?

so to follow that, we would want both contexts, but we'd want to use them differently

bpas247: Ah. So I should use AccordionContext to pass down the current active key and then use SelectableContext to pass down onSelect?

taion: yeah – that's what we do elsewhere. it's a little weird but better to be consistent, i think.

src/Accordion.js
```diff
@@ -0,0 +1,65 @@
...
+  let {
+    as: Component,
+    activeKey,
+    bsPrefix,
+    children,
+    className,
+    onClick,
+    ...controlledProps
+  } = useUncontrolled(props, {
+    activeKey: 'onClick',
```
taion: this should probably be `onSelect`, then, instead of `onClick`

src/AccordionToggle.js
```diff
@@ -0,0 +1,48 @@
...
+  children: PropTypes.element,
+};
+
+const defaultProps = {
+  as: 'button',
+};
+
+const AccordionToggle = React.forwardRef(
+  ({ as: Component, children, eventKey, onClick, ...props }, ref) => {
+    const contextOnClick = useContext(SelectableContext);
```
taion: `onSelect`

QuantumInformation:
TypeScript definitions are always nice. However, I hardly ever write them, as if you write the code in TypeScript (I wrote a few [vanilla js components](https://www.npmjs.com/package/solar-popup ) the are autogenerated for you. I guess that isn't an option here though.

bpas247:
Anything else I should look at before this is merged? I've decided I will add the TypeScript types in a separate PR.

PR #3411, opened 2019-01-13
First pass of typescript types for everything

glenjamin:
Welp, that took longer than I was expecting!

I think I've got a reasonable set of prop types for each module, but i got a bit tired of detailed checking towards the end.

I'm not sure whether the polymorphic `as` behaviour actually works correctly - although in theory the types I've got there should handle it.

Because of the generated types for the final component props, the error messages can be a bit wordy - but they do seem to also include the important information as well.

I've added dtslint into the build, but unfortunately this only checks the types against the `simple.test.tsx` file, it doesn't check the types against the actual implementation in any way.

Another mild annoyance is that although i've matched up the component filenames, the types don't work for the individual component files because of the `lib` vs `src` vs `es` part of the build process. In theory the files could be duplicated during the build for the exported code if this is desired.


jquense:
This is awesome! Thanks a ton I'll review properly on Monday

jquense:
oops can you fix the CI error?

glenjamin:
Good timing, I was just looking at that!

The failure seems to be caused by two issues, and I've just worked around it instead:

1) `<textarea>` in `@types/react` has `rows?: number`, so it doesn't accept strings
2) Either the ReplaceProps helper or the `as` inference logic doesn't quite seem to be working in typescript 3.1, and https://github.com/Microsoft/dtslint/issues/176 prevents me from telling `dtslint` to only use 3.2 or higher.

glenjamin:
Hrm, I think actually typescript 3.1 was correct in flagging an error here, and 3.2 not erroring wasn't correct 🤔

glenjamin:
That appveyor error doesn't seem to be down to anything I've changed - unless I've overlooked something?

jquense:
ya Appveyer is just been broken for a while. not your problem :P

PR #3413, opened 2019-01-15
Fix Appveyor build

glenjamin:
Using double-quotes seems to satisfy the argument parser, and gets thing working again.


taion:
does this screw up codecov in a weird way? eh, who cares

jquense:
so nice to have this fixed 🙇

PR #3414, opened 2019-01-15
chore: integrate ts types into build

jquense:


types/simple.test.tsx
```diff
@@ -31,7 +31,7 @@ import {
   Tab,
   ToggleButtonGroup,
   ToggleButton,
-} from 'react-bootstrap';
+} from 'types';
```
taion: this doesn't need to be a relative import?

jquense:
gtg?

PR #3427, opened 2019-01-29
chore: update to point to gh-pages site

jquense:
I made the latest beta latest on npm. lets do this

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "format": "npm run ci-lint . -- --fix",
     "lint": "npm run ci-lint .",
     "precommit": "lint-staged",
-    "release": "yarn changelog && rollout",
+    "release": "rollout",
     "prepublishOnly": "npm run build",
     "tdd": "karma start",
     "test": "npm run lint && npm run dtslint && npm run test-browser && npm run test-node",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "format": "npm run ci-lint . -- --fix",
     "lint": "npm run ci-lint .",
     "precommit": "lint-staged",
-    "release": "yarn changelog && rollout",
+    "release": "rollout",
```
taion: no more changelog?

www/package.json
```diff
@@ -73,7 +73,7 @@
     "develop": "gatsby develop",
     "build": "gatsby build",
     "deploy": "gatsby build --prefix-links && yarn push-docs",
-    "push-docs": "gh-pages -d public --branch next --repo \"https://github.com/react-bootstrap/react-bootstrap.github.io.git\"",
+    "push-docs": "gh-pages -d public --branch master --repo \"https://github.com/react-bootstrap/react-bootstrap.github.io.git\"",
```
taion: is this explicit `--branch` required now that it's `master`?

package.json
jquense: this wasn't working right i'm gonna follow up with something in rollout

PR #3432, opened 2019-01-30
docs: calculate SRI and show version

jquense:


jquense:
gonna merge since it's just a small doc change

PR #3364, opened 2018-11-09
Update Migration Guide

bpas247:
Fixes #3336

PR #3437, opened 2019-01-31
Export the DropdownItem from index

GabrielNicolasAvellaneda:


taion:
do we really want this instead of `Dropdown.Item`?

GabrielNicolasAvellaneda:
I don't know but other components follows this method and are exported from index so you could import many components at once:
`import { Nav, NavDropdown, Navbar, NavItem, DropdownItem } from 'react-bootstrap';`

taion:
@jquense i defer to you here; should we do this?

kt3k:
The migration guide seems implying this exists. 

I also found the comment (https://github.com/react-bootstrap/react-bootstrap/pull/3364#discussion_r232273723) which seems saying DropdownItem should be exported as DropdownItem as well.

PR #3440, opened 2019-02-01
Guard against elem.style being undefined

jeroenpeeters:
Using React Test Renderer (https://reactjs.org/docs/test-renderer.html) the element does not have a style property. Currently collapsible panels cannot be rendered correctly using the test renderer.

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -147,20 +147,24 @@ class Collapse extends React.Component {
 
   /* -- Expanding -- */
   handleEnter = elem => {
+    if (!elem.style) return;
     elem.style[this.getDimension()] = '0';
   };
 
   handleEntering = elem => {
+    if (!elem.style) return;
     const dimension = this.getDimension();
     elem.style[dimension] = this._getScrollDimensionValue(elem, dimension);
   };
 
   handleEntered = elem => {
+    if (!elem.style) return;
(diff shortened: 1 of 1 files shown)


taion:
i'm not sure this makes sense. testing transitions against a test renderer that doesn't involve a DOM seems questionable.

jeroenpeeters:
@taion I don't completely agree. I can still test transitions by looking at the CSS classes. For the test renderer this is what matters. However, I'm not actually testing the transitions. It just that codes gets invoked that leads to a transition. Currently this breaks the test. I've fixed it locally by making the code test-aware, but I think that's even worse.

mxschmitt:
@taion @jeroenpeeters whats the state of the PR? looks very stale.

taion:
does react test renderer still have this behavior? it just seems odd to me

mxschmitt:
Closed due no activity

jeroenpeeters:
I recently ran into this issue again. Please consider my fix.

Currently it is impossible to use BootstrapAccordion with react-test-renderer in the following scenario;

1. Code renders accordion with panels closed
2. Click event is simulated on panel
3. Panel opens
4. Assert current state via `expect(underTest.toJSON()).toMatchSnapshot();`

Currently #3 results into an error because react-test-renderer does not allow manipulation of style properties.

Issue #3446, opened 2019-02-04
Typescript className


madroneropaulo:
Hello!

I'm using bootstrap beta 1.0.0-beta.5 with typescript (I'm not using the Definitely Typed repo), but I'm getting an error when I try to set a className for a Navbar.Collapse component.

```
Type '{ children: Element; class: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<NavbarCollapse> & Readonly<{ children?: ReactNode; }> & Readonly<ReplaceProps<typeof Collapse, BsPrefixProps<typeof Collapse>>>'.
```
How can I add a className to react-bootstrap components and typescript?

jquense:
The type def is probably incomplete. happy to take PR's fixing them!

aMarCruz:
@jquense , you're right, typings are incomplete and has errors.
I'll be uploading PRs with the corrections and documenting them in the coming days.

jquense:
Yup 👍

webivan1:
Hello! I`ve a problem **Type '{ children: Element; id: string; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<NavbarCollapse>** This error is due to <Navbar.Collapse **id={'test-id'}**>

fredriklengstrand:
Another issue I have found with the typings is that the type for the `onChange` event on a `<Form.Control>` input does not seem to match the type `ChangeEvent<HTMLInputElement>` like it does for a regular `<input>` element:

```
Type '(event: ChangeEvent<HTMLInputElement>) => void' is not assignable to type '(event: FormEvent<ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>>) => void'.
  Types of parameters 'event' and 'event' are incompatible.
    Type 'FormEvent<ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>>' is not assignable to type 'ChangeEvent<HTMLInputElement>'.
      Types of property 'target' are incompatible.
        Type 'EventTarget' is not assignable to type 'EventTarget & HTMLInputElement'.
          Type 'EventTarget' is missing the following properties from type 'HTMLInputElement': accept, align, alt, autocomplete, and 286 more.
```

jquense:
feel free to send over any type improvements. we will merge em right in

aMarCruz:
@fredriklengstrand , in fact, it happens with several events but it has a simple solution.

[...]

Issue #3444, opened 2019-02-03
Typescript definitions: Add interfaces to exports


rvaidya:
I love all the work that's being put into getting Typescript definitions into the project.

Can some of the interfaces also be within the definitions exports?  Specifically, the top level props for each class, and perhaps the sizing declarations.

This will let us extend the interfaces when creating our own wrapper classes without needing to copy the entire interfaces into our own source and needing to keep it up to date with the definitions as they develop.

rvaidya:
Example, would like to simplify IMyCardProps with just "variant" and extend CardProps.

<details>
<summary>Code</summary>

```typescript
import React from "react";
import { Card } from "react-bootstrap";

interface IMyCardProps {
    variant?: string;
    bg?:
    | "primary"
    | "secondary"
    | "success"
    | "danger"
    | "warning"
    | "info"
    | "dark"
    | "light";
    text?:
    | "primary"
    | "secondary"
    | "success"
    | "danger"
    | "warning"
    | "info"
    | "dark"
    | "light"
    | "white"
    | "muted";
    border?:
    | "primary"
    | "secondary"
    | "success"
    | "danger"
    | "warning"
    | "info"
    | "dark"
    | "light";
    body?: boolean;
}

class MyCard extends React.Component<IMyCardProps> {
    public render() {

        if (this.props.variant !== null) {
            // Emulate the bs3 well using card
            if (this.props.variant === "well") {
                return (
                    <Card
                        bg="light"
                        body={true}
                        border={this.props.border}
                        children={this.props.children}
                    />
                );
            }
        }

        return (
            <Card
                bg={this.props.bg}
                body={this.props.body}
                border={this.props.border}
                children={this.props.children}
            />
        );
    }
}

export default MyCard;

```
</details>

jquense:
Sure! We'd be happy to take a PR addint that

mxschmitt:
@jquense this issue should be closed because this PR was merged 👍

superdopey:
When will these changes be pushed to the npm package?

PR #3448, opened 2019-02-05
Feature/typescript prop exports

rvaidya:
- Export props for each class that has them in the TypeScript definitions
- Fixed a typo in dropdowns "alightRight" vs "alignRight"
- Fixed copy/paste issues (Feedback was copy/pasted onto FormLabel and the class names were kept)


rvaidya:
#3444

types/components/Alert.d.ts
```diff
@@ -12,7 +12,7 @@ declare class AlertHeading<
   As extends React.ReactType = 'div'
 > extends BsPrefixComponent<As> {}
 
-interface AlertProps extends React.HTMLProps<Alert> {
+export interface AlertProps extends React.HTMLProps<Alert> {
```
taion: i guess we don't want the exported name for the individual modules here to be `Props`?

rvaidya: tslint recomments interfaces all start with I -> IAlertProps

if that is preferable

thejohnfreeman: @taion , your hunch is right. The general convention in the React community, including the `react` package, is to use the name `ComponentNameProps` instead of just `Props`, to make imports of multiple props types easier.

jquense:
The other pattern I've seen is to use a namespace but I'm not really familiar with what that would look like practically here

types/components/Collapse.d.ts
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { TransitionCallbacks } from './helpers';
 
-interface CollapseProps
+export interface CollapseProps
```
thejohnfreeman: This interface is missing the `id` prop seen in [the examples](https://react-bootstrap.github.io/components/navbar/#navbars-overview).

prateekkathal: It's also missing className. 

#3446 

image

rvaidya:
I noticed that too but it's not listed in the react-bootstrap docs, and also seems to work in some cases without the id prop

types/components/Collapse.d.ts
rvaidya: I noticed that too but it's not listed in the react-bootstrap docs, and also seems to work in some cases without the id prop

prateekkathal: @rvaidya The id prop is non-mandatory. Add these please 😄 

```js
id?: string;
className?: string;
```

@thejohnfreeman I believe adding these 2 lines should do it 👍

jquense: Shouldn't this extend the base html props instead of hand adding these?

rvaidya: Yeah I was gonna look into that since it wasn't done originally

thejohnfreeman: I've been wondering - is there a base props type for HTML elements? What is it called? Where does it come from? `@types/react`?

Related: is there a base props type for React elements (with `key` and `children`)?

jquense: There are yeah, in the react types on DT

rvaidya: the base proptype I see is React.HTMLProps<T>, which has been applied to some classes but not others.  Some of these decisions seem to be intentional - such as the conflict between "size" in Button:

size?: 'sm' | 'lg'

vs:

size?: number in React.HTMLProps

Based on this, I can just add id?: string to Collapse without extending anything, although the documentation for react-bootstrap should be updated to include id if the id attribute is important

prateekkathal: @rvaidya Adding the baseProps will also add the declarations for other "not-required" variables and functions. We should avoid that since the documentation doesn't say so, nor the supported methods have been added or extended by the class.

Best to add `id` and `className` in directly. 

Thanks @jquense and @thejohnfreeman for the support 👍

jquense: Oops sorry `id` doesn't belong here. The screenshot above is for `NavbarCollapse` which specifically renders a `div` the abstract Collapse doesn't render anything but it's child.

prateekkathal: @jquense But the type declarations of it exist in `Collapse.d.ts` no?

jquense: it should exist in `NavbarCollapse.d.ts` because its specific to `NavbarCollapse`, but it should be more general to `div` props not just id

jquense: its fine to not fix this in this PR but `id` shouldn't be added here

prateekkathal: @jquense I tired replacing the NavbarCollapse definition to

image
 
but it results in 

image

jquense: I would just defer this for now

prateekkathal: @jquense Giving this one last try. Let me know if you're okay with this.... It fixes the issue. Otherwise someone else can take this in the future.

```js
interface NavbarCollapseProps extends React.ClassAttributes<NavbarCollapse> {
  id?: string;
  className?: string;
}

declare class NavbarCollapse extends BsPrefixComponent<typeof Collapse, NavbarCollapseProps> {}
```

Thanks for the continued support on this 😄

thejohnfreeman: - `NavbarCollapse` renders a `Collapse` with a `div` child, and passes all its props besides `bsPrefix` through to the `Collapse`.
- The `Collapse` renders a `Transition` with a child render function that renders the child of `Collapse` (the `div`). `Collapse` takes out the `className` prop, passing it through to its child. It passes its remaining unknown props through to the `Transition`, which is imported from `react-transition-group`.
- `Transition` subtracts its props and passes the rest back to its render child. `Collapse` defines a render child that passes these props through to its child (the `div` in this case).

So it seems that `Collapse` should expect to get a mix of the props it knows it uses (already [listed as properties in `CollapseProps`](https://github.com/react-bootstrap/react-bootstrap/blob/b6651f3507ab51c63c84e13ac58d0ae0f084bf0e/types/components/Collapse.d.ts)), the props expected by `Transition` (named `TransitionCallbacks`, which `CollapseProps` already extends), and the props of its child (the `div` in this case, but could be anything). This last type is conceptually `PropsOf<children>`, but I don't know how to express that in TypeScript.

jquense: ya I think what we actually need to do long term is change the API here so each component doesn't just pass through everything to it's children

thejohnfreeman: @prateekkathal , I'm with @jquense and think you should leave the type fixing out of this PR, especially since it's so complicated. Just getting proper exports is a great step forward.

prateekkathal: Sounds good. 👍 

@rvaidya Please revert the change here 😄

jviall:
Is a different issue/PR going to be opened for the lack of `id` and `className` on the `NavbarCollapse` prop type? Really looking forward to that fix.

cyberst:
> Is a different issue/PR going to be opened for the lack of `id` and `className` on the `NavbarCollapse`

+1

PR #3454, opened 2019-02-07
Use hooks

jquense:
this is clearly an improvement for all our context usage.


src/test/ButtonGroup.js
```diff
@@ -0,0 +1,82 @@
...
+        className,
+        baseClass,
+        size && `${prefix}-${size}`,
+        toggle && `${prefix}-toggle`,
+      )}
+    />
+  );
+});
+
+ButtonGroup.displayName = 'ButtonGroup';
```
jquense: TIL you can just do this and they appear as normal components in the devtools and tests

jquense:
@taion this is done 👍

src/AbstractNavItem.js
```diff
@@ -15,76 +16,68 @@ const propTypes = {
...
+          ? navContext.activeKey === navKey
+          : active;
+    }
+
+    if (props.role === 'tab') {
+      props.tabIndex = isActive ? tabIndex : -1;
+      props['aria-selected'] = isActive;
+    }
+
+    const handleOnclick = useEventCallback(e => {
```
taion: for my benefit... what does `useEventCallback` do? i see the `useCommittedRef` thing but i don't immediately understand it

src/DropdownToggle.js
```diff
@@ -82,4 +82,5 @@ class DropdownToggle extends React.Component {
   }
 }
 
+// Needs to be a class FTM, because it needs to accept a ref that can be used with findDOMNode
```
taion: why wouldn't a forwardRef into a functional component work?

jquense: somewhere the ref doesn't get passed through, so we are actually relying on findDOMNode. I'll fix it up tho after i prune react-overlays and popper gets some hooks

src/AbstractNavItem.js
jquense: it returns a function with a stable identity that doesn't need to be invalidated or changed if the inner handler isn't stable. it's a less flexible equivalent to 

```js
someMethod() {
  this.props.handleFoo()
}
```

taion: neat/complicated!

jjustus:
This PR currently breaks ie11 because of the dependency on @restart/hooks, which currently emits functions using arrow syntax.

Issue #3481, opened 2019-02-18
Form.Label prop typings is missing xs, sm, md, lg, xl properties


codeandcats:
One cannot specify `xs`, `sm`, `md`, `lg`, or `xl` properties on `Form.Label` like in this [example](https://react-bootstrap.github.io/components/forms/#forms-input-Plaintext).

```jsx
  <Form.Group as={Row} controlId="formPlaintextEmail">
    <Form.Label column sm="2">
      Email
    </Form.Label>
    <Col sm="10">
      <Form.Control plaintext readOnly defaultValue="email@example.com" />
    </Col>
  </Form.Group>
```

This is because they are missing from the typings in `FormLabel.d.ts`, which was clearly copied from `Feedback.d.ts`.

```typescript
import * as React from 'react';

import { BsPrefixComponent } from './helpers';

interface FeedbackProps {
  htmlFor?: string;
  column?: boolean;
  innerRef?: React.LegacyRef<this>;
  srOnly?: boolean;
}

declare class Feedback extends BsPrefixComponent<'label', FeedbackProps> {}

export default Feedback;
```

jquense:
Happy to take a PR for this!

codeandcats:
@jquense Thanks. I need this soon so will gladly review it. Also happy to take it on myself if you cannot dedicate time to it immediately.

jquense:
> . Also happy to take it on myself i

Ya that's what i was suggesting :) I don't know when i'd get to it

codeandcats:
Oh sorry I misread your comment. 🤗

Actually I realised I can probably use 
`<Col sm={2} as={Form.Label>...</Col>`.

I’m not sure yet as I’m upgrading a 2 yr old project and working my way through all the breaking changes.

Seems like there might be 2 ways to make a label behave like a column. Should we remove one?

I am wondering whether we need the column property on `Form.Label`

moyuanhuang:
@codeandcats Thanks that works! I noticed that you'd also have to explicitly add the `className="col-form-label"` prop, otherwise the label wouldn't have nice paddings.

<img src="https://user-images.githubusercontent.com/14947870/53308538-07761480-3857-11e9-8333-149862643bd9.png" width="300"/>

<img src="https://user-images.githubusercontent.com/14947870/53308557-25dc1000-3857-11e9-97df-db2cc1dc9d26.png" width="240"/>

codeandcats:
[...]

PR #3500, opened 2019-02-24
Support generic `Col` props for FormLabel

moyuanhuang:
The initial motivation of this PR is to resolve the typing error reported in #3481.

IMO, we should replace the `column` prop in `FormLabel` component with `as`, also because it can make the API more consistent, since passing in the `as` prop is a common pattern that is used across the other components in this package.

### DIFF
--- a/src/FormLabel.js
+++ b/src/FormLabel.js
@@ -23,12 +23,6 @@ const propTypes = {
   htmlFor: PropTypes.string,
 
   /**
-   * Renders the FormLabel as a `<Col>` component (accepting all the same props),
-   * as well as adding additional styling for horizontal forms.
-   */
-  column: PropTypes.bool,
-
-  /**
    * The FormLabel `ref` will be forwarded to the underlying element.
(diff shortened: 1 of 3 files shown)


jquense:
Hey thanks for the PR. I think ideally this would the best API unfortunately there is a problem with this approach. The check for the 'as' being a Col component wouldn't work in a few situation, the biggest being if you wrap the col component

moyuanhuang:
@jquense I'll find some time to test it locally and dive into it more if possible, will keep this PR updated!

moyuanhuang:
Regardless, I think it'll be nice if we can at lease merge b33d4d since the naming in the type definition of `FormLabel` is just wrong.

jquense:
sure yeah it's definitely wrong oops :P I another constraint here is that even as a Col, the underlying DOM element still needs to be a `label` semantically

taion:
so should we just use a tagged union for the props here, then?

Issue #3509, opened 2019-02-28
Request: ability to change the tag/element of the <Carousel> parent div via props


jokokoloko:
Can the ability to change the tag/element of the <Carousel> parent div via props, such as to "section", be added?

taion:
we'd be happy to take a PR here to add support for the `as` prop, as on other components here

PR #3514, opened 2019-03-04
feat: added 'as' prop to Carousel (fix: #3509)

mxschmitt:
https://github.com/react-bootstrap/react-bootstrap/issues/3509

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -2,7 +2,6 @@ import classNames from 'classnames';
 import styles from 'dom-helpers/style';
 import transition from 'dom-helpers/transition';
 import React, { cloneElement } from 'react';
-import { elementType } from 'prop-types-extra';
 import PropTypes from 'prop-types';
 import uncontrollable from 'uncontrollable';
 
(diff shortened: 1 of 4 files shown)


src/Carousel.js
```diff
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import styles from 'dom-helpers/style';
 import transition from 'dom-helpers/transition';
 import React, { cloneElement } from 'react';
+import { elementType } from 'prop-types-extra';
```
jquense: I believe the newest version of `prop-types` now supports elementType, lets use that instead yeah?

mxschmitt: sure, maybe we should replace it in the whole project? (I can do it if you want)

src/Carousel.js
```diff
@@ -22,6 +22,7 @@ const propTypes = {
    * @default 'carousel'
    */
   bsPrefix: PropTypes.string,
+  as: PropTypes.element,
```
jquense: `elementType` still, `element` is different. We can change in the whole project yeah if you are up for it

Issue #2785, opened 2017-09-09
Export Props interfaces


mayerwin:
In my Typescript 2.3 project, I am trying to import ModalProps and ButtonProps to extend them for my own derived components using: 
`import { ModalProps, ButtonProps } from 'react-bootstrap';`
But this causes Intellisense errors as ModalProps and ButtonProps are not exported in the corresponding  type definition files.

Is there any workaround I could use until the definitions are fixed?

taion:
The TS defs don't live in this repo. Also, I know of no such exports, so it's not even clear what you're referring to there.

mayerwin:
I have opened the issue here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/19664

PR #3517, opened 2019-03-04
Added Properties: 'id' and 'className'

julionadal:
This two properties are missing from this types

### DIFF
--- a/types/components/Collapse.d.ts
+++ b/types/components/Collapse.d.ts
@@ -4,6 +4,8 @@ import { TransitionCallbacks } from './helpers';
 export interface CollapseProps
   extends TransitionCallbacks,
     React.ClassAttributes<Collapse> {
+  id?: string;
+  className?: string
   in?: boolean;
   mountOnEnter?: boolean;
   unmountOnExit?: boolean;


types/components/Collapse.d.ts
```diff
@@ -4,6 +4,8 @@ import { TransitionCallbacks } from './helpers';
 export interface CollapseProps
   extends TransitionCallbacks,
     React.ClassAttributes<Collapse> {
+  id?: string;
+  className?: string
```
jquense: they don't belong on this type Collapse, they belong on NavbarCollapse, and then they shouldn't be listed individually they should extend the react div props

jquense: see #2785

prateekkathal: Just FYI,

@julionadal A huge discussion was done on [this thread](https://github.com/react-bootstrap/react-bootstrap/pull/3448/commits/57278de982fb03e9b2e55a305f6f28379810706f#r256593836) in regards to the same issue. The decision to rebuild the API was made by @jquense and @thejohnfreeman

prateekkathal: Also this is a duplicate of #3502

julionadal: Thank you, i just saw @prateekkathal. Hope this changes are going to be released soon.

PR #3519, opened 2019-03-05
Add missing props to Navbar.Collapse

thejohnfreeman:
I tried testing this change in a dependent project, but installing the package from local clone just results in `Cannot find module 'react-bootstrap/Navbar'`. To complicate matters, the build setup seems closely tied to the contributors' development environment. The second commit in this PR fixes the build from a clean state; it only ever worked from a dirty state. There is no `prepare` NPM script: is that intentional?

At the least, this can serve as a starting point.

/cc @jquense

### DIFF
--- a/tools/build.js
+++ b/tools/build.js
@@ -15,7 +15,7 @@ const libRoot = path.join(__dirname, '../lib');
 const distRoot = path.join(libRoot, 'dist');
 const esRoot = path.join(libRoot, 'es');
 
-const clean = () => fse.existsSync(libRoot) && fse.remove(libRoot);
+const clean = async () => fse.existsSync(libRoot) && fse.remove(libRoot);
 
 const step = (name, root, fn) => async () => {
   console.log(cyan('Building: ') + green(name));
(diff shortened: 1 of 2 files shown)


thejohnfreeman:
I was able to release [my fork of the library](https://www.npmjs.com/package/@thejohnfreeman/react-bootstrap) today for testing, and it seems to work.

tools/build.js
```diff
@@ -15,7 +15,7 @@ const libRoot = path.join(__dirname, '../lib');
 const distRoot = path.join(libRoot, 'dist');
 const esRoot = path.join(libRoot, 'es');
 
-const clean = () => fse.existsSync(libRoot) && fse.remove(libRoot);
+const clean = async () => fse.existsSync(libRoot) && fse.remove(libRoot);
```
taion: just don't await this below; this isn't `async`

jquense: `remove` is

taion: ah shoot, we didn't merge https://github.com/react-bootstrap/react-bootstrap/pull/3501

taion: okay, can undo changes here now that we've merged

thejohnfreeman: Done.

PR #3526, opened 2019-03-07
Change default variant from primary to default

eylonronen:


jquense:
Hi there! there is no `default` variant in bootstrap 4, you may be thinking of bootstrap v3 still which this not what's supported by the latest react-bootstrap version

Issue #3520, opened 2019-03-06
Remove trailing semicolon in the form


muthusrinivasan-a:
[<Form>
  <Form.Row>
    <Col>
      <Form.Control placeholder="First name" />
    </Col>
    <Col>
      <Form.Control placeholder="Last name" />
    </Col>
  </Form.Row>
</Form>;](https://react-bootstrap.github.io/components/forms/)

Not sure why we will have a **;** in the last. Please update the code snippets

muthusrinivasan-a:
Hi Philip,

It is there in most of the form component examples. Observed that it is just updated only in the above example mentioned. Can you please update the other examples too?

ex : Sizing, Plaintext

in the Sizing, I can see only the empty parent element.

mxschmitt:
As mentioned here: https://github.com/react-bootstrap/react-bootstrap/pull/3529#issuecomment-470832914

But if we add `semi: false` it would also cause this issue: https://github.com/prettier/prettier/issues/3196
Maybe there is a workaround for that but nothing found so far.

PR #3529, opened 2019-03-08
Remove trailing semi-colon (#3520)

philipwhiuk:


mxschmitt:
Hey, we have to adjust the prettier rule for this. Because it adds it again on pre-commit time. :)
Maybe add: `semi: false`, but that would be a big PR and contains changes in the entire project. @jquense what do you think about this?

jquense:
i think it's fine to leave, it's not ideal but it's not wrong and the alternative is to remove all semi's which isn't the style we want

PR #3530, opened 2019-03-08
add `xl` to modal size enum

dleavitt:
It is a thing: https://github.com/twbs/bootstrap/issues/26952

The approach of only allowing known values is very brittle. It would be better to do:
`size?: 'sm' | 'lg' | 'xl' | string;`
for this sort of thing. Enumerate the known values but allow unknown ones. Happy to take a stab at fixing this globally if others agree.


dleavitt:
thanks!

PR #3541, opened 2019-03-14
Add support for Spinners

tjdavey:
Adds support for the [spinners components](https://getbootstrap.com/docs/4.2/components/spinners/) introduced in Bootstrap 4.2.

*Changes:*
- Adds new `Spinner` component
- Add tests for new component
- Creates new page in documentation for new component
- Upgrades documentation CSS to Bootstrap 4.2 (given the claimed 4.2 support in the documentation itself) 

*Notes:*
- I wasn't sure about the terminology for the property that switches between the two styles of spinner component. I couldn't think of a good term that wouldn't overlap with an existing property. `animation` was chosen at it was the most representative of the changes the property made, but I feel it may still be too ambiguous. I'm open to alternative suggestions.
- Whilst the documentation claims compatibility with Bootstrap 4.2 it was using a Bootstrap 4.1 CSS file. I've upgraded the documentation to a Bootstrap 4.2 version of the CSS file. I can remove this change and submit seperately if necessary.

### DIFF
--- a/src/Spinner.js
+++ b/src/Spinner.js
@@ -1,0 +1,86 @@
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import React from 'react';
+
+import { createBootstrapComponent } from './ThemeProvider';
+
+class Spinner extends React.Component {
+  static propTypes = {
+    /**
+     * @default 'spinner'
+     */
+    bsPrefix: PropTypes.string.isRequired,
+
+    /**
+     * The visual color style of the spinner
+     *
+     * @type {('primary'|'secondary'|'success'|'danger'|'warning'|'info'|'light'|'dark')}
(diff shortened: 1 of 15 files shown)


tjdavey:
> cool! can we add a type signature as well, for the TS users?

I'll add it now.

taion:
Thanks!

jquense:
this is awesome

tjdavey:
@taion I've added the Typescript support. I don't have the deepest understanding of the way the Typescript definitions are structured for this project so it might be worth just checking the details of my implementation.

davecarlson:
Any news when beta 6 will be released so we can use these without pulling in master ? 

:)

jquense:
soon!

PR #3554, opened 2019-03-19
Export DropdownItem

kt3k:
According to the migration guide and the comments in #3364, Dropdown.Item should be exported as DropdownItem, but it doesn't seem to be exported as that name. This PR fixes it.


kt3k:
sorry this is a duplicate of #3437

PR #3557, opened 2019-03-19
Update README with Components live demo

JoshK2:
I've shared react-bootstrap components with Bit into this collection:

https://bit.dev/react-bootstrap/react-bootstrap

Now people can explore the components, play with examples and install individual components as packages.
The components will continue to update automatically, or I can hand over ownership (yoshuakuttler@gmail.com) and help maintain it if you prefer :)

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,9 @@ See the [documentation][documentation] with live editable examples and API docum
 
 To find the documentation for the latest Bootstrap 3 compatible release, [go here](https://5c507d49471426000887a6a7--react-bootstrap.netlify.com/).
 
+#### Component collection
+[Explore components](https://bit.dev/react-bootstrap/react-bootstrap), play with examples and install individually.
+
 ### Migrating from Bootstrap 3 to Bootstrap 4
 
 If you would like to update React-Bootstrap within an existing project to use Bootstrap 4, please read our docs for **[migrating to React-Bootstrap V1](https://react-bootstrap.github.io/migrating/)**.


taion:
This is interesting, but we already have live editable demos for our components in the main documentation: https://react-bootstrap.github.io/components/button-group/

mxschmitt:
In my mind it's also enough to have the react play editor in the docs, so that the users can directly edit / copy / adjust components.
@jquense what do you think, if you are also against it, we can close this issue.

jonisar:
Hi all, I just want to shed a bit of light on why this is useful.
It's true that the playground and demos in the docs are great. However, that's not the main core value of this component collection. The real value lies in the installation of each component as a stand-alone reusable unit. Also, Bit lets you import every component into any project and develop it right there. Meaning, developers can now a) install individual components b) develop, modify and extend components right from the projects their building. These are two quite useful abilities which are not available to the consuming-end in the community without the collection. Does that make sense to you guys? what do you think?

jquense:
I think it makes sense as a thing that is valuable to exist, but not as a replacement for our documentation and playground which is what the Readme change reads as. As a general tool, Bit seems cool, but I don't personally have any experience with it so don't really feel comfortable recommending it on the front page of the repo.

In terms of value added it doesn't seem like a much better solution to the recommended usage patterns in the docs, which also allows importing and extending only the components you want to use, but with the benefit of also being able to only include the css/styles for the components you use.

PR #3566, opened 2019-03-24
fix: modal dialog scrollable

jquense:


Bartozzz:
Please, don't forget to add:

```typescript
scrollable?: boolean;
```

to types definitions ([types/components/Modal.d.ts](https://github.com/react-bootstrap/react-bootstrap/blob/master/types/components/Modal.d.ts)). Without this, it will generate errors for TypeScript developers when trying to add the `scrollable` prop on `<Modal />`.

PR #3571, opened 2019-03-26
fix(Modal): set display block always, fixes #3399

### LINKED ISSUE #3399: react-bootstrap 1.0.0-beta.3 disabling animation in modal not working
#### Version and browser:
react-bootstrap@^1.0.0-beta.3
Electron so chrome/chromium (v66)

#### Usage:
      <Modal animation={false} show={show} onHide={onHide}>
        <Modal.Header closeButton>
          <Modal.Title>
            Some header...
          </Modal.Title>
        </Modal.Header>
        <Modal.Body>
          Some body paragraphs...
        </Modal.Body>
        <Modal.Footer>
          Some buttons ...
        </Modal.Footer>
      </Modal>

#### Expected
Modal shown when `show` is set to `true` no matter `animation` is set to `true` or `false`

#### But
When `animation` is not set (default is `true`), modal is shown properly (btw, animation not so smooth); when `false`, screen becomes completely black without the modal.

#### Observation
By looking at developer tools, I see that the `div[role=dialog]` element which is the parent of the modal `div.modal-dialog` has style `display: none`.

kt3k:
We started setting `display: block` to div.modal at onEnter timing of Transition at Pull #3187. But when we set `animation={false}` to Modal, the Transition object is not created and therefore onEnter (=handleEnter) is not called. This causes empty modal when animation set `false` (I think this is the issue #3399 ). This PR tries to fix this behavior by setting the modal's display style to `block` always.

I think this doesn't break existing behavior because as @jquense says
> we don't allow modals to be in the DOM prior to showing 

the modal's dom element doesn't exist before rendering. So I think it's safe to set display always `block`.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -226,7 +226,6 @@ class Modal extends React.Component {
 
   handleEnter = (node, ...args) => {
     if (node) {
-      node.style.display = 'block';
       this.updateDialogStyle(node);
     }
 
@@ -241,7 +240,6 @@ class Modal extends React.Component {
   };
 
   handleExited = (node, ...args) => {
-    if (node) node.style.display = ''; // RHL removes it sometimes
     if (this.props.onExited) this.props.onExited(...args);
 
     // FIXME: This should work even when animation is disabled.
@@ -340,7 +338,7 @@ class Modal extends React.Component {
             onExiting,
             manager,
             ref: this.setModalRef,
-            style: { ...style, ...this.state.style },
+            style: { ...style, display: 'block', ...this.state.style },
             className: classNames(className, bsPrefix),
             containerClassName: `${bsPrefix}-open`,
             transition: animation ? DialogTransition : undefined,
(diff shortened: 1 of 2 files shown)


src/Modal.js
```diff
@@ -340,7 +338,7 @@ class Modal extends React.Component {
             onExiting,
             manager,
             ref: this.setModalRef,
-            style: { ...style, ...this.state.style },
+            style: { ...style, display: 'block', ...this.state.style },
```
taion: i think it'd be best to do this only when `animation` is disabled

kt3k: OK, I'll update.

prakis:
I still see this issue, my module versions are
    "react-bootstrap": "1.0.0-beta.5",
    "bootstrap": "4.3.1",

Thank you

kt3k:
@prakis This change was first included in `v1.0.0-beta.7`. You need to update to that version or above.

prakis:
@kt3k Thank you. I am testing it.
When I set animation={false}  now the background-color is dark black(not the usual semi-transparent).

PR #3581, opened 2019-03-31
Fix ListGroup/Linked example

Oedipus:
Hi there,

` alertClicked()` function in ListGroup/Linked example wasn't used.  This PR adds onClick parameter to third item.


jquense:
thanks!

PR #3584, opened 2019-04-02
Value typing for Checkbox ToggleButtonGroup

mhermher:
Checkbox ToggleButtonGroup holds array of T rather than singleton.

### DIFF
--- a/types/components/ToggleButtonGroup.d.ts
+++ b/types/components/ToggleButtonGroup.d.ts
@@ -15,7 +15,7 @@ export interface ToggleButtonCheckboxProps<T> {
   type: 'checkbox';
   name?: string;
   value?: T[];
-   defaultValue?: T[];
+  defaultValue?: T[];
   onChange?: (value: T[]) => void;
 }
 
--- a/types/components/ToggleButtonGroup.d.ts
+++ b/types/components/ToggleButtonGroup.d.ts
@@ -14,7 +14,7 @@ export interface ToggleButtonRadioProps<T> {
 export interface ToggleButtonCheckboxProps<T> {
   type: 'checkbox';
   name?: string;
-  value?: T;
+  value?: T[];
   defaultValue?: T;
   onChange?: (value: T[]) => void;
 }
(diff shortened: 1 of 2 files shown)


types/components/ToggleButtonGroup.d.ts
```diff
@@ -14,7 +14,7 @@ export interface ToggleButtonRadioProps<T> {
 export interface ToggleButtonCheckboxProps<T> {
   type: 'checkbox';
   name?: string;
-  value?: T;
+  value?: T[];
   defaultValue?: T;
```
taion:
```suggestion
   defaultValue?: T[];
```

types/components/ToggleButtonGroup.d.ts
```diff
@@ -14,8 +14,8 @@ export interface ToggleButtonRadioProps<T> {
 export interface ToggleButtonCheckboxProps<T> {
   type: 'checkbox';
   name?: string;
-  value?: T;
-  defaultValue?: T;
+  value?: T[];
+  defaultValue?: T[];
```
jquense: wait no, these can be `T` or `T[]`, depending on whether the type is `checkbox` or `radio`

types/components/ToggleButtonGroup.d.ts
```diff
@@ -14,8 +14,8 @@ export interface ToggleButtonRadioProps<T> {
 export interface ToggleButtonCheckboxProps<T> {
   type: 'checkbox';
```
jquense: this is also wrong actually, this can be `radio` as well

types/components/ToggleButtonGroup.d.ts
mhermher: Checkbox and Radio Props have separate typings. The generalized typing is a union of those two. Is the radio supposed to be single-value and the checkbox multi-value? In that case, only the checkbox would have T[]. That is from my understanding at least.

taion: yup those all come from `ToggleButtonRadioProps`

mhermher:
Sorry, it says I dismissed your review Taion. Idk what that means, but a test was failing so I pushed another commit to fix that. That might have been the cause of that.

types/components/ToggleButtonGroup.d.ts
jquense: ha sorry, didn't notice that in the github diff view. tbf tho, the component always allows T or T[] regardless of type, but i'd agree it's better to have the types be stricter

PR #3586, opened 2019-04-02
fix: flip prop not passed to BaseDropdown Component.

tzanis:
Hello, I noticed that the flip pro is actually not passed to the BaseDropDown Component. Here is my suggestion.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -96,6 +96,7 @@ const Dropdown = React.forwardRef((uncontrolledProps, ref) => {
     alignRight,
     onSelect,
     onToggle,
+    flip,
     focusFirstItemOnShow,
     as: Component,
     navbar: _4,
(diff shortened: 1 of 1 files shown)


taion:
I'm not sure that it should be. It looks like this prop is only consumed by the dropdown menu.

tzanis:
I might be missing something but even though dropdown accepts such a prop, it does nothing with it. Then what is the reason of having this props anyway?

taion:
i'm not sure if it should be here.

maybe it should be passed down via context to the menu?

it'd be relevant for e.g. `<DropdownButton>`.

tzanis:
This PR is relevant to this issue: #3589 . I 've also added a live code example to show what I am trying to solve.

taion:
I don't think this actually resolves that issue, though. `<Dropdown>` is not the component that gets the `flip` prop.

bpas247:
Looks like `flip` just gets passed to the children of the `BaseDropdown` component
https://github.com/react-bootstrap/react-bootstrap/blob/84f0f1b39519f4218665d27b6a4cefc20c651466/src/Dropdown.js#L123-L133

jquense:
yes `flip` should be passed to the `DropdownMenu` not the dropdown, in the case of DropdownButton and friends we need to explicitly handle that

PR #3608, opened 2019-04-08
fix: DropDownItem prop-types active prop (fix: #3598)

### LINKED ISSUE #3598: active on DropDown.item and NavDropDown.item is not considered a boolean
This will result in an error telling you to not use a boolean, but use boolean .toString() instead. But that's just wrong. Active IS a boolean.
```
const boolean = true;
<Dropdown.Item active={boolean}>
     {text}
</Dropdown.Item>
```
(Worked fine in beta-5)

mxschmitt:
Fixes #3598.

I hope it's okay, that I simplified these `== null` and `!= null` statements.  (active var type is `null|bool` and `makeEventKey() is null|String`)

### DIFF
--- a/src/DropdownItem.js
+++ b/src/DropdownItem.js
@@ -65,6 +65,7 @@ const DropdownItem = React.forwardRef(
       href,
       onClick,
       onSelect,
+      active,
       as: Component,
       ...props
     },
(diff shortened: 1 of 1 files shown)


taion:
should we disable the "automatically dismiss reviews" thing? assuming that was an automatic dismissal

PR #3615, opened 2019-04-09
fix: Fix up accordions a bit

### LINKED ISSUE #3614: accordion animation is not smooth
This is smooth:
https://getbootstrap.com/docs/4.3/components/collapse/#accordion-example

This jumps and glitches: (in latest chrome)
https://react-bootstrap.netlify.com/components/accordion/

taion:
Fixes #3614

@bpas247 My bad, some confusion on my end in how our APIs worked

### DIFF
--- a/src/AccordionCollapse.js
+++ b/src/AccordionCollapse.js
@@ -13,15 +13,17 @@ const propTypes = {
   children: PropTypes.element.isRequired,
 };
 
-const AccordionCollapse = ({ children, eventKey, ...props }) => {
-  const contextEventKey = useContext(AccordionContext);
+const AccordionCollapse = React.forwardRef(
+  ({ children, eventKey, ...props }, ref) => {
+    const contextEventKey = useContext(AccordionContext);
 
-  return (
-    <Collapse in={contextEventKey === eventKey} {...props}>
-      {React.Children.only(children)}
-    </Collapse>
-  );
-};
+    return (
+      <Collapse ref={ref} in={contextEventKey === eventKey} {...props}>
+        <div>{React.Children.only(children)}</div>
+      </Collapse>
+    );
+  },
+);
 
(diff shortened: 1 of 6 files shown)


src/AccordionCollapse.js
```diff
@@ -13,15 +13,17 @@ const propTypes = {
...
+    const contextEventKey = useContext(AccordionContext);
 
-  return (
-    <Collapse in={contextEventKey === eventKey} {...props}>
-      {React.Children.only(children)}
-    </Collapse>
-  );
-};
+    return (
+      <Collapse ref={ref} in={contextEventKey === eventKey} {...props}>
```
taion: i forgot that `<Collapse>` didn't render a `<div>` on its own

src/AccordionToggle.js
```diff
@@ -30,9 +30,9 @@ const AccordionToggle = React.forwardRef(
     return (
       <Component
         ref={ref}
-        onClick={() => {
-          onSelect(eventKey);
-          if (onClick) onClick(eventKey);
+        onClick={e => {
```
taion: this is how the other `onClick` and `onSelect` things work

www/src/examples/Accordion/AllCollapse.js
```diff
@@ -1,10 +1,7 @@
 <Accordion>
   <Card>
     <Card.Header>
-      <Accordion.Toggle
-        onClick={key => console.log(`I've been clicked by : ${key}`)}
-        eventKey="0"
-      >
+      <Accordion.Toggle as={Button} variant="link" eventKey="0">
```
taion: so, the default of `<button>` here is a little weird in retrospect. the TWBS example upstream uses `<Button variant="link">`, as this makes things work, but i wonder if we should just use `<SafeAnchor>` instead

PR #3624, opened 2019-04-10
Add Accordion Types

### LINKED ISSUE #3601: Add Accordion TypeScript Types
As mentioned in #3409, the Accordion component is missing TS types. In order for this issue to be resolved, we'll need a TS type for the following components:

- [ ] Accordion
- [ ] AccordionToggle
- [ ] AccordionCollapse

bpas247:
Fixes #3601

### DIFF
--- a/types/components/AccordionToggle.d.ts
+++ b/types/components/AccordionToggle.d.ts
@@ -1,14 +1,14 @@
 import * as React from 'react';
-
+import Button from './Button';
 import { BsPrefixComponent } from './helpers';
 
 export interface AccordionToggleProps {
   eventKey: string;
-  onClick: (arg0?: React.SyntheticEvent) => any;
+  onClick?: (arg0?: React.SyntheticEvent) => any;
 }
 
 declare class AccordionToggle<
-  As extends React.ReactType = 'button'
-> extends BsPrefixComponent<As, AccordionToggleProps> {}
+  As extends React.ReactType = typeof Button
+  > extends BsPrefixComponent<As, AccordionToggleProps> { }
 
 export default AccordionToggle;
--- a/types/components/AccordionToggle.d.ts
+++ b/types/components/AccordionToggle.d.ts
@@ -4,7 +4,7 @@ import { BsPrefixComponent } from './helpers';
 
 export interface AccordionToggleProps {
   eventKey: string;
-  onClick: Function;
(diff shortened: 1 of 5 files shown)


types/simple.test.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 
 import {
   Alert,
+  Accordion,
```
bpas247: This line is throwing the following error
```
ERROR: 5:3   expect     TypeScript@next compile error: 
Module '"."' has no exported member 'Accordion'.
```
I feel like I'm missing something obvious here, I'm not sure what it is though...

mxschmitt: Try to add it also there: https://github.com/react-bootstrap/react-bootstrap/blob/fbc26bbb817629aeb8a6be953603e0e26e88189b/types/components/index.d.ts

Then it should be available for importing.

types/components/AccordionToggle.d.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import Button from './Button';
+import { BsPrefixComponent } from './helpers';
+
+export interface AccordionToggleProps {
+  eventKey: string;
+  onClick?: (arg0?: React.SyntheticEvent) => any;
```
taion:
```suggestion
  onClick?: (event?: React.SyntheticEvent) => void;
```

types/components/AccordionToggle.d.ts
```diff
@@ -0,0 +1,14 @@
...
+}
+
+declare class AccordionToggle<
+  As extends React.ReactType = typeof Button
```
taion: currently this is actually `'button'`, not our `Button`

bpas247: in the accordion example we're using `Button`
https://github.com/react-bootstrap/react-bootstrap/blob/8819727d423684bb69b0b45dd7b4a97a77d715f3/www/src/examples/Accordion/Basic.js#L4

taion: right, but the actual default is `button`: https://github.com/react-bootstrap/react-bootstrap/blob/8819727d423684bb69b0b45dd7b4a97a77d715f3/src/AccordionToggle.js#L23

though as noted in that PR, we should consider changing this to `<SafeAnchor>`

Issue #3536, opened 2019-03-11
Event key is being cast to string.


illogicz:
The documentation states it is an any type, but the event receives the passed value cast to a string. This does not seem desirable. Is there a reason for this?

jquense:
yes we really should be clearer about this in the docs! it needs to be serializable to a string because in lots of casuses the key is actually stored on the DOM node as a `data-*` attribute in order to coordinate focus and selection states. Yes this was the best way to do it, the more react-y way would be a lot more code and less reliable.

scrust:
Following up from #3957, in addition to updating the docs the [type definitions for `eventKey`](https://github.com/react-bootstrap/react-bootstrap/search?l=TypeScript&q=eventKey) will also need to be updated; it seems like most are already typed as `string`, `unknown`, or `any`, but the types for `DropdownItem` are _particularly_ misleading:

https://github.com/react-bootstrap/react-bootstrap/blob/d9429a93e989c69e019bffac37b8b0eb75ff3a7c/types/components/DropdownItem.d.ts#L7-L14

PR #3636, opened 2019-04-11
fix: use general string type for eventKey property

mxschmitt:
Fixes: #3536

But maybe we should allow Strings and Numbers. What do you think @jquense and @taion?


taion:
so as the comment says, it can be anything that can be neatly cast to string, so at least any primitive type. it should not be a POJSO or an array, most likely, but i'm not sure there's much we can impose beyond that, necessarily.

PR #3644, opened 2019-04-14
chore(ci): Using latest Chrome and Firefox releases to perform CI tests

### LINKED ISSUE #3635: Travis-CI fails sometimes
```
  <SplitButton>
    ✔ should open the menu when dropdown button is clicked
    ✔ should not open the menu when other button is clicked
    ✔ should invoke onClick when SplitButton.Button is clicked (prop)
    ✖ should not invoke onClick when SplitButton.Toggle is clicked (prop)
    ✔ should not invoke onClick when SplitButton.Toggle is clicked (prop) 
  ✖ "after each" hook for "Should pass disabled to both buttons"
Finished in 6.264 secs / NaN secs @ 19:12:32 GMT+0000 (Coordinated Universal Time)
SUMMARY:
✔ 351 tests completed
✖ 2 tests failed
FAILED TESTS:
  <SplitButton>
    ✖ should not invoke onClick when SplitButton.Toggle is clicked (prop)
      Firefox 56.0.0 (Linux 0.0.0)
    uncaught exception: AssertionError: expected 'next' to equal 'prev' (:0)
  ✖ "after each" hook for "Should pass disabled to both buttons"
    Firefox 56.0.0 (Linux 0.0.0)
  console.error.expected is undefined
  ./test/index.js/<@webpack:///test/index.js:63 <- test/index.js:158200:7
```

Example: https://travis-ci.org/react-bootstrap/react-bootstrap/builds/518914585?utm_source=github_status&utm_medium=notification

mxschmitt:
Before:

| Browser | Version |
| --------- | -------- |
| Firefox   | 56 |
| Chrome | 62 |

After: 

| Browser | Version |
| --------- | -------- |
| Firefox   | 66 |
| Chrome | 73 |

Should fix #3635. I guess just a Firefox bug which is now fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
 
 addons:
   chrome: stable
-  firefox: stable
+  firefox: latest
 
 before_install:
   - export DISPLAY=:99.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: false
 
+dist: xenial
+
 language: node_js
 node_js:
   - "10"
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ cache:
     - node_modules
     - www/node_modules
 
+before_install:	
+  - export CHROME_BIN=chromium-browser	
+  - export DISPLAY=:99.0
+
 install:
   - yarn bootstrap
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 sudo: false
 
-dist: xenial
-
 language: node_js
 node_js:
-  - "10"
+  - lts/*
 
 env:
   - BROWSER=ChromeCi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: false
 
 language: node_js
 node_js:
-  - "10"
+  - lts/*
```
taion: this seems to end up as 10 anyway?

.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: false
```
taion: can you remove the `sudo: false` above?

.travis.yml
mxschmitt: yep, but so we don't have to bump it each time, once Node.js will release a new LTS version.

taion: ah cool

PR #3664, opened 2019-04-16
Add htmlFor to label when rendered as column

### LINKED ISSUE #3656: Rendered `<label>` doesn't contain `for` attribute for `<Form.Label colum>`
When `Form.Label` has a property `column` set to true then the rendered `<label>` doesn't contain a `for` attribute.

It seems the defect was introduced in 1.0.0-beta.6 as it works as expected in 1.0.0-beta.5

Renders label without `for`
```
<Form.Group controlId="foo">
  <Form.Label column={true}>Foo</Form.Label>
  <Form.Control name="foo"/>
</Form.Group>

// renders to
<div class="form-group">
  <label class="form-label col-form-label col">Foo</label>
   <input name="foo" id="foo" class="form-control">
</div>
```

Renders label correctly with `for`
```
<Form.Group controlId="foo">
  <Form.Label>Foo</Form.Label>
  <Form.Control name="foo"/>
</Form.Group>

// renders to (for is there)
<div class="form-group">
  <label for="foo" class="form-label">Foo</label>
  <input name="foo" id="foo" class="form-control">
</div>
```

You can check it on the demo page: https://react-bootstrap.github.io/components/forms/ or in this sandbox: https://codesandbox.io/s/wq1r4m2x48?fontsize=14

jansteuer:
Fixes #3656

### DIFF
--- a/src/FormLabel.js
+++ b/src/FormLabel.js
@@ -67,12 +67,12 @@ const FormLabel = React.forwardRef(
 
     if (column)
       return (
-        <Col {...props} htmlFor={htmlFor} className={classes} as="label" />
+        <Col as="label" className={classes} htmlFor={htmlFor} {...props} />
       );
 
     return (
       // eslint-disable-next-line jsx-a11y/label-has-for, jsx-a11y/label-has-associated-control
-      <label {...props} htmlFor={htmlFor} ref={ref} className={classes} />
+      <label ref={ref} className={classes} htmlFor={htmlFor} {...props} />
     );
   },
 );
--- a/src/FormLabel.js
+++ b/src/FormLabel.js
@@ -59,28 +59,20 @@ const FormLabel = React.forwardRef(
       column && 'col-form-label',
     );
 
-    if (column)
-      return (
-        <Col
-          {...props}
-          htmlFor={htmlFor || controlId}
-          className={classes}
-          as="label"
-        />
-      );
-
     warning(
       controlId == null || !htmlFor,
       '`controlId` is ignored on `<FormLabel>` when `htmlFor` is specified.',
     );
+    htmlFor = htmlFor || controlId;
+
+    if (column)
+      return (
+        <Col {...props} htmlFor={htmlFor} className={classes} as="label" />
+      );
+
     return (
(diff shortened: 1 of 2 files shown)


src/FormLabel.js
```diff
@@ -59,7 +59,15 @@ const FormLabel = React.forwardRef(
...
+      return (
+        <Col
+          {...props}
+          htmlFor={htmlFor || controlId}
+          className={classes}
+          as="label"
+        />
+      );
 
     warning(
```
taion: move this warning before the `column` check, and DRY the computation of `htmlFor`

jansteuer: Done.

src/FormLabel.js
```diff
@@ -59,20 +59,20 @@ const FormLabel = React.forwardRef(
...
-
     warning(
       controlId == null || !htmlFor,
       '`controlId` is ignored on `<FormLabel>` when `htmlFor` is specified.',
     );
+    htmlFor = htmlFor || controlId;
+
+    if (column)
+      return (
+        <Col {...props} htmlFor={htmlFor} className={classes} as="label" />
```
mxschmitt:
```suggestion
        <Col as="label" className={classes} htmlFor={htmlFor} {...props} />
```
enhance logical order of prop types a bit :)

src/FormLabel.js
```diff
@@ -59,20 +59,20 @@ const FormLabel = React.forwardRef(
...
+
     return (
       // eslint-disable-next-line jsx-a11y/label-has-for, jsx-a11y/label-has-associated-control
-      <label
-        {...props}
-        htmlFor={htmlFor || controlId}
-        ref={ref}
-        className={classes}
-      />
+      <label {...props} htmlFor={htmlFor} ref={ref} className={classes} />
```
mxschmitt:
```suggestion
      <label ref={ref} className={classes} htmlFor={htmlFor} {...props} />
```

src/FormLabel.js
jansteuer: It goes a bit beyond a simple defect fix this PR is supposed to be but ok, do it rather now than never :-)

PR #3678, opened 2019-04-18
fix: ran prettier to fix CI

mxschmitt:
- Ran prettier to fix CI
- moved prettier Eslint rule into eslintrc file, so that it will be automatically recognized by Visual Studio Code e.g. and automatically fixed

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,5 +4,8 @@
   ],
   "plugins": [
     "prettier"
-  ]
+  ],
+  "rules": {
+    "prettier/prettier": "error"
+  }
 }
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "build-docs": "yarn --cwd www run build",
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s",
     "start": "yarn --cwd www run develop",
-    "format": "eslint . -- --fix",
+    "format": "eslint . --fix",
     "lint": "eslint .",
     "release": "rollout",
     "prepublishOnly": "npm run build",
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -4,5 +4,8 @@
   ],
   "plugins": [
     "prettier"
-  ]
+  ],
+  "rules": {
+    "prettier/prettier": "error"
```
taion: @jquense explicitly took this out earlier, which is why only ci-lint errors on this. do we want it back?

package.json
```diff
@@ -16,7 +16,7 @@
     "build-docs": "yarn --cwd www run build",
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s",
     "start": "yarn --cwd www run develop",
-    "ci-lint": "eslint --rule \"prettier/prettier: 2\"",
+    "ci-lint": "eslint",
```
taion: should drop this entirely and just use the normal eslint command if we're doing this

package.json
```diff
@@ -16,9 +16,8 @@
     "build-docs": "yarn --cwd www run build",
     "changelog": "conventional-changelog -p angular -i CHANGELOG.md -s",
     "start": "yarn --cwd www run develop",
-    "ci-lint": "eslint --rule \"prettier/prettier: 2\"",
-    "format": "npm run ci-lint . -- --fix",
-    "lint": "npm run ci-lint .",
+    "format": "eslint . -- --fix",
```
taion:
```suggestion
    "format": "eslint . --fix",
```

package.json
```diff
@@ -34,7 +33,7 @@
   },
   "lint-staged": {
     "*.js": [
-      "npm run ci-lint -- --fix",
+      "eslint -- --fix",
```
taion:
```suggestion
      "eslint --fix",
```

PR #3685, opened 2019-04-21
Added Toast component

mxschmitt:
Check out for more information:
https://getbootstrap.com/docs/4.3/components/toasts/

State: Please Review :)


jquense:
@mxschmitt looks great, let us know when you want a proper review 👍

test/ToastBodySpec.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+
+import Toast from '../src/Toast';
+
+describe('Toast.Body', () => {
```
mxschmitt: Not sure, why e.g. this test fails. Referring documentation: https://airbnb.io/enzyme/docs/api/ShallowWrapper/equals.html

mxschmitt:
@jquense @taion I've added the typings and some unit tests. Please feel free now to review :)
(I don't know why Netlify fails, but I don't think that I'm the cause of it)

jquense:
image
 is the netlify error, I forget the right way to fix this in the short term @taion

src/Toast.js
```diff
@@ -0,0 +1,73 @@
...
+  animation: true,
+  autohide: false,
+  delay: 3000,
+  dialogAs: ToastDialog,
+  show: true,
+  transition: Fade,
+};
+
+const Toast = ({ children, dialogAs: Dialog, ...props }) => (
+  <Dialog {...props}>{children}</Dialog>
```
jquense: since there is no logic in this component it seems like the `dialogAs` is probably an unnecessarily customization...what about removing `ToastDialog` and moving that code here as just `Toast`?

www/src/examples/Toast/PlacementMulti.js
```diff
@@ -0,0 +1,41 @@
...
+  <div
+    style={{
+      position: 'absolute',
+      top: 0,
+      right: 0,
+    }}
+  >
+    <Toast>
+      <Toast.Header>
+        <img
```
jquense: we've got a local version of placeholder: https://github.com/react-bootstrap/react-bootstrap/blob/master/www/src/components/PlaceholderImage.js that is in scope and a bit clearer to customize, e.g. no text, diff background color. See Card examples for usage

taion:
> image
> is the netlify error, I forget the right way to fix this in the short term @taion

This is actually a non-fatal warning. But it seems fixed now anyway?

mxschmitt:
> > image
> > is the netlify error, I forget the right way to fix this in the short term @taion
> 
> This is actually a non-fatal warning. But it seems fixed now anyway?

I think I fixed it with https://github.com/react-bootstrap/react-bootstrap/pull/3685/commits/648cc15ccb5b0e9c49fc6bda56f1f1c90d1757ca

Netlify truncates the logs at a specific amount of lines of logs afaik.

mxschmitt:
I'm not sure, if we should integrate some kind of Consumer/Provider pattern as a Util component, which does integrate e.g. (toastManager will be injected as prop)
```js
toastManager.show({
    title: "test Toast",
    body: "my body"
});
```

jquense:
@mxschmitt i'd hold off on the toast manager thing for now since it's not upstream but its a good candidate for a future enhancement

mxschmitt:
So @jquense what's the state of the coding, in my option I'm finished, have you already made a proper review? :)

src/CloseButton.js
```diff
@@ -10,12 +11,21 @@ const defaultProps = {
...
-  </button>
-));
+const CloseButton = React.forwardRef(
+  ({ label, onClick, className, ...props }, ref) => (
+    <button
+      ref={ref}
+      type="button"
+      className={classNames('close', className)}
+      onClick={onClick}
+      aria-label={label}
```
jquense: what was the rationale for adding aria-label here, theoretically the label below in the button would be read by a screen reader.

types/components/ToastDialog.d.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface ToastDialogProps {
+  animation?: boolean;
+  autohide?: boolean;
+  delay?: number;
+  onClose?: () => void;
+  show: boolean;
+  transition: boolean | React.ElementType,
+}
+
+declare class ToastDialog extends BsPrefixComponent<'div', ToastDialogProps> { }
```
jquense: out of sync now that there is no ToastDialog

src/CloseButton.js
mxschmitt: I just tried to implement the exact specification from the Bootstrap 4 docs: https://getbootstrap.com/docs/4.3/components/toasts/#basic

jquense: ah so notice the upstream doesn't also do the `sr-only` with hte label in the body of the button, which should be the same thing as aria-label in this case.

```js
<button type="button" class="ml-2 mb-1 close" data-dismiss="toast" aria-label="Close">
 <span aria-hidden="true">&times;</span>
</button>
```

src/Toast.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+class Toast extends React.Component {
+  constructor(props) {
+    super(props);
+    this.toastContext = {
+      onClose: () => this.props.onClose(),
+    };
+    const { autohide } = props;
+    if (autohide) {
```
taion: this should go in `componentDidMount`

taion: it should also only run if `show` is set

src/Toast.js
```diff
@@ -0,0 +1,140 @@
...
+      this.scheduleAutohideTimeout(props.delay);
+    }
+  }
+
+  componentDidUpdate(prevProps) {
+    const { autohide, delay, show } = this.props;
+    if (!this.autohideInterval && autohide && show !== prevProps.show) {
```
taion:
```suggestion
    if (!this.autohideInterval && autohide && show && !prevProps.show) {
```

or make `scheduleAutohideTimeout` into `maybeScheduleAutohideTimeout` and put the check there

src/Toast.js
taion: this would actually be easier with a functional component and `useEffect`

mxschmitt: Very good idea, I rewrote it for `useEffect`, works great!

src/Toast.js
mxschmitt: not relevant anymore due the useEffect change.

src/Toast.js
```diff
@@ -0,0 +1,116 @@
...
+  animation,
+  delay,
+  autohide,
+  onClose,
+  innerRef,
+  ...props
+}) => {
+  useEffect(() => {
+    if (autohide && show) {
+      setTimeout(() => {
```
taion: we need to cancel this timeout on unmount

src/Toast.js
```diff
@@ -0,0 +1,116 @@
...
+  show,
+  animation,
+  delay,
+  autohide,
+  onClose,
+  innerRef,
+  ...props
+}) => {
+  useEffect(() => {
+    if (autohide && show) {
```
taion: do something like:

```js
const showRef = useRef(false);

useEffect(() => {
  if (autohide && show && !showRef.current) {
    // ...
  }

  // ...
}, [autohide, show]);

showRef.current = show;
```

taion: or maybe we don't need `showRef`... but we don't want to set a new timer on changes to other props, i think

mxschmitt: I tried to e.g. to change the children values each 100ms and added console.log statements to setTimeout and clearTimeout. That worked well as expected.

src/Toast.js
```diff
@@ -0,0 +1,120 @@
...
+      role="alert"
+      aria-live="assertive"
+      aria-atomic="true"
+    >
+      {children}
+    </div>
+  );
+
+  const toastContext = {
+    onClose: () => onClose(),
```
taion:
```suggestion
    onClose,
```

src/Toast.js
taion: okay, this is probably fine then

Tyler8oliver:
@mxschmitt  I noticed that the Toast typescript defintion is not present in the index.d.ts file in this package. Wondering if there are plans to add typescript support for this component?

taion:
Feel free to add them! (:

Issue #3595, opened 2019-04-04
Many components are no longer compatible with styled-components


nguyenquyhy:
In react-bootstrap before version 1.0.0-beta.6, I can use styled-components to augment the style of react-bootstrap's components. For example:
 ```
const StyledCard = styled(Card)`
max-width: 100px;
`
```
which would render into
```<div class="card [class name from styled-component]">...</div>```

However, after upgrading to react-bootstrap 1.0.0-beta.6. It renders into
```<div class="[class name from styled-component]">...</div>```

This same issue also happens to other components such as FormControl.

**Comment by jquense** [comment]
We'd need more info...not much has changed in the latest beta, and components all still accept className

**Comment by nguyenquyhy** [comment]
This issue comes from this change https://github.com/react-bootstrap/react-bootstrap/commit/da675dec660bd3b74db72327f7c87331e0b219c9#diff-fcf0eccbd6ec0e88e8fe1ee47f3e4cd5. 

This is due to the behavior of styled-component on `forwardRef` and `defaultProps.as`: default `as` of `'div'` will get propagated up to `styled(Card)`, and styled-component will render a bare div instead of the Card component (https://github.com/styled-components/styled-components/issues/2325).
```
const defaultProps = {
    as: 'div',
    body: false,	
};
```

Is it possible to change to use as = 'div' in parameter list instead of using defaultProps? I know it is not exactly a bug for react-bootstrap, but I think compatibility with styled-components is quite important for many developers.

I have made an example for this.
https://codesandbox.io/s/lxo3pv1ry9

I can make a PR if you think the change is proper, or please let me know if you have any better idea for this.

jquense:
ah that really sounds like a bug in Styled components. We can maybe work around this in the short term but it seems kinda bad that SC works this way, its surprising

nguyenquyhy:
Yes I would love to see styled-components to change their behavior to be nicer to other libs as this issue would occurs for anyone using default `as` and `forwardRef`.

However, from the linked issue, I don't feel that they are going to care any time soon. I see your comments, and I hope they will now do something about that.

In the mean time I would really appreciate if react-bootstrap could try to circumvent this issue. I would love to continue using the beta and reporting bugs, but this issue would force me to stay on beta5.

jquense:
We can definitely work around it, I will say tho the fastest way to get there would be to put a PR together if you're up for it

nguyenquyhy:
[...]

PR #3715, opened 2019-04-28
Bugfix/styled components

KarlGe:
Fix to #3595 based on the suggestion by @nguyenquyhy 
- Moved all defaultProps "as" to props destructuring instead of defaultProps
- Added tests for *most* instances
- - Tests that defaultProps doesn't contain "as"
- - Tests that the component still renders with the correct default element
- - I Decided to leave the defaultProps even if it's now empty.

For any component that uses createBootstrapComponent I either used or made a const for the returned object and added the defaultProps to that. For InputGroup for example I added the defaultProps to the DecoratedInputGroup, instead of to the InputGroup directly. Thats necessary to check if the class contains "as" in defaultProps. **Though I'm not sure if this is the best way!**

Any class that had the defaultProps set as a static member of the class, has been moved out of the class definition and defined as ClassName.defaultProps = {}

Didn't manage to get those tests to work with Carousel as setting defaultProps on the DecoratedCarousel lead to
` Uncaught Error: Warning: Failed prop type: You have provided a 'activeIndex' prop to 'Carousel' without an 'onSelect' handler prop. This will render a read-only field. If the field should be mutable use 'defaultActiveIndex'. Otherwise, set 'onSelect'.`

Didn't make the tests for AbstractNav either, but maybe that's fine?

### DIFF
--- a/src/AbstractNav.js
+++ b/src/AbstractNav.js
@@ -27,9 +27,7 @@ class AbstractNav extends React.Component {
     activeKey: PropTypes.any,
   };
 
-  static defaultProps = {
-    as: 'ul',
-  };
+  static defaultProps = {};
 
   constructor(...args) {
     super(...args);
@@ -112,7 +110,7 @@ class AbstractNav extends React.Component {
 
   render() {
     const {
-      as: Component,
+      as: Component = 'ul',
       onSelect: _,
       parentOnSelect: _0,
       getControlledId: _1,
--- a/src/AbstractNav.js
+++ b/src/AbstractNav.js
@@ -27,8 +27,6 @@ class AbstractNav extends React.Component {
     activeKey: PropTypes.any,
   };
 
-  static defaultProps = {};
-
   constructor(...args) {
     super(...args);
 
@@ -110,6 +108,7 @@ class AbstractNav extends React.Component {
 
   render() {
     const {
+      // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
       as: Component = 'ul',
       onSelect: _,
       parentOnSelect: _0,
--- a/src/Jumbotron.js
+++ b/src/Jumbotron.js
@@ -4,22 +4,27 @@ import classNames from 'classnames';
 
 import { createBootstrapComponent } from './ThemeProvider';
 
-const propTypes = {
-  as: PropTypes.elementType,
-  /** Make the jumbotron full width, and without rounded corners */
-  fluid: PropTypes.bool,
-  /** @default 'jumbotron' */
-  bsPrefix: PropTypes.string,
-};
-
 const defaultProps = {
-  as: 'div',
   fluid: false,
 };
 
 class Jumbotron extends React.Component {
+  static propTypes = {
+    as: PropTypes.elementType,
+    /** Make the jumbotron full width, and without rounded corners */
+    fluid: PropTypes.bool,
+    /** @default 'jumbotron' */
+    bsPrefix: PropTypes.string,
+  };
+
   render() {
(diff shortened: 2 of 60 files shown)


src/AbstractNav.js
```diff
@@ -27,9 +27,7 @@ class AbstractNav extends React.Component {
     activeKey: PropTypes.any,
   };
 
-  static defaultProps = {
-    as: 'ul',
-  };
+  static defaultProps = {};
```
taion: remove the empty object stubs?

src/Jumbotron.js
```diff
@@ -4,22 +4,27 @@ import classNames from 'classnames';
...
-  bsPrefix: PropTypes.string,
-};
-
 const defaultProps = {
-  as: 'div',
   fluid: false,
 };
 
 class Jumbotron extends React.Component {
+  static propTypes = {
```
taion: maybe undo the changes to code style for statics?

src/Navbar.js
```diff
@@ -210,13 +209,13 @@ class Navbar extends React.Component {
...
 
 Navbar.propTypes = propTypes;
-Navbar.defaultProps = defaultProps;
 
 const DecoratedNavbar = createBootstrapComponent(
   uncontrollable(Navbar, { expanded: 'onToggle' }),
   'navbar',
 );
 
+DecoratedNavbar.defaultProps = defaultProps;
```
taion: why this change?

src/Row.js
```diff
@@ -40,4 +35,10 @@ class Row extends React.Component {
   }
 }
 
-export default createBootstrapComponent(Row, 'row');
+const BootstrapRow = createBootstrapComponent(Row, 'row');
+
+BootstrapRow.defaultProps = {
```
taion: why this change?

src/Spinner.js
```diff
@@ -83,4 +78,8 @@ class Spinner extends React.Component {
   }
 }
 
-export default createBootstrapComponent(Spinner, 'spinner');
+const BootstrapSpinner = createBootstrapComponent(Spinner, 'spinner');
+
+BootstrapSpinner.defaultProps = {};
```
taion: remove, probably

test/AccordionSpec.js
```diff
@@ -89,4 +89,12 @@ describe('<Accordion>', () => {
 
     expect(onClickSpy).to.be.calledOnce;
   });
+
+  it('Should define default "as" in prop destructuring instead of deafultProps', () => {
```
taion: i'm not sure these test cases really add much. i think it's fine as long as the old tests pass.

src/AbstractNav.js
KarlGe: Yeah, was unsure about that, I can definitely remove them!

src/Navbar.js
KarlGe: In order to pass the "Should define default "as" in prop destructuring" tests. If those tests aren't necessary then this doesn't have to be there.

src/Row.js
KarlGe: Same as Navbar, to pass the test

test/AccordionSpec.js
KarlGe: You're right, I was mostly thinking about having them in order to prevent someone from inadvertently putting the "as" into the defaultProps at a later point. This whole thing can be considered a bit weird, so I thought this would be an ok way to show why they're set up like they are. What do you think?

taion: I think it'd be more effective to add a comment in the places where you set the default, honestly. Otherwise nothing stops someone from carelessly removing the test as well.

taion:
fantastic, thanks!

PR #3731, opened 2019-05-01
fix(types): add bootstrap sizes for Form.Label as column

### LINKED ISSUE #3481: Form.Label prop typings is missing xs, sm, md, lg, xl properties

dalevfenton:
I'm not sure if just extending `FormLabelProps` with `ColProps` is the preferred way to solve this issue or if you would want something that actually enforces that the column size props are only valid when `column` is `true`.

I can make the typing stricter if that's preferred. It also seems like generally types aren't imported around and reused so if that's a problem I can copy them over completely from `Col.d.ts`

### DIFF
--- a/types/components/FormLabel.d.ts
+++ b/types/components/FormLabel.d.ts
@@ -17,7 +17,7 @@ export interface FormLabelWithColProps extends FormLabelBaseProps, ColProps {
   column: true;
 }
 
-export type FormLabelProps = FormLabelWithColProps|FormLabelOwnProps
+export type FormLabelProps = FormLabelWithColProps | FormLabelOwnProps
 
 declare class FormLabel extends BsPrefixComponent<'label', FormLabelProps> {}
 
--- a/types/components/FormLabel.d.ts
+++ b/types/components/FormLabel.d.ts
@@ -1,8 +1,9 @@
 import * as React from 'react';
 
 import { BsPrefixComponent } from './helpers';
+import { ColProps } from './Col';
 
-export interface FormLabelProps {
+export interface FormLabelProps extends ColProps {
   htmlFor?: string;
   column?: boolean;
   innerRef?: React.LegacyRef<this>;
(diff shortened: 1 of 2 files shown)


types/components/FormLabel.d.ts
```diff
@@ -1,8 +1,9 @@
 import * as React from 'react';
 
 import { BsPrefixComponent } from './helpers';
+import { ColProps } from './Col';
 
-export interface FormLabelProps {
+export interface FormLabelProps extends ColProps {
```
jquense: I wonder if the right way to do this is via a type Union, e.g. `FormLabel | FormLabelWithColProps` there `column` is `true` or `false` specifically (instead of `boolean`) to allow TS to differentiate

dalevfenton: I started to do that, but didn't see that throughout the other typings so I wasn't sure if there was a preferred style. Let me write it up with the type Union.

dalevfenton:
one thing to note is that since `FormLabelProps` is now a `type` instead of an `interface` if anyone is currently using that to extend their own interface it will break their interface declaration

dalevfenton:
@jquense I think this should be good to go now.

jquense:
thanks!

Issue #3741, opened 2019-05-03
Modal type file missing scrollable flag


chrisdusyk-iq:
When developing in Typescript, I was trying to create a `Modal`. According to the docs and the prop types, `scrollable` should be a boolean flag on the [component](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Modal.js#L59), but it's missing from the [type file](https://github.com/react-bootstrap/react-bootstrap/blob/master/types/components/Modal.d.ts#L15).

taion:
We'd be happy to take a PR here.

PR #3742, opened 2019-05-03
fix(types): add scrollable to ModalProps interface to match component

chrisdusyk-iq:
This is to resolve issue #3741 where the `Modal.d.ts` file is missing the `scrollable` flag in the `ModalProps` interface, while the `proptypes` for the component have the flag.


jquense:
thanks!

PR #3849, opened 2019-05-30
shouldn't submit unless form is validated

Sbphillips19:
Need the else statement to make sure the form doesn't submit if it is not validated.


Sbphillips19:
just need to include another if statement.  Documentation is already correct.  If anyone else faces same thing try:

if (form.checkValidity() === true) {
  // this is where code would be to submit
 }

Issue #3654, opened 2019-04-15
Carousel is not correctly sliding from first index to last in 'prev' direction


TheKvist:
When on the first slide and triggering a slide event in 'prev' direction, the carousel slides to the correct element, the last slide, but with a 'next' animation. This can, for example, be observed in the documentation here:

https://react-bootstrap.github.io/components/carousel/#example

The documentation uses 1.0.0-beta.6, but I haven't had any success with 1.0.0-beta.7 or 1.0.0-beta.8.
The effect is most apparent when repeatedly sliding to the previous slide. The transition from first to last is simply off.

For comparison the expected behavior from the Bootstrap carousel:

https://getbootstrap.com/docs/4.0/components/carousel/#with-controls

Unfortunately my knowledge in this area is so limited that I can't even find the event handler responsible for the actual transition, so I don't actually know where to look for this, but maybe one of you guys know why this is happening.

taion:
it's in https://github.com/react-bootstrap/react-bootstrap/blob/cf5c2a0c8e79bca05982826bc2aa466d928a6abf/src/Carousel.js#L153-L156

we need to handle the case where `previousActiveIndex === 0` and `nextActiveIndex >= lastPossibleIndex`

we'd be happy to take a PR here

zhangyu0218:
if ( 
   (nextIndex === 0 && previousActiveIndex >= lastPossibleIndex) || 
   ( previousActiveIndex !== 0 && previousActiveIndex <= nextIndex)
 )

harsh6646:
I've added the fix above and am in the process of adding tests for it. Will create a PR for it within the next day.

jeffreymeng:
I see that the fix above has been merged, but I'm still having this issue.

kosmiq:
Closing this since a fix has been merged.

janis-rasa:
React-bootstrap version 1.0.1, the problem is still present.

When switching slides (next button) from last to first slide, css classes are incorrectly assigned. Instead of the "carousel-item-left" class, the last slide is assigned the "carousel-item-right". Also, the first slide instead of the "carousel-item-left" class gets "carousel-item-right".
Exactly the same story when switching slides in reverse (prev) order. Instead of the "carousel-item-right" class, the first and last slides get the "carousel-item-left" classes.

https://codesandbox.io/s/runtime-grass-l2cg4?file=/package.json

gabrieltcm:
[...]

PR #3853, opened 2019-05-30
Fix sliding edge-case in Carousel (#3654)

bancho22:


bancho22:
I will need to do a better job with the test here. Any good suggestions?

test/CarouselSpec.js
```diff
@@ -238,4 +238,13 @@ describe('<Carousel>', () => {
     const wrapper = mount(<Carousel>{items}</Carousel>);
     wrapper.find('div').length.should.equal(4);
   });
+
+  it('Should move backwards if left control is clicked on the first item', () => {
+    const wrapper = mount(<Carousel defaultActiveIndex={0}>{items}</Carousel>);
+    wrapper.find('a.carousel-control-prev').simulate('click');
+
+    setTimeout(() => {
+      wrapper.assertSingle('.carousel-item-left');
```
taion: This assertion isn't actually going to fire, because the test case isn't async.

Perhaps make it async? Or else take a `done` callback in the fn.

mxschmitt: You could also use 
````
const clock = sinon.useFakeTimers({
      toFake: ['setTimeout'],
});
// some logic
clock.tick(100);
// check if X has happened
```

src/Carousel.js
```diff
@@ -151,7 +151,8 @@ class Carousel extends React.Component {
       let direction;
       if (
         (nextIndex === 0 && previousActiveIndex >= lastPossibleIndex) ||
-        previousActiveIndex <= nextIndex
+        (previousActiveIndex <= nextIndex &&
+          !(previousActiveIndex === 0 && nextIndex >= lastPossibleIndex))
```
alvinbabujohn: Not a bad fix, but still will have issues when children=2, no?

mxschmitt:
@bancho22 what's the state of this PR?

Could you fix the review comments? otherwise we have to close the PR.

test/CarouselSpec.js
bancho22: Thanks! Turns out I planned the test case all wrong. I don't really see a good way to test this.

src/Carousel.js
bancho22: If children=2, then lastPossibleIndex will always be equal to 1 and so would nextIndex.
Given that nextIndex is always 1, if previousActiveIndex is indeed zero or one which it would be, then the if condition would always be satisfied.
Thus, you are correct, however, I believe it's in a better condition this way as at least we fix the most common use cases.

bancho22:
@mxschmitt There is a problem with the Travis CI build that I don't understand. Otherwise, one of the comments, I resolved, while the other one seems like it would be unresolvable, unless we agree we can live with that small issue.

mxschmitt:
> @mxschmitt There is a problem with the Travis CI build that I don't understand. Otherwise, one of the comments, I resolved, while the other one seems like it would be unresolvable, unless we agree we can live with that small issue.

The Travis CI issue is fixed.

taion:
This logic is super screwy, BTW. `handleNext` and `handlePrev` actually call `onSelect` with the direction, but the gDSFP logic just ignores that as far as I can tell.

taion:
superseded by https://github.com/react-bootstrap/react-bootstrap/pull/4917

PR #3888, opened 2019-06-10
Fixed carousels slide issue

### LINKED ISSUE #3654: Carousel is not correctly sliding from first index to last in 'prev' direction

zbc:
Closes #3654


taion:
this is actually a dupe of https://github.com/react-bootstrap/react-bootstrap/pull/3853

zbc:
@taion Sorry didn't see that one.

Issue #3889, opened 2019-06-10
Increase Test Coverage


mxschmitt:
Current State:
https://codecov.io/gh/react-bootstrap/react-bootstrap/tree/master/src

- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/BootstrapModalManager.js (checkout: #4206)
- [ ] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/OverlayTrigger.js (see #4573 for help)
- [ ] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Carousel.js
- [ ] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Row.js
- [ ] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Modal.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Toast.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Overlay.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/TabContainer.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/AbstractNav.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Nav.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/DropdownMenu.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/InputGroup.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/PageItem.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Tab.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/Tabs.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/ToggleButton.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/ToggleButtonGroup.js
- [x] https://codecov.io/gh/react-bootstrap/react-bootstrap/src/master/src/ListGroupItem.js

zdivozzo:
[...]

PR #3893, opened 2019-06-11
tests: increase test coverage on various components

mxschmitt:
#3889


test/ToastSpec.js
```diff
@@ -45,4 +45,14 @@ describe('Toasts', () => {
     clock.tick(500);
     expect(onCloseSpy).to.be.calledOnce;
   });
+
+  it('should clearTimeout after unmount', () => {
```
taion: should this assert something?

mxschmitt: not sure if it makes sense. But it will cover that: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Toast.js#L77-L79

Do you think I can check if the timeout was cleared? hmm

taion: could you manually trigger timers with jest, but check that `onClose` is not called?

taion: sinon, rather, not jest

mxschmitt: currently we won't fire `onClose` on `componentWillUnmount` not sure if this makes sense.

In the Modal it will be thrown: https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.js#L257-L259

taion: i mean, that `onClose` does _not_ get fired is a good way to ensure that the timeout gets cleared, right?

if the timeout _didn't_ get cleared, then `onClose` might get called

taion: can you add a fake clock here and make it tick, as with the above?

mxschmitt:
@taion should be ready, but I totally don't get it why some other tests are not passing.

src/Toast.js
```diff
@@ -69,17 +69,28 @@ const Toast = ({
...
-  }, [autohide, show, delay, onClose]);
+
+    const autohideHandle = setTimeout(() => {
+      onCloseRef.current();
+    }, delayRef.current);
+
+    return () => {
+      clearTimeout(autohideHandle);
+    };
+  }, [autohide, show]);
```
taion: We don't actually want to restart the timer if `delay` or `onClose` change on us, I think.

test/ToastSpec.js
```diff
@@ -28,37 +29,43 @@ describe('Toasts', () => {
...
+      mount(
+        <Toast onClose={onCloseSpy} delay={500} show autohide>
+          <Toast.Header>header-content</Toast.Header>
+          <Toast.Body>body-content</Toast.Body>
+        </Toast>,
+      );
+      clock.tick(1000);
+      expect(onCloseSpy).to.have.been.calledOnce;
+    } finally {
+      clock.restore();
```
taion: this is why things were failing; you need to clear the fake timers

test/ToastSpec.js
```diff
@@ -28,37 +29,43 @@ describe('Toasts', () => {
...
+    }
   });
 
   it('should clearTimeout after unmount', () => {
-    const clock = sinon.useFakeTimers({
-      toFake: ['setTimeout'],
```
taion: this would have failed because it wasn't mocking `clearTimeout`

mxschmitt:
Thanks @taion :)

PR #3907, opened 2019-06-15
style: Clean up FormText code and tests

taion:


test/FormTextSpec.js
```diff
@@ -23,13 +23,8 @@ describe('<FormText>', () => {
...
-        .find('small')
-        .hasClass('form-text'),
-    ).to.equal(true);
-    expect(
-      mount(<FormText muted />)
-        .find('small')
+        .find('small.form-text')
         .hasClass('text-muted'),
-    ).to.equal(true);
+    ).to.be.true;
```
jquense: unused expressions?

taion: apparently we don't lint for it? we have `.should.be.true` elsewhere. should we use dirty-chai?

taion: apparently we don't lint for it? we have `.should.be.true` elsewhere. should we use dirty-chai?

jquense: probably, I don't think anyone felt like updating

PR #3914, opened 2019-06-18
Add a section in the README for CodeSandbox examples

bhekanik:


bhekanik:
@mxschmitt What exactly should I update on the issue templates?

mxschmitt:
> @mxschmitt What exactly should I update on the issue templates?

That the link which I've sent is directly in the Issue template. But I can do it also :)

bhekanik:
Okay, I think I get you now, let me update it.

PR #3915, opened 2019-06-19
Refactor ThemeProvider

bpas247:
The ThemeProvider should be refactored to a functional component
with hooks. I decided to useMemo() the `prefixes` value, to avoid
unnecessary computations if the value doesn't change between
renders.

`prefixes` was re-written to be an object instead of a Map object,
as it allows for a more simple shallow copy implementation.

Also, duplicate code was removed in createBootstrapComponent().


src/ThemeProvider.js
```diff
@@ -1,31 +1,29 @@
...
-    Object.keys(this.props.prefixes).forEach(key => {
-      this.prefixes.set(key, this.props.prefixes[key]);
-    });
-  }
-
-  render() {
-    return <Provider value={this.prefixes}>{this.props.children}</Provider>;
-  }
-}
+const ThemeProvider = props => {
```
taion:
```suggestion
function ThemeProvider({ prefixes, children }) {
```

bpas247:
> is it that bad to just copy over the forEach to populate the map?

I wouldn't say it's "bad", it just seems a bit more complex of an implementation than it needs to be. If an Object doesn't fit all of the use cases for this component, then I can revert it back to a Map object.

src/ThemeProvider.js
```diff
@@ -1,31 +1,29 @@
...
-      this.prefixes.set(key, this.props.prefixes[key]);
-    });
-  }
-
-  render() {
-    return <Provider value={this.prefixes}>{this.props.children}</Provider>;
-  }
+function ThemeProvider({ prefixes, children }) {
+  const prefixCopy = useMemo(() => {
+    const outPrefixes = {};
```
taion: Okay, if we're not using the `Map`, then we can just simplify this as:

```js
const copiedPrefixes = useMemo(() => ({ ...prefixes }), [prefixes]);
```

PR #3982, opened 2019-07-02
Migrate doc examples to functional components

bpas247:
With the release of React hooks, it seems appropriate to migrate our simple controlled component examples in our docs to use them. This will give users straightforward examples on how to use our API with React hooks.

Note: Although this is a complete re-write of most of the class component examples, there are still class component examples in our docs that we won't be able to migrate to function components. This is mostly due to them implementing parts of our API that requires the example components to be instantiated (e.g. passing in the `container` prop for `Overlay` shown in this [example](https://github.com/react-bootstrap/react-bootstrap/blob/b0ed370f667d4d8afdfd59375f4240cc653380b1/www/src/examples/Overlays/PopoverContained.js#L23)).

### DIFF
--- a/www/.eslintrc.js
+++ b/www/.eslintrc.js
@@ -32,6 +32,7 @@ module.exports = {
       rules: {
         'comma-dangle': 'off',
         'no-console': 'off',
+        'no-undef': 'off'
       },
     },
   ],
--- a/www/.eslintrc.js
+++ b/www/.eslintrc.js
@@ -32,7 +32,6 @@ module.exports = {
       rules: {
         'comma-dangle': 'off',
         'no-console': 'off',
-        'no-undef': 'off'
       },
     },
   ],
--- a/www/src/examples/Alert/Dismissible.js
+++ b/www/src/examples/Alert/Dismissible.js
@@ -1,28 +1,19 @@
-class AlertDismissibleExample extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      show: true,
-    };
-  }
+function AlertDismissibleExample() {
+  const [show, setShow] = React.useState(true);
 
-  render() {
-    const handleDismiss = () => this.setState({ show: false });
-    const handleShow = () => this.setState({ show: true });
-    if (this.state.show) {
-      return (
-        <Alert variant="danger" onClose={handleDismiss} dismissible>
-          <Alert.Heading>Oh snap! You got an error!</Alert.Heading>
-          <p>
-            Change this and that and try again. Duis mollis, est non commodo
-            luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit.
-            Cras mattis consectetur purus sit amet fermentum.
-          </p>
-        </Alert>
-      );
-    }
-    return <Button onClick={handleShow}>Show Alert</Button>;
+  if (show) {
+    return (
+      <Alert variant="danger" onClose={() => setShow(false)} dismissible>
+        <Alert.Heading>Oh snap! You got an error!</Alert.Heading>
+        <p>
+          Change this and that and try again. Duis mollis, est non commodo
+          luctus, nisi erat porttitor ligula, eget lacinia odio sem nec elit.
+          Cras mattis consectetur purus sit amet fermentum.
+        </p>
+      </Alert>
+    );
   }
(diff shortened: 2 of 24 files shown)


www/src/examples/Alert/Dismissible.js
```diff
@@ -1,28 +1,19 @@
-class AlertDismissibleExample extends React.Component {
-  constructor(props) {
-    super(props);
-    this.state = {
-      show: true,
-    };
-  }
+function AlertDismissibleExample() {
+  const [show, setShow] = React.useState(true);
```
taion: Should we modify our setup such that e.g. `useState` can be called directly by name, rather than as `React.useState`? The former convention seems more common.

www/src/examples/Button/ToggleButtonGroupControlled.js
```diff
@@ -1,31 +1,15 @@
...
-        <ToggleButton value={1}>Option 1</ToggleButton>
-        <ToggleButton value={2}>Option 2</ToggleButton>
-        <ToggleButton value={3}>Option 3</ToggleButton>
-      </ToggleButtonGroup>
-    );
-  }
+function ToggleButtonGroupControlled() {
+  const [value, setValue] = React.useState([1, 3]);
+
+  const handleChange = val => setValue(val);
```
taion: Can we just do `onChange={setValue}` below?

www/src/examples/Carousel/Controlled.js
```diff
@@ -1,71 +1,57 @@
-class ControlledCarousel extends React.Component {
-  constructor(props, context) {
-    super(props, context);
+function ControlledCarousel() {
+  const [state, setState] = React.useState({ index: 0, direction: null });
```
taion: Is it more idiomatic to have separate `index` and `direction` state variables here?

www/src/examples/Overlays/Overlay.js
```diff
@@ -1,50 +1,39 @@
-class Example extends React.Component {
-  constructor(...args) {
-    super(...args);
+function Example() {
+  const [show, setShow] = React.useState(false);
+  const [target, setTarget] = React.useState(null);
```
taion: This should be a `useRef`, I think?

www/src/examples/Overlays/TooltipOverlay.js
```diff
@@ -1,31 +1,23 @@
-class Example extends React.Component {
-  constructor(...args) {
-    super(...args);
+function Example() {
+  const [show, setShow] = React.useState(false);
+  const [target, setTarget] = React.useState(null);
```
taion: `useRef`?

www/src/examples/Button/ToggleButtonGroupControlled.js
bpas247: hmm, I don't think that will work since `ToggleButtonGroup` passes 2 arguments to the `onChange` prop when it gets fired.
https://github.com/react-bootstrap/react-bootstrap/blob/391af529b750a106264c37871bdff6d7df00bc89/src/ToggleButtonGroup.js#L58
Technically this should be written as
```js
const handleChange = (val, event) => setValue(val);
```
or more simply
```js
const handleChange = (val, ) => setValue(val);
```

taion: ah okay

www/.eslintrc.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
       rules: {
         'comma-dangle': 'off',
         'no-console': 'off',
+        'no-undef': 'off'
```
taion: don't do this – add the new bindings to `www/examples/.eslintrc`

www/src/examples/Button/ToggleButtonGroupControlled.js
```diff
@@ -1,31 +1,15 @@
...
-        <ToggleButton value={1}>Option 1</ToggleButton>
-        <ToggleButton value={2}>Option 2</ToggleButton>
-        <ToggleButton value={3}>Option 3</ToggleButton>
-      </ToggleButtonGroup>
-    );
-  }
+function ToggleButtonGroupControlled() {
+  const [value, setValue] = useState([1, 3]);
+
+  const handleChange = (val, event) => setValue(val);
```
taion: maybe fine to change this back and just add a comment

PR #3991, opened 2019-07-03
chore: Fix and update tooling

taion:


www/config.js
```diff
@@ -1,5 +1,8 @@
+// Working around benmosher/eslint-plugin-import#1396.
+/* eslint-disable import/no-extraneous-dependencies */
```
bpas247: Hmm, I thought [this](https://github.com/react-bootstrap/react-bootstrap/pull/3977/commits/4957ab6ea9d1901604946dcbc1342064d110e200) commit fixed the `import/no-extraneous-dependencies` linting issue.

taion: Yes, but it's doing so at the cost of using deprecated modules on npm that we shouldn't be using, e.g. https://www.npmjs.com/package/crypto.

This is actually an upstream bug; we shouldn't be shadowing the built-ins.

bpas247: Maybe we can update this branch and fix the upstream bug in this PR?

taion: I mean, I looked at it a bit, but I couldn't figure out why this was happening

taion:
okay, not actually a bug – we just needed to switch to the `node` import resolver. can someone take a look?

PR #3993, opened 2019-07-03
refactor: state constructor simplifications

mxschmitt:


src/Carousel.js
```diff
@@ -120,17 +120,14 @@ const defaultProps = {
...
-    this.state = {
-      prevClasses: '',
-      currentClasses: 'active',
-    };
-    this.isUnmounted = false;
+  isUnmounted = false;
 
-    this.carousel = React.createRef();
-  }
+  carousel = React.createRef();
```
mxschmitt: not sure if something like this is OK?

src/Modal.js
```diff
@@ -183,14 +183,11 @@ function BackdropTransition(props) {
...
-    super(props, context);
+  state = { style: {} };
 
-    this.state = { style: {} };
-    this.modalContext = {
-      onHide: () => this.props.onHide(),
-    };
-  }
+  modalContext = {
+    onHide: () => this.props.onHide(),
```
mxschmitt: is the context here okay? (onHide method)

src/Navbar.js
```diff
@@ -123,8 +123,8 @@ const defaultProps = {
 };
 
 class Navbar extends React.Component {
-  constructor(...args) {
-    super(...args);
+  constructor(props) {
```
mxschmitt: changed, so that we have a consistency in the project.

src/Toast.js
```diff
@@ -79,7 +79,7 @@ const Toast = ({
       };
     }
     return () => null;
-  }, [autohide, show]);
+  }, [autohide, delay, onClose, show]);
```
mxschmitt: Was a ESLint warning

src/Carousel.js
bpas247: maybe the assignments could go under cDM? probably doesn't matter too much with `this.isUnmounted`, since it's just a primitive.

taion: this is fine

src/Modal.js
taion: yup this works

src/Navbar.js
taion: why not make `state` an instance property here as well?

src/Toast.js
taion: hmm, this was deliberate though. we didn't want to restart the timer in the event that `delay` changed. though `onClose` should probably take the latest value via a ref

src/Navbar.js
mxschmitt: because then the context of it is incorrect. So I have to create a wrap function:
`onToggle: () => this.handleToggle(),`

PR #4003, opened 2019-07-08
feat: Add Float Classes

### LINKED ISSUE #3744: Implement float classes
Toggle a float with a class as shown [here](https://getbootstrap.com/docs/4.0/utilities/float/#classes).

```html
<div class="col float-left">Float left on all viewport sizes</div><br>
<div class="col float-right">Float right on all viewport sizes</div><br>
<div class="col float-none">Don't float on all viewport sizes</div>
```

I would write it like:
```jsx
<Col float='left'>Float left on all viewport sizes</Col><br />
<Col float='right'>Float right on all viewport sizes</Col><br />
<Col>Don't float on all viewport sizes</Col>
```

bpas247:
This is an initial attempt at supporting float classes functionality as per the [upstream docs](https://getbootstrap.com/docs/4.3/utilities/float/).

The implementation attempts to do this by injecting the functionality in a utility HOC used by some of the components (`createBootstrapComponent`). However, not all components utilize this HOC, so some of the components might have to be migrated to use it, or the functionality might need to be injected into another utility HOC used by components.

Fixes #3744.

### DIFF
--- a/src/ThemeProvider.js
+++ b/src/ThemeProvider.js
@@ -36,44 +36,44 @@ function createBootstrapComponent(Component, opts) {
   // If it's a functional component make sure we don't break it with a ref
   const { prefix, forwardRefAs = isClassy ? 'ref' : 'innerRef' } = opts;
 
-  return forwardRef(
-    // eslint-disable-next-line react/prop-types
-    ({ className, float, ...props }, ref) => {
-      props[forwardRefAs] = ref;
-      const prefixes = useContext(ThemeContext);
-      let floats = null;
-      if (float && Object.keys(float).length > 0) {
-        floats = Object.entries(float).reduce(
-          (acc, [property, value]) =>
-            acc +
-            (property !== 'default'
-              ? `float-${property}-${value} `
-              : `float-${value} `),
-          '',
-        );
-      }
-      return (
-        <Component
-          {...props}
-          className={classNames(className, floats)}
-          // eslint-disable-next-line react/prop-types
-          bsPrefix={props.bsPrefix || prefixes.get(prefix) || prefix}
-        />
+  const BootstrapComponent = ({ className, float, ...props }, ref) => {
+    props[forwardRefAs] = ref;
+    const prefixes = useContext(ThemeContext);
+    let floats = null;
+    if (float && Object.keys(float).length > 0) {
+      floats = Object.entries(float).reduce(
+        (acc, [property, value]) =>
+          acc +
+          (property !== 'default'
+            ? `float-${property}-${value} `
+            : `float-${value} `),
+        '',
       );
-    },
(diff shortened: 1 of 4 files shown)


jquense:
neat idea. I'm not sure how a feel about this generally tho. We might run into feature creep and pop namespace conflicts. For instance, why `float-*` in particular here, and not the 100 other utility classes? Or is this a significant api improvement over `className='float-md-right'` ya know?

microdou:
It looks like `float` is depreciated.
https://blog.getbootstrap.com/2019/02/11/bootstrap-4-3-0/
**Deprecated: `img-retina`, `invisible`, `float`, and `size` mixins are now deprecated and will be removed in v5.**

taion:
The classes aren't deprecated – just the Sass mixins.

bpas247:
> For instance, why float-* in particular here, and not the 100 other utility classes? 

Well the issue that this PR is aiming to fix (#3744) was the initial motivation, and I personally wanted to see what our API would look like if we attempted to support the upstream utility classes.

> Or is this a significant api improvement over className='float-md-right' ya know?

After implementing the first draft of this, I realized a few things about the state of our codebase:

### Some of our components use the utility HOC `ThemeProvider.createBootstrapComponent`

This utility HOC seems to be a simple wrapper to provide consistency for the `bsPrefix` prop API, as well as providing support for forwarding refs.

#### Some components following this pattern: 
- Carousel
- Tooltip
- Toast

### Some of our components use the utility HOC `createWithBsPrefix`

This utility HOC seems to cover almost the exact same use cases as `ThemeProvider.createBootstrapComponent`, with the additional preliminary support of the `as` prop.

#### Some components following this pattern: 
- Card
- Card Columns
- Card Deck
- Dropdown

### Some of our components, use neither utility HOC

Which basically leads them to implement the use cases of both of the utility HOCs themselves. 


#### Some components following this pattern: 
- Accordion
- Row
- Col

This leads me to conclude that the **cost of migrating all of our components to use a standardized HOC utility to support this feature (`float` prop) is more costly than the value of this API improvement.**

However, this does not mean we shouldn't consider migrating our components to use a single standardized HOC utility, which I believe we should open an issue to discuss that further :)

jquense:
I do think you're right that it's worth consolidating helper usages where possible. To clarify a bit, `createWithBsPrefix` isn't really an HoC since it doesn't wrap a passed in component, instead it creates a component from some settings. If you imagine the alternative to it would be to inline the forwardRef code in that helper any place we use it, idk that saves us much. `createBootstrapComponent` is now just a `useBootstrapPrefix` alternative, for components that haven't been converted to hooks yet. Ideally we kill that one and just use hooks but some components won't convert easily, e.g. `Carousel` which depends heavily on the useState callback

bpas247:
I'm gonna go ahead and close this issue, as we're not going to be moving forward with this API change.

PR #4018, opened 2019-07-09
feat: Add Carousel touch support

### LINKED ISSUE #3683: Add touch support for carousels

bpas247:
Touch support should be provided for our Carousel component, as
per the [upstream docs for Carousel](https://getbootstrap.com/docs/4.3/components/carousel/#options) (as of Bootstrap v4.2).

Fixes #3683.

### DIFF
--- a/src/Carousel.js
+++ b/src/Carousel.js
@@ -235,8 +235,6 @@ class Carousel extends React.Component {
   };
 
   handleTouchEnd = e => {
-    const lastPossibleIndex = countChildren(this.props.children) - 1;
-
     // If the swipe is under the threshold, don't do anything.
     if (
       Math.abs(e.changedTouches[0].screenX - this.state.touchStartX) <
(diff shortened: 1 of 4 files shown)


jquense:
Doesn't look like upstream uses Hammer.js what do you think about implementing this with straight native events like they do instead of adding a new dependency?

test/CarouselSpec.js
```diff
@@ -287,4 +287,24 @@ describe('<Carousel>', () => {
...
+
+    try {
+      const onSelectSpy = sinon.spy();
+      const wrapper = mount(
+        <Carousel interval={0} onSelect={onSelectSpy} touch>
+          {items}
+        </Carousel>,
+      );
+
+      wrapper.simulate('touchStart', { changedTouches: [{ screenX: 0 }] });
```
bpas247: Hmm, can't seem to figure out why this isn't firing the touch event handler...

jquense: Are you manually attaching touch handlers or using the react event system? Enzyme will simulate at the react level, but not actually fire Dom events

src/Carousel.js
```diff
@@ -221,6 +230,39 @@ class Carousel extends React.Component {
...
+    // If the swipe is under the threshold, don't do anything.
+    if (
+      Math.abs(e.changedTouches[0].screenX - this.state.touchStartX) <
+      SWIPE_THRESHOLD
+    )
+      return;
+
+    // Going left
+    if (e.changedTouches[0].screenX < this.state.touchStartX) {
+      this.to(
```
taion: does it make sense to use `handlePrev` and `handleNext` here?

xyzr07:
thank you

Issue #3805, opened 2019-05-17
Typescript definition for ref attribute is incorrect


kumar303:
## Describe the bug

The `ref` attribute on components like `Alert` (and others) does not seem to have a correct  typescript definition.

## To Reproduce

1. Check out https://github.com/react-bootstrap/react-bootstrap/commit/84f0f1b39519f4218665d27b6a4cefc20c651466
2. Apply the following patch

```diff
diff --git a/types/simple.test.tsx b/types/simple.test.tsx
index 42e1311e..787c9ea5 100644
--- a/types/simple.test.tsx
+++ b/types/simple.test.tsx
@@ -39,6 +39,14 @@ import {
 <Alert.Link as="a" href="blah" />;
 <Alert.Heading as="h3" />;
 
+<Alert as="div" ref={(element: HTMLDivElement | null) => console.log(element)}>
+  Alert with callback ref
+</Alert>;
+
+<Alert as="div" ref={React.createRef<HTMLDivElement>()}>
+  Alert with ref object
+</Alert>;
+
 <Accordion defaultActiveKey="0">
   <Card>
     <Card.Header>
```

3. Run `cd types && ../node-modules/.bin/tsc`
4. See error (below)

<details>
<summary>tsc error</summary>

```
simple.test.tsx:42:17 - error TS2322: Type '(element: HTMLDivElement | null) => void' is not assignable to type 'string | ((instance: Alert | null) => void) | RefObject<Alert> | null | undefined'.
  Type '(element: HTMLDivElement | null) => void' is not assignable to type '(instance: Alert | null) => void'.
    Types of parameters 'element' and 'instance' are incompatible.
      Type 'Alert | null' is not assignable to type 'HTMLDivElement | null'.
        Type 'Alert' is missing the following properties from type 'HTMLDivElement': align, addEventListener, removeEventListener, accessKey, and 238 more.

42 <Alert as="div" ref={(element: HTMLDivElement | null) => console.log(element)}>
                   ~~~

  ../node_modules/@types/react/index.d.ts:97:9
    97         ref?: LegacyRef<T>;
               ~~~
    The expected type comes from property 'ref' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Alert> & Readonly<AlertProps> & Readonly<{ children?: ReactNode; }>'

[...]

Issue #3540, opened 2019-03-14
Wrong example code blocks in statically built docs pages


thatzacdavis:
On [the introduction page](https://react-bootstrap.github.io/getting-started/introduction/) the code examples that are shown for `Importing` and `Stylesheets` are identical.

I'm assuming that the stylesheets example should be updated to use the CDN as described in the text.

jquense:
this seems to be a gatsby or react bug. The code is correct and if you navigate to the page from another docs page it renders correctly

thatzacdavis:
Yes, I was looking at the docs themselves to try and submit a fix and noticed that the code looks correct.

mxschmitt:
Issue is still reproduceable: 
![image](https://user-images.githubusercontent.com/17984549/54690576-ffda2080-4b21-11e9-9dd1-d877f684a17b.png)
https://react-bootstrap.github.io/getting-started/introduction/

jquense:
likely because the docs haven't been redeployed

mxschmitt:
oh okay, I thought auto deployment is enabled from the `master` branch. So we deploy to Netlify probably after the release right? 👍

jquense:
netlify deploys continuously but the gh-pages site is a manual push

jquense:
siiigh still an issue. I'm really unsure where the bug is, but its super subtle and not in our code.

thatzacdavis:
Yeah, I tried looking into it as well. Pretty bizarre behavior happening here :/

mxschmitt:
Fixed the rendering bug due new deployment to gh-pages.

jquense:
@mxschmitt its not fixed :/

corysimmons:
I'm a firm believer that almost every project on Github should just use GitBook...

bpas247:
Hmm, I can't seem to reproduce this bug anymore. Has it been fixed?

![image](https://user-images.githubusercontent.com/18705892/56236141-c5d73e00-603d-11e9-9c06-2d5302c391be.png)

mxschmitt:
We did a major bump of our Gatsby/mdx dependencies.

@thatzacdavis @bpas247 
Please verify on: https://react-bootstrap.netlify.com/getting-started/introduction/
If the issue is still persistent.

mxschmitt:
Still reproducible on https://react-bootstrap.netlify.com/getting-started/introduction/ 
[...]

PR #4020, opened 2019-07-09
Updates

jquense:
fixes: #3540 
fixes: #3536 
fixes: #3805


www/src/pages/migrating.mdx
```diff
@@ -3,13 +3,22 @@
 React Bootstrap v1 adds full compatibility with Bootstrap 4. Because bootstrap 4 is a major rewrite of the project there
 are significant breaking changes from the pre `v1` react-bootstrap.
 
+**PLEASE FIRST READ THE UPSTREAM BOOSTRAP MIGRATION GUIDE**
+
+> https://getbootstrap.com/docs/4.3/migration/
+
+React-bootstrap _only_ contains components that are present vanilla bootstrap. If something
```
taion:
```suggestion
React-Bootstrap _only_ contains components that are present in vanilla Bootstrap. If functionality
```

www/src/pages/migrating.mdx
```diff
@@ -3,13 +3,22 @@
...
+was removed from bootstrap it was also removed from react-bootstrap. This guide does not
```
taion:
```suggestion
was removed from Bootstrap, then it was also removed from React-Bootstrap. This guide does not
```

www/src/pages/migrating.mdx
```diff
@@ -3,13 +3,22 @@
...
+repeat info found in the upstream migration guide, it's goal is to document react-bootstrap
```
taion:
```suggestion
repeat information found in the upstream migration guide. Its goal is to document React-Bootstrap-specific
```

www/src/pages/migrating.mdx
```diff
@@ -3,13 +3,22 @@
...
+specific API changes and additions
```
taion:
```suggestion
API changes and additions.
```

www/src/components/CodeBlock.js
```diff
@@ -29,8 +21,39 @@ const CodeBlock = mapProps(({ mode, codeText, ...props }) => ({
...
     }
-  `,
-);
+  }
+`;
+
+const prism = (code, language = 'jsx') => highlight(code, languages[language]);
+
+const CodeBlock = ({ mode, codeText, ...props }) => {
+  // There is a weird bug that seems to be hydration related where the wrong
```
taion: lol

www/src/components/CodeBlock.js
```diff
@@ -29,8 +21,39 @@ const CodeBlock = mapProps(({ mode, codeText, ...props }) => ({
...
+  // inner html is rendered even though the props are correct. To work around
+  // it we forcable remount the component when it looks like its rendered output
+  // is wrong
+  const [version, setVersion] = useState(1);
```
taion:
```suggestion
  const [version, setVersion] = useState(0);
```

www/src/components/CodeBlock.js
```diff
@@ -29,8 +21,39 @@ const CodeBlock = mapProps(({ mode, codeText, ...props }) => ({
...
-);
+  }
+`;
+
+const prism = (code, language = 'jsx') => highlight(code, languages[language]);
+
+const CodeBlock = ({ mode, codeText, ...props }) => {
+  // There is a weird bug that seems to be hydration related where the wrong
+  // inner html is rendered even though the props are correct. To work around
+  // it we forcable remount the component when it looks like its rendered output
```
taion:
```suggestion
  // it, we forcibly remount the component when it looks like its rendered output
```

www/src/components/CodeBlock.js
```diff
@@ -29,8 +21,39 @@ const CodeBlock = mapProps(({ mode, codeText, ...props }) => ({
...
+  // is wrong
```
taion:
```suggestion
  // is wrong.
```

types/components/Accordion.d.ts
```diff
@@ -4,13 +4,13 @@ import { BsPrefixComponent } from './helpers';
 import AccordionToggle from './AccordionToggle';
 import AccordionCollapse from './AccordionCollapse';
 
-export interface AccordionProps extends React.HTMLProps<Accordion> {
+export interface AccordionProps {
   activeKey?: string;
-  defaultActiveKey?: string
+  defaultActiveKey?: string;
```
ethanve: This should accept a `number` here as well. Better to match with React's Key type: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L78

jquense: it could but we have cast it to a string regardless b/c we add it to a `data-*` attribute, and don't try attempt to cast back. So ultimately i think its probably simpler to just require a string

PR #4031, opened 2019-07-10
fix(AbstractNav): allow passed in refs to be properly forwarded 

### LINKED ISSUE #4012: Nav ref does not work.
Using ref on Nav does not work.

Nav forwards the ref here: 

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Nav.js#L93

But AbstractNav does not and instead sets up its own ref

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/AbstractNav.js#L129

So the ref set on Nav goes nowhere.

bpas247:
`Abstract Nav` should be forwarding its ref to the underlying component, so that `Nav` (and any other components using this) can get the proper ref.

Fixes #4012.


src/AbstractNav.js
```diff
@@ -9,130 +9,117 @@ import TabContext from './TabContext';
...
-    } = this.props;
+    },
+    listNode,
+  ) => {
+    const [needsRefocus, setRefocus] = useState(false);
+
+    if (!listNode) listNode = useRef(null);
+
+    const getNextActiveChild = offset => {
+      if (!listNode) return null;
```
taion:
```suggestion
      if (!listNode.current) return null;
```

src/AbstractNav.js
```diff
@@ -9,130 +9,117 @@ import TabContext from './TabContext';
...
+      }
+      if (!nextActiveChild) return;
+
+      event.preventDefault();
+      handleSelect(nextActiveChild.dataset.rbEventKey, event);
+      setRefocus(true);
+    };
+
+    useEffect(() => {
+      if (listNode && needsRefocus) {
```
taion:
```suggestion
      if (listNode.current && needsRefocus) {
```

src/AbstractNav.js
```diff
@@ -9,130 +9,117 @@ import TabContext from './TabContext';
...
+        let activeChild = listNode.current.querySelector(
+          '[data-rb-event-key].active',
+        );
+
+        if (activeChild) activeChild.focus();
+      }
+    }, [listNode, needsRefocus]);
```
taion:
```suggestion
    }, [listNode.current, needsRefocus]);
```

src/AbstractNav.js
```diff
@@ -9,130 +9,117 @@ import TabContext from './TabContext';
...
+      role,
+      onKeyDown,
       ...props
-    } = this.props;
+    },
+    listNode,
+  ) => {
+    const [needsRefocus, setRefocus] = useState(false);
+
+    if (!listNode) listNode = useRef(null);
```
taion: Can we safely do this? What if the parent is passing down a functional ref?

bpas247: Yeah I wasn't sure if this was a safe approach to begin with, or if it's even necessary. Probably best to remove it?

taion: we might need it, though. @jquense any thoughts here?

bpas247:
https://github.com/react-bootstrap/react-bootstrap/blob/f13a1026774d2e761ceb9f216f721daa1f919a49/test/NavSpec.js#L212-L222
So I'm not exactly sure why this test is failing. The ref functionality shouldn't change anything about the assertion being made on Line 220, yet it's still failing. 

What's interesting is that if I change this line to:
```js
wrapper.assertSingle('Nav[role="tablist"]'); 
```
It passes 🤔 Could this be testing an implementation detail?

src/AbstractNav.js
jquense: I think you need like a merge ref thing here. it's not safe to assume this won't be a function: https://github.com/react-bootstrap/react-overlays/blob/master/src/utils/mergeRefs.js

bpas247:
(Just a side-note, we should consider releasing `useMergedRefs` as a standalone package, it's a pretty useful hook :eyes:)

jquense:
no reason, we should probably put it there!

taion:
i updated this to use restart/hooks; merging so i can cut a release

bpas247:
Thinking we’ll need to update `react-overlays` to use `usedMergedRefs` from `@restart/hooks` rather than its local implementation of it.

taion:
sure, but we can do that later

PR #4135, opened 2019-07-27
fix: Migrate more components to properly forward refs

bpas247:
The first pass-through of migrating more class components to properly forward their refs, via `React.forwardRef`.

Relevant to #4012 and #4031.

### DIFF
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -76,32 +76,33 @@ const fadeStyles = {
   [ENTERED]: 'show',
 };
 
-class Fade extends React.Component {
-  handleEnter = node => {
+const Fade = React.forwardRef(({ className, children, ...props }, ref) => {
+  const handleEnter = node => {
     triggerBrowserReflow(node);
-    if (this.props.onEnter) this.props.onEnter(node);
+    if (props.onEnter) props.onEnter(node);
   };
 
-  render() {
-    const { className, children, ...props } = this.props;
-
-    return (
-      <Transition addEndListener={onEnd} {...props} onEnter={this.handleEnter}>
-        {(status, innerProps) =>
-          React.cloneElement(children, {
-            ...innerProps,
-            className: classNames(
-              'fade',
-              className,
-              children.props.className,
-              fadeStyles[status],
-            ),
-          })
-        }
-      </Transition>
-    );
-  }
-}
+  return (
+    <Transition
+      ref={ref}
+      addEndListener={onEnd}
+      {...props}
+      onEnter={handleEnter}
+    >
+      {(status, innerProps) =>
+        React.cloneElement(children, {
+          ...innerProps,
+          className: classNames(
+            'fade',
+            className,
+            children.props.className,
+            fadeStyles[status],
+          ),
+        })
+      }
+    </Transition>
+  );
+});
 
 Fade.propTypes = propTypes;
 Fade.defaultProps = defaultProps;
--- a/src/Fade.js
+++ b/src/Fade.js
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import React from 'react';
+import React, { useCallback } from 'react';
 import PropTypes from 'prop-types';
 import Transition, {
   ENTERED,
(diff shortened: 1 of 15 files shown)


bpas247:
I'm gonna mark this as ready for review, as the diff for this PR is getting quite large. We should open a new issue to keep track of the remaining components that should be migrated.

src/Fade.js
```diff
@@ -76,32 +76,33 @@ const fadeStyles = {
   [ENTERED]: 'show',
 };
 
-class Fade extends React.Component {
-  handleEnter = node => {
+const Fade = React.forwardRef(({ className, children, ...props }, ref) => {
+  const handleEnter = node => {
```
jquense: `useCallback` probably?

src/ListGroupItem.js
```diff
@@ -73,16 +64,31 @@ class ListGroupItem extends React.Component {
...
       eventKey,
+      onClick,
       ...props
-    } = this.props;
+    },
+    ref,
+  ) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'list-group-item');
+
+    const handleClick = event => {
```
jquense: `useCallback`

src/Fade.js
```diff
@@ -76,32 +76,33 @@ const fadeStyles = {
...
+            children.props.className,
+            fadeStyles[status],
+          ),
+        })
+      }
+    </Transition>
+  );
+});
 
 Fade.propTypes = propTypes;
```
jquense: for any `forwardRef` component we need to also define a `displayName` property so they show up nicely in the React devtools. like: `Fade.displayName = 'Fade'`

test/SafeAnchorSpec.js
```diff
@@ -72,18 +72,14 @@ describe('SafeAnchor', () => {
...
 
     mount(
       <SafeAnchor disabled href="#foo" onClick={clickSpy}>
         Title
       </SafeAnchor>,
     ).simulate('click');
 
-    expect(spy).to.have.been.calledOnce;
     expect(clickSpy).to.have.not.been.called;
-    expect(spy.getCall(0).args[0].isDefaultPrevented()).to.equal(true);
```
jquense: can we keep these asserts but make them on the `clickSpy`?

test/DropdownItemSpec.js
```diff
@@ -98,13 +98,6 @@ describe('<Dropdown.Item>', () => {
       .should.not.have.property('onSelect');
   });
 
-  it('does not pass onSelect to children', () => {
-    mount(<Dropdown.Item onSelect={() => {}}>Item</Dropdown.Item>)
```
jquense: shouldn't remove, adding a displayName to the components would make this work I think

src/Toast.js
```diff
@@ -56,76 +56,79 @@ const defaultProps = {
...
+    bsPrefix = useBootstrapPrefix('toast');
+    const delayRef = useRef(delay);
+    const onCloseRef = useRef(onClose);
+
+    // We use refs for these, because we don't want to restart the autohide
+    // timer in case these values change.
+    delayRef.current = delay;
+    onCloseRef.current = onClose;
+
+    useEffect(() => {
```
jquense: lets use `@restart/hooks/useTimeout` maybe

src/Tabs.js
```diff
@@ -93,8 +93,8 @@ function getDefaultActiveKey(children) {
   return defaultActiveKey;
 }
 
-class Tabs extends React.Component {
-  renderTab(child) {
+const Tabs = React.forwardRef((props, ref) => {
+  const renderTab = child => {
```
jquense: can we hoist this out of render?

test/SafeAnchorSpec.js
bpas247: @jquense I'm not sure if I'm following, since the whole point of this test was to see if `clickSpy` was called, no? How would these assertions pass if `clickSpy` was never called?

src/Fade.js
bpas247: Fixed in c3fab88.

src/Toast.js
bpas247: Fixed in c3fab88.

test/SafeAnchorSpec.js
jquense: oops yes sorry :P

PR #4170, opened 2019-08-03
tests: Increase Test coverage

mxschmitt:
#3889

### DIFF
--- a/src/Tab.js
+++ b/src/Tab.js
@@ -8,7 +8,7 @@ import TabPane from './TabPane';
 /* eslint-disable react/require-render-return, react/no-unused-prop-types */
 class Tab extends React.Component {
   static propTypes = {
-    title: PropTypes.node,
+    title: PropTypes.node.isRequired,
   };
 
   render() {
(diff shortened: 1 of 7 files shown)


src/Tab.js
```diff
@@ -8,7 +8,7 @@ import TabPane from './TabPane';
 /* eslint-disable react/require-render-return, react/no-unused-prop-types */
 class Tab extends React.Component {
   static propTypes = {
-    title: PropTypes.node.isRequired,
+    title: PropTypes.node,
```
taion: It seems like requiring this is good given the semantics below, though.

Can we just use `shouldWarn` for that test case below?

test/DropdownSpec.js
```diff
@@ -257,4 +257,26 @@ describe('<Dropdown>', () => {
     const wrapper = mount(<Dropdown />);
     expect(wrapper.find('div').length).to.equal(1);
   });
+
+  it('Should also accept a custom component', () => {
+    const customComponent = React.forwardRef(
+      (
+        {
+          // eslint-disable-next-line no-unused-vars
+          show,
```
mxschmitt: Would be neat if we could test the attributes for correctness. But I don't see there any easy way to do this via Jest, because we would have to pass them here via DOM attributes to the actual DOM component. Then we could check for the correctness via expect etc. and compare the props.
Any ideas?

PR #4175, opened 2019-08-04
Change Navbar expand prop TypeScript definition to allow false value.

### LINKED ISSUE #4174: Always collapsed navbar
Hi,

Upstream bootstrap supports having a navbar that is always collapsed.
[Example](https://www.codeply.com/go/dvCH985QU0/bootstrap-4-navbar-always-collapsed)

Right now reac-bootstrap navbar only either always expands or expands if the browser window is particular size (xs...xl)

I didn't find a way to make react-bootstrap behave like the upstream in the example linked above.

Could you add support for expand={false} please.

Cheers

rafalkrupinski:
It's already supported by the JS code and it allows the navbar never to
expand.

Fix #4174

### DIFF
--- a/types/components/Navbar.d.ts
+++ b/types/components/Navbar.d.ts
@@ -12,7 +12,7 @@ declare class NavbarText<
 
 export interface NavbarProps {
   variant?: 'light' | 'dark';
-  expand?: true | 'sm' | 'md' | 'lg' | 'xl';
+  expand?: false | true | 'sm' | 'md' | 'lg' | 'xl';
   bg?: string;
   fixed?: 'top' | 'bottom';
   sticky?: 'top' | 'bottom';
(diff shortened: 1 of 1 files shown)


types/components/Navbar.d.ts
```diff
@@ -12,7 +12,7 @@ declare class NavbarText<
 
 export interface NavbarProps {
   variant?: 'light' | 'dark';
-  expand?: true | 'sm' | 'md' | 'lg' | 'xl';
+  expand?: false | true | 'sm' | 'md' | 'lg' | 'xl';
```
mxschmitt:
```suggestion
  expand?: boolean | 'sm' | 'md' | 'lg' | 'xl';
```
Should also work :)

rafalkrupinski: That's `true` :)

rafalkrupinski:
Cheers

Issue #4190, opened 2019-08-06
Modal: Setting animation to false prevents dialog from showing up


derwaldgeist:
If I set `animation` to `false` on a `Modal`, the dialog does not show up. It only works if `animation` is `true`. 

Environment:
- macOS
- Chrome
- React-Bootstrap Version [e.g. 1.0.0-beta.10]

I would also like to set custom animation classes.

derwaldgeist:
Oh, cool, so this is fixed? Awesome!

PR #4192, opened 2019-08-06
fix(Modal): directly show backdrop if no animation (fix: #4190)

mxschmitt:
https://github.com/react-bootstrap/react-bootstrap/issues/4190

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -273,12 +273,14 @@ class Modal extends React.Component {
   }
 
   renderBackdrop = props => {
-    const { bsPrefix, backdropClassName } = this.props;
+    const { bsPrefix, backdropClassName, animation } = this.props;
 
     return (
       <div
         {...props}
-        className={classNames(`${bsPrefix}-backdrop`, backdropClassName)}
+        className={classNames(`${bsPrefix}-backdrop`, backdropClassName, {
+          show: !animation,
+        })}
       />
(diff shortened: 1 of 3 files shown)


src/Modal.js
```diff
@@ -273,12 +273,14 @@ class Modal extends React.Component {
...
-    const { bsPrefix, backdropClassName } = this.props;
+    const { bsPrefix, backdropClassName, animation } = this.props;
 
     return (
       <div
         {...props}
-        className={classNames(`${bsPrefix}-backdrop`, backdropClassName)}
+        className={classNames(`${bsPrefix}-backdrop`, backdropClassName, {
+          show: !animation,
+        })}
```
jquense: we generally prefer the `!animation && 'show'` style over the object form.

www/src/pages/components/modal.js
```diff
@@ -85,6 +86,14 @@ export default withLayout(function ModalSection({ data }) {
         content.
       </p>
       <ReactPlayground codeText={ModalBasic} />
+      <LinkedHeading h="3" id="modals-live">
+        Without Animation
+      </LinkedHeading>
+      <p>
+        A Modal can also be without an animation. For that just set the
```
jquense:
```suggestion
        A Modal can also be without an animation. For that set the
```

PR #4221, opened 2019-08-09
fix(types): make column prop optional for Label

Maxim-Mazurok:
In IntelliJ IDEA, using `<Form.Label>Email address</Form.Label>` results in `Warning: Element Form.Label doesn't have required attribute column`.
Making `column` optional in `FormLabelWithColProps` fixes it.

### DIFF
--- a/types/components/FormLabel.d.ts
+++ b/types/components/FormLabel.d.ts
@@ -14,7 +14,7 @@ export interface FormLabelOwnProps extends FormLabelBaseProps {
 }
 
 export interface FormLabelWithColProps extends FormLabelBaseProps, ColProps {
-  column: true;
+  column?: true;
 }
 
 export type FormLabelProps = FormLabelWithColProps | FormLabelOwnProps


taion:
This doesn't make sense typing-wise though. This should match the non-column branch of the union.

Maxim-Mazurok:
I agree, probably, the best way to approach this problem is from both sides: report IDEA bug and apply this fix to react-bootstrap, while we're waiting for the fix from IDEA. It won't hurt typing-wise AFAIK.

taion:
Why won't it hurt typing-wise? It would make the column branch on the union match even when it shouldn't.

Maxim-Mazurok:
Reported IDEA bug here: https://youtrack.jetbrains.com/issue/IDEA-220380
There's also a [similar issue](https://youtrack.jetbrains.com/issue/WEB-39080) that is not being fixed in three months.



> Why won't it hurt typing-wise? It would make the column branch on the union match even when it shouldn't.

Well, currently TS allows me to add `ColProps` without `column: true`: `<Form.Label xs={'3'}>Email address</Form.Label>`
And it also allows me to add `column={true}` without any `ColProps`: `<Form.Label column={true}>Email address</Form.Label>`

So, it seems to me that this branching is kind of useless anyway...

Also, with my changes applied, `<Form.Label column={false} xs={'1'}>Password</Form.Label>` is still invalid (as it should be).

So, I don't see any real issues with it. What do you think?

taion:
This seems like a different bug with our types, then? We shouldn't allow unknown `<Col>` props when the label is not a column.

Maxim-Mazurok:
Right, I'll try to fix it.

Maxim-Mazurok:
Well, I got it working, but it only works if I define props outside of the component.
Here is the typings:
```ts
interface FormLabelBaseProps {
  htmlFor?: string;
  innerRef?: React.LegacyRef<this>;
  srOnly?: boolean;
}

export interface FormLabelOwnProps extends FormLabelBaseProps {
  column: false;
}

export interface FormLabelWithColProps extends FormLabelBaseProps, ColProps {
  column: true;
}

export type FormLabelProps = FormLabelWithColProps | FormLabelOwnProps | FormLabelBaseProps;
```

And here's my code:
```tsx
const props: FormLabelProps = {
  column: true,
  xs: '1'
};

<Form.Label {...props}>Label</Form.Label>
```

If I put invalid props into jsx, it will still pass validation:
```tsx
<Form.Label xs={'1'}>Password:</Form.Label>
```

I think it might be React or Typescript bug.

Otherwise, all tests are passing:
Invalid:
```
const props: FormLabelProps = {
  column: false,
  xs: '1'
};
```
```
const props: FormLabelProps = {
  xs: '1'
};
```
and valid:
```
const props: FormLabelProps = {};
```
```
const props: FormLabelProps = {
  column: false,
};
```
```
const props: FormLabelProps = {
  column: true,
  xs: '1'
};
```

P.S. Turns out, that swapping `FormLabelWithColProps` and `FormLabelOwnProps` removed warning from IDEA.

mxschmitt:
Closed because it's stale

PR #4225, opened 2019-08-09
fix: Fix DropdownToggleProps

taion:


types/components/DropdownToggle.d.ts
```diff
@@ -4,15 +4,14 @@ import Button from './Button';
 
 import { BsPrefixComponent } from './helpers';
 
-export interface DropdownToggleProps
-  extends React.ComponentPropsWithRef<typeof Button> {
```
taion: This was breaking things because it was picking up the `as`. It's not needed because we inject `as` defaulting to `Button` later anyway.

PR #4228, opened 2019-08-09
feat(types): add useAccordionToggle types

### LINKED ISSUE #4191: How to use useAccordionToggle with Typescript?
Can you give me example how do it?

bpas247:
I'm not actually sure if this is correct, since we don't really have a standardized way of writing unit tests for our TypeScript types (other than dtslint and `types/simple.test.tsx`).

fixes #4191


types/components/useAccordionToggle.d.ts
```diff
@@ -0,0 +1,5 @@
+export default interface useAccordionToggle {
```
taion: Maybe it'd make the most sense to just make this a named export from `./AccordionToggle`? And in `src/` as well.

bpas247: hmm, it would be a breaking change if we changed the export in `src/`, no?

taion: yeah, but we're still in betas :p

PR #4246, opened 2019-08-13
fix(types): Toast component

mxschmitt:
I don't know how to fix this, but that's at least the test for it.
Would be cool if someone could help me :)

### DIFF
--- a/types/components/Toast.d.ts
+++ b/types/components/Toast.d.ts
@@ -3,15 +3,15 @@ import * as React from 'react';
 import ToastBody from './ToastBody';
 import ToastHeader from './ToastHeader';
 
-import { BsPrefixComponent, TransitionCallbacks } from './helpers';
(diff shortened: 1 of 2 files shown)


taion:
`show` and `transition` aren't marked optional in https://github.com/react-bootstrap/react-bootstrap/blob/04aed51c6597596fe0652d085b0455dbbd316ae4/types/components/Toast.d.ts#L13-L14

Is that intentional? That's the error you're seeing.

mxschmitt:
> `show` and `transition` aren't marked optional in
> 
> https://github.com/react-bootstrap/react-bootstrap/blob/04aed51c6597596fe0652d085b0455dbbd316ae4/types/components/Toast.d.ts#L13-L14
> 
> Is that intentional? That's the error you're seeing.

I don't think so, because we provide `defaultProps` there. So `true` for `show` and `Fade` for `transition`. I've adjusted it.

PR #4258, opened 2019-08-14
fix(navbar): resolve lifecycle warning issue

### LINKED ISSUE #4240: Navbar Uses componentWillMount Still
## Describe the bug
I was getting a warnings after setting up some unit tests in my project. After digging in it looks like the Navbar component still uses componentWillMount. It looks like other components are being updated to use the hook form of uncontrollable (useUncontrollable). Are there plans and/or a timeline for Navbar to be updated to no longer depend on the legacy versions of these lifecycle hooks? It's not a pressing issue, just a bit of an eyesore!

## Screenshots

image

image

bpas247:
This migrates the Navbar component to be a function component, which should resolve the lifecycle deprecation warnings.

fixes #4240

### DIFF
--- a/src/Navbar.js
+++ b/src/Navbar.js
@@ -2,19 +2,19 @@ import classNames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
 
-import { uncontrollable } from 'uncontrollable';
+import { useUncontrolled } from 'uncontrollable';
 
 import createWithBsPrefix from './utils/createWithBsPrefix';
 import NavbarBrand from './NavbarBrand';
 import NavbarCollapse from './NavbarCollapse';
 import NavbarToggle from './NavbarToggle';
-import { createBootstrapComponent } from './ThemeProvider';
+import { useBootstrapPrefix } from './ThemeProvider';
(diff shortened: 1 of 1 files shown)


src/Navbar.js
```diff
@@ -122,103 +122,82 @@ const defaultProps = {
...
+  // "nulling" of the role if the user really doesn't want one
+  if (controlledProps.role === undefined && Component !== 'nav') {
+    controlledProps.role = 'navigation';
   }
-}
+  let expandClass = `${bsPrefix}-expand`;
+  if (typeof expand === 'string') expandClass = `${expandClass}-${expand}`;
+
+  return (
+    <NavbarContext.Provider value={{ handleToggle, bsPrefix, expanded }}>
```
jquense: we should `useMemo` the `value` here, otherwise React will trigger an update of any Consumers of this every time Navbar renders, since the object will not `===` the last object.

bpas247:
I'm not exactly sure why the [deploy preview failed](https://app.netlify.com/sites/react-bootstrap/deploys/5d54b7c437168e000c0bd374), but it should be ready for review now 🚀

Issue #4226, opened 2019-08-09
Add a switch component


Simpleqode:
The switch component was added in Bootsrtap v4.2: https://getbootstrap.com/docs/4.3/components/forms/#switches. Would love to see it here as well.

PR #4268, opened 2019-08-16
feat: add Switches

jquense:
closes: #4226

### DIFF
--- a/src/FormCheck.js
+++ b/src/FormCheck.js
@@ -1,12 +1,12 @@
 import classNames from 'classnames';
-import React, { useContext, useMemo } from 'react';
 import PropTypes from 'prop-types';
-
-import { useBootstrapPrefix } from './ThemeProvider';
-import FormContext from './FormContext';
+import all from 'prop-types-extra/lib/all';
+import React, { useContext, useMemo } from 'react';
 import Feedback from './Feedback';
 import FormCheckInput from './FormCheckInput';
 import FormCheckLabel from './FormCheckLabel';
+import FormContext from './FormContext';
+import { useBootstrapPrefix } from './ThemeProvider';
 
 const propTypes = {
   /**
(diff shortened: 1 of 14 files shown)


www/src/pages/components/forms.js
```diff
@@ -264,6 +264,21 @@ export default withLayout(function FormControlsSection({ data }) {
       </p>
       <ReactPlayground codeText={CheckCustom} />
 
+      <LinkedHeading h="3" id="forms-custom-switch">
+        Switches
+      </LinkedHeading>
+      <p>
+        A switch has the markup of a custom checkbox but uses{' '}
+        <code>type="switch"</code> to render a toggle switch. Switches also
+        support the same custom.
```
taion: the same custom what?

src/Switch.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import FormCheck from './FormCheck';
+
+const Switch = React.forwardRef((props, ref) => (
+  <FormCheck {...props} ref={ref} custom type="switch" />
```
taion:
```suggestion
  <FormCheck {...props} ref={ref} type="switch" />
```

redundant?

www/src/examples/Form/Switch.js
```diff
@@ -0,0 +1,13 @@
+<Form>
+  <Form.Check // prettier-ignore
```
taion: why?

src/FormCheck.js
```diff
@@ -1,19 +1,26 @@
...
    * @default 'form-check'
    */
   bsPrefix: PropTypes.string,
 
+  /**
+   * A seperate bsPrefix used for custom controls
+   *
+   * @default 'custom-control'
+   */
+  bsCustomPrefix: PropTypes.string,
```
taion: How is this intended to be used? I don't see it getting hit at all.

jquense: it's used below in the useBootstrap thing

taion: i mean i don't see this ever getting overridden or anything demonstrating this getting overridden

jquense: O it's more an escape hatch for a user, same as bsPrefix, since we are using `custom-control` in the same way it seemed good to also expose it

taion: might be good to document/demo

jquense: It documented i guess. I didn't want to make a big deal out of it and confuse normal users i guess!

JeromeDeLeon:
is this already push to the current version?

mxschmitt:
> is this already push to the current version?

Nope, will be included in the next one.

neylsongularte:
+1 for release to beta

PR #4269, opened 2019-08-17
chore: use cherry-pick to build esm/cjs

jquense:


tools/build.js
```diff
@@ -4,6 +4,7 @@ const webpack = require('webpack');
 const path = require('path');
 const fse = require('fs-extra');
 const execa = require('execa');
+const cherryPick = require('../../cherry-pick').default;
```
bpas247:
```suggestion
const cherryPick = require('cherry-pick').default;
```

jquense:
It's absurd. I'm finally resigning myself to it tho

PR #4290, opened 2019-08-21
docs: Fix PopoverContained sample

afrz:
Fix example here => https://react-bootstrap.github.io/components/overlays/#overlays-container

Event shall be destructured into a "target" variable in order handle click to work properly.

### DIFF
--- a/www/src/examples/Overlays/PopoverContained.js
+++ b/www/src/examples/Overlays/PopoverContained.js
@@ -3,9 +3,9 @@ function Example() {
   const [target, setTarget] = useState(null);
   const ref = useRef(null);
 
-  const handleClick = ({ target }) => {
+  const handleClick = event => {
     setShow(!show);
-    setTarget(target);
+    setTarget(event.target);
   };
 
   return (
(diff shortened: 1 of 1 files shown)


www/src/examples/Overlays/PopoverContained.js
```diff
@@ -3,9 +3,9 @@ function Example() {
   const [target, setTarget] = useState(null);
   const ref = useRef(null);
 
-  const handleClick = ({ targ }) => {
+  const handleClick = event => {
```
mxschmitt:
```suggestion
  const handleClick = ({ target }) => {
```

www/src/examples/Overlays/PopoverContained.js
```diff
@@ -3,9 +3,9 @@ function Example() {
...
     setShow(!show);
-    setTarget(targ);
+    setTarget(event.target);
```
mxschmitt:
```suggestion
    setTarget(target);
```

www/src/examples/Overlays/PopoverContained.js
afrz: What you suggest is what I did in first place, but then CI will fail because **"target"** is already defined in component scope (no-shadowed-variable rule).

Two possibilities : 
- do no destructure event into **"target"** variable
or
- rename **"target"** local state variable into something else

www/src/examples/Overlays/PopoverContained.js
afrz: same as above

PR #4295, opened 2019-08-21
feat: Add the as prop to Badge to support actionable badges

### LINKED ISSUE #4227: Support actional bagdes by adding the as property to the Badge component
According to Bootstrap, we can turn simple badges to actionable badges by using them with an `a` tag: https://getbootstrap.com/docs/4.3/components/badge/#links. Currently, the badge component lacks the `as` prop. Would be great to have it.

mhatvan:
Fixes #4227

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -20,6 +20,9 @@ const propTypes = {
    * some additional horizontal padding
    */
   pill: PropTypes.bool.isRequired,
+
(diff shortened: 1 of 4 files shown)


mhatvan:
> fine, but add this to the TypeScript def as well

unsure about how they should look like, but I added another commit trying ;)

Issue #3497, opened 2019-02-23
TabPane with unmountOnExit should retain transition


sisp:
I would have expected `TabPane` to retain the transition even when `unmountOnExit` is `true`, but looking at the [the source](https://github.com/react-bootstrap/react-bootstrap/blob/08510be5802874ae975c9fbc851cc2b3d61b16e9/src/TabPane.js#L112) `null` is returned immediately rather than when `onExited` is triggered. Is this intended?

jquense:
It should retain the transition, active isn't toggled to false until the transition calls it's onExited callback, at which point we render null so it's unmounted

sisp:
I can't confirm that the transition is happening and unless I'm missing anything I don't see how it should happen given the source code.

The following [snippet](https://github.com/react-bootstrap/react-bootstrap/blob/08510be5802874ae975c9fbc851cc2b3d61b16e9/src/TabPane.js#L162-L165) shows how the `active` prop is determined from the context:
```js
active:
  props.active == null && key != null
    ? makeEventKey(activeKey) === key
    : props.active,
```
The `<Transition />` component animates the transition when the `active` prop changes. But when `unmountOnExit === true` the `<Transition />` component is not rendered with `active === false` because `null` is returned immediately (see [here](https://github.com/react-bootstrap/react-bootstrap/blob/08510be5802874ae975c9fbc851cc2b3d61b16e9/src/TabPane.js#L112)).

jquense:
Sorry are you saying it is working but you don't understand why or it's not working?

sisp:
It's not working for me and the source code doesn't look like it's doing the right thing either.

jquense:
Ok probably a bug then

sisp:
I've put together a small example based on the [custom tabs layout](https://react-bootstrap.netlify.com/components/tabs/#tabs-custom-layout) example from the docs: https://codesandbox.io/s/xl5n993np4

Change the `unmountOnExit` prop of the `App` component to see the difference and notice that `unmountOnExit === false` prevents animated transitions. (I need to open the preview in a new window in order for the tabs to be clickable - not sure why.)

jquense:
yup you're right, I think that one line should be `!active && !Transition && unmountOnExit`

fungjj92:
[...]

PR #4312, opened 2019-08-22
fix: Allow tab to un/mount with transition

fungjj92:
## Overview

The Tabpane component has a bug where when set to unmount on exit, any transitions are ignored. The solution was already discussed in issue #3497 by @sisp and @jquense ; I am shepherding it to the library.

## Demo
 
@sisp created a [codesandbox](https://codesandbox.io/s/xl5n993np4) showing how in the latest release of react-bootstrap `<TabPane ... unmountOnExit=true />` on does not execute the transition between tabs.

Refs #3497

### DIFF
--- a/src/TabPane.js
+++ b/src/TabPane.js
@@ -27,9 +27,9 @@ const propTypes = {
   active: PropTypes.bool,
 
   /**
-   * Use animation when showing or hiding `<TabPane>`s. Use `false` to disable,
-   * `true` to enable the default `<Fade>` animation or
-   * a react-transition-group v2 `<Transition/>` component.
+   * Use animation when showing or hiding `<TabPane>`s. Defaults to `<Fade>`
+   * animation, else use `false` to disable or a react-transition-group
+   * `<Transition/>` component.
    */
   transition: PropTypes.oneOfType([PropTypes.bool, PropTypes.elementType]),
 
@@ -136,7 +136,7 @@ const TabPane = React.forwardRef((props, ref) => {
 
   const prefix = useBootstrapPrefix(bsPrefix, 'tab-pane');
 
-  if (!active && unmountOnExit) return null;
+  if (!active && !Transition && unmountOnExit) return null;
 
   let pane = (
     <Component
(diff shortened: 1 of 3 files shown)


taion:
would be good to add a test here

fungjj92:
Thanks for your speedy response, meanwhile pardon my delays! I agree a test would be good. I'm not familiar with enzyme. Ideally, I'd be able to test the className progression (the entering and exiting phases) to ensure the transition is being applied. I don't get the sense this can be done, rather a test will only be able to naively check the final classNames. Please advise. Thanks!

taion:
Take a look at e.g. https://github.com/react-bootstrap/react-bootstrap/blob/b81eb86420189ad4c7c019a97c0c33ccb61d4798/test/ToastSpec.js#L54 and other things that interact with the fake timer system. It seems like it should be possible to use those to delay the transition finishing, to assert on the intermediate classes.

src/TabPane.js
```diff
@@ -27,9 +27,9 @@ const propTypes = {
   active: PropTypes.bool,
 
   /**
-   * Use animation when showing or hiding `<TabPane>`s. Use `false` to disable,
-   * `true` to enable the default `<Fade>` animation or
-   * a react-transition-group v2 `<Transition/>` component.
+   * Use animation when showing or hiding `<TabPane>`s. Defaults to `<Fade>`
+   * animation, else use `false` to disable or a react-transition-group v2
```
jquense: as an aside we should probably scrub the version from RTG references now that it's on like version 4

kosmiq: Apart from this, is this PR ready to be merged?

fungjj92: Thanks for bumping this @kosmiq . I'll commit this text change. As far as tests go, I was having trouble faking the timing and behavior

fungjj92:
I was reminded this PR was still hanging out 🐌 I just rebased and pushed changes. I had taken a look at writing a test, which I agree would be helpful, but struggled to fake the desired behavior. Would gladly accept a test commit to this branch if it is deemed necessary.

kosmiq:
I tested it and the fix works. The lack of a test is a bit unfortunate, however we do not currently have tests for animation of tabs as far as I can tell.

We could probably merge this fix and add the test case to the ongoing issue #3889 ?

PR #4324, opened 2019-08-23
ci: added GitHub Action workflow for CI

mxschmitt:
Basic CI workflow. Other workflow for deploying the documentation will follow once I've tested it myself. Also we are now using the headless browsers, which are a) more faster and b) does have less requirements.

**Needs `CODECOV_TOKEN` from https://codecov.io/gh/react-bootstrap/react-bootstrap/settings set in the Repository Secret settings.**


taion:
Do we definitely want both GH Actions and Travis running?

mxschmitt:
> Do we definitely want both GH Actions and Travis running?

I would maybe remove Appveyor 🤔

taion:
The point of AppVeyor was to make sure that we don't break stuff for Windows users...

Maybe we can use the GH Actions Windows environment?

mxschmitt:
> The point of AppVeyor was to make sure that we don't break stuff for Windows users...
> 
> Maybe we can use the GH Actions Windows environment?

Sure, I will add it!

bpas247:
> Do we definitely want both GH Actions and Travis running?

Yeah, if we’re planning on migrating entirely to GH Actions then we probably won’t need Travis anymore. Should we go ahead and remove the Travis config file here then?

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
mxschmitt: fixes a issue that on Windows all text based files were `CRLF` instead of `LF`

.travis.yml
```diff
@@ -6,13 +6,10 @@ jobs:
   include:
     - addons:
         chrome: stable
-      env: BROWSER=ChromeCi
+      env: BROWSER=ChromeHeadless
```
bpas247: The GH Actions CI is already running a Chrome instance, yeah? IMO we should just remove the Travis config altogether :stuck_out_tongue:

mxschmitt: maybe let it there until GitHub Actions is out of Beta?

taion: maybe drop firefox from travis, then? and just keep chrome there?

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Install Dependencies
+      run: yarn bootstrap
+    - name: Run Tests
+      run: yarn test
+      env:
+        BROWSER: ${{ matrix.browser }}
+    - name: Upload Test Coverage to codecov
+      # Only upload it a single time
+      if: matrix.os == 'ubuntu-latest' && matrix.browser == 'ChromeHeadless'
+      run: yarn run codecov
```
taion: Codecov can handle receiving multiple sets of coverage results, though. This actually comes in handy for places where different browsers hit different branches.

mxschmitt:
I think it's mergable now once @taion will set the codecov token as an environment variable.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,32 @@
...
+      if: matrix.os == 'macOS-latest' && matrix.browser == 'FirefoxHeadless'
+      run: brew cask install firefox
+    - name: Install Dependencies
+      run: yarn bootstrap
+    - name: Run Tests
+      run: yarn test
+      env:
+        BROWSER: ${{ matrix.browser }}
+    - name: Upload Test Coverage to codecov
+      run: yarn run codecov
```
bpas247: Should we use the [CodeCov Action](https://github.com/codecov/codecov-action) in this job, rather than have the CI build run it locally?

mxschmitt: Regarding the token issue: https://community.codecov.io/t/whitelist-github-action-servers-to-upload-without-a-token/491/2

https://github.community/t5/GitHub-API-Development-and/Whitelist-Github-Workflow/td-p/15916

mxschmitt: https://github.com/react-bootstrap/react-overlays/pull/372#issuecomment-524907547

Issue #4270, opened 2019-08-17
Update changelog to include last two beta releases


bpas247:
## Describe the bug

We are missing releases `v1.0.0-beta.11` and `v1.0.0-beta.12` from the changelog.

## To Reproduce

1. Navigate to `CHANGELOG.md`.
2. Observe that the last two releases are missing.

## Additional context

We've had previous discussions about automated releases (#3546, #3753), although I'm not sure what our current stance is on that for the beta releases.

PR #4330, opened 2019-08-23
chore: added missing versions to changelog (fix: #4270)

mxschmitt:
regenerated by running `yarn changelog -r 3`


bpas247:
I’m wondering how well this would work with the [changelog plugin for semantic-release](https://github.com/semantic-release/changelog), if we’re still planning on integrating `semantic-release` for `v1.0.0`.

mxschmitt:
> did you have to update any of the tooling to make this work?

🤦‍♂I did all the adjustments to the changelog myself. Now after running the command it does make sense. 👍

Issue #4194, opened 2019-08-07
Migrate remaining components to forward their refs properly


bpas247:
## Is your feature request related to a problem? Please describe
For examples on how to migrate these components, please read #4031, #4135, and #3992.
## A clear and concise description of what the problem is.
Currently, not all of our components properly forward the passed in `ref` prop to the underlying component. This leads to end-users not being able to access the underlying DOM node (#3923 and #4012).
## Describe the solution you'd like

Migrate the remaining class components to be ref forwarders. The components that need to be migrated are:
- [ ] [Modal](https://github.com/react-bootstrap/react-bootstrap/blob/0d02bf6d279d227f92703520b3e45727bfded3bf/src/Modal.js)
- [x] [DropdownToggle](https://github.com/react-bootstrap/react-bootstrap/blob/89693c7967594d4713149a17aca0993b45373cc0/src/DropdownToggle.js) (#4661)
- [x] [ToggleButton](https://github.com/react-bootstrap/react-bootstrap/blob/7bd64defe2d0390397fab6938cc84f1191bc80ec/src/ToggleButton.js) (#4640)
- [x] [NavDropdown](https://github.com/react-bootstrap/react-bootstrap/blob/04184495ce111d9e0a9d100cec8f8b8122015ccf/src/NavDropdown.js) (#4626)
- [x] [PageItem](https://github.com/react-bootstrap/react-bootstrap/blob/7e0f4c2579ce578c929042c3945a07e2f2e1971c/src/PageItem.js)
- [x] [ToggleButtonGroup](https://github.com/react-bootstrap/react-bootstrap/blob/8dff07dba3c1bccb39abf32192b00a9f8e846a7a/src/ToggleButtonGroup.js) (#4605)
- [x] [Pagination](https://github.com/react-bootstrap/react-bootstrap/blob/04184495ce111d9e0a9d100cec8f8b8122015ccf/src/Pagination.js) (#4593)
- [x] [AbstractNav](https://github.com/react-bootstrap/react-bootstrap/blob/438cdfcd7d6054d0197e97f760df5bc31e46b19c/src/AbstractNav.js) (#4031)
- [x] [ProgressBar](https://github.com/react-bootstrap/react-bootstrap/blob/8179e12692c7682888032b9f22c1bc063ef0a755/src/ProgressBar.js)
[...]

PR #4334, opened 2019-08-25
feat(refs): migrate more components to properly forward their refs

bpas247:
Relevant to #4194.

### DIFF
--- a/src/Image.js
+++ b/src/Image.js
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import React, { useMemo } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 
 import { useBootstrapPrefix } from './ThemeProvider';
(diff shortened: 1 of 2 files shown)


src/Image.js
```diff
@@ -1,68 +1,72 @@
...
+    ref,
+  ) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'img');
 
-    const classes = classNames(
-      fluid && `${bsPrefix}-fluid`,
-      rounded && `rounded`,
-      roundedCircle && `rounded-circle`,
-      thumbnail && `${bsPrefix}-thumbnail`,
+    const classes = useMemo(
```
taion: no need to memoize this i think

taion: this is a quick calculation, but also it returns a string, so `===` equality checks on the return value will still pass even without memoization

bpas247: Totally right, I pre-optimized here 😬

PR #4370, opened 2019-08-30
fix(TabContainer): Resolve lifecycle deprecation

### LINKED ISSUE #4368: Please update the following components Unctrolled(TabContainer)
## Describe the bug
Closely related to https://github.com/react-bootstrap/react-bootstrap/issues/4306 , this issue is however related to TabContainer and not NavBar.

Using TabContainer creates the following warning:
> backend.js:1 Warning: componentWillMount has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.
> 
> * Move code with side effects to componentDidMount, and set initial state in the constructor.
> * Rename componentWillMount to UNSAFE_componentWillMount to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.
> * Please update the following components: Uncontrolled(TabContainer)

## Expected behavior
No warning caused by using TabContainer.

## Environment 
"react-bootstrap": "^1.0.0-beta.12",
"react-bootstrap": "^1.0.0-beta.12",
 "react-dom": "^16.8.6",

## Additional context

Updated to react bootstrap 1.0.0-beta.12 a couple days ago, and removed the same error for NavBar. Would it be possible to apply the same fix for TabContainer?

nortonwong:
Replace `uncontrollable` with equivalent `useUncontrolled`.

Fixes #4368.

### DIFF
--- a/src/TabContainer.js
+++ b/src/TabContainer.js
@@ -1,138 +1,133 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
-import { uncontrollable } from 'uncontrollable';
+import { useUncontrolled } from 'uncontrollable';
 
 import TabContext from './TabContext';
 import SelectableContext from './SelectableContext';
 
-class TabContainer extends React.Component {
-  static propTypes = {
-    /**
-     * HTML id attribute, required if no `generateChildId` prop
-     * is specified.
-     *
-     * @type {string}
-     */
-    id(props, ...args) {
-      let error = null;
+const propTypes = {
+  /**
+   * HTML id attribute, required if no `generateChildId` prop
+   * is specified.
+   *
+   * @type {string}
+   */
+  id(props, ...args) {
+    let error = null;
 
-      if (!props.generateChildId) {
-        error = PropTypes.string(props, ...args);
+    if (!props.generateChildId) {
+      error = PropTypes.string(props, ...args);
 
-        if (!error && !props.id) {
-          error = new Error(
-            'In order to properly initialize Tabs in a way that is accessible ' +
-              'to assistive technologies (such as screen readers) an `id` or a ' +
-              '`generateChildId` prop to TabContainer is required',
-          );
-        }
+      if (!error && !props.id) {
+        error = new Error(
+          'In order to properly initialize Tabs in a way that is accessible ' +
+            'to assistive technologies (such as screen readers) an `id` or a ' +
+            '`generateChildId` prop to TabContainer is required',
+        );
       }
+    }
 
-      return error;
-    },
+    return error;
+  },
 
-    /**
-     * Sets a default animation strategy for all children `<TabPane>`s. Use
-     * `false` to disable, `true` to enable the default `<Fade>` animation or
-     * a react-transition-group v2 `<Transition/>` component.
-     *
-     * @type {{Transition | false}}
-     * @default {Fade}
-     */
-    transition: PropTypes.oneOfType([
-      PropTypes.oneOf([false]),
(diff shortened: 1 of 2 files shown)


taion:
note the CI junk around lint though \:

src/TabContainer.js
```diff
@@ -1,138 +1,134 @@
...
-        error = PropTypes.string(props, ...args);
-
-        if (!error && !props.id) {
-          error = new Error(
-            'In order to properly initialize Tabs in a way that is accessible ' +
-              'to assistive technologies (such as screen readers) an `id` or a ' +
-              '`generateChildId` prop to TabContainer is required',
-          );
-        }
+/* eslint-disable react/no-unused-prop-types */
```
taion: @jquense bleh, is there a better way to do this? alternatively, why isn't this triggering for e.g. `<Alert>`? does this rule just not work for `React.forwardRef`s?

taion: hah, i guess it just doesn't work for `ForwardRef` things...

src/TabContainer.js
```diff
@@ -1,138 +1,134 @@
...
+
+  /**
+   * The `eventKey` of the currently active tab.
+   *
+   * @controllable onSelect
+   */
+  activeKey: PropTypes.any,
+};
+
+const TabContainer = props => {
```
bpas247: We should be using `React.forwardRef` here (#4194).

src/TabContainer.js
```diff
@@ -1,138 +1,134 @@
...
+    generateChildId: generateCustomChildId,
+    onSelect,
+    activeKey,
+    transition,
+    mountOnEnter,
+    unmountOnExit,
+    children,
+  } = useUncontrolled(props, { activeKey: 'onSelect' });
+
+  const generateChildId = useMemo(
```
bpas247: I don’t think this is an expensive computation? Seems like a pre-optimization to do this here.

src/TabContainer.js
taion: Does this actually render a DOM node though?

src/TabContainer.js
taion: This isn't for cost... it's to avoid unnecessarily invalidating the `tabContext` object below and forcing context updates.

bpas247: > This isn't for cost... it's to avoid unnecessarily invalidating the tabContext object below and forcing context updates.

Hmm, how would it invalidate the context object if that’s also being memoized? Would it basically trigger a re-render and re-calculate the memoized value even though the deps for `tabContext` hasn’t changed?

taion: `generateChildId` is a dependency for the `useMemo` for `tabContext`, so if we don't memoize it, then it will change on every render, preventing memoization of `tabContext`.

src/TabContainer.js
nortonwong: If https://github.com/reactjs/rfcs/pull/97 lands, I think the children could be wrapped in a fragment with ref, but that proposal is still open.

src/TabContainer.js
nortonwong: After some investigation, it's not `forwardRef` at fault. If a component spreads props into a child, the lint rule passes. Even the following example passes:
```js
const Alert = _props => <div {...{}} />;
```

Most `forwardRef` components spread remainder props, so they already trigger this behavior.

taion: ah okay. well, whatever. thanks!

PR #4372, opened 2019-08-30
fix: navs have the wrong role, as well as tabs navigation visual glitch

jquense:


www/gatsby-node.js
```diff
@@ -17,7 +17,7 @@ exports.onCreateWebpackConfig = function onCreateWebpackConfig({
 }) {
   actions.setWebpackConfig({
     devtool: stage.includes('develop')
-      ? 'cheap-inline-module-source-map'
+      ? 'inline-module-source-map'
```
taion: ?

jquense: the cheap one doesn't work well you can't set line breaks :/

PR #4373, opened 2019-08-30
docs: Fix playground

taion:


www/src/components/ReactPlayground.js
```diff
@@ -114,172 +117,168 @@ const StyledExample = styled('div')`
...
+
+  return (
+    <div className="position-relative">
+      <StyledEditor
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        onKeyDown={handleKeyDown}
+        onMouseDown={handleMouseDown}
+        ignoreTabKey={ignoreTab}
+        aria-describedby={showMessage ? id : null}
```
taion: I don't think we want `aria-describedby` to be for an ID not in the DOM.

www/src/components/ReactPlayground.js
```diff
@@ -114,172 +117,168 @@ const StyledExample = styled('div')`
...
-    this.setState({
-      ignoreTab: !this.mouseDown,
-      keyboardFocused: !this.mouseDown,
-    });
-  };
-
-  handleMouseDown = () => {
-    this.mouseDown = true;
+function Editor() {
+  const [focused, setFocused] = useState(false);
```
taion: So we just track this in-JS rather than with CSS?

Alternatively we can change the above to `:focus-within`.

www/src/components/ReactPlayground.js
```diff
@@ -114,172 +117,168 @@ const StyledExample = styled('div')`
...
-      if (key === 'Enter') event.preventDefault();
-      this.setState({ ignoreTab: false });
-    }
-    if (!this.state.ignoreTab && key === 'Escape') {
-      this.setState({ ignoreTab: true });
-    }
-  };
-
-  handleFocus = e => {
-    if (e.target !== e.currentTarget) return;
```
taion: I think the same styling changes that broke the CSS also make this check no-longer-necessary.

www/src/components/ReactPlayground.js
```diff
@@ -114,172 +117,168 @@ const StyledExample = styled('div')`
...
-        this.hjs = hjs;
-        hjs.addTheme('gray', {
-          bg: background,
-          fg: foreground,
-          font: 'Helvetica',
-          fontweight: 'normal',
-        });
-        this.runHolder();
-      });
+  useIsomorphicEffect(() => {
```
taion: \\:

this is the best way to use `useLayoutEffect`, right?

PR #4377, opened 2019-08-31
chore: added Stalebot

mxschmitt:
Might be a good idea to get rid of inactive issues.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false


.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
```
bpas247: Should we include how long before it will close if no further activity occurs?

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - security
```
bpas247: I think we should create a new label that will be used to mark PRs that are currently blocked (waiting on dependency updates, waiting on other PRs to merge first, etc).

.github/stale.yml
```diff
@@ -0,0 +1,17 @@
...
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false
```
bpas247: Do we want to not mark issues as stale? We have quite a few issues that are pretty outdated atm.

taion:
Are there reasonable defaults here, or do we need to explicitly configure all the bot values?

Also should we take a quick audit and pin long-standing old issues before updating?

.github/stale.yml
mxschmitt: makes sense

krichter722:
This is a bad idea on all imaginable projects. You'll have to triage manually one day in order to figure out which issues are active and which are not. The assumptions that issues go away is just not true even if stale and similar bots support this line of thinking.

Issue #4410, opened 2019-09-04
Horizontal ListGroup


YurySolovyov:
Bootstrap has [Horizontal](https://getbootstrap.com/docs/4.3/components/list-group/#horizontal) ListGroup component, it would be nice to have it exposed as boolean or string (layout="vertical"/"horizontal") prop on the ListGroup component.

jquense:
we'd gladly take a PR adding a `horizontal` prop!

malsf21:
Mind if I take a shot at this with a PR? Should be pretty simple!

malsf21:
Just wanted to update, submitted a PR (#4421).

bpas247:
#4421 closes this.

PR #4421, opened 2019-09-06
feat: adds support for horizontal list groups via layout prop

malsf21:
This PR aims to implement the requested feature in #4410, exposing a prop to allow for [horizontal list groups](https://getbootstrap.com/docs/4.3/components/list-group/#horizontal). The prop, `horizontal`, is stringly-typed and supports six values:
* `null` - vertical flow. This is the default
* `true`/boolean attribute - list group items flow horizontally across all breakpoints
* `sm` - list group items flow horizontally for `sm` viewports and above
* `md` - list group items flow horizontally for `md` viewports and above
* `lg` - list group items flow horizontally for `lg` viewports and above
* `xl` - list group items flow horizontally for `xl` viewports and above

In addition, there's a styling conflict with horizontal list groups and using the `flush` variant, as noted in [the Bootstrap documentation](https://getbootstrap.com/docs/4.3/components/list-group/#horizontal), so I added a warning when that happens (and a test to make sure the warning gets logged).

I documented the changes to the API, added a section in the documentation explaining the feature, and wrote some pretty minor tests.

Note: I'm new to contributing to `react-bootstrap`, so I'm not sure if this solution (the implementation or the documentation) is the best way to do it. Please let me know if I should change anything!

### DIFF
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -58,7 +58,8 @@ const ListGroup = React.forwardRef((props, ref) => {
 
   let horizontalVariant;
   if (horizontal) {
-    horizontalVariant = horizontal === true ? 'horizontal' : `horizontal-${horizontal}`;
+    horizontalVariant =
+      horizontal === true ? 'horizontal' : `horizontal-${horizontal}`;
   } else {
     horizontalVariant = null;
   }
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -24,18 +24,11 @@ const propTypes = {
   /**
    * Changes the flow of the list group items from vertical to horizontal.
    * A value of `null` (the default) sets it to vertical for all breakpoints;
-   * `horizontal` sets it go horizontal for all breakpoints, while `horizontal-{sm|md|lg|xl}`
+   * Just including the prop sets it for all breakpoints, while `{sm|md|lg|xl}`
    * makes the list group horizontal starting at that breakpoint’s `min-width`.
-   * @type {('horizontal'|'horizontal-sm'|'horizontal-md'|'horizontal-lg'|'horizontal-xl')}
+   * @type {(true|'sm'|'md'|'lg'|'xl')}
    */
-  layout: PropTypes.oneOf([
-    'horizontal',
-    'horizontal-sm',
-    'horizontal-md',
-    'horizontal-lg',
-    'horizontal-xl',
-    null,
-  ]),
+  horizontal: PropTypes.oneOf([true, 'sm', 'md', 'lg', 'xl', null]),
 
   /**
    * You can use a custom element type for this component.
@@ -45,7 +38,7 @@ const propTypes = {
 
 const defaultProps = {
   variant: null,
-  layout: null,
+  horizontal: null,
 };
 
 const ListGroup = React.forwardRef((props, ref) => {
@@ -53,7 +46,7 @@ const ListGroup = React.forwardRef((props, ref) => {
     className,
     bsPrefix,
     variant,
-    layout,
+    horizontal,
     // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
     as = 'div',
     ...controlledProps
@@ -62,6 +55,10 @@ const ListGroup = React.forwardRef((props, ref) => {
   });
 
   bsPrefix = useBootstrapPrefix(bsPrefix, 'list-group');
+  let horziontalString = 'horizontal';
+  if (horizontal && horizontal !== true) {
+    horziontalString = `horizontal-${horizontal}`;
+  }
 
   return (
     <AbstractNav
@@ -72,7 +69,7 @@ const ListGroup = React.forwardRef((props, ref) => {
         className,
         bsPrefix,
         variant && `${bsPrefix}-${variant}`,
-        layout && `${bsPrefix}-${layout}`,
+        horizontal && `${bsPrefix}-${horziontalString}`,
       )}
     />
   );
--- a/src/ListGroup.js
+++ b/src/ListGroup.js
@@ -55,9 +55,12 @@ const ListGroup = React.forwardRef((props, ref) => {
   });
 
   bsPrefix = useBootstrapPrefix(bsPrefix, 'list-group');
-  let horziontalString = 'horizontal';
-  if (horizontal && horizontal !== true) {
-    horziontalString = `horizontal-${horizontal}`;
+
+  let horizontalVariant;
+  if (horizontal) {
+    horizontalVariant = horizontal === true ? 'horizontal' : `horizontal-${horizontal}`;
+  } else {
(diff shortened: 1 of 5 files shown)


malsf21:
Oops, forgot to add tests. Committed some prelim ones.

YurySolovyov:
Nice, I remember trying to use it with `flush`, and it didn't work as expected, but that's probably a bug of bootstrap itself. Can you confirm? Should we add a note and/or fill an issue for bootstrap?

malsf21:
Yeah, [Bootstrap doesn't support using `flush` with horizontal](https://getbootstrap.com/docs/4.3/components/list-group/#horizontal). You can probably submit an issue there, though I don't think they intend for those two to work together.

src/ListGroup.js
```diff
@@ -54,6 +74,7 @@ const ListGroup = React.forwardRef((props, ref) => {
         className,
         bsPrefix,
         variant && `${bsPrefix}-${variant}`,
+        layout && `${horizontalPrefix}-${layout}`,
```
taion: why is `horizontalPrefix` different from just `bsPrefix` above? i think they should be the same?

malsf21: Ah not exactly, in this case it's because horizontalPrefix just prepends `list-group-` to each class, so the resulting classes are `list-group-horizontal*`. Should I rename it/make it more clear?

taion: But `horizontalPrefix` is just `'list-group'`. It's the same as `bsPrefix` here. The `horizontal-` part is in `layout`.

malsf21: Ah, I looked it over again and totally my fault - I misunderstood what you meant. I'll update the PR now!

src/ListGroup.js
```diff
@@ -21,6 +21,22 @@ const propTypes = {
...
+   * @type {('horizontal'|'horizontal-sm'|'horizontal-md'|'horizontal-lg'|'horizontal-xl')}
+   */
+  layout: PropTypes.oneOf([
+    'horizontal',
+    'horizontal-sm',
+    'horizontal-md',
+    'horizontal-lg',
+    'horizontal-xl',
+    null,
+  ]),
```
jquense: I think i would change this API to 

```js
horiztonal: PropTypes.oneOf([
    true,
    'sm',
    'md',
    'lg',
    'xl',
    null,
  ]),
```

I think that's probably more inline with how w'eve done this elsewhere in the codebase, e.g. `navbar.expand`

malsf21: Makes sense! Just pushed a commit to the fork that updates that + changes the docs.

src/ListGroup.js
```diff
@@ -44,6 +55,10 @@ const ListGroup = React.forwardRef((props, ref) => {
   });
 
   bsPrefix = useBootstrapPrefix(bsPrefix, 'list-group');
+  let horziontalString = 'horizontal';
```
taion:
```suggestion

  let horizontalVariant;
```

src/ListGroup.js
```diff
@@ -44,6 +55,10 @@ const ListGroup = React.forwardRef((props, ref) => {
...
+  if (horizontal && horizontal !== true) {
```
taion:
```suggestion
  if (horizontal) {
```

src/ListGroup.js
```diff
@@ -54,6 +69,7 @@ const ListGroup = React.forwardRef((props, ref) => {
         className,
         bsPrefix,
         variant && `${bsPrefix}-${variant}`,
+        horizontal && `${bsPrefix}-${horziontalString}`,
```
taion:
```suggestion
        horizontalVariant && `${bsPrefix}-${horizontalVariant}`,
```

src/ListGroup.js
```diff
@@ -44,6 +55,10 @@ const ListGroup = React.forwardRef((props, ref) => {
   });
 
   bsPrefix = useBootstrapPrefix(bsPrefix, 'list-group');
+  let horziontalString = 'horizontal';
+  if (horizontal && horizontal !== true) {
+    horziontalString = `horizontal-${horizontal}`;
```
taion:
```suggestion
    horizontalVariant = horizontal === true ? 'horizontal' : `horizontal-${horizontal}`;
```

src/ListGroup.js
```diff
@@ -44,6 +55,10 @@ const ListGroup = React.forwardRef((props, ref) => {
...
+  }
```
taion:
```suggestion
  } else {
    horizontalVariant = null;
  }
```

taion:
* with readability notes as above

malsf21:
Thanks @taion, committed those changes.

malsf21:
@jquense just checking in, does the new API make more sense?

www/src/pages/components/list-group.mdx
```diff
@@ -78,6 +80,22 @@ edge-to-edge in a parent container [such as a `Card`](/components/cards/#list-gr
   exampleClassName={styles.listGroup}
 />
 
+### Horizontal
+
+Use the `horizontal` prop to make the ListGroup render horizontally. Currently **horizontal list groups cannot be combined with flush list groups.**
```
bpas247: Are we actually enforcing this in the implementation? We should at least log a warning when the end user tries to do this.

malsf21: No, I haven't actually implemented it - do you want me to throw an error, or just log the warning? The "unsupportedness" mostly comes from clashing CSS styling.

malsf21: Also, is there a specific logger I should use? Or just `console.log()`?

bpas247: > does it make more sense to throw an error, or just log the warning? The "unsupportedness" mostly comes from clashing CSS styling.

IMO a warning should be good enough, I'm not sure if a hard fail would provide any more benefit.

> Also, is there a specific logger I should use? Or just console.log()?

We log warnings via the `warning` package. There's an example of its usage in the [FormLabel](https://github.com/react-bootstrap/react-bootstrap/blob/d6026eb0f4c7db98cc676dab119f8c4715775f6a/src/FormLabel.js#L62-L65) implementation.

malsf21: Gotcha, thanks for the pointer! I added a warning on both being used, and added a test in `ListGroupSpec.js` - let me know if it makes sense!

bpas247:
Looks like this should be good for merge 🚀 

@malsf21 Do you mind if I go ahead and merge? Are you planning on pushing any more commits?

malsf21:
> Are you planning on pushing any more commits?

Nope, unless there are any other things I should be implementing for the PR. Thank you for the help!

bpas247:
> > Are you planning on pushing any more commits?
> 
> Nope, unless there are any other things I should be implementing for the PR. Thank you for the help!

Thank you for contributing 😄

Issue #3583, opened 2019-04-01
Declaration for FormControlProps.onChange is incorrect


AndriySvyryd:
```TS
@observer
class InputForm extends React.Component<IInputFormProps, {}> {
    @observable outgoingMessage: string = '';

    @action.bound
    handleChange(event: ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>) {
        this.outgoingMessage = event.target.value;
    }

    @action.bound
    handleSubmit(event: React.FormEvent<HTMLFormElement>) {
        this.props.sendMessage(this.outgoingMessage);
        this.outgoingMessage = '';
        event.preventDefault();
    }

    render() {
        return <Form inline noValidate onSubmit={this.handleSubmit} autoComplete="off">
            <Form.Control
                type="text"
                name="outgoingMessage"
                value={this.outgoingMessage}
                onChange={this.handleChange}
            />
            <Button type="submit">Send</Button>
        </Form>;
    }
}

interface IInputFormProps {
    sendMessage: (outgoingMessage: string) => void;
}
```

>TS2339: Property 'target' does not exist on type 'ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>'.

mattvonb:
I'm facing this problem as well. (I think.)

With react-bootstrap 0.32.4, I would define the `onChange` handler with type:

    (event: React.FormEvent<FormControl & HTMLInputElement>) => void

If I try this with react-boostrap 1.0.0-beta.6, I get the following compiler error:

    Type '(event: FormEvent<Form<"input"> & HTMLInputElement>) => void' is not assignable to type '(event: FormEvent<ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>>) => void'.
      Types of parameters 'event' and 'event' are incompatible.
        Type 'FormEvent<ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>>' is not assignable to type 'FormEvent<Form<"input"> & HTMLInputElement>'.
          Type 'ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>' is not assignable to type 'Form<"input"> & HTMLInputElement'.
[...]

Issue #3568, opened 2019-03-24
createRef type for FormControl


hamanovich:
I'm using react-bootstrap together with typescript.
What's the proper type in createRef for <Form.Control> element? (instead of any)

`private textInput = createRef<any>();`

`<Form.Control
                      name='textarea_name'
                      as="textarea"
                      ref={this.textInput}
                    />`

Once I replace <any> to anything else e.g. <FormControl>, I always see the error in a console like so:
The expected type comes from property 'ref' which is declared here on type 'IntrinsicAttributes & IntrinsicClassAttributes<Form<"input">> & Readonly<ReplaceProps<"input", BsPrefixProps<"input"> & FormControlProps>> & Readonly<{ children?: ReactNode; }>...

taion:
It's whatever the `as` component is (in this case `<textarea>`) – though I'm not sure our types are set up quite correctly.

benkaiser:
Yeah there's definitely an issue with the types for Form.Control

kbsanders:
Ran into the same issue.

Until it's fixed, I have found that you can do the following to at least allow your RefObject to maintain a type for usage in the component class:

`private textInput = React.createRef<HTMLTextAreaElement>();`

`<Form.Control name='textarea_name' as="textarea" ref={this.textInput as React.RefObject<any>} />`

Note the cast to RefObject<any> inside the ref prop assignment.

denis-n:
This worked for me:

`
private commentRef = React.createRef<
        FormControl<"textarea"> & HTMLTextAreaElement
    >();

    public componentDidMount() {
        if (this.commentRef.current) {
            this.commentRef.current.focus();
        }
    }`

Also had to change `rows` in `Form.Component as='textarea'` from string to number.

denis-n:
Does anyone know if `innerRef` is working?

https://github.com/react-bootstrap/react-bootstrap/blob/master/types/simple.test.tsx#L211

When I'm usin `innerRef` I get this error:

> React does not recognize the `innerRef` prop on a DOM element

Aaronik:
@denis-n, how did you import that `FormControl` type from 

```
private commentRef = React.createRef<
        FormControl<"textarea"> & HTMLTextAreaElement
    >();
```

??

I think if I had that I'd be able to type this correctly, but I can't find where that's exported.

onichandame:
> @denis-n, how did you import that `FormControl` type from
> 
> ```
> private commentRef = React.createRef<
[...]

Issue #2781, opened 2017-09-04
Typing issue with FormControl onChange


Alan-Hinton:
I'm not quite sure if this issue belongs here or on the Definitely Typed repository, but I'll try here first.

When using the FormControl in React with typescript I am forced to cast the `currentTarget` to `any` in the event handler in order to access the value of the FormControl.  This gives me linter error and just seems wrong.

Either the FormControl should have a value property that can be accessed through the currentTarget or there should be an entierly new event that just passes me back the value string.  The later would follow the react methodology more closely.

Here is an example of the code with the offending cast.

```
class MyComponent extends React.Component<Props, {}> {
  queryChange(event: React.FormEvent<FormControl>) {
    // tslint:disable-next-line:no-any  
    this.props.queryChanged((event.currentTarget as any).value);
  }

  render() {
    return (
      <div>
            <form>
              <FormGroup>
                <ControlLabel>Site</ControlLabel>
                <InputGroup>
                  <FormControl
                    type="text"
                    value={this.props.query}
                    placeholder="Search site"
                    onChange={e => this.queryChange(e)}
                  />
                </InputGroup>
              </FormGroup>
            </form>
          )}
      </div>
    );
  }
}
```

jquense:
This seems like a TS type def issue. The event in the change handler is the unmodified React SyntheticEvent object not anything specific to react bootstrap

patsissons:
The main problem is that `DOMAttributes<T>` is using the same generic type param as `ClassAttributes<T>` which is technically incorrect. While it is sensible to do this for very simple components, any composite components may nor longer make any sense as you cannot presume that all DOM events are generated by the react component, instead of a sub element bubbling events up to the react component, `FormControl` is one of these exceptions. As these typings are defined in `@types/react` there is little `react-bootstrap` can do to mitigate this on their end.

I have come to two solutions to deal with this react typing deficiency, either aggregate sub element events within the component (using a `handleBlah` function pattern works well) and deal with the type variance within the internal function, or you can type your `onChange` parameter with a union type...

```
<FormControl type='text' onChange={ (x: React.FormEvent<FormControl & HTMLInputElement>) => { alert(x.currentTarget.value); } } />
```

The first option is probably the best option, but if you want to inline without upcasting to `any` then the 2nd option is handy.

modmatic:
@patsissons  Still having problems when I try the second approach:

```
[...]

PR #4435, opened 2019-09-09
fix(FormControl): onChange and innerRef ts decls (#3583,#3568,#2781)

craigpg:
These changes seemed get the types sorted out for my project.  I added some tests to `simple.test.tsx` to confirm and demonstrate my usage.

May want to consider adding a test-types script to package.json and including it in the main test script.

```
"test": "npm run lint && npm run dtslint && npm run test-browser && npm run test-node && npm run test-types"
"test-types": "tsc --noEmit -p types"
```

### DIFF
--- a/types/components/FormControl.d.ts
+++ b/types/components/FormControl.d.ts
@@ -2,14 +2,19 @@ import * as React from 'react';
 import Feedback from './Feedback';
 import { BsPrefixComponent } from './helpers';
 
+type FormControlElement =
+  | HTMLInputElement
+  | HTMLSelectElement
+  | HTMLTextAreaElement;
+
 export interface FormControlProps {
-  innerRef?: React.LegacyRef<this>;
+  innerRef?: React.LegacyRef<FormControlElement>;
   size?: 'sm' | 'lg';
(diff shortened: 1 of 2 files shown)


taion:
we do test the types, i thought?

we should get rid of the last few `innerRef` instances

craigpg:
@taion  - The `ref` types are coming from:

1. The generic class that `FormControl` extends to get the props of the wrapped `input` element.  This will add a `ref` prop with type `HTMLInputElement`, which is really the only ref we want.
2. When TS transforms the JSX, it creates a `FormControl` element and the underlying TS declarations automatically add a `ref` prop with type `FormControl` by intersecting it with the rest of the props.
3. Since the `ref` prop already exists, albeit with a different type, the prop types are intersected.

There is a [workaround](https://github.com/react-bootstrap/react-bootstrap/issues/3568#issuecomment-487276679)

taion:
thanks for digging that up.

i'm comfortable moving forward with this. we'll likely want to switch everything to `forwardRef` just for non-TS API concerns before we cut 1.0.0 final, but this is a strict improvement as things are now

PR #4464, opened 2019-09-13
fix(types): add missing Popover and Modal props

p-smith:
This PR updates the type definitions for the Popover and Modal components.

Over time new props have been added to the Modal and Popover components, but these new props are not yet included in the respective typescript type definitions.


mxschmitt:
Thanks!

PR #4512, opened 2019-09-23
remove magic in Typescript definitions to avoid false positives

olmobrutall:
As part of the general update of [Signum Framework](https://github.com/signumsoftware/framework) to TypeScript 3.6, React 16.9 and functional components I'm happy to take again a dependency to react-bootstrap (I removed a few years ago due to lacking Bootstrap 4 update). 

Unfortunately, I realize that the auto-completion of attributes was missing in many cases and that extra attributes where not being marked as errors... making the typing quite useless. 

After diving deep in the react-bootstrap / react typings I found the problem is in the inference in the pattern: 

```TS
declare class Tab<	
  As extends React.ElementType = 'div'	
> extends BsPrefixComponent<As, TabProps>
```

The intention here is to say: take all the attributes in As (typically 'div') and override them with the ones in TabProps in case on conflict. 

But inference doesn't work like this. If you add some extra attributes (without changing `as`) somehow TS ends up inferring 'any' (or some custom object type with the wrong attributes included on it) and produces no error. 

This means almost any value in statically typing react-bootstrap is lot. 

I have confirmation from private email with @glenjamin about this behaviour. 

> You are correct that they never really worked, but AllHTMLAttributes also wouldn’t work with custom elements.

As I don't think that TS is able to express the complicated relationship between 'as' attribute and all the other ones reliably (much less easy to understand...), I've change the code to the more simple and limited version. 

```TS
declare class Tab extends BsPrefixComponent<'div', TabProps> {
```

This means the type will allow all the props in 'div' overriden by TabProps, if you want to use a different component, and set some custom attribute that is not also in 'div' you will need to hack it with `any`: 

```TS
<Tab as="Button" {...{variant: "secondary"} as any}>
</Tab>
```

but for the general case, the works very well in practice, with reliable autocomplete and detecting more than 40 errors in my codebase that were previously silent.

### DIFF
--- a/types/components/Accordion.d.ts
+++ b/types/components/Accordion.d.ts
@@ -10,9 +10,7 @@ export interface AccordionProps {
   defaultActiveKey?: string;
 }
 
-declare class Accordion<
-  As extends React.ElementType = 'div'
-> extends BsPrefixComponent<As, AccordionProps> {
+declare class Accordion extends BsPrefixComponent<'div', AccordionProps> {
   static Toggle: typeof AccordionToggle;
   static Collapse: typeof AccordionCollapse;
 }
--- a/types/components/AccordionToggle.d.ts
+++ b/types/components/AccordionToggle.d.ts
@@ -7,14 +7,6 @@ export interface AccordionToggleProps {
   onClick?: (event?: React.SyntheticEvent) => void;
 }
 
-export interface useAccordionToggle {
-  (eventKey: string, onClick: (event?: React.SyntheticEvent) => void): (
-    event?: React.SyntheticEvent,
-  ) => void;
-}
-
-declare class AccordionToggle<
-  As extends React.ElementType = 'button'
-> extends BsPrefixComponent<As, AccordionToggleProps> {}
+declare class AccordionToggle extends BsPrefixComponent<'button', AccordionToggleProps> {}
 
 export default AccordionToggle;
--- a/types/components/Badge.d.ts
+++ b/types/components/Badge.d.ts
@@ -1,7 +1,6 @@
 import * as React from 'react';
-import { BsPrefixComponent } from './helpers';
 
-export interface BadgeProps {
+export interface BadgeProps extends React.HTMLProps<HTMLSpanElement> {
   bsPrefix?: string;
   variant?:
     | 'primary'
@@ -15,8 +14,6 @@ export interface BadgeProps {
   pill?: boolean;
 }
 
-declare class Badge<
-  As extends React.ElementType = 'span'
-> extends BsPrefixComponent<As, BadgeProps> {}
+declare const Badge: React.ForwardRefExoticComponent<BadgeProps>;
 
 export default Badge;
--- a/types/components/Breadcrumb.d.ts
+++ b/types/components/Breadcrumb.d.ts
@@ -9,9 +9,7 @@ export interface BreadcrumbProps {
   listProps?: React.OlHTMLAttributes<any>;
 }
 
-declare class Breadcrumb<
-  As extends React.ElementType = 'nav'
-> extends BsPrefixComponent<As, BreadcrumbProps> {
+declare class Breadcrumb extends BsPrefixComponent<'nav', BreadcrumbProps> {
   static Item: typeof BreadcrumbItem;
 }
 
--- a/types/components/BreadcrumbItem.d.ts
+++ b/types/components/BreadcrumbItem.d.ts
@@ -9,8 +9,6 @@ export interface BreadcrumbItemProps {
   title?: React.ReactNode;
 }
 
-declare class BreadcrumbItem<
-  As extends React.ElementType = 'li'
-> extends BsPrefixComponent<As, BreadcrumbItemProps> {}
+declare class BreadcrumbItem extends BsPrefixComponent<'li', BreadcrumbItemProps> {}
 
 export default BreadcrumbItem;
--- a/types/components/Button.d.ts
+++ b/types/components/Button.d.ts
@@ -2,35 +2,38 @@ import * as React from 'react';
 
 import { BsPrefixComponent } from './helpers';
 
+export type ButtonVariant =
+  | 'primary'
+  | 'secondary'
+  | 'success'
+  | 'danger'
+  | 'warning'
+  | 'info'
+  | 'dark'
+  | 'light'
+  | 'link'
+  | 'outline-primary'
+  | 'outline-secondary'
+  | 'outline-success'
+  | 'outline-danger'
+  | 'outline-warning'
+  | 'outline-info'
+  | 'outline-dark'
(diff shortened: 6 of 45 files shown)


taion:
Seems to work fine here: http://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdwAdjFlAXlnAEICuMMCEwAKxMOgDedODLipgALywB+AFzlUIZABttZOAB9y25FADmWWnQC+DGAE8wXbGBO4soiOIA8ASSZMbHBYAB6sTAAm6Nh4MAB0AKLaWCBYLAAqjlgANHDpnuIAfHAAvHAA8iDAMN7SmFTxAMIkkIEsBagA6tUAFtgEfgFshdl1ANZY9hAEeR10xQBkdfliqPR0oZCwcMys7JxwAMoOyRHN4EJpMB3eAILooeFR9bGJyakZWcVSMmhqcPd6LYNiEtvBcCZUOhjvZTudWldajJ7sEwmlnjF8G8UldMk5RjIVl5UPNUU9og04vDLixvC43B5VndULkYXCWjTrkz7otZqtCt9gXQIWh0HwBEIkQCHmjIhTXkkcZ8nKVyAAjfiCJhkUmPdHQk5YM4cto1e65cVajqChi7NgcdxwABiEAgHTgPzgBFd6lQMCgzDMQIYBF4THwwCEztdAAowKt1C63asAJQeuqUGC8KBMOBMXi6YN0byWqPeiAlABEatMlbgAHpCvQS5qo2gShIk9YG03i6Xc+3O67u+WqzWoHXG83+3JFFgq5odNpJ72WxLc-IlAuQCvp62B6gO12va6x7XZ1vK4vdLu+-uUIehxBu5v51etDee3v1w+j8OL2+mi7kAA

olmobrutall:
Interesting... looks like the error is only when defaulting to class components and not intrinsic, here is my counter example: 

http://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdWAHpLHDAJ5hZwDyIwGAB4AKgBo4AVQB8cALxwACsFwBrURICiTXABsArgBMsQ1Vg4QCccXDMWrM6fUYto8Tt0xYwuvFkXEYKhCAJIAdmFYUHDMMFhhhujYeDAAdJq6WCDxMCJcWBKKsgr8gkJ0cF4pqQDCJJCRYTABEEEA6oIAFtgEoRFR0mIVSnSyAGTDis7MrPDATVEEfnAAQqgBWATATC1BQgCC6LHxiVX46ZnZTXncsgDew2gA-ABccIf0lQBG65RbTK84KgYFB5gBzegAXwYMzccD0aHQaw2-zq4AgjWEw0OMSYcQSSSoaQyWRyNwKk3kcDu0MqsmOBLOaTRDRyQmwPj8u2ChwkyL+225B1Q4yU0nucGhMNcbHmcSgS1wPAAcsgAG7c6nDMLqr7IKD8zbbQHA0FhCHDXD6wwACRwxgNvyNALepvBnzgav1wGQTUBZBgyB+ZDgAB9yGBgLpdKhaI98MA1VgANLmQH6MKqMIQADuYQ9xiW+l0MH2CaTqY46cz2bzHq20cBXwgEEyvo9ACt9MDgAQq29m62cPnhsRMiaQe7tbr9U2W22R5UMZWACK5sKA5LnSvN62aJNNG2+wyZKBCGCdYAiqEMYwIyjw3yodCqtXlSo4hmnLfEy5k-JUmQhiJmQwz0niJxIk6qL1BibK8nAr7cvcdBSrCsoLAqywiEGvytOgDyVCkiYpmmbwZlm64FpsyDFqW5akf2cAUbWi6et6vowP6gbBmGEZRjGcaVCCvqoIIwAYoCSwxjw4Y-hcpLXPkHrAIYrqTuaHogBAGYwLwYSaJhc5Du2wwZtpun6dogjGQuN50HevgPgiz7WLh77vEcEGMvJJJXLkAEKB4WCWIh6pgbi+KnIaMHopiwoSDhPzIdSkoMHQoi4XAqlyAARMguVwAQLZ5XqUC5QA9LIFUIMVEBAp0Om6IYMRQMQ0RoHA2bwIW8xYIYCAVUAA

Using your slightly different StyledComponent / Omit variant also doesn't help: 

http://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdMAnmFplmADZ5YAKxYVAB4AkgDsxWKHCwAPGFjEATdNjwwAdAFFOWEIpgAVFlgA0cQ-wiCAfHAC8cAPIhgMIXTjt1GgMIlISTEYK0EAdTcAC2wCUQkpG1NPOABrLCYIAgtQ1Do7ADJkywFUejo5SFg4YGCpAh44AGVmXSV-cAggkJKhAEF0OQVlVSpNHT0DY1Y7AG9ktAB+AC44fvoAXwYK6HhcblR0ZqZW9sCDDy9+mXlFFW98bV19YKmzIpzku0HbkZ9TzvO2C4PByfVQ5iOJwCAOCoP6BWyJRsszgmwY5VklXgNQUUHquDYADlkAA3HJwOZeMSkgBGyCgACFUPwsARgLJlnBUDAoDUAOb0Ly4elKAASOCUUiZLLZHJW3N5YgFyRJ9OAyGCnLIMGQNNQZDgAB9yGBgJxOPrBSh8MASVgANLpTkAVzEKTEEAA7mIrZL6s7ODBeja7Y6mC63R7vVa2ebOTSIBBdBqrQArZ3c4AEcMrBNJnA+5LEXSchX8q3Ukl0qDxxPJwteTphgAiXrEnLUDzDCZFWjtwVFGqUuigQhgkWAqBsGwYkr29LY84OcGJJIuqwGN2G9zGT0mJgc5CUtrIn2uQzukKwbWhXTB5lXOVmdDR2yqOLqDUMuuZJQp8xDB0nRWV13TbX1WWQAMg0AsMIzA6MVTVDUYC1HU9QNY0yFNc1LWSHkNVQNxgE6Tl6gtNhjU7XcJheEwrWAJR5R5ctkhACBXRgJwxC0Woa1zOsCytV12M47itFkNxa3zFMX1nLB50oOAl3Qb89XXK5vm3ajHloowD0cZhWEyFdSTPLTLxaa9-jvfpzDU39rCnClUXRIQHOqJR7AAImQby4AIRMfOrbyAHo7FChBAogLlIg4zglDgGk2ElBR8GvFB0FdP0amvBBQroIA

Thanks for the quick answer :)

olmobrutall:
This PR is better https://github.com/react-bootstrap/react-bootstrap/pull/4524

PR #4524, opened 2019-09-24
fix typings for Components passing props to other component

olmobrutall:
This is the real solution to solve the problem in https://github.com/react-bootstrap/react-bootstrap/pull/4512. 

With this change the `As` generic argument and the dependent props are correctly inferred for both Intrinsics ('div', 'a', ...) that where already working as well as other react-bootstrap Components ('SafeAnchor',  'Button') **new fix!!**

Here is a playground snippet comparing the two solutions. 

http://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcA5FDvmQNwBQdWAHpLHDAJ5hZwDyIwGAB4AKgBo4AVQB8cALxwACsFwBrURICiTXABsArgBMsQ1Vg4QCccXDMWrM6fUYto8Tt0xYwuvFkXEYKhCAJIAdmFYUHDMMFhhhujYeDAAdJq6WCDxMCJcWBKKsgr8gkJ0cF4pqQDCJJCRYTABEEEA6oIAFtgEoRFR0mIVSnSyAGTDis7MrPDATVEEfnAAQqgBWATATC1BQgCC6LHxiVX46ZnZTXncsgDew2gA-ABccIf0lQBG65RbTK84KgYFB5gBzegAXwYMzccD0aHQaw2-zq4AgjWEw0OMSYcQSSSoaQyWRyNwKk3kcDu0MqsmOBLOaTRDRyQmwPj8u2ChwkyL+225B1Q4yU0nucGhLlm7Hyq1+m22LIxORqvlQwWxRzxJ0J1RJV1y+SGlUUVJpw2KTNq9RVTTViPZ3l8uH8gR5qD5Cv+QsOoqKEqlsLY8ziUCWrrgADlkAA3bnU4ZhONfZBQfmKgFvYGgsIQ4a4NOGAASOGM6e920BOfBnzgsbTwGQTUBZBgyB+ZDgAB9yGBgLpdKhaI98MBY1gANLmQH6MKqMIQADuYTrxiW+l0MH2Y4n044s-ni5Xda2g8BXwgEEyzbrACt9MDgAQD29L9ecKvhsRMtWQbWkxTNMLyvG8v0qDF9wAEWXMJAWSc590vItNAnJpi2bQxMigIQYE6YARShBhjARSh4XVdAY1jcpKhxBlTgQ4lLjJOUFDIQxxzIS1cXxU4M1RW1MWFCQqO5e46CDVwQwWcNlhEDtfladAHkqFJxynGc3jnBdYLXTZkE3bddw0184G049wPrRtmxgVt207Hs+wHIcR0qEFm1QQRgAxQEliHHhe0Yi5SWufI62AQxs3-PM6xACA5xgXgwk0GSQI-W9hjnOKEqS7RBDSsCiLoEjfDIhENWsBS2mIPMaPebVeL1c4DRYzwFA8LBLGjONuPopFKyYZUhN5SqfjE6lJWIrBSJ4cr0Hkn4MGAMFOixWiGt1a0WtCtr5RRJVBNVCihHYziRLjMTep1Rl+IO9Fhs9UbFKCQMGDoBtomQKlRCqmqwTgCK5AAImQIG4AIK9gdTKAgYAeicOBYYQCGICBTp4t0QwYigYhPvQRd4HXeYsEMBBYfetM4C+b6FtQJaVrmQxgdB8HIaB6G4YRpGUbgMBiEMfRXSOHHoBQbVuHwEmyaAA

I've checked with a relatively big codebase and works ok.

### DIFF
--- a/types/components/Spinner.d.ts
+++ b/types/components/Spinner.d.ts
@@ -19,7 +19,7 @@ export interface SpinnerProps {
 }
 
 declare class Spinner<
-  As extends React.ElementType = typeof Spinner
+  As extends React.ElementType = 'div'
 > extends BsPrefixComponent<As, SpinnerProps> {}
 
 export default Spinner;
--- a/types/components/Alert.d.ts
+++ b/types/components/Alert.d.ts
@@ -5,6 +5,7 @@ import SafeAnchor, { SafeAnchorProps } from './SafeAnchor';
 import { BsPrefixComponent, BsPrefixComponentClass } from './helpers';
 
 export class AlertLink<
+  // Need to use BsPrefixComponentClass to get proper type checking.
   As extends React.ElementType = BsPrefixComponentClass<'a', SafeAnchorProps>
 > extends BsPrefixComponent<As> {}
 
--- a/types/components/Alert.d.ts
+++ b/types/components/Alert.d.ts
@@ -1,11 +1,11 @@
 import * as React from 'react';
 
-import SafeAnchor from './SafeAnchor';
+import SafeAnchor, { SafeAnchorProps } from './SafeAnchor';
 
-import { BsPrefixComponent } from './helpers';
+import { BsPrefixComponent, BsPrefixComponentClass } from './helpers';
 
 export class AlertLink<
-  As extends React.ElementType = typeof SafeAnchor
+  As extends React.ElementType = BsPrefixComponentClass<'a', SafeAnchorProps>
 > extends BsPrefixComponent<As> {}
 
 export class AlertHeading<
--- a/types/components/DropdownItem.d.ts
+++ b/types/components/DropdownItem.d.ts
@@ -2,7 +2,11 @@ import * as React from 'react';
 
 import SafeAnchor, { SafeAnchorProps } from './SafeAnchor';
 
-import { BsPrefixComponent, SelectCallback, BsPrefixComponentClass } from './helpers';
+import {
+  BsPrefixComponent,
+  SelectCallback,
+  BsPrefixComponentClass,
+} from './helpers';
 
 export interface DropdownItemProps {
   active?: boolean;
--- a/types/components/DropdownItem.d.ts
+++ b/types/components/DropdownItem.d.ts
@@ -1,8 +1,8 @@
 import * as React from 'react';
 
-import SafeAnchor from './SafeAnchor';
+import SafeAnchor, { SafeAnchorProps } from './SafeAnchor';
 
-import { BsPrefixComponent, SelectCallback } from './helpers';
+import { BsPrefixComponent, SelectCallback, BsPrefixComponentClass } from './helpers';
 
 export interface DropdownItemProps {
   active?: boolean;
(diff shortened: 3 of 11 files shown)


types/components/Spinner.d.ts
```diff
@@ -19,7 +19,7 @@ export interface SpinnerProps {
 }
 
 declare class Spinner<
-  As extends React.ElementType = typeof Spinner
+  As extends React.ElementType = 'div'
```
taion: oops

olmobrutall: You mean is my error? I don’t think makes any sense that Spinner passes props to another Spinner, so my fix ok. Am I right?

taion: no, i mean on our end. this was a silly typo

glenjamin:
I think I just-about understand this.

Could you add some cases to https://github.com/olmobrutall/react-bootstrap/blob/be32ad50859407cd9e1dd00bea55ec3ed594b2ce/types/simple.test.tsx which fail before and pass now?

olmobrutall:
Not so easy because the problem before was that the type checked was **accepting every prop argument as valid** for react elements of type AlerLink,  DropdownItem, NavLink, Spinner, ToggleButton, ToggleButtonGroup and Tabs (new), because they had `As extends React.ElementType = typeof SomeOtherReactComponent`.

`simple.test.tsx` can only check for valid programs that do not typecheck, not for invalid programs that typecheck.

In the TS playground snippet above you can see how `TabsRight` (at the end) using the new `BsPrefixComponentClass<'div', NavProps>` is able to reject `foo` as invalid, while `TabsWrong` using the current `typeof Nav` takes it without complains (producing at the end a run-time error. 

About fully understating it... I'm with you. With conditional types things get crazy complicated fast. 

That's why my first PR was more like "kill it all and make it simple". But by using a fully determined ComponentClass instead of a generic one with default type parameters the inference works. 

To get full understating we need some TS masters: @ahejlsberg @DanielRosenwasser, @weswigham, some idea?

glenjamin:
> `simple.test.tsx` can only check for valid programs that do not typecheck, not for invalid programs that typecheck.

Ah yes, this rings a bell now - Flow has an annotation for "this should be a type error" for use in tests like this, but I don't recall typescript having an equivalent.

jquense:
you can use  `// $ExpectERror` and `// $ExpectType Foo` comments in the type tests like in astroturf: https://github.com/4Catalyzer/astroturf/blob/master/types/test.tsx

olmobrutall:
Done :)

olmobrutall:
I added a few examples using `as`, positive and negative examples. The inference mechanism didn't work for the example `<Button as="img" bla="foo" />`, producing no errors, but by explicitly setting it `<Button<'img'> as="img" bla="foo" />` then works. 

It's still a much better situation because before didn't produce any error for non-as cases.

Issue #4502, opened 2019-09-21
Option to copy import statement on click 


kushmisra:
## Feature request
While users go through the documentation and components which they need to use in their projects, they want to quickly copy the import statement. Currently, import statements required are explicitly mentioned but there is no easy way to copy them.

## Solution
 We should add a small icon there clicking to which, will copy the import statements. 

## Additional context
I will like to take up this addition.

bpas247:
closed by #4546

PR #4546, opened 2019-09-30
docs: Added copy to clipboard option for import statements in API doc…

jacruzca:
This PR aims to implement the requested feature in #4502. This is what I did:

- Added `copy-text-to-clipboard` lib as a dependency in `www` folder
- Created a new `CopyImport` component which displays a copy icon that performs the mentioned functionality
- Added `CopyImport` to `ComponentApi` next to the import statement `code` section.

This is how it looks:

imagen

Please let me know if you need something else.

### DIFF
--- a/www/src/components/ComponentApi.js
+++ b/www/src/components/ComponentApi.js
@@ -1,32 +1,13 @@
 import { graphql } from 'gatsby';
 import kebabCase from 'lodash/kebabCase';
 import React from 'react';
-import styled from 'astroturf';
-
-import Heading from './Heading';
 import Anchor from './Anchor';
+import Heading from './Heading';
+import ImportApi from './ImportApi';
 import LinkToSource from './LinkToSource';
 import PropTable from './PropTable';
 
-const Keyword = styled('span')`
-  color: #a626a4;
-`;
-const Code = styled('code')`
-  padding: 0;
-  display: block;
-  color: #50a14f;
-  background-color: transparent;
-  margin-bottom: 1rem;
-`;
-
 const propTypes = {};
-
-const Import = ({ name }) => (
-  <Code aria-label={`Import code for the ${name} component`}>
-    <Keyword>import</Keyword> {name} <Keyword>from</Keyword> 'react-bootstrap/
-    {name}'
-  </Code>
-);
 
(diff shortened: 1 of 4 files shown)


www/src/components/ComponentApi.js
```diff
@@ -22,10 +22,13 @@ const Code = styled('code')`
 const propTypes = {};
 
 const Import = ({ name }) => (
```
taion: Move this into `CopyImport` to remove the duplication?

www/src/components/CopyImport.js
```diff
@@ -0,0 +1,30 @@
...
+import Tooltip from 'react-bootstrap/Tooltip';
+
+const Link = styled('span')`
+  font-size: 1rem;
+  padding: 0 0.5rem;
+  cursor: pointer;
+`;
+
+export default ({ name }) => {
+  const textToCopy = `import ${name} from 'react-boostrap/${name}'`;
```
taion:
```suggestion
  const textToCopy = `import ${name} from 'react-bootstrap/${name}'`;
```

www/src/components/ComponentApi.js
jacruzca: Sure. I encapsulated both display and copy feature in that component and renamed it `ImportApi`. Is that okay?

www/src/components/ImportApi.js
```diff
@@ -0,0 +1,52 @@
...
+  margin-bottom: 1rem;
+`;
+
+const CopyImport = ({ name }) => {
+  const textToCopy = `import ${name} from 'react-bootstrap/${name}'`;
+
+  return (
+    <OverlayTrigger
+      overlay={
+        <Tooltip id={`copy-${name}-import-tooltip`}>Copy import code</Tooltip>
```
bpas247: Could we change the overlay text to indicate to the end-user that it's been copied after they click on it? E.g. changing the text from `Copy import code` to `Copied!`.

jacruzca: Done :) Thanks for the review. The text is now changed to `Copied!` when the button is clicked and restored back to the original text after 2 seconds.

taion: @bpas247 can you handle merging this then?

Issue #312, opened 2014-12-16
Modal cannot be scrolled if content overflows viewport


cmwelsh:
On the live demo there is an issue with the modal component:

http://react-bootstrap.github.io/components.html

Scroll down to Modals and then click Launch Demo Modal. When scrolling with the modal open, the modal content does not scroll and the greyed out page content behind it scrolls instead.

Tested in Firefox 34 and Chrome 39.

Compare with behavior on official Bootstrap website: http://getbootstrap.com/javascript/#modals

mtscout6:
This was fixed by #321.

PR #4567, opened 2019-10-02
docs: updating example in overlay-trigger documentation

malsf21:
Hey there! This is a super quick PR that implements a fix for the problem in [react-bootstrap/react-overlays#312](https://github.com/react-bootstrap/react-overlays/issues/312). It now uses the `Tooltip` component, which properly handles down the props that were causing problems!

### DIFF
--- a/www/src/examples/Overlays/OverlayTrigger.js
+++ b/www/src/examples/Overlays/OverlayTrigger.js
@@ -1,28 +1,10 @@
-const renderTooltip = ({
-  placement,
-  scheduleUpdate,
-  arrowProps,
-  outOfBoundaries,
-  show,
-  ...props
-}) => (
-  <div
-    {...props}
-    style={{
-      backgroundColor: 'rgba(0, 0, 0, 0.85)',
-      padding: '2px 10px',
-      color: 'white',
-      borderRadius: 3,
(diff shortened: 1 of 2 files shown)


www/src/examples/Overlays/OverlayTrigger.js
```diff
@@ -1,4 +1,11 @@
-const renderTooltip = props => (
+const renderTooltip = ({
```
taion: we can do simpler than this:

```js
function renderTooltip(props) {
  return <Tooltip {...props}>Simple tooltip</Tooltip>;
}
```

and change `placement` below to just `right` instead of `right-start`

malsf21: Thank you for the suggestion! Implemented it. Do you know if this still resolves the issue in #312?

malsf21:
> neither necessarily fixes the root issue (which i would argue is that the docs in react-overlays are bad... or maybe that the API there needs to be improved)

Gotcha, thanks for explaining it! Is there anything I can do in this PR to help fix the root issue (e.g. rewrite the docs)?

taion:
oops misclicked.

i'm not sure how you could if it's a problem in a different repo

PR #4573, opened 2019-10-03
Increase overlay trigger test coverage

fr-ser:
Increase test coverage of Overlay(Trigger) by checking all trigger events and their respective hide events (and small refactoring and bug fixing).

Issue: https://github.com/react-bootstrap/react-bootstrap/issues/3889

One thing I would have liked to do better (help/input appreciated):
I could not really get the tooltip element to vanish entirely. I think it has something to do with the transition, but I am not sure.
While the hidden state is triggered and the underlying Overlay element is updated the tooltip is not removed from the DOM.

I also had problems testing the delay case, which delayed the reaction, but actually never fired to the transition to the correct state :man_shrugging: 

The test I would have liked to write but does not work: 

```
  it('Delays the trigger event if the delay prop is passed', (done) => {
    const wrapper = mount(
      <OverlayTrigger
        defaultShow
        delay={1}
        trigger="hover"
        overlay={<Div className="test" />}
      >
        <button type="button">button</button>
      </OverlayTrigger>,
    );
    wrapper.assertSingle('div.test.show');

    wrapper.find('button').simulate('mouseOut');

    wrapper.assertSingle('div.test.show');

    setTimeout(() => {
      wrapper.assertNone('div.test.show');
      done()
    }, 2);
  });
```

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -72,7 +72,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  defaultOverlayShown: false,
+  defaultShow: false,
   trigger: ['hover', 'focus'],
 };
 
@@ -149,19 +149,18 @@ class OverlayTrigger extends React.Component {
 
   handleFocus = e => {
     const { onFocus } = this.getChildProps();
-    this.handleShow(e);
+    this.handleShow();
     if (onFocus) onFocus(e);
   };
 
   handleBlur = e => {
     const { onBlur } = this.getChildProps();
-    this.handleHide(e);
+    this.handleHide();
     if (onBlur) onBlur(e);
   };
 
   handleClick = e => {
     const { onClick } = this.getChildProps();
-
     if (this.state.show) this.hide();
     else this.show();
 
@@ -184,7 +183,7 @@ class OverlayTrigger extends React.Component {
     const related = e.relatedTarget || e.nativeEvent[relatedNative];
 
     if ((!related || related !== target) && !contains(target, related)) {
-      handler(e);
+      handler();
     }
   }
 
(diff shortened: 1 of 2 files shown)


fr-ser:
Also what about this line:
`let triggers = trigger ? [].concat(trigger) : [];`

We try to make sure that trigger is set before assigning the triggers variable.
But we have defaultProps for trigger... Personally I would remove this check since we have defaultProps.
Is there a guideline for this in this repo?

fr-ser:
Also a maybe bug:

While focus and hover call handleShow
```
  handleBlur = e => {
    const { onBlur } = this.getChildProps();
    this.handleHide();
    if (onBlur) onBlur(e);
  };
```

The click trigger calls hide directly (avoiding the possibility of a delay/transition)

```
  handleClick = e => {
    const { onClick } = this.getChildProps();
    if (this.state.show) this.hide();
    else this.show();
 
    if (onClick) onClick(e);
  };
```

Is this intended?

src/OverlayTrigger.js
```diff
@@ -212,18 +211,18 @@ class OverlayTrigger extends React.Component {
 
     const triggerProps = {};
 
-    let triggers = trigger == null ? [] : [].concat(trigger);
+    let triggers = trigger ? [].concat(trigger) : [];
 
-    if (triggers.indexOf('click') !== -1) {
+    if (triggers.includes('click')) {
```
taion: i'm not sure we can use`Array.prototype.includes` without modifying our transpilation settings; this transpiles as-is to

```js
    if (triggers.includes('click')) {
      triggerProps.onClick = this.handleClick;
    }
```

but `Array.prototype.includes` is not available on IE full stop https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

and i don't think we want to require our users to pull in polyfills

taion:
On click delay, this matches the upstream logic: https://github.com/twbs/bootstrap/blob/60559d44a2166708a4dc2f6ccb835052da08ab65/js/src/tooltip.js#L534-L560

I think it makes sense. If the user deliberately clicks on something, there's no reason to hide the overlay. The user probably wants to see it

taion:
For triggers, I think we actually _do_ want to allow the user to pass in `null` to get "no triggers". I'm guessing that's why that logic is there

taion:
And it's simpler in the rest of the code if we always coerce it to an array

taion:
on the timeouts... yeah, they're a bit fuzzy. maybe use fake timers like in the carousel PR?

src/OverlayTrigger.js
fr-ser: While I would normally go for adjusting the transpilation, for this PR it totally makes sense to leave it as is then.

fr-ser:
Alright. The difference for the click trigger makes sense then.

Regarding the array concatenation. Wouldn't this line make more sense then:

let triggers = [].concat(trigger) ;

The null case results in no triggers, we always get an array, but we do not check for something that is  set by default...

fr-ser:
I still can't get the delay to work, though... :man_shrugging: 

```
    it.only('Delays the trigger event if the delay prop is passed', () => {
      const clock = sinon.useFakeTimers();
      try {
        const wrapper = mount(
          <OverlayTrigger
            defaultShow
            delay={10}
            trigger="hover"
            overlay={<Div className="test" />}
          >
            <button type="button">button</button>
          </OverlayTrigger>,
        );
        wrapper.assertSingle('div.test.show');

        wrapper.find('button').simulate('mouseOut');

        wrapper.assertSingle('div.test.show');

        clock.tick(50);
        wrapper.assertNone('div.test.show');
      } finally {
        clock.restore();
      }
    });
```

mxschmitt:
Closed because repo was removed.

Issue #4576, opened 2019-10-03
Import react-bootstrap 1.0.0-beta.13 Module not found: Can't resolve 'react-bootstrap'


xAgustin93:
## Describe the bug
When importing react-bootstrap returns the error Module not found: Can't resolve 'react-bootstrap' in the version 1.0.0-beta.13.

In the version 1.0.0-beta.12 everything works well

## To Reproduce
yarn add react-bootstrap

Open react componente and import one componen the react-bootstrap.
import { Container } from "react-bootstrap";

## Reproducible Example

Example in [CodeSandbox](https://codesandbox.io/s/rough-glitter-2fl8d).

taion:
@jquense file-butler appears to not be buttle-ing `index.js`... is there a config option we're missing here?

taion:
FWIW we strongly recommend using imports like

```js
import Foo from 'react-bootstrap/Foo';
```

which should still work

xAgustin93:
Yes taion the imports if they work 
import Foo from 'react-bootstrap/Foo';

but if you want to use more components it's more comfortable to do it the other way
import { Container, Row, Col, Spinner } from 'react-bootstrap';

I think they must work both ways

bgok:
> FWIW we strongly recommend using imports like
> 
> ```js
> import Foo from 'react-bootstrap/Foo';
> ```
> 
> which should still work

Unfortunately, the alternate import format doesn't work either.

taion:
works fine for me: https://codesandbox.io/embed/ecstatic-water-npx6s

taion:
fix released in beta 14

PR #4577, opened 2019-10-04
fix: Fix main and module

taion:
Fixes https://github.com/react-bootstrap/react-bootstrap/issues/4576

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   },
   "homepage": "https://react-bootstrap.github.io/",
   "sideEffects": false,
-  "main": "lib/index.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,8 +8,8 @@
   },
   "homepage": "https://react-bootstrap.github.io/",
   "sideEffects": false,
-  "main": "lib/index.js",
-  "module": "lib/es/index.js",
+  "main": "lib/cjs/index.js",
```
bgok: I think this should be `cjs/index.js` and the following line is `esm/index.js`

taion: our build process strips the prefix

PR #4592, opened 2019-10-04
fix(Table): allow passed in refs to be properly forwarded

DoctorHowser:
#4194

Updates Table component to play nice with a passed ref.

### DIFF
--- a/src/Table.js
+++ b/src/Table.js
@@ -1,64 +1,62 @@
 import classNames from 'classnames';
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useBootstrapPrefix } from './ThemeProvider';
 
-import { createBootstrapComponent } from './ThemeProvider';
+const propTypes = {
+  /**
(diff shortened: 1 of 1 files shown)


src/Table.js
```diff
@@ -66,24 +65,25 @@ class Table extends React.Component {
       size,
       variant,
       responsive,
-      ...props
-    } = this.props;
-
+      ...otherProps
```
taion: just `...props` i think... unless there’s a name collision

DoctorHowser: It certainly could be. Mostly I was following a different example  (https://github.com/react-bootstrap/react-bootstrap/pull/3992/files) that also had otherProps, and to at least delineate that it is other props than what is being used with this stuff. 

Would require a change here, and a change on line 84.

taion: let’s do that. the example there already has a `props` so we needed another name

PR #4593, opened 2019-10-04
fix(Pagination): support for ref attribute

mxschmitt:
Related to #4194

### DIFF
--- a/src/Pagination.js
+++ b/src/Pagination.js
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React from 'react';
 
-import { createBootstrapComponent } from './ThemeProvider';
+import { useBootstrapPrefix } from './ThemeProvider';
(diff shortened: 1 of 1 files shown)


src/Pagination.js
```diff
@@ -13,44 +13,44 @@ import PageItem, { First, Prev, Ellipsis, Next, Last } from './PageItem';
...
+   * */
+  bsPrefix: PropTypes.string,
+
+  /**
+   * Set's the size of all PageItems.
+   *
+   * @type {('sm'|'lg')}
+   */
+  size: PropTypes.string,
+};
```
bpas247: I think to keep our code base consistent we should define our propTypes *before* we define the component. Some examples include [Accordion](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Accordion.js), [Alert](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Alert.js), [Nav](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Nav.js), etc. I'm not sure if we want to change this convention in the future though

PR #4596, opened 2019-10-04
fix(Popover): support for ref attribute

mxschmitt:
Related to #4194


src/Popover.js
```diff
@@ -57,40 +54,47 @@ const defaultProps = {
...
+    {
+      bsPrefix,
+      placement,
+      className,
+      style,
+      children,
+      content,
+      arrowProps,
+      scheduleUpdate: _,
+      outOfBoundaries: _1,
```
bpas247: should these just be removed? or are we destructuring them for backwards-compatibility?

taion: These come from the react-overlays `<Overlay>`. We're removing them to avoid passing unknown props to DOM children I think.

PR #4603, opened 2019-10-05
tests: added more Nav/Item tests

fr-ser:
Issue: #3889

Most of the not covered statements in the (Abstract)Nav are null check of the reference or the navItem key.
I honestly could not figure out the scenarios when those would be null.
If you could tell me the scenarios I would also add tests for those

### DIFF
--- a/test/NavSpec.js
+++ b/test/NavSpec.js
@@ -35,7 +35,7 @@ describe('<Nav>', () => {
     assert.notOk(items.at(1).is('.active'));
   });
 
-  it('should adds variant class', () => {
+  it('should add variant class', () => {
     mount(
       <Nav variant="tabs">
         <Nav.Link eventKey={1}>Pill 1 content</Nav.Link>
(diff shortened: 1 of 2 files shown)


test/NavSpec.js
```diff
@@ -30,7 +35,7 @@ describe('<Nav>', () => {
     assert.notOk(items.at(1).is('.active'));
   });
 
-  it('Should adds variant class', () => {
+  it('should adds variant class', () => {
```
taion:
```suggestion
  it('should add variant class', () => {
```

test/NavSpec.js
```diff
@@ -39,7 +44,7 @@ describe('<Nav>', () => {
     ).assertSingle('div.nav.nav-tabs');
   });
 
-  it('Should adds justified class', () => {
+  it('should adds justified class', () => {
```
taion:
```suggestion
  it('should add justified class', () => {
```

test/NavSpec.js
```diff
@@ -48,7 +53,7 @@ describe('<Nav>', () => {
     ).assertSingle('div.nav.nav-justified');
   });
 
-  it('Should adds fill class', () => {
+  it('should adds fill class', () => {
```
taion:
```suggestion
  it('should add fill class', () => {
```

PR #4606, opened 2019-10-06
Add `navbar` to `DropdownProps` type

hassankhan:


taion:
This is a private prop. Use `<NavDropdown>` to render a nav dropdown.

hassankhan:
Thanks for pointing that out, was clearly having a bit of a moment 😅

Issue #3508, opened 2019-02-28
Class "carousel-item" listed twice per item div


jokokoloko:
The "carousel-item" class is listed twice per item div.

![image](https://user-images.githubusercontent.com/5240653/53549092-29d29100-3ae8-11e9-8d56-d05947aa9b9b.png)

taion:
getting injected from parent as well... i guess it doesn't make sense to go fully context-based here, though

bpas247:
#3508 closes this issue,

PR #4609, opened 2019-10-06
#3508 fix duplicate class "carousel-item"

MKrupauskas:
Regarding issue #3508.

The class "carousel-item" is duplicated because it gets passed from the parent as well as the carousel item itself:

`export default createWithBsPrefix('carousel-item');`

We fix this by no longer passing the class from the parent.


jquense:
thanks!

PR #4618, opened 2019-10-07
feat(AbstractNavItem): expose generated props id

### LINKED ISSUE #3940: <Nav.Item> in <Nav> ignores the id prop
## Describe the bug

Can not set id attribute to ListGroupItem

## To Reproduce

Just type something like this and look into browser DevTools - ListGroup has an id, but list items doesn't
```   
	  <ListGroup id="items">
        <ListGroupItem id="item1">Pit</ListGroupItem>
        <ListGroupItem id="item2">John</ListGroupItem>
        <ListGroupItem id="item3">Ivan</ListGroupItem>
      </ListGroup>
```

Screenshot from 2019-06-22 12-39-14

## Reproducable Example

https://codesandbox.io/embed/frosty-ride-252jt

baktiaditya:
fix #3940

### DIFF
--- a/src/AbstractNavItem.js
+++ b/src/AbstractNavItem.js
@@ -49,7 +49,10 @@ const AbstractNavItem = React.forwardRef(
     if (navContext) {
       if (!props.role && navContext.role === 'tablist') props.role = 'tab';
 
-      id = navContext.getControllerId(navKey);
+      const controllerId = navContext.getControllerId(navKey);
+      if (controllerId) {
+        id = controllerId;
+      }
 
       props['data-rb-event-key'] = navKey;
       props['aria-controls'] = navContext.getControlledId(navKey);
--- a/src/AbstractNavItem.js
+++ b/src/AbstractNavItem.js
@@ -35,6 +35,7 @@ const AbstractNavItem = React.forwardRef(
       onSelect,
       onClick,
       as: Component,
+      id: _id,
       ...props
     },
     ref,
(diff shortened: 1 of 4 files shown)


taion:
I don't think this is the right approach. I think `AbstractNav` should default `getControlledId` and `getControllerId` to `null` instead of `noop` in its provided context, and that those two context values should only be used when provided.

baktiaditya:
updated, hope this is approach is correct

baktiaditya:
updated, hope this approach is the correct one

src/AbstractNavItem.js
```diff
@@ -34,20 +35,26 @@ const AbstractNavItem = React.forwardRef(
...
     const navKey = makeEventKey(eventKey, props.href);
     const parentOnSelect = useContext(SelectableContext);
     const navContext = useContext(NavContext);
+    let id = _id;
 
     let isActive = active;
     if (navContext) {
       if (!props.role && navContext.role === 'tablist') props.role = 'tab';
 
+      const controllerId = navContext.getControllerId(navKey);
```
taion: we want something like:

```js
if (navContext.getControllerId) {
  props.id = navContext.getControllerId(navKey);
}
```

with the default removed in `AbstractNav`

atsolberg:
Any update on this PR? Not having id's is brutal for the analytics/tracking I need to do.

taion:
My comment on https://github.com/react-bootstrap/react-bootstrap/pull/4618/files#r332227426 still stands – this PR doesn't do what it's supposed to do.

bpas247:
Since this has been stale for quite awhile, I'm going to go ahead and close it 🙂 

@baktiaditya please feel free to re-open this if you make any new progress on incorporating @taion's feedback.

PR #4626, opened 2019-10-08
Add ref Forwarding to NavDropdown

DoctorHowser:
Adds in ref forwarding for NavDropdown per
#4194


src/NavDropdown.js
```diff
@@ -51,28 +51,27 @@ class NavDropdown extends React.Component {
...
-      </Dropdown>
-    );
-  }
-}
+      <Dropdown.Menu role={menuRole} rootCloseEvent={rootCloseEvent}>
+        {children}
+      </Dropdown.Menu>
+    </Dropdown>
+  ),
+);
```
jquense: needs a `displayName` explicitly added here

DoctorHowser: I'd be happy to add it, judging from other components it would be
`NavDropdown.displayName = 'NavDropdown';`

Do you have resources/reasoning for why it's needed? I didn't see anything in the component that mentioned it before, and all I can infer is you want to avoid ForwardedRef being the component that shows in an error?

DoctorHowser:
Added

DoctorHowser:
Doh. I was being so careful. Will do.

jquense:
thanks!

Issue #4398, opened 2019-09-03
React 16.9.x deprecation warnings with 0.32.4


lvpro:
React 16.9 is now spamming the console with warnings regarding the use of componentWillMount, componentWillReceiveProps, and componentWillUpdate.

Apparently rbs 0.32.4 still uses these in most of the components.  Since seamlessly moving to BS4 is no small feat styling-rule wise, I'm curious .... is there any chance the BS3 based release will be updated to replace these calls to make React happy?

I believe these calls will stop working all together in React 17.

jquense:
I believe someone recently did the work of swapping out for `UNSAFE_` prefixed versions, and we need to release that. In terms of proper migration via switching to gDSFP, that is less likely from the maintainers here, since most (all?) of us are using v4 now. We would happily accept PR's tho!

MatthewHerbst:
@jquense is there anything blocking that release from going out?

mturley:
+1 for releasing the version with `UNSAFE_` prefixes. My team is maintaining a legacy version of our component library that depends on react-bootstrap and bootstrap v3, and we need to support newer React versions. @jquense is there anything we can do to help with that release? Our alternative is to make a copy of the source in-house, which obviously isn't ideal.

mxschmitt:
We also use the v3 variant of react-bootstrap in production and would benefit of a new release.

wheeler:
`UNSAFE_` prefixes were introduced in `react@16.3` so we would need to bump the min version in `package.json`. I'm happy to prepare a PR against the `bs3-dev` branch.

taion:
sounds good

wheeler:
React required version PR here: https://github.com/react-bootstrap/react-bootstrap/pull/4530

nortonwong:
The remaining lifecycle warnings are from the higher-order-component `Uncontrollable`. I've submitted a pull request at https://github.com/jquense/uncontrollable/pull/40 which migrates those usages.

nortonwong:
Opened a `react-overlays` upgrade PR at https://github.com/react-bootstrap/react-bootstrap/pull/4639. Some breaking changes are involved.

kimyu92:
@taion @bpas247 would it be possible to publish an official npm package release v0.32.5 which include changes of https://github.com/react-bootstrap/react-bootstrap/commits/bs3-dev up till c9c298f1a5b941d1b39a595c9a5440a890d8c58f before breaking changes from #4639

[...]

PR #4639, opened 2019-10-09
Fix(Bs3): Remove unsafe lifecycles by upgrading react-overlays to ^1.2.0

nortonwong:
Closes https://github.com/react-bootstrap/react-bootstrap/issues/4398.

Use `react-overlays` 1.x due to fewer breaking changes. In 2.x, **RootCloseWrapper** has been replaced by the [`useRootClose` hook](https://github.com/react-bootstrap/react-overlays/blob/v2.0.0/src/useRootClose.js). This change set should still be compatible with `react` ^16.3.0.

#### Adapt **Modal**:
* Maintain backwards-compatibility for `props.backdropClassName` and `props.backdropStyle` by using a fallback value for `props.renderBackdrop`.
* Convert **ModalDialog** into a `forwardRef`, because **BaseModal** [expects the ref to be a DOM node](https://github.com/react-bootstrap/react-overlays/blob/v1.2.0/src/Modal.js#L371).

#### Adapt **Overlay** component family:
* Take changes from master for components integrated with `react-popper`: **Overlay**, **OverlayTrigger**, **Tooltip**, and **Popover**.
* Accept breaking prop changes from master.
* Drop changes where `props.bsClass` was converted into `props.bsPrefix`, since the new utilities require [`useContext`](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ThemeProvider.js#L19) and a branch-wide prop rename.
* Drop **Popover.Title** and **Popover.Content** as out of scope.

#### Adapt tests:
* Fix tests broken by using [ReactTestUtils on `forwardRef` components](https://github.com/facebook/react/issues/13455).

### DIFF
--- a/package.json
+++ b/package.json
@@ -122,7 +122,7 @@
     "keycode": "^2.2.0",
     "prop-types": "^15.6.1",
     "prop-types-extra": "^1.0.1",
-    "react-overlays": "^0.8.0",
+    "react-overlays": "^1.2.0",
     "react-prop-types": "^0.4.0",
     "react-transition-group": "^2.0.0",
     "uncontrollable": "^7.0.2",
--- a/src/DropdownMenu.js
+++ b/src/DropdownMenu.js
@@ -3,7 +3,7 @@ import keycode from 'keycode';
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
-import RootCloseWrapper from 'react-overlays/lib/RootCloseWrapper';
+import RootCloseWrapper from 'react-overlays/RootCloseWrapper';
 
 import {
   bsClass,
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -6,8 +6,8 @@ import getScrollbarSize from 'dom-helpers/util/scrollbarSize';
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
-import BaseModal from 'react-overlays/lib/Modal';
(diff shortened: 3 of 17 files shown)


nortonwong:
Unfortunately, `yarn run build-docs` on bs3-dev gives me
```js
ReferenceError: React is not defined
```
even without the changes in this PR.

nortonwong:
Turns out the `raw-loader` regex wasn't cross-OS. 🤷

Because the setup didn't understand `forwardRef` components, the ModalDialog graphql result was null. I upgraded `react-docgen` past [3.0.0](https://github.com/reactjs/react-docgen/releases/tag/v3.0.0), and there were a few other failures until I bumped the `gatsby` packages.

mturley:
@nortonwong anything I can do to help you get this merged? My team has some folks bugging us to fix lifecycle method warnings, we would love to help get the fixes released if we can.

nortonwong:
@mturley Thanks. I'm waiting on feedback at the moment.

The code coverage has dropped for Overlay and OverlayTrigger, so if this gets merged eventually those files will need more tests. In the meantime I'm waiting for the maintainers to review this.

nortonwong:
In the meantime, you can silence all the other warnings by using the latest from the branch in your package.json:
```js
    "react-bootstrap": "react-bootstrap/react-bootstrap#bs3-dev",
```

wheeler:
Also interested to help if possible. Looks like this PR pulls most of its hefty changes out of the bs4 master branch. Master has approximately the same coverage so we can probably dismiss the `codecov/patch`?

@taion @jquense anything that would be helpful for review?

PR #4640, opened 2019-10-10
[#4194] - ref forwarding for ToggleButton

zhuber:
This is a first pass at handling the ref forwarding work for the ToggleButton.

# Related issue:
https://github.com/react-bootstrap/react-bootstrap/issues/4194

# Note:
- I think that `innerRef` can be removed for this component, as well as the previously added "wrapping ref" that was exported. I'm new to this project so that could totally be wrong.

### DIFF
--- a/src/ToggleButton.js
+++ b/src/ToggleButton.js
@@ -65,13 +65,9 @@ const ToggleButton = React.forwardRef(
   ) => {
     const [focused, setFocused] = useState(false);
 
-    const handleFocus = e => {
-      if (e.target.tagName === 'INPUT') setFocused(true);
-    };
-
-    const handleBlur = e => {
-      if (e.target.tagName === 'INPUT') setFocused(false);
-    };
+    const toggleFocus = useCallback(e => {
+      if (e.target.tagName === 'INPUT') setFocused(!focused);
+    }, []);
 
     return (
       <Button
@@ -94,8 +90,8 @@ const ToggleButton = React.forwardRef(
           autoComplete="off"
           checked={!!checked}
           disabled={!!disabled}
-          onFocus={handleFocus}
-          onBlur={handleBlur}
+          onFocus={toggleFocus}
+          onBlur={toggleFocus}
           onChange={onChange || noop}
         />
 
--- a/src/ToggleButton.js
+++ b/src/ToggleButton.js
@@ -1,6 +1,6 @@
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import React from 'react';
+import React, { useState } from 'react';
 
 import Button from './Button';
 
--- a/src/ToggleButton.js
+++ b/src/ToggleButton.js
@@ -29,20 +29,10 @@ const propTypes = {
   disabled: PropTypes.bool,
 
   /**
-   * The focused state of the input
-   */
-  focused: PropTypes.bool,
-
-  /**
    * A callback fired when the underlying input element changes. This is passed
    * directly to the `<input>` so shares the same signature as a native `onChange` event.
    */
   onChange: PropTypes.func,
-
-  /**
-   * A callback fired when the underlying input element is focused or blurred.
-   */
-  toggleFocus: PropTypes.func,
 
   /**
    * The value of the input, should be unique amongst it's siblings when nested in a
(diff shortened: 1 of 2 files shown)


src/ToggleButton.js
```diff
@@ -44,42 +55,39 @@ const propTypes = {
...
       children,
       name,
       className,
       checked,
       type,
       onChange,
+      toggleFocus,
       value,
       disabled,
+      focused,
```
jquense: this should be state, or is there a larger plan that i'm not seeing here?

taion: there's no reason to control focused state. add a `[focus, setFocus] = useState(...)`

src/ToggleButton.js
```diff
@@ -71,15 +59,24 @@ class ToggleButton extends React.Component {
...
       disabled,
       inputRef,
-      innerRef,
       ...props
-    } = this.props;
-    const { focused } = this.state;
+    },
+    ref,
+  ) => {
+    const [focused, setFocus] = useState(false);
```
taion:
```suggestion
    const [focused, setFocused] = useState(false);
```

src/ToggleButton.js
```diff
@@ -71,15 +59,24 @@ class ToggleButton extends React.Component {
...
+
+    const handleFocus = e => {
+      if (e.target.tagName === 'INPUT') setFocus(true);
```
taion:
```suggestion
      if (e.target.tagName === 'INPUT') setFocused(true);
```

src/ToggleButton.js
```diff
@@ -71,15 +59,24 @@ class ToggleButton extends React.Component {
...
-      innerRef,
       ...props
-    } = this.props;
-    const { focused } = this.state;
+    },
+    ref,
+  ) => {
+    const [focused, setFocus] = useState(false);
+
+    const handleFocus = e => {
```
taion: add `useCallback`

src/ToggleButton.js
```diff
@@ -71,15 +59,24 @@ class ToggleButton extends React.Component {
...
+      if (e.target.tagName === 'INPUT') setFocus(true);
+    };
+
+    const handleBlur = e => {
```
taion: add `useCallback`

src/ToggleButton.js
```diff
@@ -71,15 +59,24 @@ class ToggleButton extends React.Component {
...
+      if (e.target.tagName === 'INPUT') setFocus(false);
```
taion:
```suggestion
      if (e.target.tagName === 'INPUT') setFocused(false);
```

types/components/ToggleButton.d.ts
```diff
@@ -9,10 +9,11 @@ export interface ToggleButtonProps {
   name?: string;
   checked?: boolean;
   disabled?: boolean;
+  focused?: boolean;
```
taion:
```suggestion
```

types/components/ToggleButton.d.ts
```diff
@@ -9,10 +9,11 @@ export interface ToggleButtonProps {
...
   onChange?: React.ChangeEventHandler<this>;
+  onFocus?: React.FocusEventHandler<this>;
```
taion:
```suggestion
```

types/components/ToggleButton.d.ts
```diff
@@ -9,10 +9,11 @@ export interface ToggleButtonProps {
...
   value: unknown;
   inputRef?: React.LegacyRef<this>;
```
taion:
```suggestion
```

zhuber: I actually did not touch the `inputRef` because I'm not 100% sure I can scrap it. I did remove `innerRef` which was the one previously in use on the wrapper component.

src/ToggleButton.js
zhuber: I combined both focus/blur to a single toggle handler, since really it is just switching it between true/false. Let me know if this should be reverted to be more explicit (it seems like we should only be `onFocus` when it is currently blurred (`focused: false`) and `onBlur` when it is currently focused `focused: true`.

src/ToggleButton.js
```diff
@@ -97,19 +90,17 @@ class ToggleButton extends React.Component {
           autoComplete="off"
           checked={!!checked}
           disabled={!!disabled}
-          onFocus={this.handleFocus}
-          onBlur={this.handleBlur}
+          onFocus={toggleFocus}
```
taion: use separate `handleFocus` and `handleBlur`. those callbacks need fewer dependencies, plus it's more explicit – the action on focus is specifically to set focused to true, not to toggle the state

types/components/ToggleButton.d.ts
taion: it's just cruft; somehow that prop definition got duplicated

zhuber: So should I remove the `inputRef` in the component as well?

taion: my bad; you're right. i misread

src/ToggleButton.js
```diff
@@ -97,19 +94,17 @@ class ToggleButton extends React.Component {
...
 
         {children}
       </Button>
     );
-  }
-}
+  },
+);
 
 ToggleButton.propTypes = propTypes;
```
taion:
```suggestion
ToggleButton.displayName = 'ToggleButton';
```

taion: just added this myself

PR #4660, opened 2019-10-11
Updated Overlay Trigger to support flip property

### LINKED ISSUE #4262: Add flip prop to Overlay(Trigger)
If I add the flip property it works and is passed all the way to Overlay in react-overlays. It would be nice if it was officially typed and documented.

Now I have to use `<OverlayTrigger {...{ flip: true }}`... to bypass typescript errors.

Tarun047:
Fixes #4262


Tarun047:
Updated TS defs as well to support flip property

jquense:
Thanks!

PR #4661, opened 2019-10-11
chore: react overlays

jquense:
BREAKING CHANGE: custom Dropdown Menu's and Toggles can no longer be class components unless they are wrapped in forwardRef()

depends on https://github.com/react-bootstrap/react-overlays/pull/616/files


test/DropdownMenuSpec.js
```diff
@@ -33,4 +32,32 @@ describe('<Dropdown.Menu>', () => {
...
+
+  //   expect(() =>
+  //     mount(
+  //       <Parent>
+  //         <DropdownMenu show as={Menu}>
+  //           <DropdownItem>Item</DropdownItem>
+  //         </DropdownMenu>
+  //       </Parent>,
+  //     ),
+  //   ).to.throw();
```
jquense: IDK how to catch this error, anyone got ideas?

src/useWrappedRefWithWarning.js
```diff
@@ -0,0 +1,23 @@
+import invariant from 'invariant';
+import { useCallback } from 'react';
+import useMergedRefs from '@restart/hooks/useMergedRefs';
+
+const isReactClassComponent = c => c && c.isReactComponent;
```
taion: surely you can just inline this :p

src/useWrappedRefWithWarning.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default function useWrappedRefWithWarning(ref, componentName) {
+  if (!__DEV__) return ref;
+
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  const warningRef = useCallback(
+    refValue => {
+      invariant(
+        refValue == null || !isReactClassComponent(refValue),
```
taion:
```suggestion
        refValue == null || !refValue.isReactComponent
```

src/useWrappedRefWithWarning.js
```diff
@@ -0,0 +1,23 @@
...
+        `${componentName} injected a ref to a provided \`as\` component that resolved to a component instance instead of a DOM element. ` +
+          'Use `React.forwardRef` to provide the injected ref to the class component as a prop in order to pass it directly to a DOM element',
+      );
+    },
+    [componentName],
+  );
+  // eslint-disable-next-line react-hooks/rules-of-hooks
+  return useMergedRefs(warningRef, ref);
```
taion: I don't understand the value of the wrap/merge approach here. Why not just print out a warning?

test/DropdownMenuSpec.js
taion: Can we use `shouldWarn`?

Do we really want this to hard throw?

taion: In the worst case, you could put an error boundary around this guy, and assert on the error boundary having been hit

src/useWrappedRefWithWarning.js
jquense: yes, i thought originally the sigil was on the constructor

src/useWrappedRefWithWarning.js
jquense: The warning needs to happen in the ref handler itself, not in the hook call. It's effectively a callback ref that doesn't do anything with the value.

test/DropdownMenuSpec.js
jquense: this has an error boundary that doesn't seem to be hit. The error also still gets thrown crashing the test, which is confusing.

jquense: it's gonna hard error somewhere in Popper if we only warn

src/useWrappedRefWithWarning.js
taion: okay, i see... why can't we just use `findDOMNode` here?

jquense: We could but I'm attempting to remove the need here, to close out strict mode warnings

test/DropdownMenuSpec.js
taion: What happens if you try either using react test utils directly, or else doing shallow?

src/useWrappedRefWithWarning.js
taion: What if we fall back to using `findDOMNode` if the ref is a react component? Then maybe emit some appropriate warning?

jquense: idk that seems unnecessary, since this is a deprecated pattern we should probably use the moment for making breaking changes to switch over. It's an easy enough workaround in user code if folks want the old behavior

test/DropdownMenuSpec.js
jquense: shallow doesn't work with hooks really, and i've not had much luck doing anything directly. you wanna take a shot? I may just be dense

jquense:
ping

src/useWrappedRefWithWarning.js
```diff
@@ -0,0 +1,21 @@
+import invariant from 'invariant';
+import { useCallback } from 'react';
+import useMergedRefs from '@restart/hooks/useMergedRefs';
+
+export default function useWrappedRefWithWarning(ref, componentName) {
+  if (!__DEV__) return ref;
```
mxschmitt: can we not test for `process.env.NODE_ENV === "test"`?
So we can omit this __DEV__ variable.

jquense: this is actually shorthand for `process.env.NODE_ENV === 'production'` Babel replaces instances of `__DEV__` with the process.env check via the `dev-expression` plugin, as well as rewriting some of the invariant or warning calls so they can be dead-code eliminated when applicable

PR #4666, opened 2019-10-13
fix(Collapse): allow passed in refs to be properly forwarded 

jrwiegand:
My PR for Collapse forward ref. I am getting an error on the `AccordinCollapse` test without any idea on how to fix it. Help would be much appreciated.

Below is the error:
```
Error: Warning: React does not recognize the `getDimensionValue` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `getdimensionvalue` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
        in div (at AccordionCollapse.js:22)
        in Transition (at Collapse.js:191)
```

Thanks in advance to anyone who can point me in the right direction. 😄

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -125,7 +125,6 @@ const defaultProps = {
   mountOnEnter: false,
   unmountOnExit: false,
   appear: false,
-
   dimension: 'height',
   getDimensionValue,
 };
@@ -142,10 +141,11 @@ const Collapse = React.forwardRef((props, ref) => {
     ...otherProps
   } = props;
 
+  delete otherProps.dimension;
+  delete otherProps.getDimensionValue;
+
   const getDimension = () =>
-    typeof otherProps.dimension === 'function'
-      ? otherProps.dimension()
-      : otherProps.dimension;
+    typeof props.dimension === 'function' ? props.dimension() : props.dimension;
 
   // for testing
   const _getScrollDimensionValue = (elem, dimension) => {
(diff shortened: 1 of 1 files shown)


nortonwong:
You removed [these lines](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Collapse.js#L186):
```js
delete props.dimension;
delete props.getDimensionValue;
```

So `otherProps` now contains `dimension` and `getDimensionValue`.

jrwiegand:
@nortonwong I made some progress but I am still getting a couple more errors. Thoughts?

nortonwong:
[CollapseSpec.js](https://github.com/react-bootstrap/react-bootstrap/blob/787a19ed22f5292e493d494816ae1cb9cc9b28d6/test/CollapseSpec.js)

✖ [Should default to collapsed](https://github.com/react-bootstrap/react-bootstrap/blob/787a19ed22f5292e493d494816ae1cb9cc9b28d6/test/CollapseSpec.js#L33)

You need to set the `displayName` on Collapse, because `forwardRef` components have different names by default

✖ ["before each" hook for "Should have collapsing class"](https://github.com/react-bootstrap/react-bootstrap/blob/787a19ed22f5292e493d494816ae1cb9cc9b28d6/test/CollapseSpec.js#L46)
✖ [Defaults to height](https://github.com/react-bootstrap/react-bootstrap/blob/787a19ed22f5292e493d494816ae1cb9cc9b28d6/test/CollapseSpec.js#L177)
✖ [Uses getCollapsibleDimension if exists](https://github.com/react-bootstrap/react-bootstrap/blob/787a19ed22f5292e493d494816ae1cb9cc9b28d6/test/CollapseSpec.js#L187)

These tests used to grab the Collapse instance and call or stub the instance's methods. Since `forwardRef` components have no instance, you'll have to find a different way to express these tests.

jquense:
Sorry for the lack of response here. I'm not actually sure that Collapse _should_ support forwardRef since it doesn't actually render any ref-able thing. You are always required to provide a child so you could just attach the ref there directly?

jrwiegand:
I am closing this PR as `Collapse` was removed from the issue referenced.
https://github.com/react-bootstrap/react-bootstrap/issues/4194#issuecomment-575644811

PR #4690, opened 2019-10-15
fix(drop-down-toggle): migrate to be ref forwarder

hetpatel33:
Relevant to #4194 
@bpas247 @taion @mxschmitt

### DIFF
--- a/src/DropdownToggle.js
+++ b/src/DropdownToggle.js
@@ -7,7 +7,7 @@ import BaseDropdownToggle from 'react-overlays/DropdownToggle';
 import React from 'react';
 
 import Button from './Button';
-import { createBootstrapComponent } from './ThemeProvider';
+import { useBootstrapPrefix } from './ThemeProvider';
 
 const wrapRef = props => {
   const { ref } = props;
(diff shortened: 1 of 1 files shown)


src/DropdownToggle.js
```diff
@@ -7,80 +7,82 @@ import BaseDropdownToggle from 'react-overlays/DropdownToggle';
...
-    } = this.props;
+const defaultProps = {
+  as: Button,
+};
 
-    if (childBsPrefix !== undefined) {
-      props.bsPrefix = childBsPrefix;
-    }
+const DropdownToggle = React.forwardRef((props, ref) => {
+  const {
```
taion: can you spread this out in the function signature, and name the rest param `props` instead of `otherProps`?

hetpatel33: Sure @taion 
Can you also help me with adding the ref to the`BaseDropdownToggle` which is a functional component?
Should I pass it directly to the children?

jquense:
heads up this is already handled in https://github.com/react-bootstrap/react-bootstrap/pull/4661

hetpatel33:
@jquense @taion  Why no one is updating #4194?
Can you help me finding the issues which are up for pick?

src/DropdownToggle.js
taion: i think we need #4661, then, sorry

mxschmitt:
Closed because it will be already implemented in #4661. But @hetpatel33 thanks for your work anyway.

Issue #4314, opened 2019-08-22
Vertical navs of tabs variant use horizontal styling


akhilpai:
## Describe the bug

[Vertical navs](https://react-bootstrap.github.io/components/navs/#vertical) of the [tabs variant](https://react-bootstrap.github.io/components/navs/#tabs) use the default horizontal styling. 

## Reproducible Example

https://codesandbox.io/embed/code-sandbox-examples-9f8hc

## Expected behavior

Either there should be a warning in the documentation not to use the tabs nav variant with the flex-column css or there should be a way to define which direction the styling should be applied.

jquense:
Yeah the styles don't support that layout really. I agree that the docs should allude to it at least. It's also unclear in the upstream bootstrap docs as well

malsf21:
This seems like a pretty simple PR, I can handle it. Should I also implement a console-level warning, or just something in the docs?

taion:
Hmm, would it be possible for a Bootstrap theme to implement this sort of styling? If so, an unsilencable warning in code might be a bit much.

malsf21:
Maybe, but I haven't come across one? Basically, the theme has to override `nav-tabs` when it sees `flex-column` applied to the same element. I can just add it to the docs then with no warning!

kyletsang:
Looks like we can close this since docs have been updated with a note regarding vertical tabs

PR #4703, opened 2019-10-18
docs(Navs): mentions that vertical navs styled as tabs are unsupported by Bootstrap

malsf21:
Hey there! This is a very small PR that follows up on #4314, which is that vertical navs (created by `class="flex-column"` are not styled properly when used on Tabs (`variant="tabs"`), which is an issue core to Bootstrap's stylesheet.

Let me know if I should change the language of the docs!

### DIFF
--- a/www/src/pages/components/navs.mdx
+++ b/www/src/pages/components/navs.mdx
@@ -84,7 +84,9 @@ responsive versions to stack in some viewports but not others (e.g.
 ## Tabs
 
 Visually represent nav items as "tabs". This style pairs nicely with
-tabbable regions created by our [Tab components](../tabs/)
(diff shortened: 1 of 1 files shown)


malsf21:
Hm, it's failing a test off of `<OverlayTrigger>` - "Should pass transition callbacks to Transition". I don't think that's from a change that I implemented though.

bpas247:
> Hm, it's failing a test off of `<OverlayTrigger>` - "Should pass transition callbacks to Transition". I don't think that's from a change that I implemented though.

You're right, it isn't an issue with your changes. Our sometimes fail for no reason 😛 I restarted the CI build and the tests are passing now.

PR #4722, opened 2019-10-22
[FEATURE] added `target` and `rel` attrs in NavLink

shhdharmen:
I am adding [target](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target) and [rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-rel) attributes in [NavLink](https://react-bootstrap.github.io/components/navs/#nav-link-props).


taion:
The `...props` spread here already handles this. Plus the typing for `target` isn't correct here anyway.

PR #4763, opened 2019-10-31
modal slideup from bottom to top addded

jyotishman:


taion:
The point of this repo is to implement Bootstrap functionality in a natively React context. Additional functionality like this doesn't belong here.

Issue #4176, opened 2019-08-05
onClick stopPropagation() ignored on Dropdown, DropdownButton


wheezymustafa:
Hi all, I'm new to React and react-bootstrap, so bear with me.  I'm building a form that is sectioned off by Accordion components.  Within an Accordion, I have a Card with a body, and within that body I have a DropdownButton.  When I select a DropdownButton value, the entire Accordion section collapses.  This can be reproduced on the following simple example:

```
import React from 'react'
import {Accordion, Dropdown, Card} from 'react-bootstrap'

const Test = props => {
    return (
        <Accordion defaultActiveKey="0">
            <Accordion.Toggle as={Card.Header} eventKey="0"><strong>Transaction Details</strong></Accordion.Toggle>
            <Accordion.Collapse eventKey="0">
                <div >
                    <Card body>
                    <Dropdown >
                        <Dropdown.Toggle variant="success" id="dropdown-basic">
                            Dropdown Button
                        </Dropdown.Toggle>

                        <Dropdown.Menu  >
                            <Dropdown.Item href="#/action-1">Action</Dropdown.Item>
                            <Dropdown.Item href="#/action-2">Another action</Dropdown.Item>
                            <Dropdown.Item href="#/action-3">Something else</Dropdown.Item>
                        </Dropdown.Menu>
                    </Dropdown>
                    </Card>
                </div>
            </Accordion.Collapse>
        </Accordion>
    )
}

export default Test
```

I believe this has to do with event bubbling, and that my Accordion component's onClick event is also firing.  However, I've tried to set my `stopPropagation()` method on just about every component, but the event continues to make it up to the Accordion.  Note: I've tried using DropdownButton and Dropdown with explicit inner components (Dropdown.Item, Menu, etc) with no difference in result.

I found a similar post here: https://github.com/react-bootstrap/react-bootstrap/issues/2327.  OP attempted to wrap the DropdownButton with a div with success, however I'm still having the same problem even after wrapping with a div.  Is it something I'm not setting correctly?  Thanks in advance.

jaril5976:
Same Issue, is there any solution?

BenStadnick:
[...]

PR #4772, opened 2019-11-05
Bugfix: Prevent Dropdown from calling SelectContext function of surrounding components (#4176)

BenStadnick:
This pull request is to fix the bug described in issue #4176.

When a DropdownItem is selected, the Dropdown component calls a function provided by SelectableContext. This allows a nested Dropdown to close parent Dropdowns but also causes Dropdown to call functions provided by other components SelectableContext as well, e.g. an Accordion component.

This fix is internal to Dropdown and has no API changes. It includes a unit tests to check that Dropdown does not call functions provided by SelectableContext unless they come from a parent Dropdown. It also includes a unit test to check that selecting a DropdownItem in a nested Dropdown will close the parent Dropdowns as well.


src/Dropdown.js
```diff
@@ -112,10 +112,11 @@ const Dropdown = React.forwardRef((uncontrolledProps, ref) => {
   );
 
   const handleSelect = useEventCallback((key, event) => {
-    if (onSelectCtx) onSelectCtx(key, event);
+    if (onSelectCtx && onSelectCtx.isDropdownSelCtx) onSelectCtx(key, event);
     if (onSelect) onSelect(key, event);
     handleToggle(false, event, 'select');
   });
+  handleSelect.isDropdownSelCtx = true;
```
taion: how does mutating the result of `useEventCallback` do anything to the value of `useContext(SelectableContext)` above?

taion: in other words you're just hard-disabling this functionality, which is wrong

BenStadnick: The idea is to modify the the context it provides to child components via
```JavaScript
<SelectableContext.Provider value={handleSelect}>
```
This way onSelectCtx will only fire if the parent is a Dropdown, closing that parent. I checked and this fix does break NavDropdown, so I will try another approach. I also found that selecting a Nav.Item will cause an Accordion to collapse as well. So perhaps Accordion would be a better place to fix this issue.

I still think a unit test for nested Dropdown behavior is good to have as this behavior is provided by bootstrap via the dropdown-submenu class.

taion: you're mutating the return value of a `useEventCallback` hook. and you shouldn't be mutating used context anyway

taion:
did you see my PR comment above?

the point of this logic was to make `NavDropdown` work correctly. the right fix is to move the consumption of the selectable context to `NavDropdown`

BenStadnick:
I’d like to stop the Dropdown from collapsing the Accordion as in issue #4176. Moving consumption to NavDropdown would accomplish this. However, other components which consume SelectContext, such as AbstractNav, also collapse a parent Accordion. For example, selecting an item in Nav will collapse the accordion in the example code bellow:
```JavaScript
import React from "react";
import { Nav, Card, Accordion } from "react-bootstrap";
const Test = props => {
  return (
    <Accordion>
      <Card>
        <Card.Header>
          <Accordion.Toggle eventKey="0">Toggle 1</Accordion.Toggle>
        </Card.Header>
          <Card.Body>
            <Nav activeKey="1">
              <Nav.Item>
                <Nav.Link eventKey="1">
                  NavLink 1 content
                </Nav.Link>
              </Nav.Item>
              <Nav.Item>
                <Nav.Link eventKey="2">
                  NavLink 2 content
                </Nav.Link>
              </Nav.Item>
            </Nav>
          </Card.Body>
        </Accordion.Collapse>
      </Card>
    </Accordion>
  );
}

export default Test
```
TabPane blocks SelectContext in a similar way for nested Nav components. Excerpt from line 168-177:
```JavaScript
  // We provide an empty the TabContext so `<Nav>`s in `<TabPane>`s don't
  // conflict with the top level one.
  return (
    <TabContext.Provider value={null}>
      <SelectableContext.Provider value={null}>
        {pane}
      </SelectableContext.Provider>
    </TabContext.Provider>
  );
});
```

Moving consumption to NavDropdown will solve the Dropdown case but it would be good to handle other cases as well.

taion:
abstract nav provides its own selectable context, so it wouldn't interact with the parent accordion

BenStadnick:
Yes, AbstractNav provides context to its children but it also grabs the surrounding context, line 49,
```JavaScript
const parentOnSelect = useContext(SelectableContext);
```
When I run the code posted above, selecting an item in the Nav collapses the accordion.

kyletsang:
Fixed in #5201

PR #4820, opened 2019-11-14
Enable to install react-bootstrap from GitHub

hiroto7:
Enables to install react-bootstrap by typing the following command
```
npm install react-bootstrap/react-bootstrap#master
```


taion:
I don’t think this works. We don’t publish from root. Plus you’d need all the dev deps for this to even have a chance of working.

PR #4847, opened 2019-11-20
chore: Update Renovate config

taion:


renovate.json
```diff
@@ -4,8 +4,11 @@
     ":disablePeerDependencies",
     ":semanticCommitTypeAll(chore)",
     "group:allNonMajor",
-    "schedule:daily"
+    "schedule:weekly"
```
taion: probably a better schedule?

renovate.json
```diff
@@ -4,8 +4,11 @@
...
   ],
+  "major": {
```
taion: maybe?

renovate.json
```diff
@@ -4,8 +4,11 @@
...
+    "masterIssueApproval": true,
+  },
   "rangeStrategy": "update-lockfile",
```
taion: we should probably bump the version in package.json for dev deps

renovate.json
jquense: what's this do?

taion: i'm trying to get a master issue like https://github.com/gatsbyjs/gatsby/issues/16840

where you have to tick off individual major deps to upgrade, instead of having renovate create them for you automatically

on second thought, though, i don't think i want this. i'd like to eagerly do major bumps, at least here

PR #4848, opened 2019-11-20
fix: dom-helper api change

jquense:


taion:
ignore comment; i fixed formatting

PR #4857, opened 2019-11-22
Fix type of Overlay's onHide prop

dsgkirkby:
The type is incorrect, and is a regression from v0.x; react-overlays supplies the event to the `onHide` callback.

Reference:
- `onHide` is called from the `useRootClose` hook: https://github.com/react-bootstrap/react-overlays/blob/master/src/Overlay.js#L74
- `useRootClose` passes the event to the `onClose` function: https://github.com/react-bootstrap/react-overlays/blob/master/src/useRootClose.js#L59


jquense:
thanks!

PR #4861, opened 2019-11-25
chore: Clean up dependencies

taion:


taion:
dropping some unused deps and moving www deps to dev

PR #4866, opened 2019-11-28
fix(Modal): TypeScript manager typings

voidale:
The docs have a manager property but it's missing from the types

### DIFF
--- a/types/components/Modal.d.ts
+++ b/types/components/Modal.d.ts
@@ -27,7 +27,7 @@ export interface ModalProps extends TransitionCallbacks {
   container?: any;
   scrollable?: boolean;
   onEscapeKeyDown?: () => void;
-  manager: any;
+  manager?: any;
 }
 
 declare class Modal<
--- a/types/components/Modal.d.ts
+++ b/types/components/Modal.d.ts
@@ -27,6 +27,7 @@ export interface ModalProps extends TransitionCallbacks {
   container?: any;
   scrollable?: boolean;
   onEscapeKeyDown?: () => void;
+  manager: any;
 }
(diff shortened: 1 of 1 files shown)


taion:
Note the CI failure. Also, I think we should at least have a nominal type there for the manager.

kosmiq:
What should the manager accept? The docs says object, so it would be enough with a type that is object?

taion:
It should be an instance of `ModalManager` from React Overlays: https://github.com/react-bootstrap/react-bootstrap/blob/3f981a42815603a111a630c49365ab9bba81dd6f/src/BootstrapModalManager.js
https://github.com/react-bootstrap/react-overlays/blob/c29dc76f66c7eb74f5123ca9d5f531918ea7134a/src/ModalManager.js

Unfortunately there are no typings for that, so... not sure what we can do

mxschmitt:
> It should be an instance of `ModalManager` from React Overlays: [/src/BootstrapModalManager.js@`3f981a4`](https://github.com/react-bootstrap/react-bootstrap/blob/3f981a42815603a111a630c49365ab9bba81dd6f/src/BootstrapModalManager.js)
> [react-bootstrap/react-overlays:src/ModalManager.js@`c29dc76`](https://github.com/react-bootstrap/react-overlays/blob/c29dc76f66c7eb74f5123ca9d5f531918ea7134a/src/ModalManager.js)
> 
> Unfortunately there are no typings for that, so... not sure what we can do

@taion maybe for now type any?

taion:
@mxschmitt probably, yeah! unless something like `{}` works...

PR #4869, opened 2019-12-01
fix tabs prop types

ericvergnaud:


bpas247:
@ericvergnaud what is this actually fixing? doesn't really seem that much different to assign `defaultActiveKey` to `null` initially rather than `undefined` IMO

taion:
ah this ended up being a dupe of https://github.com/react-bootstrap/react-bootstrap/pull/4875 i think – @bpas247 could you review that?

PR #4875, opened 2019-12-02
fix(Tabs): added defaultActiveKey PropType

ericvergnaud:


src/Tabs.js
```diff
@@ -81,7 +82,7 @@ const defaultProps = {
 };
 
 function getDefaultActiveKey(children) {
-  let defaultActiveKey;
+  let defaultActiveKey = null;
```
bpas247: What issue is this fixing?

src/Tabs.js
```diff
@@ -20,6 +20,7 @@ const propTypes = {
    * @controllable onSelect
    */
   activeKey: PropTypes.any,
+  defaultActiveKey: PropTypes.any,
```
bpas247: Needs prop type comments for documentation purposes.

taion: is this actually a prop, even?

bpas247: it's returned from `uncontrollable`, no?

taion: ah, consumed by uncontrollable, even. i got a little confused given the `getDefaultActiveKey` function, which is doing something somewhat unrelated

taion: can copy the comment from accordion

ericvergnaud: There is no docs provided, see https://5c507d49471426000887a6a7--react-bootstrap.netlify.com/components/tabs/, although the examples use it, and my experience shows it is necessary...

src/Tabs.js
ericvergnaud: To my knowledge, the default value for a variable declared by `let's unspecified. As a result, the line 87 `if (defaultActiveKey == null)`has unpredictable behavior. This eliminates the uncertainty.

src/Tabs.js
bpas247: > There is no docs provided
okay, so add the comments for it then.

src/Tabs.js
ericvergnaud: and it's very possible that is why it's mandatory to provide a defaultActiveKey property in order to have a tab active by default

taion: the default value for an uninitialized variable in JS is perfectly well-defined – it's always `undefined`.

ericvergnaud: then it makes it clear that this fix is necessary for line 87 to match the condition

src/Tabs.js
ericvergnaud: done

src/Tabs.js
bpas247: but it’s already clear, and it’s not necessary to have this line for it to be clear...

ericvergnaud: I wrote "this fix is necessary for line 87 to match the condition"
If `defaultActiveKey` is undefined, then `if (defaultActiveKey == null)`never matches, and the method always returns `undefined`.
This is probably the cause of the bug I've observed, where no tab is active unless you specify the defaultActiveKey _property_.

taion: null == undefined in JS. it doesn’t pass ===, but the use of == is intentional here

ericvergnaud: then I guess my original comment 'clarify intent' is a good enough reason for this harmless change, coz all this is only clear to a small subset of JS developers...

bpas247: but the main problem is that it’s not adding any *value* here. There’s not really a point in adding this change if there’s nothing significant that is being added here.

> coz all this is only clear to a small subset of JS developers

I believe that’s making an unfair assumption on the people who will read this portion of the code. At most, we could add a comment here.

ericvergnaud: just proposed an entirely new version which is clear, concise and should run faster (iteration stops of first candidate key)

src/Tabs.js
```diff
@@ -81,14 +83,8 @@ const defaultProps = {
...
 function getDefaultActiveKey(children) {
-  let defaultActiveKey;
-  forEach(children, child => {
-    if (defaultActiveKey == null) {
-      defaultActiveKey = child.props.eventKey;
-    }
-  });
-
-  return defaultActiveKey;
+  const first = children.find(child => child.props.eventKey);
```
taion: you can't do this. `children` is a `React.ReactNode`, not e.g. an array

src/Tabs.js
ericvergnaud: > but the main problem is that it’s not adding any _value_ here. There’s not really a point in adding this change if there’s nothing significant that is being added here.
> 
> > coz all this is only clear to a small subset of JS developers
> 
> I believe that’s making an unfair assumption on the people who will read this portion of the code. At most, we could add a comment here.

Sorry to hear that we have such different perceptions of _clarity_ and _value_...

ericvergnaud: > the default value for an uninitialized variable in JS is perfectly well-defined – it's always `undefined`.

Probably true, but can you point to an mdn spec that says this is true for `let`? I couldn't find one, hence my say that it's unspecified.

taion: it's right here: https://www.ecma-international.org/ecma-262/6.0/#sec-let-and-const-declarations

> If a LexicalBinding in a let declaration does not have an Initializer the variable is assigned the value undefined when the LexicalBinding is evaluated.

ericvergnaud:
Thanks

Envoyé de mon iPhone

ericvergnaud:
Hi, any action required on my side?

src/Tabs.js
```diff
@@ -81,13 +83,12 @@ const defaultProps = {
 };
 
 function getDefaultActiveKey(children) {
-  let defaultActiveKey;
+  let defaultActiveKey = null;
   forEach(children, child => {
-    if (defaultActiveKey == null) {
+    if (defaultActiveKey === null) {
```
taion: this change is _not_ correct. if the `eventKey` on a child is `undefined` (because it's not defined), then we end up calculating a `defaultActiveKey` here of `undefined`, because this check will now fail.

please undo the changes to this function. it's good to have documentation for the `defaultActiveKey` prop, but this change is wrong

src/Tabs.js
```diff
@@ -20,6 +20,8 @@ const propTypes = {
    * @controllable onSelect
    */
   activeKey: PropTypes.any,
+  /** The default active key that is expanded on start */
```
taion:
```suggestion
  /** The default active key that is selected on start */
```

as this is not an accordion

src/Tabs.js
ericvergnaud: not sure how it can be correct to return an undefined key but agree this would be a breaking change

ericvergnaud:
Hi there,
this small change has been pending a 2nd review for 3 months...

kosmiq:
So should this be proptypes.any or should it rather be "string or number" as allowed proptypes?

I see it follows activeKey so that might be a discussion to have elsewhere.

Other than that it looks good.

taion:
the test is failing because we've made the github token a secret, which makes it unavailable for third-party builds ):

so all PRs coming in from forks fail at the codecov step

ericvergnaud:
> Looks good to me.
> 
> A question however for @taion @jquense or whomever can answer. Should activeKey and defaultActiveKey have PropTypes.any? Shouldn't it be "PropTypes.string | PropTypes.Number" or am I missing something here?
> 

It needs to be consistent with TabsPane eventKey which is PropTypes.any.

PR #4893, opened 2019-12-13
Add size prop for FormLabel component - Fixes #4886

### LINKED ISSUE #4886: Form.Label does not have size prop
## Describe the bug
I would like to set the size for Form.Control and its corresponding Form.Label components.
While its working as expected for Form.Control prop, Form.Label component does not have a similar size prop. 

## Reproducible Example

Minimal example by using [CodeSandbox](https://codesandbox.io/s/react-bootstrap-3kswo).

## Expected behavior

The size of the label should be applied accordingly.

- Operating System: [Ubuntu]
- Browser, Version [Chrome 78]
- React-Bootstrap Version [1.0.0-beta.16]

raghavi92:


src/FormLabel.js
```diff
@@ -24,6 +24,13 @@ const propTypes = {
    */
   column: PropTypes.bool,
 
+  /**
+   * Label size variants. Only applicable if column is true
+   *
+   * @type {('sm'|'lg')}
+   */
+  size: PropTypes.string,
```
taion: This should be merged with the `column` prop.

raghavi92:
Can you review this @taion ?

src/FormLabel.js
taion: as in `column` should take `true` or `"sm"` or `"lg"`

raghavi92: I see. But I thought it would be more consistent with <FormControl> this way?

taion: the issue is that the prop doesn't do anything if you're not specifying column...

www/src/examples/Form/InputSizes.js
```diff
@@ -1,7 +1,14 @@
-<>
+<Form.Group>
+  <Form.Label size="lg" col>
```
taion: can you fix this to show the actual API, and merge this in with the horizontal form groups section, or else make a new section paralleling the upstream TWBS docs?

src/FormLabel.js
```diff
@@ -56,7 +56,8 @@ const FormLabel = React.forwardRef(
       className,
       bsPrefix,
       srOnly && 'sr-only',
-      column && 'col-form-label',
+      typeof column === 'boolean' && 'col-form-label',
```
taion: this would set `col-form-label` even if we had `column={false}`. see e.g. how we calculate and use `expandClass` in https://github.com/react-bootstrap/react-bootstrap/blob/1bb3f41d2a9650b06402c4d3dcce50b9c19a860f/src/Navbar.js#L164-L165

kosmiq:
Checked into this. Looking on regular bootstrap, looking at this code and what @raghavi92 wrote in his issue I have a few thoughts.
- The size prop would be needed as well to cover sizing of the label without making it a column. This would cover the case where the user wants to set a label size, but not make it a column: https://getbootstrap.com/docs/4.4/components/forms/#sizing
- the column prop with either true, sm or large (as in this PR) covers the use case for this functionality: https://getbootstrap.com/docs/4.4/components/forms/#horizontal-form

They are separate layouts within the same component in bootstrap (and uses different classes as well).

So to cover the OP original request, as I have read it and after looking at Bootstrap 4.4, the Form.Label should have the size prop for specifying size when it is used in a vertial layout and use the column prop for sizing when in a horizontal layout (as the column prop currently does).

jquense:
thanks!

raghavi92:
Thank you!

Issue #4887, opened 2019-12-12
Support for row-cols in <Row>


kosmiq:
## Is your feature request related to a problem? Please describe
Bootstrap 4.4 added .row-cols-x classes to the row class (https://getbootstrap.com/docs/4.4/layout/grid/#row-columns)

This would be an awesome addition to react-bootstrap.

## Describe the solution you'd like
I'd imagine something like
`<Row xs={[number]}></Row>`

Thus having a similar usage as <Col> does when defining widths of columns.

The docs would of course need to be updated to reflect this.

## Describe alternatives you've considered
I did briefly consider that it could be set as
`<Row cols={[number]}></Row>`
but then we would need
`<Row cols-xs={[number]}></Row>`
or similar to define breakpoints and that feels less clean.

## Additional context
Bootstrap docs for row-cols: https://getbootstrap.com/docs/4.4/layout/grid/#row-columns

taion:
We'd be happy to take a PR for this. I think e.g. a `colsXs` prop would be pretty reasonable, actually. It'd match what we have for `<Col>`.

bcobra:
When will this be released?

kyletsang:
This is available in 1.0.1:  https://react-bootstrap.github.io/layout/grid/#row-layout-col-sizing

This can be closed

PR #4900, opened 2019-12-16
#4887 Add support for setting columns in <Row>

kosmiq:
This PR extends <Row> to accept sizing prop similar to how <Col> does it.

I created an issue for this, #4887 , and here is my attempt at an PR for it.

- Extends Row.js with logic from Col.js
- Updated RowSpec.js with what it should handle with the above changes
- Updated docs with some examples and text that (hopefully) explains how it works

There are some gotcha's I found while creating this:
- The sizing prop into `<Row>` doesn't set the column widths, rather it sets the number of columns that should fit next to each other
- Setting `<Row md={4}>` will override `<Col xs={6}>` on medium devices (Row-sizing props takes precedence based on breakpoint as it does in Bootstrap)
- Setting `<Row md={4}>` will NOT override `<Col md={6}>` on medium devices

I'm unsure if this is clear enough in the docs.

I am sure that you can propose better names for a few of the files and headings I added for the docs.

### DIFF
--- a/src/Row.js
+++ b/src/Row.js
@@ -66,14 +66,19 @@ const defaultProps = {
 };
 
 const Row = React.forwardRef(
-  // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
   (
-    { bsPrefix, className, noGutters, as: Component = 'div', ...props },
+    {
+      bsPrefix,
+      className,
+      noGutters,
+      // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
+      as: Component = 'div',
+      ...props
+    },
     ref,
   ) => {
     const decoratedBsPrefix = useBootstrapPrefix(bsPrefix, 'row');
     const sizePrefix = `${decoratedBsPrefix}-cols`;
-    const rowCols = [];
     const classes = [];
 
     DEVICE_SIZES.forEach(brkPoint => {
@@ -89,11 +94,11 @@ const Row = React.forwardRef(
 
       let infix = brkPoint !== 'xs' ? `-${brkPoint}` : '';
 
-      if (cols != null) rowCols.push(`${sizePrefix}${infix}-${cols}`);
+      if (cols != null) classes.push(`${sizePrefix}${infix}-${cols}`);
     });
 
-    if (!rowCols.length) {
-      rowCols.push(decoratedBsPrefix); // plain 'row'
+    if (!classes.length) {
+      classes.push(decoratedBsPrefix); // plain 'row'
     }
 
     return (
@@ -104,7 +109,6 @@ const Row = React.forwardRef(
           className,
           decoratedBsPrefix,
           noGutters && 'no-gutters',
-          ...rowCols,
           ...classes,
         )}
       />
--- a/src/Row.js
+++ b/src/Row.js
@@ -5,6 +5,16 @@ import React from 'react';
 
 import { useBootstrapPrefix } from './ThemeProvider';
 
+const DEVICE_SIZES = ['xl', 'lg', 'md', 'sm', 'xs'];
+const colSize = PropTypes.oneOfType([PropTypes.number, PropTypes.string]);
+
+const column = PropTypes.oneOfType([
+  colSize,
+  PropTypes.shape({
+    size: colSize,
+  }),
+]);
+
 const propTypes = {
   /**
    * @default 'row'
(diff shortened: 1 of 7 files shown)


src/Row.js
```diff
@@ -5,6 +5,16 @@ import React from 'react';
 
 import { useBootstrapPrefix } from './ThemeProvider';
 
+const DEVICE_SIZES = ['xl', 'lg', 'md', 'sm', 'xs'];
+const colSize = PropTypes.oneOfType([PropTypes.number, PropTypes.string]);
+
+const column = PropTypes.oneOfType([
+  colSize,
+  PropTypes.shape({
+    size: colSize,
```
taion: this isn't a size thing, i don't think?

src/Row.js
```diff
@@ -14,36 +24,101 @@ const propTypes = {
...
+    const spans = [];
+    const classes = [];
+
+    DEVICE_SIZES.forEach(brkPoint => {
+      let propValue = props[brkPoint];
+      delete props[brkPoint];
+
+      let span;
+      if (propValue != null && typeof propValue === 'object') {
+        ({ span = true } = propValue);
```
taion: but above you have `size`?

taion:
if we want to support an object thing, the key should probably be `cols`

kosmiq:
> if we want to support an object thing, the key should probably be `cols`

Agreed. I have updated the PR.

The only reason I kept the object is because the <Col> accepts it. It could be removed since we only accept one value. Let me know what you think.

src/Row.js
```diff
@@ -14,36 +24,101 @@ const propTypes = {
...
-  );
-});
+const Row = React.forwardRef(
+  // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
+  (
+    { bsPrefix, className, noGutters, as: Component = 'div', ...props },
+    ref,
+  ) => {
+    const decoratedBsPrefix = useBootstrapPrefix(bsPrefix, 'row');
+    const sizePrefix = useBootstrapPrefix(
```
taion: there's no point to doing `useBootstrapPrefix` here. you'll always end up with `${decoratedBsPrefix}-cols`.

src/Row.js
```diff
@@ -14,36 +24,101 @@ const propTypes = {
...
+    const rowCols = [];
+    const classes = [];
+
+    DEVICE_SIZES.forEach(brkPoint => {
+      let propValue = props[brkPoint];
+      delete props[brkPoint];
+
+      let cols;
+      if (propValue != null && typeof propValue === 'object') {
+        ({ cols = true } = propValue);
```
taion: why is the default here `true`? that's not a valid value here

src/Row.js
```diff
@@ -14,36 +24,101 @@ const propTypes = {
...
+      } else {
+        cols = propValue;
+      }
+
+      let infix = brkPoint !== 'xs' ? `-${brkPoint}` : '';
+
+      if (cols != null)
+        rowCols.push(
+          cols === true
```
taion: as above; `cols` must be a number or string, no?

src/Row.js
```diff
@@ -5,6 +5,16 @@ import React from 'react';
 
 import { useBootstrapPrefix } from './ThemeProvider';
 
+const DEVICE_SIZES = ['xl', 'lg', 'md', 'sm', 'xs'];
+const rowColWidth = PropTypes.oneOfType([PropTypes.number, PropTypes.string]);
+
+const rowColumns = PropTypes.oneOfType([
+  rowColWidth,
+  PropTypes.shape({
+    size: rowColWidth,
```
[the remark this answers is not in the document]
kosmiq: Will fix

src/Row.js
kosmiq: True. I'll clean it up.

src/Row.js
kosmiq: Leftover from copying some code from Cols.js and not cleaning up properly. Will fix.

kosmiq:
@taion all fixed I think.

taion:
oops forgot to send in my PR feedback yesterday

kosmiq:
@taion I have made new commits for this with cleaned code, comment moved to the right location and updated the docs a bit.

src/Row.js
```diff
@@ -14,36 +24,93 @@ const propTypes = {
...
+      ...props
+    },
+    ref,
+  ) => {
+    const decoratedBsPrefix = useBootstrapPrefix(bsPrefix, 'row');
+    const sizePrefix = `${decoratedBsPrefix}-cols`;
+    const classes = [];
+
+    DEVICE_SIZES.forEach(brkPoint => {
+      let propValue = props[brkPoint];
```
taion:
```suggestion
      const propValue = props[brkPoint];
```

bcobra:
Would love to see this soon

types/components/Row.d.ts
```diff
@@ -2,8 +2,29 @@ import * as React from 'react';
...
+type RowColSpec = NumberAttr | { cols?: NumberAttr };
+
 export interface RowProps {
   noGutters?: boolean;
+  xs?: RowColSpec;
+  sm?: RowColSpec;
+  md?: RowColSpec;
+  lg?: RowColSpec;
+  xl?: RowColSpec;
 }
```
taion:
```suggestion
type RowColWidth =
  | number
  | '1'
  | '2'
  | '3'
  | '4'
  | '5'
  | '6'
  | '7'
  | '8'
  | '9'
  | '10'
  | '11'
  | '12';
type RowColumns = RowColWidth | { cols?: RowColWidth };

export interface RowProps {
  noGutters?: boolean;
  xs?: RowColumns;
  sm?: RowColumns;
  md?: RowColumns;
  lg?: RowColumns;
  xl?: RowColumns;
}
```

Issue #4884, opened 2019-12-10
Support for Responsive Containers (Bootstrap 4.4)


davidjb:
## Is your feature request related to a problem? Please describe

Bootstrap 4.4 added responsive containers (https://getbootstrap.com/docs/4.4/layout/overview/#responsive),  with the syntax `.container-{breakpoint}`.

## Describe the solution you'd like

Add support for setting a breakpoint prop on `Container` components, which then sets the resultant `className`.

## Describe alternatives you've considered

A workaround is to use `<div className="container-lg">` and manually create the container.

Using `<Container className="container-lg">` isn't suitable as a solution given how the [classes operate](https://getbootstrap.com/docs/4.4/layout/overview/#containers) (eg it has no effect)

taion:
We'd be happy to take a PR for this!

kosmiq:
@taion this can be closed since the PR has been merged.

PR #4902, opened 2019-12-17
#4884 add responsive containers

kosmiq:
An attempt at extending `<Container>` with breakpoints as in #4884

**I have not updated docs yet since I want to make sure we find consensus about how this prop should work for the Container component.** I will add docs once we have decided if this is a good idea or not.

- Adds "breakpoint" prop to `<Container>`
- Permits only "sm", "md", "lg" and "xl" (same as Bootstrap classes)
- if fluid is set it takes precedence since fluid and container-[breakpoint] should not be combined
- updated test for `<Container>` to include breakpoint

I first considered setting sm, md, lg and xl as booleans, but realised that would allow for all kinds of fails since then you could set more than one as true at the same time (not good).

And since we already have fluid I left that prop alone and extended on it.

**Regarding docs**
I'm a bit unsure where to add them. Comparing to the regular Bootstrap docs we do not have their Overview section. Should we add it and place the Container documentation in it like they do?

### DIFF
--- a/src/Container.js
+++ b/src/Container.js
@@ -4,8 +4,8 @@ import PropTypes from 'prop-types';
 
 import { useBootstrapPrefix } from './ThemeProvider';
 
-const brkPoints = PropTypes.oneOfType([
-  PropTypes.oneOf(['sm', 'md', 'lg', 'xl']),
+const sizes = PropTypes.oneOfType([
+  PropTypes.oneOf(['sm', 'md', 'lg', 'xl', 'fluid']),
 ]);
 
 const propTypes = {
@@ -26,7 +26,7 @@ const propTypes = {
    * You can set responsive container width.
    * @type {("sm"|"md"|"lg"|"xl")}
    */
-  breakpoint: brkPoints,
+  size: sizes,
 };
 
 const defaultProps = {
@@ -36,11 +36,11 @@ const defaultProps = {
 const Container = React.forwardRef(
   // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
   (
-    { bsPrefix, fluid, breakpoint, as: Component = 'div', className, ...props },
+    { bsPrefix, fluid, size, as: Component = 'div', className, ...props },
     ref,
   ) => {
     const prefix = useBootstrapPrefix(bsPrefix, 'container');
-    let suffix = breakpoint != null ? `-${breakpoint}` : '';
+    let suffix = size != null ? `-${size}` : '';
     return (
       <Component
         ref={ref}
--- a/src/Container.js
+++ b/src/Container.js
@@ -24,6 +24,7 @@ const propTypes = {
   as: PropTypes.elementType,
   /**
    * You can set responsive container width.
+   * @type {("sm"|"md"|"lg"|"xl")}
    */
   breakpoint: brkPoints,
 };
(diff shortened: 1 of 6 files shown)


kosmiq:
Honestly I think the best approach would be to remove fluid and replace it with "size" or similar. And make it accept sm/md/lg/xl/fluid as a string.

But that would be a breaking change. Perhaps replace breakpoint with size and add fluid as an option to it. And keep fluid around for a while longer, but mark it deprecated and remove in a future version?

And inputting strings into the fluid prop feels weird imo. Especially since they are separate things.

taion:
Does `size` do anything when `fluid` isn't set?

kosmiq:
> Does `size` do anything when `fluid` isn't set?

I imagine it working like this:

- `<Container size="fluid">` will output <div class="container-fluid"> (thus being the same as the current `<Container fluid>`
- `Container size="sm">` will output <div class="container-sm">

If we look at the Bootstrap documentation container-fluid and container-sm are classes that both set the container width in different ways. Quoting from their docs:

> - .container, which sets a max-width at each responsive breakpoint
> - .container-fluid, which is width: 100% at all breakpoints
> - .container-{breakpoint}, which is width: 100% until the specified breakpoint

kosmiq:
I updated the PR with my thoughts above, hopefully making it a bit clearer what we want to achieve. Unsure though what the prop should be named, I think size is better than breakpoint. I'm open to any suggestions.

This commit adds fluid as a valid value to the size prop.

jquense:
I don't ahve a strong opinion here but it does seem like the responsiveness is related to `fluid` or am i reading it wrong? i.e. IT seems like `container-md` is `container-fluid` until the `md` breakpoint, in which case it does seem like it makes sense to have just the `fluid` prop take a string and boolean, WDT?

kosmiq:
@jquense and @taion since you both seem to agree on how it should be I have updated the PR.
I re-read the docs over at Bootstrap and took a look at my code again. I weighed in your comments and I see no issues with going your way and avoiding adding new props/renaming any props.

- Size prop has been removed
- Breakpoints (sm, md, lg, xl) have been added as possible strings to the fluid prop (along with true/false as before)
- Docs have been updated
- Tests have been updated

src/Container.js
```diff
@@ -25,14 +31,27 @@ const defaultProps = {
...
+      fluid,
+      // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
+      as: Component = 'div',
+      className,
+      ...props
+    },
+    ref,
+  ) => {
     const prefix = useBootstrapPrefix(bsPrefix, 'container');
+    let suffix = fluid === false ? '' : `-${fluid}`;
```
taion:
```suggestion
    const suffix = typeof fluid === 'string' ? `-${fluid}` : '-fluid';
```

src/Container.js
```diff
@@ -25,14 +31,27 @@ const defaultProps = {
...
     return (
       <Component
         ref={ref}
         {...props}
-        className={classNames(className, fluid ? `${prefix}-fluid` : prefix)}
+        className={classNames(
+          className,
+          fluid === true ? `${prefix}-fluid` : `${prefix}${suffix}`,
```
taion:
```suggestion
          fluid ? `${prefix}${suffix}` : prefix,
```

PR #4903, opened 2019-12-18
Fix issue with OverlayTrigger, hover and disabled child element

fredriklengstrand:
This pull request fixes an issue with the fact that `onMouseOut` does not trigger when the mouse leaves a disabled element (In Chrome and Safari). This causes the `show` state to still be `true` when leaving the element and the overlay will re-appear when the button is enabled again even if the mouse no longer is above the element.

The fix is to always set the `show` state to `false` if the childs `disabled` property has changed to `true` and the `show` state is still `true`. If you prefer to solve this issue in another way just let me know but it would be nice if this bug is fixed in some way. :)

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -110,6 +110,14 @@ class OverlayTrigger extends React.Component {
     };
   }
 
+  componentDidUpdate(prevProps) {
+    const disabled = React.Children.only(this.props.children).props.disabled;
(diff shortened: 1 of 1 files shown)


taion:
This doesn't seem safe to me... what if e.g. the disabled element is wrapped in other components or something? I'd think we'd at least need to inspect the DOM node here

fredriklengstrand:
I see your point, but I'm not sure how to best implement that safety and since the component has been refactored in the master branch I will close this pull request.

PR #4914, opened 2019-12-28
docs: Fix component spacing in examples

taion:


www/src/examples/Dropdown/Dropup.js
```diff
@@ -1,26 +0,0 @@
-<>
```
taion: This was unused. It seemed pointless because we already have the "directions" example.

www/src/components/ReactPlayground.js
```diff
@@ -67,15 +67,6 @@ const StyledExample = styled('div')`
   }
 
   :global {
-    .badge + .badge,
```
taion: This ends up being more trouble than it's worth. People get confused at this being just a quirk of our docs page.

PR #4917, opened 2019-12-31
fix(Carousel): support for ref attribute

muzakparov:
Related to #4194


src/Carousel.js
```diff
@@ -126,315 +134,350 @@ const defaultProps = {
   touch: true,
 };
 
-class Carousel extends React.Component {
-  state = {
+const Carousel = React.forwardRef((uncontrolledProps, ref) => {
+  let {
+    activeIndex: activeIndexProp,
+    as: Component = 'div',
```
taion: copy over the comment from the other places where we have this

src/Carousel.js
```diff
@@ -126,315 +134,350 @@ const defaultProps = {
...
-    touchStartX: 0,
-  };
-
-  isUnmounted = false;
+  });
+  let activeIndex = useRef(undefined);
+  let previousActiveIndex = useRef(undefined);
+  let direction = useRef('');
+
+  const isUnmounted = useRef(true);
```
taion: use `@restart/hooks/useMounted`

src/Carousel.js
```diff
@@ -126,315 +134,350 @@ const defaultProps = {
...
-        this.handleNext(event);
+        handleNext(event);
         break;
       default:
         break;
     }
   };
 
-  handleNextWhenVisible = () => {
+  useLayoutEffect(() => {
```
taion: we generally use `useIsomorphicEffect` instead of `useLayoutEffect` for SSR safety

src/Carousel.js
muzakparov: thank you for the prompt review! :) 
unfortunately I wasn't able to use `useMounted` because it works in a bit different way, I tried including it, but it caused a lot of tests to fail. 
I checked the `useMounted` source code and it seems to be setting `mounted = true` initially. I am not quite if it was a bug or intentional.

taion: eh? why shouldn't it start out as "true"? isn't the component initially mounted?

taion: in the old implementation, `isUnmounted` is set to `false` at class construction time. this seems to be different now?

muzakparov: yes, it is different now. Basically I needed a way to skip running `useIsomorphicEffect` on the first render and I reused `isUnmounted` with initial value set to `true` and then setting it to `false` inside `useEffect(()=>{ ... } , [ ])`. So `isUnmounted=false` only after first render. Perhaps `didMount` would be better name for it, but I didn't want to introduce a new variable. Also worth noting that I won't be able to use `useMounted` even if it is renamed to `didMount`.
I can't see how to make it work with the current `useMounted`, sorry. Any suggestions?

taion:
Shoot!

My bad here – I just realized that this _shouldn't_ just forward the ref. `<Carousel>` is one of the few components that actually has internal logic. As such, this is one of the places where it's appropriate to use `useImperativeHandle` to expose e.g. `next`, `prev`, and other functionality to the parent ref holder.

My bad – should have realized earlier.

muzakparov:
No worries. Just to clarify, do you mean exposing refs for `controls` such as `next` and `prev` and for `indicators`? I have some ideas on how to implement it and it shouldn't be hard(for `controls`) to do if that is the case. It might be a bit tricky for the `indicators` though because they are dynamic.
Do you want me to submit another commit to this PR?

taion:
I think we want to expose:

- `prev()`
- `next()`
- `pause()`
- `cycle()`
- err... either `to()` or `select()`... not sure which

muzakparov:
added `useImperativeHandle`. Now it is possible to get handle of the _controls_ and _indicators_ with _**refs**_. Usage would look something like:
```
function ControlledCarousel() {
  ...
  const carouselRef = useRef(null);
  ...
  useEffect(() => {
    console.log('access carouselNode', carouselRef.current.carouselNode);
    console.log('access previous control', carouselRef.current.prevControl);
    console.log('access next control', carouselRef.current.nextControl);
    console.log('access all indicators list', carouselRef.current.indicatorsList);
    console.log('access first child of indicators list', carouselRef.current.indicatorsList.childNodes[0]);
   });
  ...
  return (
    <Carousel activeIndex={index} onSelect={handleSelect} ref={carouselRef}>
    ...
  )
```
I tried this on `ControlledCarousel` and seems to work. 
What do you think? :)

muzakparov:
@jquense  thank you for taking time to review! I might be wrong, but I guess `ref` is just an escape hatch to access dom elements imperatively from what I undersood reading  react documentation. As for the benefit I am not sure if people will use it hehe. Rewriting to hooks definitely made animation logic more linear imho

taion:
sorry, been really busy – will get to this soon

muzakparov:
no worries :)

taion:
note to self – ref #5003 and check that this also fixes that

taion:
sorry for the long delay here. i revised the implementation to make better use of the hook functionality. take a look!

taion:
this should also fix

https://github.com/react-bootstrap/react-bootstrap/issues/5003
https://github.com/react-bootstrap/react-bootstrap/issues/4956
https://github.com/react-bootstrap/react-bootstrap/issues/3654

src/Carousel.js
```diff
@@ -100,441 +124,433 @@ const propTypes = {
...
 
-  componentDidMount() {
-    this.cycle();
-  }
+  return (
+    elementStyle.display !== 'none' &&
+    elementStyle.visibility !== 'hidden' &&
+    getComputedStyle(element.parentNode).display !== 'none'
+  );
+}
```
taion: @jquense is there a better way to do this with dom-helpers?

taion:
@muzakparov can you take a look here as well?

types/components/Carousel.d.ts
```diff
@@ -7,22 +7,24 @@ import { BsPrefixComponent } from './helpers';
 
 export interface CarouselProps {
   bsPrefix?: string;
-  slide?: boolean;
+  animate?: boolean;
```
muzakparov: I guess it should be renamed to `animate` in component code too

src/Carousel.js
```diff
@@ -100,441 +124,430 @@ const propTypes = {
...
+    children,
+    ...props
+  } = useUncontrolled(uncontrolledProps, {
+    activeIndex: 'onSelect',
+  });
+
+  const prefix = useBootstrapPrefix(bsPrefix, 'carousel');
+
+  const nextDirectionRef = useRef(null);
+  const [direction, setDirection] = useState(null);
```
muzakparov: `orderClassName` has an initial value of  `carousel-item-null`, probably making it `next` makes sense here:
```
const [direction, setDirection] = useState('next');
```

src/Carousel.js
```diff
@@ -100,441 +124,430 @@ const propTypes = {
...
+  );
 
-  to(index, event) {
-    const { children } = this.props;
-    if (index < 0 || index > countChildren(children) - 1) {
-      return;
-    }
+  const handleMouseOut = useCallback(
+    event => {
+      setPausedOnHover(false);
```
muzakparov: similar to `handleMouseOver` in case `pause` is `false`, this will avoid extra rerender
```
if (pause === 'hover') {
  setPausedOnHover(false);
}
```
but I guess you did it for a reason and there might be an edge case that I am not seeing :)

taion: `setState` for a value that's the same value (`===`) won't cause another render when it's done in a callback, I think.

It _will_ cause an active render function to bail out, but that's not at issue with callbacks.

types/components/Carousel.d.ts
taion: oh, oops. cruft on my part. the upstream has the class as just "slide", so i figured we'd keep the same naming here.

taion:
```suggestion
  slide?: boolean;
```

src/Carousel.js
taion:
```suggestion
  const [direction, setDirection] = useState('next');
```

types/components/Carousel.d.ts
taion:
```suggestion
  slide?: boolean;
```

taion:
@jquense @bpas247 @kosmiq @mxschmitt any interest in taking a look?

PR #4922, opened 2020-01-06
increase test coverage with minor refactor

ShunyaWatanabe:
- I wrote test cases to increase coverage for Toast.js and TabContainer.js
- I removed a fallback for id in Tabcontainer.js since there was no way (I could think of) that can happen. Please let me know if I should revert the refactoring.

related #3889

### DIFF
--- a/test/TabContainerSpec.js
+++ b/test/TabContainerSpec.js
@@ -41,6 +41,52 @@ describe('<TabContainer>', () => {
     onSelect.should.have.been.calledOnce;
   });
 
+  it('should let generateChildId function create id', () => {
+    const generateChildIdSpy = sinon.spy(() => 'test-id');
+
+    let instance = mount(
+      <TabContainer generateChildId={generateChildIdSpy}>
+        <div>
+          <Nav>
+            <Nav.Item>
+              <Nav.Link eventKey="1">One</Nav.Link>
+            </Nav.Item>
+          </Nav>
+          <TabContent>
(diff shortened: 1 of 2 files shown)


src/TabContainer.js
```diff
@@ -94,9 +94,7 @@ const TabContainer = props => {
   } = useUncontrolled(props, { activeKey: 'onSelect' });
 
   const generateChildId = useMemo(
-    () =>
-      generateCustomChildId ||
-      ((key, type) => (id ? `${id}-${type}-${key}` : null)),
+    () => generateCustomChildId || ((key, type) => `${id}-${type}-${key}`),
```
taion: why remove the null fallback?

ShunyaWatanabe: Since at least one of id and generateChildId must be defined, id cannot be a falsy value if generateCustomChildId is falsy, as validated in propTypes. So the null fallback is unnecessary in my opinion.

taion: they're semantically required, yes, but the issue is that there are no hard checks preventing users from omitting one of them.

what we're looking for, then, is the least broken behavior possible in the event that users specify neither

ShunyaWatanabe: I can revert the refactor commit

taion: yes please; thanks!

ShunyaWatanabe: Done!

PR #4926, opened 2020-01-10
docs: fix element type used for Breadcrumb

davidjb:
A very minor change to the docs for Breadcrumb -- it uses an `<ol>` vs a `<ul>`.


jquense:
thanks!

Issue #4913, opened 2019-12-28
Actually support defaultChecked in <ToggleButton>


taion:
https://react-bootstrap.github.io/components/buttons/#checkbox--radio is broken.

billkabanga:
I would like to take up this issue and raise a PR for it. @taion

taion:
We'd appreciate it. Thanks!

billkabanga:
@taion I've finished working on this but I am unable to push my branch as permission is denied. Your help on this would be much appreciated. Thank You.

taion:
Fork the repo and send a PR from your fork

PR #4930, opened 2020-01-13
#4913: add support for defaultChecked in <ToggleButton>

billkabanga:
#### What does this PR do?
Adds support for defaultChecked in `<ToggleButton>`

#### Description of tasks completed.
- add defaultChecked prop to the ToggleButtonProps interface
- use defaultChecked as value for active prop for the Button element
- write a test for this to make sure it is working fine

#### What is the related issue?
[#4913](https://github.com/react-bootstrap/react-bootstrap/issues/4913)

#### Screenshots

Before:
Screenshot 2020-01-13 at 00 40 39

After:
Screenshot 2020-01-13 at 00 40 31

### DIFF
--- a/src/ToggleButton.js
+++ b/src/ToggleButton.js
@@ -2,9 +2,8 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React, { useState, useCallback } from 'react';
 
+import { useUncontrolled } from 'uncontrollable';
 import Button from './Button';
-
-const noop = () => {};
 
 const propTypes = {
   /**
@@ -20,6 +19,8 @@ const propTypes = {
 
   /**
    * The checked state of the input, managed by `<ToggleButtonGroup>` automatically
+   *
+   * @controllable onChange
    */
   checked: PropTypes.bool,
 
@@ -31,6 +32,8 @@ const propTypes = {
   /**
    * A callback fired when the underlying input element changes. This is passed
    * directly to the `<input>` so shares the same signature as a native `onChange` event.
+   *
+   * @controllable  checked
    */
   onChange: PropTypes.func,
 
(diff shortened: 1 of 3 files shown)


src/ToggleButton.js
```diff
@@ -83,7 +89,7 @@ const ToggleButton = React.forwardRef(
           disabled && 'disabled',
         )}
         type={null}
-        active={!!checked}
+        active={!!checked || !!defaultChecked}
```
jquense: this wouldn't work, the button would remain active when once checked changed, since `defaultChecked` is never updated

jquense: A proper implementation i think would require controlling `checked` here and passing `defaultChecked` as the initial value

billkabanga: @jquense Thank You for this feedback. I am going to work on fixing it.

billkabanga: @jquense I have updated the PR kindly check it out. Thank You

src/ToggleButton.js
```diff
@@ -45,65 +50,74 @@ const propTypes = {
...
+        name={name}
+        type={type}
+        value={value}
+        ref={inputRef}
+        autoComplete="off"
+        checked={!!checked}
+        disabled={!!disabled}
+        onFocus={handleFocus}
+        onBlur={handleBlur}
+        onChange={handleChange || noop}
```
taion:
```suggestion
        onChange={handleChange}
```

noop case isn't possible any more

src/ToggleButton.js
```diff
@@ -45,65 +50,74 @@ const propTypes = {
...
-    );
-  },
-);
+const ToggleButton = React.forwardRef((props, ref) => {
+  let {
+    children,
+    name,
+    className,
+    defaultChecked,
+    checked = defaultChecked,
```
taion: there's no reason to pass through `defaultChecked` or to explicitly default `checked = defaultChecked`; uncontrollable takes care of that for us

src/ToggleButton.js
```diff
@@ -45,65 +50,74 @@ const propTypes = {
...
+    if (e.target.tagName === 'INPUT') setFocused(true);
+  }, []);
+
+  const handleBlur = useCallback(e => {
+    if (e.target.tagName === 'INPUT') setFocused(false);
+  }, []);
+
+  const handleChange = event => {
+    onChange(!checked, event);
+  };
```
taion: this would be a breaking change, as it would change the behavior of explicitly-configured `onChange` callbacks.

@jquense do we need to change the API here?

billkabanga: I have reverted this so that we maintain the behaviour of the explicitly configured onChange callbacks. 
@taion

taion: does checking/unchecking still work when using `defaultChecked`, though?

billkabanga: It works but the user would have to manually handle the checking/unchecking for the Toggle buttons.

taion:
that's not really what `defaultChecked` is supposed to do, right? it's supposed to make things "just work" without requiring an explicit handler...

billkabanga:
> that's not really what `defaultChecked` is supposed to do, right? it's supposed to make things "just work" without requiring an explicit handler...
That's true @taion

jquense:
Taking a step back (which i should have done earlier), I think the reason defaultChecked wasn't supported here is that it's really unclear what it's behavior should even be. In the context of ToggleButtonGroup it really shouldn't do _anything_, the default value should go on ToggleButtonGroup since that is the component controlling the actual value. So functionally you get the `defaultChecked` behavior by adding `defaultValue` to the parent instead.

The other case would be using the ToggleButton's alone and preserving the underlying checkbox api, but is there a use case for that?

kosmiq:
So I started looking through this just to have an opinion. This entire discussion and PR should probably be closed.

The PR is set out to add behavior to ButtonGroup (**NOTE: not ToggleButtonGroup**). In the linked example there are three radio buttons within a ButtonGroup. The use-case is already managed by the ToggleButtonGroup component where you can actually set defaultChecked etc.

So we should probably updated the docs and remove defaultChecked from the example https://react-bootstrap.github.io/components/buttons/#checkbox--radio.

jquense:
IDK, the example is explicitly showing the usage with ButtonGroup because that is a supported bootstrap component. ToggleButtonGroup is more opinionated and added to help simplify usage for common cases, but `<ButtonGroup toggle>` is a first case API.

The other thing is that `defaultChecked` really doesn't work with ToggleButtonGroup, since all the children are explicitly controlled and the value is passed from the group down. Overall i think there is reasonable use of this component in an uncontrolled state, same as with normal checkboxes

src/ToggleButton.js
```diff
@@ -2,10 +2,9 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import React, { useState, useCallback } from 'react';
 
+import { useUncontrolled } from 'uncontrollable';
```
jquense: this should use `uncontrolledProp` now fyi

kosmiq:
Well in that case I guess we have more work to do. ButtonGroup.js does not currently support either defaultValue or value as props. And it has no functionality to forward the values towards underlying buttons of any type.

I guess that would mean adding checks to make sure that value/defaultValue on the ButtonGroup can only bet set when toggle is true.

jquense:
> Well in that case I guess we have more work to do. ButtonGroup.js does not currently support either defaultValue or value as props. And it has no functionality to forward the values towards underlying buttons of any type.

it doesn't need those things actually, the goal would be for it function like a native set of checkboxes or radios, but with button styling. It already does that, you can stick into a `<form>` and submit it and the values will be correct. 

The only thing that really doesn't work here is [defaultChecked](https://reactjs.org/docs/uncontrolled-components.html#default-values) which is both React and DOM native thing. That is because the state of the button (just the button) depends on the value of the input and so we are forced to control `checked` on the input, even tho to the user it's still uncontrolled, which breaks defaultChecked, which needs to be added back.

jquense:
And to clarify the role of each of these versions:

- `<ButtonGroup toggle>` with `ToggleButtons` is the equivalent to: https://getbootstrap.com/docs/4.4/components/buttons/#checkbox-and-radio-buttons which has no fancy state management and is purely presentational sugar on top of native HTML checkbox groups
- `ToggleButtonGroup` is an opinionated addition that implements _one_ possible method of serializing/deserializing checkbox/radio value state in a more idiomatic React API.

Issue #4932, opened 2020-01-15
Open NavDropdown on mouse over


amsterdamharu:
## Is your feature request related to a problem? Please describe
Would like to open NavDropdown on mouse over but can't because there is [no prop](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/NavDropdown.js#L8-L40) controlling the open state of the NavDropdown

## Describe the solution you'd like

Tried the following

```javascript
<NavDropdown
  title={category.name[LANGUAGE]}
  onMouseEnter={handleOpen}
  onMouseLeave={handleClose}
  open={isOpen}
>
```

Would like to have a prop that can control if the menu is open or not if such a prop exist then it's not documented anywhere nor can I find it in the source code.

taion:
The prop is `show`, and it works just fine. Though we'd be happy to take a PR to update the docs, perhaps by including NavDropdown's props table somewhere.

amsterdamharu:
@taion  I can make a pull request with`show` added to the [proptypes](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/NavDropdown.js#L8-L40)

taion:
let's track this in https://github.com/react-bootstrap/react-bootstrap/issues/4946 instead, as that issue more clearly states the problem

PR #4935, opened 2020-01-16
Issue #4932 missing prop type

amsterdamharu:
Added the proptype `show`


taion:
This is Navbar, not NavDropdown.

Also, there's a bigger problem where we're not showing the NavDropdown props table full stop.

PR #4938, opened 2020-01-17
refactor(OverlayTrigger): convert to function component

muzakparov:
Related to #4194


taion:
First, great job refactoring this away from being a class component!

Second, sorry, but unfortunately, this looks like it was a bug in #4194. The `<OverlayTrigger>` itself probably shouldn't accept a ref – the `<RefHolder>` itself is just so we can attach a ref to find a DOM node. It's otherwise uninteresting and there's no particular reason why a user would _want_ a ref there. If anything, the user would want a ref to the wrapped component – but he or she can just drop a ref on that component directly.

I don't think we want to change `<OverlayTrigger>` to require the rendered child to accept a ref.

So... I think it'd be justifiable to just drop the `useMergedRefs` thing and make this a non-ref-forwarding component, but still a functional component that uses hooks (as it is now), rather than a class component.

muzakparov:
@taion thank you for detailed explanation! :) 
could you please help me with tests? Since component was refactored to react hooks, some of the test related to  react `state` fail:
```
expect(wrapper.state('show')).to.equal(...)  // - fails
```
etc

should I just remove them?

taion:
Can you instead look for the `<Overlay>` component instance and assert on the value of its `show` prop as passed down? Should be the same thing.

muzakparov:
@taion sorry, I am bad with test and I couldn't come up with a way to access `show` prop. `<Overlay>` component turned out to be a function Component and I guess it doesn't have an instance.

muzakparov:
@taion removed `forwardRef` along with `useMergedRef`

src/OverlayTrigger.js
```diff
@@ -1,8 +1,7 @@
 /* eslint-disable max-classes-per-file */
```
taion:
```suggestion
/* eslint-disable max-classes-per-file */

```

src/OverlayTrigger.js
```diff
@@ -82,189 +81,175 @@ const defaultProps = {
 };
 
 // eslint-disable-next-line react/no-multi-comp
```
taion:
```suggestion
```

no longer need this pragma as we only have one class component now

src/OverlayTrigger.js
```diff
@@ -82,189 +81,175 @@ const defaultProps = {
...
-        const target = this.getTarget();
-        if (!this.state.show || !target) return data;
-
-        const role = popper.getAttribute('role') || '';
-        if (popper.id && role.toLowerCase() === 'tooltip') {
-          target.setAttribute('aria-describedby', popper.id);
-        }
-        return data;
-      },
+const OverlayTrigger = ({
```
taion: to match the other components, could we have this as:

```js
function OverlayTrigger({ ... })
```

i.e. we don't use arrow functions except with `React.forwardRef`

src/OverlayTrigger.js
```diff
@@ -82,189 +81,175 @@ const defaultProps = {
...
+        onHide={handleHide}
+        target={getTarget}
+      >
+        {overlay}
+      </Overlay>
+    </>
+  );
+};
+
+OverlayTrigger.displayName = 'OverlayTrigger';
```
taion:
```suggestion
```

no longer needed as this is not a `React.forwardRef` any more

test/OverlayTriggerSpec.js
```diff
@@ -232,111 +232,112 @@ describe('<OverlayTrigger>', () => {
...
-        wrapper.find('button').simulate('click');
-
-        // Need to click this way for it to propagate to document element.
-        document.getElementById('replace-overlay').click();
-
-        wrapper.state('show').should.be.true;
-      });
-    });
-  });
+  // We need different tests that doesn't rely on state
```
bpas247: Any chance we can re-write these to get them to work? otherwise we should just remove them.

muzakparov: removing them for now

src/OverlayTrigger.js
```diff
@@ -77,192 +100,168 @@ const propTypes = {
...
+  defaultShow,
+  delay: propsDelay,
+  ...props
+}) {
+  const triggerNodeRef = useRef(null);
+  const timeoutRef = useRef();
+  const hoverStateRef = useRef();
+  const [show, setShow] = useState(!!defaultShow);
+
+  useEffect(() => {
```
jquense: `useWillUnmount()` simplifies this in restart/hooks

src/OverlayTrigger.js
```diff
@@ -77,192 +100,168 @@ const propTypes = {
...
-    clearTimeout(this._timeout);
-  }
+  const child = React.Children.only(children);
+  const { onFocus, onBlur, onClick } = child.props;
 
-  getChildProps() {
-    return React.Children.only(this.props.children).props;
-  }
+  const getTarget = useCallback(
+    () => ReactDOM.findDOMNode(triggerNodeRef.current),
```
jquense: we should consider dropping the `findDOMNode` before v1 like we did for dropdowns

src/OverlayTrigger.js
```diff
@@ -77,192 +100,168 @@ const propTypes = {
...
+  trigger,
+  overlay,
+  children,
+  popperConfig = {},
+  defaultShow,
+  delay: propsDelay,
+  ...props
+}) {
+  const triggerNodeRef = useRef(null);
+  const timeoutRef = useRef();
```
jquense: also in restart/hooks `useTimeout` simplifies the logic here a bit

src/OverlayTrigger.js
```diff
@@ -77,192 +96,164 @@ const propTypes = {
...
+}) {
+  const triggerNodeRef = useRef(null);
+  const timeout = useTimeout();
+  const hoverStateRef = useRef();
+  const [show, setShow] = useState(!!defaultShow);
 
-  componentWillUnmount() {
-    clearTimeout(this._timeout);
-  }
+  useWillUnmount(() => timeout.clear());
```
taion: oh, we can actually remove this entirely, if that's all we need from unmounting

`useTimeout` will clear itself on unmount

bpas247: It even calls `useWillUnmount` internally.
https://github.com/react-restart/hooks/blob/master/src/useTimeout.ts#L41

src/OverlayTrigger.js
bpas247: @jquense would that be out of scope for this PR? Or should we do this in this PR?

taion: not in scope for this PR, i don't think

src/OverlayTrigger.js
muzakparov: removed it, thanks!

taion:
thanks!

Issue #4934, opened 2020-01-15
Broken aria-describedby reference on OverlayTrigger to tooltip element


tragessere:
## Describe the bug

The OverlayTrigger adds an `aria-desribedby` attribute to it's target when showing (if the popper has a tooltip role). When the tooltip is no longer showing the `aria-describedby` attribute is left

## To Reproduce

1. Create an OverlayTrigger with an overlay that has the 'tooltip' role and and id. Add an element inside the OverlayTrigger to activate it such as a button (any trigger type is fine).
2. The element correctly does not have an `aria-describedby` attribute at the start.
2. Activate the popover.
3. The element now correctly has an `aria-describedby` attribute referencing the popover.
4. Close the popover.
5. The element now incorrectly still has the `aria-describedby` attribute which no longer references anything in the DOM.

## Reproducible Example

Go to https://react-bootstrap.github.io/components/overlays/#popover-examples. Look at the attributes for the 'Click me to see' example button before showing the popover and after closing it.

## Expected behavior

The `aria-describedby` attribute should be removed from the OverlayTrigger's target element when not showing the popover. Alternatively the popover could be left in the DOM and shown/hid dynamically instead and the `aria-describedby` attribute could always be on the target element.

## Environment (please complete the following information)

- Operating System: Windows
- Browser, Version Chrome 79
- React-Bootstrap Version 1.0.0-beta.16

## Additional context

The dynamic `aria-describedby` attribute was added in a merge request as part of this old issue: https://github.com/react-bootstrap/react-bootstrap/issues/1827

taion:
Looks like we're missing a cleanup function for the ARIA modifier. We'd be happy to take a PR to address this.

ocBruno:
A little new to this environement but i'm going to give this one a shot, if anyone already working on this give me a heads up!

tragessere:
Closing this issue. It was fixed as part of https://github.com/react-bootstrap/react-bootstrap/pull/5017

PR #4940, opened 2020-01-19
(fix): remove associated aria on overlay hide

ocBruno:
fix for issue #4934 

solution:
-get target from element during onHide event and use removeAttribute

First PR for this repo and still need to dive more into react so please spot me on any mistakes or incorrect design patterns!
Any feedback and constructive critcism is always welcome!

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -166,9 +166,11 @@ class OverlayTrigger extends React.Component {
 
   handleClick = e => {
     const { onClick } = this.getChildProps();
-
-    if (this.state.show) this.hide();
-    else this.show();
+    if (this.state.show) {
+      const target = e.currentTarget;
+      target.removeAttribute('aria-describedby');
+      this.hide();
+    } else this.show();
 
     if (onClick) onClick(e);
   };
--- a/test/OverlayTriggerSpec.js
+++ b/test/OverlayTriggerSpec.js
@@ -97,6 +97,24 @@ describe('<OverlayTrigger>', () => {
     });
   });
 
+  it('Should remove aria-describedby for tooltips if the state is hide', done => {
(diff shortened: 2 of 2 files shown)


taion:
sorry, what i mean is – it looks like popper itself gives us a way to create and clean up side effects from modifiers: https://popper.js.org/docs/v2/modifiers/#effect

given this, can we use this functionality?

ocBruno:
> sorry, what i mean is – it looks like popper itself gives us a way to create and clean up side effects from modifiers: https://popper.js.org/docs/v2/modifiers/#effect
> 
> given this, can we use this functionality?

I think we could use this functionality if it was to improve performance but I don't know if it would, it would be nice if someone with more knowledge could chime in on this for us, any changes I'm willing to apply!

tragessere:
@taion I don't think that cleanup function is available to here since react-bootstrap is using Popper v1.16 and not v2.x. The latest update of [react-overlays](https://github.com/react-bootstrap/react-overlays/blob/master/package.json) is still using popper v1.15 as well.

https://popper.js.org/docs/v1/
I'm not immediately seeing a good cleanup function provided in the first version.

Not sure if upgrading this library should be another task before this one.

tragessere:
Actually I do see that there is a pull request in React-Overlays to upgrade to Popper v2. I'll just add a link here since this will probably wait on that one.

https://github.com/react-bootstrap/react-overlays/pull/766

taion:
looks like this will be handled by https://github.com/react-bootstrap/react-bootstrap/pull/5017/files?utf8=%E2%9C%93&diff=split&w=1#diff-0d27b0102f68505cf339e89dfe5c6893R198-R202

jquense:
thanks for this! I rolled it into #5017

Issue #3797, opened 2019-05-15
Event handler types seem to be missing in 1.0.0-beta.8


statico:
## Describe the bug

First, thank you for making this wonderful library.

It seems that event handlers don't have types in 1.0.0-beta.8. When I create a `<Button>` with an `onClick` handler like `event => ...`, and my TypeScript project has `strict` enabled, VS Code shows me an error, `Parameter 'event' implicitly has an 'any' type.ts(7006)`.

## To Reproduce

1. Create a file, `Test.tsx`, like this:

```typescript
import * as React from 'react'
import { Button } from 'react-bootstrap'

export default () => (
  <Button onClick={event => alert('hello, ' + String(event))} />
)
```

2. Run `tsc --jsx react --noEmit --strict components/Test.tsx`
3. See error

## Reproducable Example

https://codesandbox.io/s/condescending-curran-lz9k6

Notice the warning on line 8.

## Expected behavior

I would expect that `onClick` attributes be typed correctly, such as when using `<button onClick={...}>` (the native button component). These are provided via the `@types/react` package: https://www.npmjs.com/package/@types/react

## Screenshots

![image](https://user-images.githubusercontent.com/137158/57814999-7b270f80-772a-11e9-9579-b7a2197398ff.png)

## Environment (please complete the following information)

- Operating System: macOS 10.14.4
- React-Bootstrap Version 1.0.0-beta.8
- TypeScript version 3.4.5

bpas247:
https://github.com/react-bootstrap/react-bootstrap/blob/7bd64defe2d0390397fab6938cc84f1191bc80ec/types/components/Button.d.ts#L32-L34
Hmm, I would think that the declaration here would inherit the type for `onClick`, but it doesn't seem like that's the case. 

Maybe we just need to add an explicit type for `onClick` to `ButtonProps`? Would be happy to take a PR for this.

0xc0d3r:
@bpas247 I think, adding this `onClick: (event: React.MouseEvent) => void` to `ButtonProps` will resolve the typescript error.

Shall I create a PR?

bpas247:
> @bpas247 I think, adding this `onClick: (event: React.MouseEvent) => void` to `ButtonProps` will resolve the typescript error.
> 
> Shall I create a PR?

That was my initial thought for a possible solution, yeah. I'd say go for it 😃

0xc0d3r:
@bpas247 I'm unable to push the code. Please add me as a contributor.

[...]

PR #4961, opened 2020-02-01
Fix onClick typescript error in Button (#3797)

0xc0d3r:


mxschmitt:
It's already there because we in the end extend from the button HTML element interface. You can try to use the official code sandbox Typescript example with React-Bootstrap and add a `onClick` handler there: https://codesandbox.io/s/github/react-bootstrap/code-sandbox-examples/tree/master/basic-ts

(In the showcases directory)

image

kosmiq:
This works fine for me as well in a current project without any updates to types etc.
Should we close this @mxschmitt

mxschmitt:
Yep @kosmiq

PR #4968, opened 2020-02-05
docs: add example for preselecting option select

donkeyDau:


taion:
I think we're trying to follow https://getbootstrap.com/docs/4.4/components/forms/#form-controls pretty closely there. Is there a particular reason to document what should be a fairly standard prop on `<select>` there?

donkeyDau:
I had a hard time to determine how the preselection of a `<select>` works. I tried a `selected` prop on the options and a few other things. In the end I got aware that the `value` prop is the key.

I was no aware that it's that close to the bootstrap docs. But there are examples using `selected` on `<option>` of the `<select>` (e.g. https://getbootstrap.com/docs/4.4/components/forms/#form-row -> State `<select>` [even though I think it's not really pointing out the main functionality by preselecting the first option ...]) that are not represented in the the react-bootstrap docs.

taion:
sure, so if you want to edit/update a different example in a way that matches the upstream docs, we'd be happy to have that

donkeyDau:
New PR https://github.com/react-bootstrap/react-bootstrap/pull/4976

PR #4970, opened 2020-02-05
feat: add support for className mapping useful for CSS Modules

adamayres:
## Background

[Css Modules](https://github.com/css-modules/css-modules) is popular library for adding CSS to one's application typically used in conjunction with a CSS loader such as [css-loader](https://github.com/webpack-contrib/css-loader), [sass-loader](https://github.com/webpack-contrib/sass-loader), [postcss-loader](https://github.com/postcss/postcss-loader), etc.

This feature and functionality is supported by many frameworks by default, such as in [NextJs](https://nextjs.org/docs/basic-features/built-in-css-support#adding-component-level-css), [GatsbyJs](https://www.gatsbyjs.org/packages/gatsby-plugin-react-css-modules/), [Meteor](https://github.com/akanix42/meteor-css-modules), and others.

From [CSS Modules](https://github.com/css-modules/css-modules#css-modules) documentation:

> A **CSS Module** is a CSS file in which all class names and animation names are scoped locally by default.

## Use Case

CSS Modules are useful for sandboxing component styles from styles that are contributed by a third party, such as the case in a white label application.

CSS Modules are also a useful way of developing CSS, see more here:

* [What are CSS Modules and why do we need them?](https://css-tricks.com/css-modules-part-1-need/)
* [CSS Modules, Welcome to the future](https://glenmaddern.com/articles/css-modules)

## Issue

The `react-bootstrap` library uses hard-coded class names, which currently allow for a prefix, via the `bsPrefix` prop (per-component, or via the `ThemeProvider`), however, they do not allow for a full re-mapping of the class names as-is needed in order to support CSS Modules.

## Approach

Similar to how `bsPrefix` support was implemented, a `classNameMap` prop was added to `ThemeProvider` and each component.

There was a new hook added in `ThemeProvider` called `useClassNameMapper` that each functional component uses in place of the previous `classnames` library. The `useClassNameMapper` returns an API identical to that of `classnames`, but decorated with support for mapping the class names used in the component to the ones provided by the `classNameMap` prop. In the case no `classNameMap` is provided at the component level or by the `ThemeProvider`, it will leave the class names unmapped - as they are by default.

**Note:** For the remaining components that use Classes, notably `Modal` and `Collapse`, a `<ThemeConsumer>` is used to build the decorated `classnames` object as opposed to the `useClassNameMapper` hook.

Furthermore, every place that adds a `className` attribute, or passes a class name onto a third party library, has been updated to pass through the decorated `classnames` object.

**Note:** I discovered a handful of places where the `bsPrefix` was not being applied to all of the outputted classes. For example: the `active` class applied in `Button`, the classes applied by `Fade` in `Alert`, any place the `sr-only` was being used, the `collapseStyles` used in `Collapse`, the `show` class used in `Modal`, as well as some others.

Lastly, in order to support when CSS Modules is configured to output the compiled class names using [`camelCaseOnly`](https://webpack.js.org/loaders/css-loader/#localsconvention), support was added to the `ThemeProvider` to allow for a `classNameConverter` callback function to be provided. An example usage has been documented in the `theming` doc's page. Docs and tests have been added for the new functionality added.

**Note:** While this may seem like a large change set, most of the changes made per-component were relatively small.

Thanks for the great library! Feedback is appreciated.


taion:
Right, so on the utility classes, I think those are intentionally unprefixed. The `bsPrefix` support is on e.g. changing `form` to `my-form`, but e.g. `show` is the same everywhere.

As for all of this, I'm not completely sure about the approach here. We in fact happen to use Bootstrap and CSS modules, but I'm not sure that there's much point to using CSS modules for Bootstrap itself, given that Bootstrap also does stuff like run a page-level styling reset that by construction can't be modularized.

adamayres:
@taion Thanks for the feedback. To address you specific feedback:

> Right, so on the utility classes, I think those are intentionally unprefixed. The bsPrefix support is on e.g. changing form to my-form, but e.g. show is the same everywhere.

That makes sense, however for my use case, I would like to change all classes that are applied by `react-bootstrap` as to avoid any potential overlap with CSS that might be added to the page by a third party. In our use case, we allow customers to add CSS to our pages.

> As for all of this, I'm not completely sure about the approach here. We in fact happen to use Bootstrap and CSS modules, but I'm not sure that there's much point to using CSS modules for Bootstrap itself, given that Bootstrap also does stuff like run a page-level styling reset that by construction can't be modularized.

Yes, that is a fair point. As noted, the Bootstrap library contains the `_reboot.scss` partial that will restyle root elements that will be added globally even when using CSS Modules. However, there are ways around this, such as building Bootstrap without the `_reboot.scss` partial, or replacing `_reboot.scss` with your own version that is scoped under a class name that is used on the html or body element.

I think there is value in supporting the remapping of the classes used in Bootstrap for CSS Modules. I am not sure if this will help my case or hurt it, but reactstrap has support for this functionality and my approach was borrowed from their implementation, see this [PR](https://github.com/reactstrap/reactstrap/pull/206) for more details.

Thanks again for the consideration.

taion:
@jquense wdyt?

kosmiq:
One thing to have in mind regarding the reboot.css.

If you omit this CSS, buttons etc might suddenly look different in some browsers depending on their defaults, giving you some unexpected appearances. 

An example is the Bootstrap docs sidemenu nav. Reboot in Bootstrap defines
```
*, ::after, ::before {
    box-sizing: border-box;
}
```
If we disable this rule (i.e. the user does not include reboot.css or any other reset framework that defines this behaviour for example), the menu looks very different in terms of width:
bootstrap-reboot-box-sizing
(left: rule enabled, right: rule disabled)

And this is just one example. There might be other inheritances that are at play as well (I haven't really dug into it, I think Bootstrap is pretty good in that regard these days).

The thing is, unless you really make sure to include the basic rules that defines some of the Bootstrap layouts and removes browser quirks, you could end up with a Bootstrap implementation that looks different than intended.

jquense:
Can we take a step back and outline the specific use-case for this? I understand the value of css-modules generally, but it does seem like there isn't much to gain here from enabling them.  Bootstrap as a framework really isn't meant to be used super piecemeal, nor would it make a lot of sense as styling choice for "embedded" components, where you'd have to worry about existing classes conflicting. 

Practically is the expectation that a user would run `bootstrap.css` though css-modules and pass the class map to each component? I'm not opposed to this at all but it seems to serve a very niche use-case with a non-trivial amount of code/complexity. Would love to hear more.

jquense:
@adamayres sorry for drop off. I think this is probably a reasonable thing to do but I want to try and scope it more minimally if possible

- don't allow class mappings to be specified on components, the root provider seems like a sufficient place to do it. (it was also probably overkill for bsPrefix but that is more legacy)
- See if it's possible to combine the class mapping and bsPrefix code paths more so the later can be described in terms of the classMapping feature

How does that sound? feedback?

adamayres:
@jquense No worries, thanks for getting back to me. Regarding your bullet points:

> don't allow class mappings to be specified on components

That makes sense, I think I fell into the trap of trying to emulate the `bsprefix` too much and ended up with a per-component override which has not been useful in practice. I will remove this and only use the global one.

> See if it's possible to combine the class mapping and `bsPrefix` code paths more so the later can be described in terms of the classMapping feature

I think it would be possible to express the `bsprefix` in terms of the class name mapping, but I was trying to be sensitive to changing the API for `bsprefix` as to make my changes backwards compatible. I think this should be possible, I will look into it.

sebacruz:
Hi! Any news on this?

anton6:
Hi, is this feature still under consideration?

PR #4975, opened 2020-02-07
Add Comment to AccordionCollapse.js

HPS-RobRice:
Accordion.Collapse cannot have more than one child else the application will crash with the error `Unhandled Rejection (Error): React.Children.only expected to receive a single React element child.` Added a comment to reflect that.


src/AccordionCollapse.js
```diff
@@ -10,6 +10,7 @@ const propTypes = {
    */
   eventKey: PropTypes.string.isRequired,
 
+  /** Children prop should only contain a single child, and  is enforced as such */
```
taion:
```suggestion
  /** Children prop should only contain a single child, and is enforced as such */
```

PR #4864, opened 2019-11-26
Add missing horizontal in ListGroupProps

mustangore:
This pull request fixes #4862

PR #4979, opened 2020-02-07
Added type string to ListGroupProps horizontal prop

mikecsmith:
Hi,

This PR fixes a minor oversight with PR #4864 - the ListGroup component has a horizontal prop which is either a boolean (makes it horizontal) or a string (this makes the ListGroup horizontal at a given breakpoint) - PR #4864 only covers the boolean use case.

Thanks!

### DIFF
--- a/types/components/ListGroup.d.ts
+++ b/types/components/ListGroup.d.ts
@@ -6,7 +6,7 @@ import { BsPrefixComponent, SelectCallback } from './helpers';
 
 export interface ListGroupProps {
   variant?: 'flush';
-  horizontal?: boolean;
+  horizontal?: boolean | string;
   activeKey?: unknown;
   defaultActiveKey?: unknown;
   onSelect?: SelectCallback;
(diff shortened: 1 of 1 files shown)


types/components/ListGroup.d.ts
```diff
@@ -6,7 +6,7 @@ import { BsPrefixComponent, SelectCallback } from './helpers';
 
 export interface ListGroupProps {
   variant?: 'flush';
-  horizontal?: boolean;
+  horizontal?: boolean | string;
```
taion: actually, this should be the union of known breakpoints like in https://github.com/react-bootstrap/react-bootstrap/blob/87ea3552e0e31e550dc896dca3e6ac2967d82ec2/types/components/Navbar.d.ts#L15

mikecsmith: Good point - lazy on my part - I'll go tweak and push now

Issue #5008, opened 2020-02-27
Allow `SplitButton` primary button `type` to be set


WillJW:
## Is your feature request related to a problem? Please describe

Adding a SplitButton as the main submit button on a form, I noticed hitting 'enter' in the form wouldn't submit it. This is because the primary SplitButton button is `type="button"`.

## Describe the solution you'd like

I would like to be able to set the primary button's type.

Example:

```jsx
<SplitButton title="Save" variant="primary" type="submit">
  <Dropdown.Item onClick={this.doSomething}>
    Delete
  </Dropdown.Item>
</SplitButton>
```

mxschmitt:
Closed because it was implemented in #5009

PR #5009, opened 2020-02-27
Allow type to be set for SplitButton non-toggle Button

WillJW:
https://github.com/react-bootstrap/react-bootstrap/issues/5008


WillJW:
@mxschmitt @taion Thanks for merging this. I don't suppose you know when it will be tagged for release?

PR #5015, opened 2020-02-28
Toast uses opacity 0 instead of display none

VinitSarvade:
Added class hide when the toast is manually closed or autoclosed. If the toast uses fade transition then the hide class gets added after completion of the transition

### DIFF
--- a/src/Toast.js
+++ b/src/Toast.js
@@ -79,13 +79,17 @@ const Toast = React.forwardRef(
     bsPrefix = useBootstrapPrefix('toast');
     const delayRef = useRef(delay);
     const onCloseRef = useRef(onClose);
-    const [transitionComplete, setTransitionComplete] = useState(false);
+    const [exited, setExited] = useState(!show);
 
-    useEffect(() => {
-      if (show && transitionComplete) {
-        setTransitionComplete(false);
+    const useAnimation = Transition && animation;
+
+    if (show) {
+      if (exited) {
+        setExited(false);
       }
-    }, [show, transitionComplete]);
+    } else if (!useAnimation && !exited) {
+      setExited(true);
+    }
 
     useEffect(() => {
       // We use refs for these, because we don't want to restart the autohide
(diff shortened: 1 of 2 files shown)


taion:
Do we really want the `'hide'` class as opposed to removing from the DOM entirely?

VinitSarvade:
> Do we really want the `'hide'` class as opposed to removing from the DOM entirely?

Well yeah, that makes more sense. I will make the necessary changes.

AparnaPrasad:
Also, component Fade, has unmountOnExit prop, wouldn't using that be simpler than having callback on transition end?

taion:
ah, you're right – it would make more sense to use `unmountOnExit`

taion:
so i think the right thing to do here is to just pass through `unmountOnExit` to the transition. we might want to default it to `true`, though

src/Toast.js
```diff
@@ -73,6 +79,17 @@ const Toast = React.forwardRef(
...
+
+    const useAnimation = Transition && animation;
+
+    if (show) {
+      if (exited) {
+        setExited(false);
+      }
+    } else if (!useAnimation && !exited) {
+      setExited(true);
+    }
```
bpas247: This should be in a `useEffect` call, and it should also be in a functional update.
```suggestion
useEffect(() => {
  setExited(prevExited => {
    if (show && prevExited) {
      return false;
    } else if (!useAnimation && !prevExited) {
      return true;
    } else {
      return prevExited;
    }
  });
}, [show, useAnimation]);
```

src/Toast.js
```diff
@@ -73,6 +79,17 @@ const Toast = React.forwardRef(
     bsPrefix = useBootstrapPrefix('toast');
     const delayRef = useRef(delay);
     const onCloseRef = useRef(onClose);
+    const [exited, setExited] = useState(!show);
+
+    const useAnimation = Transition && animation;
```
bpas247: Out of curiosity, why was this originally named `useAnimation`? seems like `hasAnimation` would be more appropriate here, since it's not a custom hook.

rjray:
I've run into this same issue. Any feeling for when it'll be reviewed and merged?

v0lume:
also need this functionality to remove current workarounds

VinitSarvade:
Successsor: https://github.com/react-bootstrap/react-bootstrap/pull/5123

PR #5017, opened 2020-02-28
Bump react-overlays

jquense:


src/DropdownMenu.js
```diff
@@ -43,7 +43,7 @@ const propTypes = {
   as: PropTypes.elementType,
 
   /**
-   * A set of popper options and props passed directly to react-popper's Popper component.
+   * A set of popper options and props passed directly to Popper's.
```
jquense:
```suggestion
   * A set of popper options and props passed directly to Popper.
```

src/Overlay.js
```diff
@@ -26,7 +26,7 @@ const propTypes = {
   show: PropTypes.bool,
 
   /**
-   * A set of popper options and props passed directly to react-popper's Popper component.
+   * A set of popper options and props passed directly to PopperJs.
```
jquense:
```suggestion
   * A set of popper options and props passed directly to Popper.
```

taion:
re-running tests... concerned about possible test flake though

jquense:
unsure what the test issue is here. can't repro

jquense:
anyone have an idea why the test is failing her? Could someone else try this out locally and see if they can reproduce?

kosmiq:
> anyone have an idea why the test is failing her? Could someone else try this out locally and see if they can reproduce?

Working fine locally here as well. No errors reported at all. Checked out the branch in a brand new folder, ran `yarn start bootstrap` and then `yarn test`
MacOS 10.15.3
Chrome 80.0.3987

I am however curious about what is actually reported in the test here. I did you (frankly not very pleasant) work of copying the entire test log from the build tools. And pasted them into VSCode and started looking for fails the manual way.

The fail is placed within, and I'm not sure what is up with that. Formatting error of console output? Does it run async?
```
<SplitButton>
	✔ should open the menu when dropdown button is clicked
	✔ should not open the menu when other button is clicked
	✔ should invoke onClick when SplitButton.Button is clicked (prop)
	✔ should not invoke onClick when SplitButton.Toggle is clicked (prop)
	✔ Should pass disabled to both buttons
	✔ Should set target attribute on anchor
	✔ should set accessible label on toggle
	✔ should set aria-label on toggle from toggleLabel
	✖ "after each" hook in "{root}"
	✔ should set type attribute from type
```

taion:
it's something toxic in overlayspec... it's making mocha bail out locally and run too few tests

taion:
in chrome headless

taion:
i took a look and updated the tests such that the correct one is failing (though perhaps it should be a rAF instead of a setTimeout)

two issues here, one easily fixable and one not:
- r-o's `<Overlay>` is still passing down the typoed `outOfBoundries`, while the components are only stripping out `outOfBoundaries`
- `<Overlay>` is passing in `...popper` into the child, which is propagating down into the `<div>`. we can't really strip out those entries, because we don't know what they are. and this is leading to some "bad props" warning on the `<div>` (largely from popper state, i think)

PR #5022, opened 2020-03-02
feat: can customize BreadcrumbItem's inner link

### LINKED ISSUE #4472: Cannot customize the `a` inside `Breadcrumb.Item`
when I write something like this

```
<Breadcrumb.Item as={Link} href={href} active={active}>
  {text}
</Breadcrumb.Item>
```

error occurs to say Link didn't get href prop.

after looking through the source code, I found the `as` prop only applies for `li` component, so I have no way to replace the inside `a` tag with proper router link component.

see https://github.com/react-bootstrap/react-bootstrap/blob/master/src/BreadcrumbItem.js#L38

KoltesDigital:
I propose the following change to be able to customize BreadcrumbItem's inner link.

The new property is `linkAs` as suggested.

I split the inner component in a new function because making two ternary comparisons is ugly, and forbidden by the linter anyway. However I don't know the project's good practices regarding naming and visibility.

Fixes #4472.

### DIFF
--- a/src/BreadcrumbItem.js
+++ b/src/BreadcrumbItem.js
@@ -5,6 +5,35 @@ import PropTypes from 'prop-types';
 import SafeAnchor from './SafeAnchor';
 import { useBootstrapPrefix } from './ThemeProvider';
 
+const linkPropTypes = {
+  active: PropTypes.bool,
+  as: PropTypes.elementType,
+  href: PropTypes.string,
+  target: PropTypes.string,
+  title: PropTypes.node,
+};
+
+const linkDefaultProps = {};
+
+const BreadcrumbItemLink = ({ active, as: Component, ...props }) => {
+  const { href, title, target, ...elementProps } = props;
+  const linkProps = { href, title, target };
+
+  if (active) {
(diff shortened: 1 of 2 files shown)


taion:
i mean, it's not ridiculous to do that, but if we're going to go that far, why not also let the user customize the `<span>`?

KoltesDigital:
Well I can't tell, it's only a little step further for sure, but I'm afraid this would add fine-grained customization that would equal writing the original code in the first place.

Besides, [Bootstrap doc illustrates](https://getbootstrap.com/docs/4.0/components/breadcrumb/) the active `li` with no `span` at all:

```html
<ol class="breadcrumb">
  <li class="breadcrumb-item"><a href="#">Home</a></li>
  <li class="breadcrumb-item active" aria-current="page">Library</li>
</ol>
```

So maybe the `span` should be replaced by `React.Fragment`. Or if there's an upvote for adding `activeAs`, it should default to a fragment.

KoltesDigital:
As simple as that. LGTM. I'll check that tests do not break too much.

KoltesDigital:
Superseded in #5043.

PR #5029, opened 2020-03-06
Fix whitespace on homepage

davidjb:
Previously, the hyphen at the end of the line caused whitespace to appear between `React-` and `Bootstrap` on the homepage

### DIFF
--- a/www/src/pages/index.js
+++ b/www/src/pages/index.js
@@ -152,9 +152,9 @@ export default withLayout(
                   ecosystem.
                 </p>
                 <p>
-                  By relying entirely on the Bootstrap stylesheet, React-
-                  Bootstrap just works with the thousands of bootstrap themes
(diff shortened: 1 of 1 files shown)


www/src/pages/index.js
```diff
@@ -152,9 +152,9 @@ export default withLayout(
                   ecosystem.
                 </p>
                 <p>
-                  By relying entirely on the Bootstrap stylesheet, React-
-                  Bootstrap just works with the thousands of bootstrap themes
-                  you already love.
+                  By relying entirely on the Bootstrap stylesheet,
+                  React-Bootstrap just works with the thousands of bootstrap
```
taion:
```suggestion
                  React-Bootstrap just works with the thousands of Bootstrap
```

jquense:
thanks!

PR #5035, opened 2020-03-09
Update CHANGELOG.md for beta 17

kosmiq:
Include responsive containers in changelog. Only the types fix was included.

Since beta 17 seem to be unpublished, this adds responsive containers to the changelog


taion:
huh, it's on npm, but the tag is missing for some reason. @jquense can you push the tag up?

PR #5036, opened 2020-03-10
feat: add file input component

kosmiq:
A file input component that also supports the custom prop.

A few remaning to do's
- [x] Add FormFile,  FormFileInput, FormFileLabel
- [x] Add Types
- [ ] Clean up use of useContext and useMemo?
- [x] Update docs
- [x] Update docs with "long form" instructions
- [ ] Mention in docs that  isValid/isInvalid styles only works on the custom element
- [x] Update docs with order of label/input when using custom and in the "long form"
- [x] Add tests
- [x] Remove bs-custom-file-input script
- [x] Add buttonText prop for data-browse attribute on FormFileLabel
- [x] Add lang as prop to FormFile and FormFileInput
- [x] Test valid/inValid props

**A few things that should be discussed and probably changed before this is merged:**
1. ~~The prop "buttontext" should probably have a better name~~
2. ~~Docs should probably mention that if using the longer form with custom prop set the order of the label and input matters if you want to use the buttontext prop.~~
```
<FormCheck custom>
<FormCheck.Label>Label<FormCheck.Label>
<FormCheck.Input />
</FormCheck>
```
3. ~~I copied and cleaned up the FormCheck components. I am sure more things can be removed from the new FormFile-files to further clean up and simplify them.~~
4. I am pretty unfamiliar with useContext and useMemo so anyone with more knowledge please weigh in we can remove these if they have no function in this use case.
5. ~~I'm sure you can find other improvements to do.~~
6. ~~I haven't fully tested inValid/Valid functionality yet. But setting the "inValid" prop and adding a `<Feedback>` works. You probably have to setup your own states though.~~


src/FormFile.js
```diff
@@ -0,0 +1,189 @@
...
+
+  /** Manually style the input as invalid */
+  isInvalid: PropTypes.bool.isRequired,
+
+  /** A message to display when the input is in a validation state */
+  feedback: PropTypes.node,
+
+  /** The string for the button when using custom file input */
+  // buttontext: PropTypes.string,
+  buttontext: all(PropTypes.string, ({ custom, buttontext }) =>
```
jquense: `buttonText`

www/src/pages/components/forms.js
```diff
@@ -292,6 +294,16 @@ export default withLayout(function FormControlsSection({ data }) {
       <h3>Inline</h3>
       <ReactPlayground codeText={CheckCustomInline} />
 
+      <LinkedHeading h="3" id="forms-custom-file">
+        File
+      </LinkedHeading>
+      <p>A custom styled File uploader.</p>
+
+      <ReactPlayground codeText={File} />
```
jquense: ultimately want to add a more about how to customize the inner markup like https://react-bootstrap.github.io/components/forms/#forms-check-api

package.json
```diff
@@ -67,6 +67,7 @@
     "@restart/context": "^2.1.4",
     "@restart/hooks": "^0.3.11",
     "@types/react": "^16.8.23",
+    "bs-custom-file-input": "^1.3.4",
```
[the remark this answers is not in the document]
kosmiq: Well without it the input will not even say what file has been added/uploaded. It will be stuck saying "select a file" or whatever label you have defined. I think as a default behavior it is lacking maybe to much without the script included?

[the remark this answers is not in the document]
jquense: hmm I agree that the experience isn't great but adding a third party library we don't know anything about isn't good either, especially for optional behavior, anyone using any Form field is going to end up with this in their bundle, even tho it's small.

The expectation wouldn't be to include this since upstream doesn't. I don't see a lot wrong with doing the same thing as them and including it the docs but not the bundle

@taion @bpas247 what do you think?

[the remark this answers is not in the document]
kosmiq: With your arguments above I'm going to remove it and add it to the docs as bootstrap does. Thank you for the input! :)

kosmiq:
Last commits extends the docs, removes the script, has isValid/isInvalid tested (working on custom layout for file).

I'd very much like feedback on wording etc in the docs for the "customizing" example.

I might have screwed a bit the merge since I attempted to use Github Desktop instead of an trusty git app. Should I redo it or do we simply squash the commits here if/when merging?

src/FormFile.js
```diff
@@ -155,7 +147,7 @@ const FormFile = React.forwardRef(
                 <>
                   {input}
                   {hasLabel && (
-                    <FormFileLabel buttontext={buttontext}>
+                    <FormFileLabel buttontext={buttonText}>
```
jquense:
```suggestion
                    <FormFileLabel buttontText={buttonText}>
```

src/FormFileLabel.js
```diff
@@ -40,7 +40,7 @@ const FormFileLabel = React.forwardRef(
         ref={ref}
         htmlFor={htmlFor || controlId}
         className={classNames(className, bsPrefix)}
-        data-browse={buttontext}
+        data-browse={buttonText}
```
jquense: what uses this data attribute?

kosmiq: The data attribute is used to set the text of the "button" (really an ::after pseudo element) in the custom file uploader label.

It can then be used to translate the text in the "button" (either hardcoded or with i18next or similar). It can also be done via CSS but that falls outside the scope of what we are doing here :)

Bootstrap docs here: https://getbootstrap.com/docs/4.4/components/forms/#translating-or-customizing-the-strings-with-html

Screenshot 2020-03-11 at 00 12 34

www/src/pages/components/forms.js
kosmiq: This has been fixed and committed.

package.json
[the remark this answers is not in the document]
kosmiq: The script has been removed and I updated the docs. I do not add the script into the www docs either. Instead I added some text that should guide whomever reads it towards achieving it with the plugin.

src/FormFile.js
```diff
@@ -0,0 +1,190 @@
...
+  isValid: PropTypes.bool.isRequired,
+
+  /** Manually style the input as invalid */
+  isInvalid: PropTypes.bool.isRequired,
+
+  /** A message to display when the input is in a validation state */
+  feedback: PropTypes.node,
+
+  /** The string for the button when using custom file input */
+  buttonText: all(PropTypes.string, ({ custom, buttonText }) =>
```
kosmiq: Been thinking about this some more after some discussion (not on GitHub) and looking at the Bootstrap docs again. Should this prop be remove and the data attribute be passed via ...props instead? This would better align with the official Bootstrap documentation.

However it would, in the case of using
```
  <Form.File
    id="custom-file-translate"
    label="Custom file input"
    data-browse="Button text"
    custom
  />
```
Also be passed to the input element and adding unneeded data attributes to that element.

test/FormFileSpec.js
```diff
@@ -0,0 +1,80 @@
...
+    const wrapper = mount(<FormFile custom label="My label" />);
+
+    wrapper
+      .assertSingle('div.custom-file')
+      .assertSingle('input.custom-file-input');
+
+    wrapper.assertSingle('label.custom-file-label');
+  });
+
+  it('should supports lang when custom', () => {
```
kosmiq: Is there any way to test with enzyme that the CSS that *should* match based on the lang tag being there?

test/FormFileSpec.js
```diff
@@ -0,0 +1,80 @@
...
+    const wrapper = mount(<FormFile custom lang="en" label="My label" />);
+
+    expect(wrapper.prop('lang')).to.equal('en');
+  });
+
+  it('should supports buttonText when custom', () => {
```
kosmiq: Is there any way to test with enzyme that the CSS content attribute that is used appears correctly?

In this case setting the buttonText
```
  <Form.File
    id="custom-file-translate"
    label="Custom file input"
    buttonText="Button text"
    custom
  />
```

Will  match this CSS
```
.custom-file-input~.custom-file-label[data-browse]::after {
    content: attr(data-browse);
}
```

And will thus output like this:
Screenshot 2020-03-11 at 02 11 30

We would probably like to test that "Button Text" in this case actually appears?

src/FormFileLabel.js
taion: interesting... i guess this is the JS that TWBS pulls in?

kosmiq: Actually it is not. The JS they add is only used for displaying the uploaded files name within the custom file uploader. See the end for what the JS does :)

**What the CSS does/how it works**
The data-attribute to button text is a bunch of CSS selector magic and pseudo-elements
Looking att TWBS and how they do it (and how we replicate it with the buttonText prop) is that on the `<label>` element you set the data-browse attribute to a string like so:
`<label class="custom-file-label" for="customFileLangHTML" data-browse="My Custom Button Text">My Regular Label Text</label>`

And with that set the CSS magic works:
```
.custom-file-input~.custom-file-label[data-browse]::after {
    content: attr(data-browse);
}
```

This is why the order of the label and input elements becomes important. When we have an element with the class `.custom-file-input` followed by an element with the class `.custom-file-label` with the data attribute `data-browse` set an `::after` pseudo-element is created. And that pseudo-elements content is set to the data-attribute by using `content: attr(data-browse)`

All according to how `attr()` is supposed to be used by the spec. More over at MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/attr

Totally done without any JS involved at all. The same goes for the `lang` prop that I added in another commit, useable without any JS involved at all.


_**This is what the JS does**_
**With JS**
Choose file becomes [filename]
customFileJS

**Without JS**
Choose file does not change (but the value for the underlying input element does)
customFileNoJS

test/FormFileSpec.js
kosmiq: I can answer this myself. EnzymeJS does not currently support testing ::after pseudo-elements. There is an open issue over at their GitHub: https://github.com/enzymejs/enzyme/issues/2102

test/FormFileSpec.js
kosmiq: I can answer this myself. EnzymeJS does not currently support testing ::after pseudo-elements. There is an open issue over at their GitHub: https://github.com/enzymejs/enzyme/issues/2102

kosmiq: I can answer this myself. EnzymeJS does not currently support testing ::after pseudo-elements. There is an open issue over at their GitHub: enzymejs/enzyme#2102

src/FormFile.js
kosmiq: Removed buttonText prop and instead uses data-browse attribute as TWBS to keep consistency with them.

jquense:
i'm a bit conflicted on `data-browse` vs `buttonText`, i think you are right that just going with `data-browse` is probably the best, since a user may have to engage with the API directly for translation, and it being a plain data attribute doesn't create confusion as to why _only_ strings are allowed (instead of react nodes)

on the other hand it's uglier :/

kosmiq:
I agree on the ugliness. On the other hand it avoids confusion like you said, and it follows TWBS way of doing it and with the same way of doing it using the same attributes.

src/FormFile.js
```diff
@@ -0,0 +1,186 @@
...
+      feedback,
+      className,
+      style,
+      label,
+      children,
+      custom,
+      lang,
+      'data-browse': dataBrowse,
+      // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
+      as = 'input',
```
jquense: I wonder how valuable this `as` is...it's consistent with FormCheck i know, but overall it's weird (it's weird on FormCheck also).  Usually `as` if for the outer, root, component not some inner component. This would be better names as `inputAs`. The other thing is that why would want to customize this? it's basically a hidden input.

@taion

src/FormFileInput.js
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * The underlying HTML element to use when rendering the FormFileInput.
+   *
+   * @type {('input'|elementType)}
+   */
+  as: PropTypes.elementType,
+
+  /** A HTML id attribute, necessary for proper form accessibility. */
+  id: PropTypes.string,
```
jquense: usually we have a requiredForA11y proptype for this

kosmiq: You are correct. Looking into this now and I think we should do like this:

- FormFile
  - Component
    -   id: isRequiredForA11y(
    PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
  ),
  - Types
    - id: string | number;
- FormFileInput
  - Component
    -   id: isRequiredForA11y(
    PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
  ),
  - Types
    - id: string | number;
  - FormFileLabel
    - do we do the same thing here for htmlFor?

There seems to be a mix of how ID is defined between components as well. Do we have a set bet practice? I see things such as:
```
  id: isRequiredForA11y(
    PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
  ),
```
as well as
`id: requiredForA11y(PropTypes.string),`

kosmiq: Looking at FormCheck/Input/Label again. There it seems as if you are allowed to omit id and if you do controlId will be set to FormContext. Id is not forced. Perhaps the type description should be updated instead of forced?

`const { controlId, custom } = useContext(FormContext);`
and then
```
<Component
{...props}
ref={ref}
id={id || controlId}
className={classNames(
	className,
	bsPrefix,
	isValid && 'is-valid',
	isInvalid && 'is-invalid',
	isStatic && 'position-static',
)}
/>
```

src/FormFile.js
```diff
@@ -0,0 +1,186 @@
...
+  bsCustomPrefix: PropTypes.string,
+
+  /**
+   * The FormFile `ref` will be forwarded to the underlying input element,
+   * which means it will be a DOM node, when resolved.
+   *
+   * @type {ReactRef}
+   * @alias ref
+   */
+  _ref: PropTypes.any,
```
bpas247: This component is already forwarding its ref properly, so why do we need this as a part of its API?

types/components/FormFile.d.ts
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import FormFileInput from './FormFileInput';
+import FormFileLabel from './FormFileLabel';
+import { BsPrefixComponent } from './helpers';
+
+export interface FormFileProps {
+  bsCustomPrefix?: string;
+  innerRef?: React.LegacyRef<this>;
```
bpas247: We don't have a propType for this and this component is already forwarding its ref properly, and we probably don't need to support it so we can just remove this.

types/components/FormFileInput.d.ts
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
+
+import { BsPrefixComponent } from './helpers';
+
+export interface FormFileInputProps {
+  id?: string;
+  isValid?: boolean;
+  isInvalid?: boolean;
+  innerRef?: React.LegacyRef<this>;
```
bpas247: Same comment as above.

www/src/pages/components/forms.js
```diff
@@ -330,6 +334,75 @@ export default withLayout(function FormControlsSection({ data }) {
...
+
+      <h4>Translating or customizing the strings with HTML</h4>
+      <p>
+        Bootstrap also provides a way to translate the “Browse” text in HTML
+        with the <code>data-browse</code> attribute which can be added to the
+        custom input label (example in Dutch):
+      </p>
+      <Callout>
+        Note that the <code>data-browse</code> attribute does not to anything
+        unless the <code>custom</code> prop is set.
```
bpas247: we should display a warning if our end user ends up doing this.

www/src/pages/components/forms.js
```diff
@@ -330,6 +334,75 @@ export default withLayout(function FormControlsSection({ data }) {
...
+        <ul>
+          <li>
+            If you are not setting the <code>custom</code> prop the
+            <code>label</code> should be before the <code>input</code>.
+          </li>
+          <li>
+            If you are setting the custom prop the <code>input</code> element
+            has to be placed before the <code>label</code> or the{' '}
+            <code>buttonText</code> prop will not work.
+          </li>
```
bpas247: These sound like very specific rules that our end users can definitely get caught up in. Would it be better not to allow our end users to be able to set their own `FormFile` structure? I'm not sure how much customization they would be able to do here anyways.

types/components/FormFileLabel.d.ts
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
+import { BsPrefixComponent } from './helpers';
+
+export interface FormFileLabelProps {
+  htmlFor?: string;
+  innerRef?: React.LegacyRef<this>;
```
bpas247: Same comment as above.

www/src/pages/components/forms.js
kosmiq: Well I have a specific use-case for myself actually. In a project using react-bootstrap.

We are looking to use FormFile when it is available. We will be using isValid and isInvalid. However we will also be adding "is-warning" through usage of classes and some SCSS.

And since the valid/invalid messages are displayed through some Bootstrap CSS magic you can extend that with your own variants. You do this by adding more states in their $form-validation-stages (https://github.com/twbs/bootstrap/blob/v4.4.1/scss/_variables.scss#L670)

Making it look like this:
```
$form-validation-states: map-merge(
  (
    "valid": (
      "color": $form-feedback-valid-color,
      "icon": $form-feedback-icon-valid
    ),
    "invalid": (
      "color": $form-feedback-invalid-color,
      "icon": $form-feedback-icon-invalid
    ),
    "warning": (
      "color": $form-feedback-warning-color,
      "icon": $form-feedback-icon-warning
    ),
  ),
  $form-validation-states
);
```
This will allow you to do like this, and this will require being able to customize the FormFile markup. Of course the "is-warning" class is managed through a function to mimic isValid/isInvalid that exists on the component already.
```
<Form.File id="formcheck-api-custom" custom>
	<Form.File.Input className="is-warning" />
	<Form.File.Label data-browse="Button text">
		Custom file input
	</Form.File.Label>
	<Form.Control.Feedback className="warning-feedback">You did it!</Form.Control.Feedback>
</Form.File>
```

www/src/pages/components/forms.js
kosmiq: Yeah, had a warning for when it was buttonText. Just figured out how to do it properly for a data-attribute so it will be committed soon.

src/FormFile.js
kosmiq: Probably no reason at all. Left from FormCheck that I used as a basis for this. Removing since refs seems to be working without it.

types/components/FormFile.d.ts
kosmiq: Probably no reason at all. Left from FormCheck that I used as a basis for this. Removing since refs seems to be working without it.

types/components/FormFileInput.d.ts
kosmiq: Probably no reason at all. Left from FormCheck that I used as a basis for this. Removing since refs seems to be working without it.

types/components/FormFileLabel.d.ts
kosmiq: Probably no reason at all. Left from FormCheck that I used as a basis for this. Removing since refs seems to be working without it.

src/FormFileInput.js
kosmiq: @jquense I took another look again.

If, for any reason, the Id is not set the label will not work properly when custom is not used. So I guess we should always force the Id.

And I guess we should do it like this?
```
  id: isRequiredForA11y(
    PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
  ),
```

Plus update the types

src/FormFileLabel.js
kosmiq: @jquense are you okay with the new data-browse attribute and the overview of how it works above? Can we resolve this? :)

www/src/pages/components/forms.js
kosmiq: I have expanded the docs for Form.File in the API section to make this even clearer
Screenshot 2020-03-13 at 09 30 23

Do you think this will be enough @bpas247 ?

bpas247: Yeah that's probably fine then 😃 great work!

src/FormFile.js
taion: well, idk. well, we had to do:

```js
const HiddenFileInput = React.forwardRef((props, ref) => (
  <input
    ref={ref}
    {...props}
    type="file"
    value={
      // XXX: Need to specify a value of '' to get onChange to fire again when
      //  selecting the same file. This component isn't really controlled in
      //  any event.
      props.onChange ? '' : undefined
    }
    className="sr-only"
  />
));
```

so i can see people needing to do that for controlled file inputs, maybe

taion: but i do agree that it would make more sense as `inputAs`, with `as` here allowing customization of the wrapping `<div>`

www/src/pages/components/forms.js
```diff
@@ -330,6 +334,75 @@ export default withLayout(function FormControlsSection({ data }) {
...
+
+      <h4>Translating or customizing the strings with SCSS</h4>
+      <p>
+        Please refer to the official{' '}
+        <a href="https://getbootstrap.com/docs/4.4/components/forms/#translating-or-customizing-the-strings-with-scss">
+          Bootstrap documentation for translating via SCSS
+        </a>
+        . The <code>lang</code> prop can be used to pass the language.
+      </p>
+      <ReactPlayground codeText={FileButtonTextSCSS} />
```
taion: i think per how we do SCSS, this should be `FileButtonTextScss`, and the file should also be named that way

src/FormFileInput.js
taion: this looks right to me... especially given that the examples often show the file inputs in form groups

hmac2222:
Yay, this is great!!

PR #5043, opened 2020-03-13
feat: can customize BreadcrumbItem's inner link

### LINKED ISSUE #4472: Cannot customize the `a` inside `Breadcrumb.Item`

KoltesDigital:
As discussed with @taion in #5022, in addition to `linkAs`, the inner `span` which was added when inactive should be dropped. The tests have been adapted in consequence.

Fixes #4472. Supersedes #5022.

### DIFF
--- a/src/BreadcrumbItem.js
+++ b/src/BreadcrumbItem.js
@@ -20,6 +20,10 @@ const propTypes = {
    */
   href: PropTypes.string,
   /**
+   * You can use a custom element type for this component's inner link.
+   */
+  linkAs: PropTypes.elementType,
+  /**
    * `title` attribute for the inner `a` element
    */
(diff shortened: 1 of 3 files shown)


kosmiq:
Would you like to update the types as well?

KoltesDigital:
Sure, I missed that.

PR #5055, opened 2020-03-20
feat(BreadcrumbItem): pass props to li element

### LINKED ISSUE #5054: Cannot pass microdata attribute to Breadcrumb.Item
To add [schema microdata](https://schema.org/BreadcrumbList#itemlist-1) to breadcrumb, there is a option to pass custom attributes to `<ol>` via Breadcrumb & `listProps` but, there is no option to pass custom attributes to `<li>`  via Breadcrumb.Item

mxschmitt:
Closes #5054

### DIFF
--- a/src/BreadcrumbItem.js
+++ b/src/BreadcrumbItem.js
@@ -31,12 +31,17 @@ const propTypes = {
    * `target` attribute for the inner `a` element
    */
   target: PropTypes.string,
+  /**
+   * Additional props passed as-is to the underlying `<li>` element
+   */
+  listItemProps: PropTypes.object,
 
   as: PropTypes.elementType,
 };
 
 const defaultProps = {
   active: false,
+  listItemProps: {},
 };
 
 const BreadcrumbItem = React.forwardRef(
@@ -55,7 +60,7 @@ const BreadcrumbItem = React.forwardRef(
   ) => {
     const prefix = useBootstrapPrefix(bsPrefix, 'breadcrumb-item');
 
-    const { href, title, target, ...elementProps } = props;
+    const { href, title, target, listItemProps, ...elementProps } = props;
     const linkProps = { href, title, target };
 
     return (
@@ -63,6 +68,7 @@ const BreadcrumbItem = React.forwardRef(
         ref={ref}
         className={classNames(prefix, className, { active })}
         aria-current={active ? 'page' : undefined}
+        {...listItemProps}
       >
         {active ? children : <LinkComponent {...elementProps} {...linkProps} />}
       </Component>
--- a/src/BreadcrumbItem.js
+++ b/src/BreadcrumbItem.js
@@ -34,14 +34,14 @@ const propTypes = {
   /**
    * Additional props passed as-is to the underlying `<li>` element
    */
-  listItemProps: PropTypes.object,
+  linkProps: PropTypes.object,
 
   as: PropTypes.elementType,
 };
 
 const defaultProps = {
   active: false,
-  listItemProps: {},
+  linkProps: {},
 };
 
 const BreadcrumbItem = React.forwardRef(
@@ -54,23 +54,24 @@ const BreadcrumbItem = React.forwardRef(
       // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
       as: Component = 'li',
       linkAs: LinkComponent = SafeAnchor,
+      linkProps,
       ...props
     },
     ref,
   ) => {
     const prefix = useBootstrapPrefix(bsPrefix, 'breadcrumb-item');
 
-    const { href, title, target, listItemProps, ...elementProps } = props;
-    const linkProps = { href, title, target };
+    const { href, title, target, ...elementProps } = props;
+    const linkComponentProps = { href, title, target, ...linkProps };
 
     return (
       <Component
         ref={ref}
         className={classNames(prefix, className, { active })}
         aria-current={active ? 'page' : undefined}
-        {...listItemProps}
+        {...elementProps}
       >
-        {active ? children : <LinkComponent {...elementProps} {...linkProps} />}
+        {active ? children : <LinkComponent {...linkComponentProps} />}
       </Component>
     );
   },
--- a/src/BreadcrumbItem.js
+++ b/src/BreadcrumbItem.js
@@ -55,23 +55,34 @@ const BreadcrumbItem = React.forwardRef(
       as: Component = 'li',
       linkAs: LinkComponent = SafeAnchor,
       linkProps,
+      href,
+      title,
+      target,
       ...props
     },
     ref,
   ) => {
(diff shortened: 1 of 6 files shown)


www/yarn.lock
```diff
@@ -5137,17 +5137,10 @@ functional-red-black-tree@^1.0.1:
   resolved "https://registry.yarnpkg.com/functional-red-black-tree/-/functional-red-black-tree-1.0.1.tgz#1b0ab3bd553b2a0d6399d29c0e3ea0b252078327"
   integrity sha1-GwqzvVU7Kg1jmdKcDj6gslIHgyc=
 
-<<<<<<< HEAD
```
mxschmitt: there was a merge conflict merged by jquense

src/BreadcrumbItem.js
```diff
@@ -55,14 +60,15 @@ const BreadcrumbItem = React.forwardRef(
...
-    const { href, title, target, ...elementProps } = props;
+    const { href, title, target, listItemProps, ...elementProps } = props;
     const linkProps = { href, title, target };
 
     return (
       <Component
         ref={ref}
         className={classNames(prefix, className, { active })}
         aria-current={active ? 'page' : undefined}
+        {...listItemProps}
```
taion:
```suggestion
    const { href, title, target, ...props } = props;
    const linkProps = { href, title, target };

    return (
      <Component
        ref={ref}
        {...props}
        className={classNames(prefix, className, { active })}
        aria-current={active ? 'page' : undefined}
```

maybe we can just do this?

mxschmitt: would break backwards compatibility right? thats why I made this approach.

taion: we're in betas :D

post-https://github.com/react-bootstrap/react-bootstrap/pull/5043 i think it's a reasonable change to make...

mxschmitt: maybe we should release v1 soon, so that we have at least something for Bootstrap 4.

Then for Bootstrap 5 we can use v2 or instantly jump to v5 idk. (seems to be the user friendly approach for me)

taion: yeah... i guess my concern with matching bootstrap major versions is that it doesn't let us make potentially necessary breaking changes on our end.

part of the question might be how often upstream cuts breaking changes. if it will happen on a semi-regular cadence, then not too big a deal.

if it will be years and years, then we have a bit of a problem

src/BreadcrumbItem.js
```diff
@@ -31,12 +31,17 @@ const propTypes = {
    * `target` attribute for the inner `a` element
    */
   target: PropTypes.string,
+  /**
+   * Additional props passed as-is to the underlying `<li>` element
```
taion:
```suggestion
   * Additional props passed as-is to the underlying link for non-active items.
```

src/BreadcrumbItem.js
```diff
@@ -49,22 +54,24 @@ const BreadcrumbItem = React.forwardRef(
       // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
       as: Component = 'li',
       linkAs: LinkComponent = SafeAnchor,
+      linkProps,
```
taion:
```suggestion
      href,
      title,
      target,
      linkProps,
```

src/BreadcrumbItem.js
```diff
@@ -49,22 +54,24 @@ const BreadcrumbItem = React.forwardRef(
...
 
     return (
       <Component
         ref={ref}
         className={classNames(prefix, className, { active })}
         aria-current={active ? 'page' : undefined}
+        {...elementProps}
       >
-        {active ? children : <LinkComponent {...elementProps} {...linkProps} />}
+        {active ? children : <LinkComponent {...linkComponentProps} />}
```
taion:
```suggestion
        {active ? (
          children
        ) : (
          <LinkComponent
            {...linkProps}
            href={href}
            title={title}
            target={target}
          />
            {children}
          </LinkComponent>
        )}
```

taion: am i mis-reading or did we somehow drop `children` as an input to `linkProps`?

src/BreadcrumbItem.js
```diff
@@ -49,22 +54,24 @@ const BreadcrumbItem = React.forwardRef(
...
+      linkProps,
       ...props
     },
     ref,
   ) => {
     const prefix = useBootstrapPrefix(bsPrefix, 'breadcrumb-item');
 
     const { href, title, target, ...elementProps } = props;
-    const linkProps = { href, title, target };
+    const linkComponentProps = { href, title, target, ...linkProps };
```
taion: you can remove these definitions entirely

src/BreadcrumbItem.js
```diff
@@ -49,22 +54,24 @@ const BreadcrumbItem = React.forwardRef(
...
 
     return (
       <Component
         ref={ref}
         className={classNames(prefix, className, { active })}
         aria-current={active ? 'page' : undefined}
+        {...elementProps}
```
taion:
```suggestion
        {...props}
        className={classNames(prefix, className, { active })}
        aria-current={active ? 'page' : undefined}
```

src/BreadcrumbItem.js
taion: yeah, woof:
image

test/BreadcrumbItemSpec.js
```diff
@@ -137,10 +137,30 @@ describe('<Breadcrumb.Item>', () => {
...
+      .should.eq('bar');
+  });
+
+  it('Should be able to pass attributes to the li element', () => {
+    const instance = mount(<Breadcrumb.Item foo="bar">Crumb</Breadcrumb.Item>);
+    instance
+      .find('li')
+      .prop('foo')
+      .should.eq('bar');
+  });
```
bpas247: Could we re-write these tests to make assertions on actual use cases?

types/components/BreadcrumbItem.d.ts
```diff
@@ -8,6 +8,7 @@ export interface BreadcrumbItemProps {
   linkAs?: React.ElementType;
   target?: string;
   title?: React.ReactNode;
+  linkProps?: React.LinkHTMLAttributes<any>;
```
bpas247: Is there a better generic that we can use here other than `any`?

test/BreadcrumbItemSpec.js
mxschmitt: you mean dont use `should.eq`?
I adjusted the test names

mxschmitt:
Should be okay now. But unsure regarding the `id` handling. Before we applied the ID to the link element what we dont do anymore with this implementation. But we have linkProps for that, so idk.

types/components/BreadcrumbItem.d.ts
mxschmitt: should be okay now. thx

taion:
@jquense last blocker for v1 here

PR #5067, opened 2020-03-25
fix(types): more accurate form types

jquense:
Does this seem better? They types are fairly off right now b/c nothing is actually a class anymore.

### DIFF
--- a/types/simple.test.tsx
+++ b/types/simple.test.tsx
@@ -173,8 +173,9 @@ import {
     <Form.Control
       type="email"
       placeholder="name@example.com"
-      innerRef={React.createRef<HTMLInputElement>()}
-      onChange={(e: React.FormEvent<HTMLInputElement>) => {
+      ref={React.createRef<HTMLInputElement>()}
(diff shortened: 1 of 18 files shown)


types/simple.test.tsx
```diff
@@ -210,7 +211,7 @@ import {
     <Form.Control
       as="textarea"
       rows={3}
-      innerRef={React.createRef<HTMLTextAreaElement>()}
```
jquense: IDK why these were added innerRef isn't supported on these

kosmiq:
The changes themselves looks good to me. I do however wonder about the error the test throws.
```
    <Form.Control
      type="email"
      placeholder="name@example.com"
      ref={React.createRef<HTMLInputElement>()}
      onChange={(e) => {
        // $ExpectType React.FormEvent<FormControlElement>
        e;
      }}
    />
```

It is set to expect React.FormEvent but actually returns FormEvent. I see that @jquense  added the ExceptType line. Typo or do we get the wrong type back?

Issue #5057, opened 2020-03-21
Form.File is missing in v1.0.0-beta.17


wombat86:
## Describe the bug

`Form.File` is missing in both code and typings for package version `1.0.0-beta.17`. However, it is present in `master` branch and is mentioned in documentation for version `1.0.0-beta.17`.

## To Reproduce

1. Setup a React/TypeScript project
1. Install npm package `1.0.0-beta.17`
1. Add import: ```import Form from "react-bootstrap/Form";```
1. Create any react component and add ```<Form.File/>``` to TSX.
1. Compile
1. Observe compilation error `TS2339: Property 'File' does not exist on type 'typeof Form'.`

## Expected behavior

* At least the documentation should be updated accordingly to not confuse developers.
* Also, it would be nice to have this feature actually released since it's already in master branch and documentation.

kosmiq:
Where is it mentioned in the current docs? I skimmed through what's published and can't seem to find it?

It is, just like you noticed, coming in the next version. The pr is not that old though and I guess a few more features/Prs will be merged/finished before it comes around.

wombat86:
My bad, I didn't provide a URL. Here it is: https://react-bootstrap.netlify.com/components/forms/#forms-custom.
Providing the screenshot as well:
<img width="1919" alt="https___react-bootstrap_netlify_com_components_forms__forms-custom" src="https://user-images.githubusercontent.com/5948134/77224202-e8d27200-6b6b-11ea-832a-bc79af1b4170.png">

kosmiq:
Huh. I had to run incognito to see it. The docs were added in the same pr as the component itself.

@jquense are the docs automatically from master?

taion:
Netlify auto-deploys from master. It's the "live" docs page, though, and not the "official" one, which is at https://react-bootstrap.github.io/, and correctly omits `<Form.File>` in https://react-bootstrap.github.io/components/forms/.

We should probably add something there indicating that the docs are for `master` and not for the latest release.

wombat86:
@taion tbh I found this URL via Google, so you might get more developers like me who read that "unofficial" page. Perhaps you might want to restrict access to it for public users.

mxschmitt:
Maybe we should add a "if URL ends with `netlify.com` then show a banner at the top right with the text: "current master branch and not guaranteed to be published" or something like that.

altwarg:
[...]

PR #5078, opened 2020-03-28
Create master docs alert

kosmiq:
This PR adds a simple component that checks if the current window.location.host contains "netlify.com" and if so shows the following Alert (not sticky):

Screenshot 2020-03-28 at 01 03 40

I got no idea if it's running SSR or similar on Netlify so I added a check to check that window is not undefined.

Any improvements or better ideas how to do this is welcome. I took the fast and easy route here :)

This issue with the Netlify docs display features not yet in a release was raised in #5057.

### DIFF
--- a/www/src/components/MasterDocsAlert.js
+++ b/www/src/components/MasterDocsAlert.js
@@ -1,18 +1,24 @@
 import React from 'react';
 import Alert from 'react-bootstrap/Alert';
+import Container from 'react-bootstrap/Container';
+import Row from 'react-bootstrap/Row';
 
 function MasterDocsAlert() {
   return (
     typeof window !== 'undefined' &&
     window.location.host.substr(window.location.host.indexOf('.') + 1) ===
       'netlify.com' && (
-      <Alert variant="warning">
-        You are currently viewing the auto-generated docs from the
-        master-branch. The docs for the current release are available at{' '}
-        <Alert.Link href="https://react-bootstrap.github.io/">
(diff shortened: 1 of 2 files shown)


taion:
so... netlify as set up runs a custom build command of our choice. i think it's currently running `build-docs`, but we could have it run some new `build-docs:netlify` that does whatever it needs to bake in this banner

i think @jquense set up the netlify bit? i can't find it in my netlify thing

mxschmitt:
> so... netlify as set up runs a custom build command of our choice. i think it's currently running `build-docs`, but we could have it run some new `build-docs:netlify` that does whatever it needs to bake in this banner
> 
> i think @jquense set up the netlify bit? i can't find it in my netlify thing

Currently it runs `cd www && yarn && yarn build`. Only @jquense and me have access. So jquense could you add taion too? Thanks :)
(We've got the unlimited Open Source plan since a few month, so thats quite nice actually)

taion:
More helpful to add @kosmiq who's actually looking at this :p

jquense:
can add, @kosmiq i need an email to send an invite too

kosmiq:
@jquense I sent you an email :)

bpas247:
Is this still being worked on? I wouldn't mind stepping in and helping if that's the case 🙂

kosmiq:
@bpas247 feel free to. I'm so loaded at work I'm unlikely to pick up this again for the next 2-3 months.

bpas247:
superseded by #5318

PR #5080, opened 2020-03-30
Fix type error in overlay example

akx:
`Overlay.target` does not accept `null`; it does accept `undefined` though, which is the default for `useRef`.

### DIFF
--- a/www/src/examples/Overlays/Overlay.js
+++ b/www/src/examples/Overlays/Overlay.js
@@ -4,7 +4,7 @@ function Example() {
 
   return (
     <>
-      <Button variant="danger" ref={target} onClick={() => setShow(!show)}>
+      <Button variant="danger" ref={target || undefined} onClick={() => setShow(!show)}>
         Click me to see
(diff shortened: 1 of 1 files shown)


taion:
That's not what that code is doing. `target` is the ref, not the value of the ref. It's always defined.

akx:
@taion I beg to differ... In Typescript, the example, copied verbatim, leads to a type error since https://github.com/react-bootstrap/react-bootstrap/blob/d26573de99f412d944a4b8cadc1a2740ba3cacaa/types/components/Overlay.d.ts#L42 does not accept `null`.

Screenshot 2020-03-31 at 10 29 12

taion:
No, I mean, your code literally doesn't do anything. `target.current` can be falsy, but `target` itself can't.

akx:
https://github.com/react-bootstrap/react-bootstrap/commit/02484efb372450f3ba3c6c8dcb4e41e281d50175 apparently addressed this as a side effect...

PR #5098, opened 2020-04-07
export the UMD global.

heroboy:


taion:
is this quite appropriate? we don't want to declare this global for users of CJS, say, for whom that global will not exist

heroboy:
The declared global is not visible to the users of CJS. Unless they set the https://www.typescriptlang.org/v2/en/tsconfig#allowUmdGlobalAccess to true.

PR #5104, opened 2020-04-08
Solving issue 4911

### LINKED ISSUE #4911: Inactive <ListGroup.Item> nested in active <ListGroup.Item> is invisible
Nesting `<ListGroup>`s inside of a `<ListGroup.Item>` isn't supported if the outer item is `active` and the inner item is not.

Screen Shot 2019-12-26 at 1 24 58 PM
Screen Shot 2019-12-26 at 1 25 04 PM

harsh6646:
Closes #4911


taion:
This seems like an odd approach. Surely we could just reset the context as necessary here?

harsh6646:
Hey, what do you mean by that? What kind of approach would you recommend?
The issue was that the `active` prop doesn't cascade to child components so basically we fixed that by checking the children in the parent and setting the props for them as needed.

taion:
isn't the issue that the inner items becomes active?

KostyaIliouk:
The wording of the issue is a little odd, but from what we found is that if there is any ListGroup.Item component within a ListGroup.Item parent that is set to active, the inner child renders with white text. The issue creator was interested in having any children to also be rendered as active, thus we had to add the active prop to any children ListGroup.Item found within this parent.

kosmiq:
See my comment in #4911. This is an TWBS issue.

taion:
Per https://github.com/react-bootstrap/react-bootstrap/issues/4911#issuecomment-613049083, I don't think we want to proceed with this kind of change.

If you want to do this in user space, I'd recommending adding your own component that wraps `<ListGroup.Item>` and both provides and consumes a context value with the current "active" setting, and uses that to enable "active" in children of active list group items.

PR #5112, opened 2020-04-14
fix: consider margins in Popper calculations

jquense:
This doesn't address the warning but it does fix the issue. Not quite sure if it's a good idea, but seems fine?


src/Overlay.js
```diff
@@ -120,12 +130,73 @@ function wrapRefs(props, arrowProps) {
...
+      };
+    },
+    fn: ({ state, name }) => {
+      const { popper: popperRect } = state.rects;
+      const { popper, arrow } = state.elements;
+
+      const popperMargins = getMargins(popper);
+      const arrowMargins = arrow && getMargins(arrow);
+
+      state.rects.popper = {
```
jquense, line 150: this needs to be set very early since all the other modifiers use it for calculations

src/Overlay.js
```diff
@@ -120,12 +130,73 @@ function wrapRefs(props, arrowProps) {
...
+    requiresIfExists: ['arrow'],
+    fn({ state }) {
+      if (!state.elements.arrow) return;
+
+      const { arrow: arrowMargin } = state.modifiersData.marginOffsets;
+
+      const offsets = state.modifiersData.arrow;
+
+      if (offsets.x != null) offsets.x -= arrowMargin.left;
+      if (offsets.y != null) offsets.y -= arrowMargin.top;
```
jquense, line 177: this is not strictly correct because it assumes margin on either side is the same, but that is true for bootstrap. I think anyone modifying styles should do it correctly via popper

jquense:
We should consider scoping this to _just_ bootstrap components, maybe but guarding based on elements having the base bootstrap classes?

kosmiq:
I'd be fine with a solution like this. The warning is a warning and if this solves any issues I don't see any reason not to go this route. I'll check the pr a bit closer later.

src/Overlay.js
kosmiq: Should we add anything to the docs about this?

src/Overlay.js
kosmiq: Yup, very early it is. Tested the other phases as well but no go on those.

src/Overlay.js
```diff
@@ -120,12 +130,73 @@ function wrapRefs(props, arrowProps) {
...
+const marginModifiers = [
+  {
+    name: 'marginOffsets',
+    enabled: true,
+    phase: 'beforeRead',
+    effect: ({ state }) => {
+      return () => {
+        state.elements.reference.removeAttribute('aria-describedby');
+      };
+    },
```
taion, line 142: is this replacing something else?

src/Overlay.js
taion: might want to add a comment here, at least

jquense:
Ok different approach. This jumps in on the refs before popper runs and stores the margins for use in the offsetModifier. 

I realized tho that we do need to include the modifiers all upstream because of the common js issue with cherry-picking them that required bundling in RO

kosmiq:
@jquense I get an error on the disabled elements in the docs when running locally (http://localhost:8001/components/overlays/#overlays-disabled)

Other than that everything seems to work and having the expected placements of everything for dropdowns, popovers etc.

Screenshot 2020-04-20 at 21 41 35

src/usePopperMarginModifiers.js
```diff
@@ -0,0 +1,57 @@
...
+        return;
+
+      margins.current = getMargins(overlay);
+      overlay.style.margin = 0;
+      overlayRef.current = overlay;
+    }, []),
+    [
+      useMemo(() => {
+        return {
+          name: 'offset',
```
jquense, line 33: this is already included upstream, maybe i added it?

src/usePopperMarginModifiers.js
```diff
@@ -0,0 +1,57 @@
...
+
+export default function usePopperMarginModifiers() {
+  const overlayRef = useRef(null);
+  const margins = useRef(null);
+
+  return [
+    useCallback((overlay) => {
+      if (
+        !overlay ||
+        !(hasClass(overlay, 'popover') || hasClass(overlay, 'dropdown-menu'))
```
taion, line 22: shouldn't this be the bsClass from the component itself?

src/usePopperMarginModifiers.js
```diff
@@ -0,0 +1,57 @@
...
+      )
+        return;
+
+      margins.current = getMargins(overlay);
+      overlay.style.margin = 0;
+      overlayRef.current = overlay;
```
taion, line 28: is layout effect too late? it seems like it might be nice to remove the ref at the end... though i guess it doesn't matter?

maybe leave a comment on why we're doing this, at least, somewhere in this module

Issue #5111, opened 2020-04-14
Tooltips with position auto are placed on top of the element


kosmiq:
## Describe the bug

It seems as if tooltips that uses `placement="auto"` are no longer working since popperjs2 upgrade in react-overlays.

There seems to be other issues as well with placements of tooltips. Look at the linked example on the buttons and the headers. :(

## To Reproduce

1. Create a button
2. Wrap it in an overlaytrigger with placement auto
3. Tooltip is rendered inside the button

## Reproducible Example

Example: https://codesandbox.io/s/bold-chaum-4xhmr?file=/src/App.js

## Expected behavior

That tooltips works as in beta 17.

## Screenshots

v1.0.0
![Screenshot 2020-04-14 at 11 50 29](https://user-images.githubusercontent.com/322759/79211205-39c14780-7e46-11ea-89e5-394ac39e17fc.png)
![Screenshot 2020-04-14 at 11 50 34](https://user-images.githubusercontent.com/322759/79211209-3c23a180-7e46-11ea-95e6-662375805b55.png)

beta17
![Screenshot 2020-04-14 at 11 50 47](https://user-images.githubusercontent.com/322759/79211246-4cd41780-7e46-11ea-8d3e-4085f37aa960.png)
![Screenshot 2020-04-14 at 11 50 49](https://user-images.githubusercontent.com/322759/79211254-4fcf0800-7e46-11ea-8e47-c54ad31c2cd3.png)

kosmiq:
@jquense @taion should we revert to PopperJS 1.x?

jquense:
This is less a bug and more a configuration difference i think. the `auto` placement _requires_ the `flip` modifier to be enabled. maybe we should should automatically default it to true when auto placements are picked

kosmiq:
That would probably be a good idea. We should otherwise mention in the docs that `flip` is required when using `placement=auto`

PR #5113, opened 2020-04-14
#5111 Set flip to true if placement is auto in OverlayTrigger.js

kosmiq:
Sets flip to true if placement is auto. Discussed in #5111. As an added bonus the possible placements are now displayed on the OverlayTrigger API in the docs.

### DIFF
--- a/src/OverlayTrigger.js
+++ b/src/OverlayTrigger.js
@@ -92,6 +92,27 @@ const propTypes = {
    * @private
    */
   show: PropTypes.oneOf([null]),
+
+  /**
+   * The placement of the Overlay in relation to it's `target`.
+   */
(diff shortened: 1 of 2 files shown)


src/OverlayTrigger.js
```diff
@@ -106,6 +127,8 @@ function OverlayTrigger({
   popperConfig = {},
   defaultShow,
   delay: propsDelay,
+  placement,
+  flip = placement === 'auto',
```
jquense, line 131: `placement.indexOf('auto') !== -1` probably?

kosmiq: You are correct, totally forgot about auto-start and auto-end... :(

kosmiq:
@taion added test

PR #5125, opened 2020-04-20
Add Button `as` prop to typescript definition

abeMedia:
The `as` prop was missing from typescript definition.


taion:
That's not quite how the typing is set up to work. `as` is added by the generic template class below, that also handles props from the `as` class.

PR #5129, opened 2020-04-21
Clarifying Dismissible Alerts

TurekBot:
When I found myself confused about dismissible alerts, I ran into a bunch of other people who also seemed confused.

That spurred me to improve the documentation for the Alert component, clarifying that in order to get an Alert to dismiss, `onClose` needs to be implemented.

Feel free to offer suggestions.

Thanks!


www/src/examples/Alert/BasicDismissible.js
```diff
@@ -1 +1,36 @@
...
+                    {this.props.children}
+                </Alert>
+            );
+        }
+        return null;
+    }
+}
+
+
+export default DismissibleAlert;
```
TurekBot: This solution requires everyone to add this class to their project. Would it be acceptable to have this class added to React Bootstrap to save everyone an extra few steps on a very common use of Alerts?

www/src/examples/Alert/BasicDismissible.js
```diff
@@ -0,0 +1,36 @@
...
+
+/*
+ * Add this component to your project, if you just need a simply dismissible Alert.
+ * Usage:
+ *   <DismissibleAlert variant="primary">
+ *      You're just going to dismiss me, aren't you.
+ *   </DismissibleAlert>
+ */
+class DismissibleAlert extends React.Component {
+    constructor(props) {
```
mxschmitt, line 12: Would be awesome if you could rewrite it using React Hooks.

www/src/pages/components/alerts.mdx
```diff
@@ -47,12 +47,16 @@ Alerts can contain whatever content you like. Headers, paragraphs, dividers, go
 
 ### Dismissing
 
-Add the `dismissible` prop to add a functioning dismiss
-button to the Alert.
+Add the `dismissible` prop to add a dismiss button to the Alert.
 
-<ReactPlayground codeText={Dismissible} />
```
mxschmitt, line 53: Seems like we already have an example for that, let's eather keep it or enhance it since that would end up in an unused file.

PR #5132, opened 2020-04-22
Refactor modal

jquense:


src/Modal.js
```diff
@@ -337,83 +231,215 @@ class Modal extends React.Component {
...
+    const [modalStyle, setStyle] = useState({});
+    const waitingForMouseUpRef = useRef(false);
+    const ignoreBackdropClickRef = useRef(false);
+
+    const [modal, setModalRef] = useCallbackRef();
+    const handleHide = useEventCallback(onHide);
+
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'modal');
+
+    useImperativeHandle(
```
taion, line 252: should this expose anything else? like a DOM node or whatever?

jquense: i don't think so not sure what help that would be, we can revisit tho if folks complain

Issue #5148, opened 2020-04-27
All-in TypeScript?


akx:
## Is your feature request related to a problem? Please describe

The TypeScript types being not-quite-perfectly-in-sync with the JavaScript implementation is starting to be a little frustrating.

Could we just go all-in and convert the source to TypeScript? I think the strictness benefits would be worth it.

Downstream, there'd still be the esm/plain-JS modules, of course, for plain-JS consumers.

## Describe the solution you'd like

* Convert all .js to .ts
* Inline the types
* Set up the build pipeline(s) to do the Right Things regarding packaging
* Fix All The Bugs! 😄 

## Additional context

If the project leadership is amenable to this, I'm willing to do at least initial legwork to get things going. I'd rather not, though, if such a PR would never pass.

jquense:
I'm up for it. I feel fairly confident we could type the weird `as` props correctly now (well as good as is possible anyway). What do others think?

Also any setup for TS should be consistent with what we did in react-overlays

mxschmitt:
Thats definitely true.
Also thought about that, should be feasible but requires a lot of typing effort. Maybe we can do it in several iterations to just start with allowJS and then go component by component to have a strict type required codebase in the end.

taion:
We should probably add the TS downlevel stuff?

jquense:
isn't that a compilation target thing vs a type thing? Or does it also downlevel the types for iterators and what not?

kosmiq:
This would be awesome IMO.

taion:
well, whatever we're doing for react-overlays is not quite right: see e.g. the use of `Omit` in https://unpkg.com/browse/react-overlays@3.1.3/esm/Overlay.d.ts

akx:
Well, I'll take these comments as a green light to go ahead and try TypeScript conversion :)

akx:
Hmm... What do you think, should PropTypes be left in though? They could be useful for JS developers, but will be extra headache to keep in sync with TS interfaces.

akx:
Okay, now that #5150 was merged into the `typescript` branch, what's the plan ahead? I honestly think we should just merge it into master and deal with the breakage; otherwise it's just not going to get done, I'm afraid.

cc @jquense

jquense:
As much as it's tempting to just run with it we do need to move carefully RB is used by many many people and we need to do our best to try not to break them, or limit our ability to release bug fixes while getting there. 

[...]

PR #5150, opened 2020-04-27
TypeScript all the things

akx:
To follow up on #5148, this PR starts to convert `src/` to TypeScript.

I grabbed the TS/eslint configuration from react-overlays, loosened it up a little for development purposes. A little Python script (enclosed) merged .js and .d.ts files into .tsx files, and I then began working in the dim light of `while true; yarn tsc --noEmit 2>&1 >tmp.log; clear; head -n50 tmp.log; sleep 5; end`...

# 23:56, 27 April

* some helper functions and such, used by multiple components, have either been strictly typed or any-typed to make things easier for the time being
* a small part of components pass `tsc` checks,
  * ⚠️ **but** the `as="element"` polymorphism is very likely lost on them right now. I'm having a somewhat hard time wrapping my head around the typings required to make that work in a strict enough way; it'll likely require some deeper type magic than I'm accustomed to so far! (Any ideas? 🤔) 
  * in the same vein, assigning `propTypes` onto these components requires an ugly `as any` thing 🤢 – I expect there to be no `as any`s when we're collectively finished with this...

Anyway! It's a start, and judging by stricter typing finding us #5149, I think it's worth it 😄 

# 22:59, 5 May

Okay, the test suites pass now!

* There are lots of `as any`s, `as unknown as X`s and some `// @ts-ignore`s. 
* There's a handful of TODO comments, mostly for types and such that weren't quite documented.
* I haven't checked whether all `propTypes` declarations match all `*Props` interfaces.
* The `as` polymorphism is now implemented with a `type ...`; I'm not sure if that's the right way, but it seems to work and type-checks alright.
* build seems to result in something sane, but I'm not sure if it's equivalent to what was there before.
* The typing for `Card.Body` etc. feels evil (it's where I couldn't think of any other ways to do things but that `as unknown as X` thing).
* Components created with `createWithBsPrefix` are utterly poorly typed. :(
* The generated .d.ts files look pretty horrible, since they include all of react's builtin HTML element names...


jquense:
the `as` typing is pretty tricky. The pattern i've found that works best is to not try and type `as` as part of the component definition, but declare it separately and cast the component to it. The Form components show the right pattern for that. Something like:

```ts
export interface FormControlProps {
  ...
}

declare interface FormControl
  extends BsPrefixRefForwardingComponent<'input', FormControlProps> {
  Feedback: typeof Feedback;
}

const RBFormControl: FormControl = React.forwardRef((props: FormControlProps, ref: Ref) => { 

  return (...)
})
```

What happens is that _inside_ the FormControl Props is _just_ the defined FormControlProps, which is what you want, you can only rely on props you know exist. But a user of FormControl is getting the correct type that will respect their `as` prop

akx:
> the `as` typing is pretty tricky. The pattern i've found that works best is to not try and type `as` as part of the component definition, but declare it separately and cast the component to it. The Form components show the right pattern for that. 

Cheers for the pointer. I'll look into that :)

akx:
@jquense et al.: any idea why Travis isn't building anymore?

src/BootstrapModalManager.tsx
```diff
@@ -10,21 +10,28 @@ const Selector = {
...
+  private adjustAndStore<T extends keyof CSSStyleDeclaration>(
+    prop: T,
+    element: HTMLElement,
+    adjust: number,
+  ) {
     const actual = element.style[prop];
     element.dataset[prop] = actual;
     css(element, {
-      [prop]: `${parseFloat(css(element, prop)) + adjust}px`,
+      [prop]: `${parseFloat(css(element, prop as any)) + adjust}px`,
```
jquense, line 21: the `T` isn't doing a much here if we have to cast to any right below yeah?

src/BootstrapModalManager.tsx
```diff
@@ -36,10 +43,10 @@ export default class BootstrapModalManager extends ModalManager {
       this.adjustAndStore('paddingRight', el, size),
     );
     qsa(container, Selector.STICKY_CONTENT).forEach((el) =>
-      this.adjustAndStore('margingRight', el, -size),
```
jquense, line 39: o wow this is just straight up a bug ooops

src/Carousel.tsx
```diff
@@ -202,25 +242,27 @@ const Carousel = React.forwardRef((uncontrolledProps, ref) => {
...
   const prefix = useBootstrapPrefix(bsPrefix, 'carousel');
 
-  const nextDirectionRef = useRef(null);
+  const nextDirectionRef = useRef<string | null>(null);
   const [direction, setDirection] = useState('next');
 
   const [isSliding, setIsSliding] = useState(false);
-  const [renderedActiveIndex, setRenderedActiveIndex] = useState(activeIndex);
+  const [renderedActiveIndex, setRenderedActiveIndex] = useState<number>(
+    activeIndex || 0,
```
jquense, line 250: a little concerned with the coercing to `0` having unintended consequences

src/Collapse.tsx
```diff
@@ -206,23 +226,25 @@ class Collapse extends React.Component {
         onExit={handleExit}
         onExiting={handleExiting}
       >
-        {(state, innerProps) =>
-          React.cloneElement(children, {
+        {(state, innerProps) => {
+          return React.cloneElement(children as any, {
```
jquense, line 230: can children be typed as `React.ReactElement` to avoid this?

src/DropdownButton.tsx
```diff
@@ -2,6 +2,31 @@ import React from 'react';
...
 
 import Dropdown from './Dropdown';
+import DropdownToggle from './DropdownToggle';
+import { BsPrefixRefForwardingComponent } from './helpers';
+import DropdownMenu from './DropdownMenu';
+
+type PropsFromToggle = Partial<
+  Pick<
+    React.ComponentPropsWithRef<typeof DropdownToggle>,
+    'href' | 'size' | 'variant' | 'disabled'
```
jquense, line 12: maybe we can just export this subset from DropdownToggle?

src/DropdownMenu.tsx
```diff
@@ -109,22 +132,20 @@ const DropdownMenu = React.forwardRef(
 
     // For custom components provide additional, non-DOM, props;
     if (typeof Component !== 'string') {
-      menuProps.show = show;
-      menuProps.close = close;
-      menuProps.alignRight = alignEnd;
+      (menuProps as any).show = show;
+      (menuProps as any).close = close;
+      (menuProps as any).alignRight = alignEnd;
```
jquense, line 137: I think we need to cast this to a more correct type vs casting to any below

src/Feedback.tsx
```diff
@@ -13,12 +22,15 @@ const propTypes = {
 };
 
 const defaultProps = {
-  type: 'valid',
+  type: 'valid' as 'valid',
```
jquense, line 25: `as const`

bpas247:
@jquense Looks like the last build seems to be failing due to the [prettier step](https://travis-ci.com/github/react-bootstrap/react-bootstrap/jobs/324017082), although I'm not sure why it's stopped kicking off for those new commits after (also strange that GitHub actions isn't kicking off _at all_ here).

bpas247:
Update: This might be due to how many times the git history has been re-written for this PR (especially considering the last build references a commit that doesn't exist anymore), I did notice that there is a few merge conflicts so hopefully the CI builds will kick off again with a merge commit

src/BootstrapModalManager.tsx
akx: It kind of is... `T` allows us to read `element.style` safely, but `css(...)` uses a different typing, so... maybe some sort of intersection type between `keyof CSSStyleDeclaration` and `keyof CSS.Properties`?

src/BootstrapModalManager.tsx
akx: Yeah, heh, was fixed separately in #5149 :)

src/Carousel.tsx
akx: I guess, yeah, but there are places doing arithmetic like `renderedActiveIndex - 1` without checking whether it is might happen to be `undefined`... What do you think, fix those to assume things or..?

src/Collapse.tsx
akx: Not without refactoring Collapse into an FC. ClassComponents' `props` are typed like
```
readonly props: Readonly<P> & Readonly<{ children?: ReactNode }>;
```

src/DropdownButton.tsx
akx: Yeah, or rather `PropsFromToggle` altogether – this is used in multiple places, it looks like :)

src/DropdownMenu.tsx
akx: Yeah, I suppose, but how do we know the props of the custom `Component`? 🤔

akx:
@bpas247 Rebasing on origin/master seemed to do the trick. :)

merge_ts.py
```diff
@@ -0,0 +1,64 @@
+import glob
```
mxschmitt, line 1: do we still need this script?

package.json
```diff
@@ -35,7 +34,8 @@
   },
   "lint-staged": {
     "*.js": "eslint --fix",
-    "types/**/*.ts": "prettier --write"
+    "*.ts": "eslint --fix",
```
mxschmitt, line 37:
```suggestion
    "*.{js,ts,tsx}": "eslint --fix",
```
then you can get rid of the others

src/DropdownMenu.tsx
jquense: we don't, but that's not important here, the Component _must_ accept these props, e.g. the interface for whatever Component is, is "accepts at least these props". That said i think that's less important than just correctly typing menuProps so that it's safe to use _here_ not just so Component knows it.

src/DropdownMenu.tsx
```diff
@@ -1,12 +1,32 @@
...
+  show?: boolean;
+  renderOnMount?: boolean;
+  flip?: boolean;
+  alignRight?: boolean;
+  onSelect?: SelectCallback;
+  rootCloseEvent?: 'click' | 'mousedown';
+  popperConfig?: { modifiers?: any };
+}
+
+type DropdownMenu = BsPrefixRefForwardingComponent<'a', DropdownMenuProps>;
```
jquense, line 29:
```suggestion
type DropdownMenu = BsPrefixRefForwardingComponent<'div', DropdownMenuProps>;
```

src/DropdownItem.tsx
```diff
@@ -3,10 +3,26 @@ import PropTypes from 'prop-types';
...
+export interface DropdownItemProps extends BsPrefixPropsWithChildren {
+  active?: boolean;
+  disabled?: boolean;
+  eventKey?: string;
+  href?: string;
+  onClick?: React.MouseEventHandler<this>;
+  onSelect?: SelectCallback;
+}
+
+type DropdownItem = BsPrefixRefForwardingComponent<'a', DropdownItemProps>;
```
jquense, line 25:
```suggestion
type DropdownItem = BsPrefixRefForwardingComponent<typeof SafeAnchor, DropdownItemProps>;
```

src/DropdownButton.tsx
```diff
@@ -2,6 +2,24 @@ import React from 'react';
...
+  extends React.PropsWithChildren<PropsFromToggle> {
+  id?: string;
+  title: React.ReactNode;
+  menuRole?: string;
+  renderMenuOnMount?: boolean;
+  rootCloseEvent?: 'click' | 'mousedown';
+  bsPrefix?: string;
+}
+
+type DropdownButton = BsPrefixRefForwardingComponent<
```
jquense, line 19: this isn't relevant here since it doesn't accept `as`, the name of this helper is a bit misleading

src/DropdownItem.tsx
```diff
@@ -93,6 +109,7 @@ const DropdownItem = React.forwardRef(
     });
 
     return (
+      // @ts-ignore
```
jquense, line 112: should add a note why this is here

src/DropdownToggle.tsx
```diff
@@ -4,9 +4,30 @@ import isRequiredForA11y from 'prop-types-extra/lib/isRequiredForA11y';
...
+  BsPrefixPropsWithChildren,
+  BsPrefixRefForwardingComponent,
+} from './helpers';
+
+export interface DropdownToggleProps
+  extends BsPrefixPropsWithChildren,
+    ButtonProps {
+  split?: boolean;
+  childBsPrefix?: string;
+  eventKey?: null; // TODO: Used by NavDropdown?
```
jquense, line 20: why is this `null`?

src/Figure.tsx
```diff
@@ -3,9 +3,14 @@ import createWithBsPrefix from './createWithBsPrefix';
...
-const Figure = createWithBsPrefix('figure', {
+type Figure = React.ComponentType & {
+  Image: typeof FigureImage;
+  Caption: typeof FigureCaption;
+};
+
+const Figure: Figure = (createWithBsPrefix('figure', {
   Component: 'figure',
-});
+}) as unknown) as Figure;
```
jquense, line 13: usually we deal with this by using `Object.assign` which handles adjusting the  type

src/Feedback.tsx
```diff
@@ -1,6 +1,15 @@
...
 import PropTypes from 'prop-types';
+import { BsPrefixProps, BsPrefixRefForwardingComponent } from './helpers';
+
+export interface FeedbackProps extends BsPrefixProps {
+  className?: string;
+  bsPrefix?: never;
+  type?: 'valid' | 'invalid';
+}
+
+type Feedback = BsPrefixRefForwardingComponent<'div', FeedbackProps>;
```
taion, line 12: **nitpick:** it's a little confusing, maybe, to have two things e.g. both named `Feedback` here? maybe we can just inline the type def?

src/NavbarCollapse.tsx
```diff
@@ -1,23 +1,34 @@
...
+
+export interface NavbarCollapseProps
+  extends CollapseProps,
+    React.HTMLAttributes<HTMLDivElement>,
+    BsPrefixProps {}
+
+// type NavbarCollapse = BsPrefixRefForwardingComponent<
+//   'div',
+//   NavbarCollapseProps
+// >;
```
taion, line 17: **nitpick:** unused? remove?

src/Switch.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+type Switch = BsPrefixRefForwardingComponent<FormCheck, SwitchProps> & {
+  Input: typeof FormCheck.Input;
+  Label: typeof FormCheck.Label;
+};
+
+// declare interface Switch
+//   extends BsPrefixRefForwardingComponent<FormCheck, SwitchProps> {}
+//
+// declare const Switch: Switch;
```
taion, line 15: ?

src/AbstractNav.tsx
```diff
@@ -58,15 +74,17 @@ const AbstractNav = React.forwardRef(
       getControllerId = tabContext.getControllerId;
     }
 
-    const listNode = useRef(null);
+    const listNode = useRef<HTMLElement>();
```
bpas247, line 77: Why are we changing the initial value here?

src/Accordion.tsx
```diff
@@ -36,25 +54,23 @@ const Accordion = React.forwardRef((props, ref) => {
...
-          className={classNames(className, bsPrefix)}
+          className={classNames(className, useBootstrapPrefix(bsPrefix, 'accordion'))}
         >
           {children}
         </Component>
       </SelectableContext.Provider>
     </AccordionContext.Provider>
   );
-});
+}) as unknown) as Accordion;
```
bpas247, line 70: Is there a better way of typing this?

src/AccordionContext.tsx
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export default React.createContext<string | null>(null);
```
bpas247, line 3: nit: _technically_ this file doesn't have any JSX in it, so I'm not sure if it should have the `.tsx` extension.

src/Accordion.tsx
```diff
@@ -36,25 +54,23 @@ const Accordion = React.forwardRef((props, ref) => {
...
   return (
-    <AccordionContext.Provider value={activeKey}>
-      <SelectableContext.Provider value={onSelect}>
+    <AccordionContext.Provider value={activeKey || null}>
+      <SelectableContext.Provider value={onSelect || null}>
         <Component
           ref={ref}
           {...controlledProps}
-          className={classNames(className, bsPrefix)}
+          className={classNames(className, useBootstrapPrefix(bsPrefix, 'accordion'))}
```
bpas247, line 63: Should we be calling hooks inside the return statement of the component? I'm not sure if this is safe to do.

mxschmitt:
Hey @akx, would be awesome if you could go over the comments if you have time, because once this is merged, we can start with making it Bootstrap 5 compatible.

akx:
Okay, rebased on origin/master again and upgraded some dependencies required, addressed most if not all review comments...

@mxschmitt: Thank you! I also agree this should be merged sooner rather than later. There will likely be some breakage for both TS and JS consumers since the typing likely isn't 100% watertight, but it won't get fixed without merging this... Also, I think the Gatsby site build is definitely broken right now, and I don't quite have enough Gatsby-fu to look into it.

@jquense: Yeah, there actually is a `BsPrefixAndClassNameOnlyProps` type now, but that's one unwieldy name! :) 

@taion: I was actually surprised TS allows you to name a variable the same as a type, but apparently so! Those confusions could be fixed up in a later PR, yeah?

@jquense, @bpas247: Re the `as unknown as any` hacks, I think those could be replaced with Object.assigns in a subsequent PR.

jquense:
I think we should merge this into a typescript branch here, for followups and so @akx doesn't have to keep it up to date with master himself. I'm concerned that pushing this right to master will leave us in a bad spot releasing, when i'd like to test this out and do a prerelease for folks to try before we push it out to latest, given the general risk involved. also we need to fix the docs

jquense:
I changed the base!

akx:
> I changed the base! we can merge once the yarn.lock conflict is fixed

@jquense Rebased & fixed conflicts!

jquense:
awesome work @akx Thanks a ton!

akx:
By the way, I happened upon this TS implementation of ref forwarding: https://github.com/reach/reach-ui/blob/52d721cd643d1a6d24c7253ab24e114be262d092/packages/utils/src/index.tsx#L193-L217

Not sure if it's safer or less so than ours though :)

PR #5165, opened 2020-05-06
feat(Modal): Add modal static backdrop animation

### LINKED ISSUE #5163: Missing modal static backdrop animation
## Describe the bug

There's an animation missing from a modal with static backdrop when trying to click on the backdrop or by pressing escape when `keyboard` is set to `false`.

See Bootstrap docs for a working example of the animation.
https://getbootstrap.com/docs/4.4/components/modal/#static-backdrop

## To Reproduce

1. Create a modal with `backdrop` set to `static`
2. Click the backdrop or press escape
3. Notice there is no animation

## Reproducible Example

https://codesandbox.io/s/cranky-bas-9m1zh?file=/src/App.js

## Expected behavior

Modal should animate.

- Operating System: Windows
- Browser, Version: Chrome 81.0.4044.129
- React-Bootstrap Version: 1.0.1

kyletsang:
Resolves #5163 

This adds the animation using the bootstrap class `modal-static`. The animation will appear on static backdrop modals when clicking the backdrop or pressing escape when `keyboard` is set to `false`.

### DIFF
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -441,7 +441,7 @@ const Modal = React.forwardRef(
         className={classNames(
           className,
           bsPrefix,
-          animateStaticModal && 'modal-static',
+          animateStaticModal && `${bsPrefix}-static`,
         )}
         onClick={
           backdrop === true || backdrop === 'static' ? handleClick : undefined
--- a/src/Modal.js
+++ b/src/Modal.js
@@ -4,6 +4,8 @@ import canUseDOM from 'dom-helpers/canUseDOM';
 import ownerDocument from 'dom-helpers/ownerDocument';
 import removeEventListener from 'dom-helpers/removeEventListener';
 import getScrollbarSize from 'dom-helpers/scrollbarSize';
+import transitionEnd from 'dom-helpers/transitionEnd';
+import listen from 'dom-helpers/listen';
 import PropTypes from 'prop-types';
 import React, {
   useCallback,
@@ -241,8 +243,11 @@ const Modal = React.forwardRef(
     ref,
   ) => {
     const [modalStyle, setStyle] = useState({});
+    const [animateStaticModal, setAnimateStaticModal] = useState(false);
     const waitingForMouseUpRef = useRef(false);
     const ignoreBackdropClickRef = useRef(false);
+    const removeKeydownListenerRef = useRef(null);
+    const removeStaticModalAnimationRef = useRef(null);
 
     const [modal, setModalRef] = useCallbackRef();
     const handleHide = useEventCallback(onHide);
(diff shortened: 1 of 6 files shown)


kyletsang:
Strange. All current tests succeed on my machine.

src/Modal.js
```diff
@@ -379,8 +438,14 @@ const Modal = React.forwardRef(
         role="dialog"
         {...dialogProps}
         style={baseModalStyle}
-        className={classNames(className, bsPrefix)}
-        onClick={backdrop === true ? handleClick : undefined}
+        className={classNames(
+          className,
+          bsPrefix,
+          animateStaticModal && 'modal-static',
```
taion, line 444: **suggestion:** i think this should use the prefix calculated above, rather than hard-coded "modal"?

kyletsang:
> 
> 
> **suggestion:** the upstream `<Modal>` from react-overlays exposes methods to listen to backdrop clicks and escape keydown events already
> 
> it'd be better to use those than to add another set of listeners here: https://github.com/react-bootstrap/react-overlays/blob/10b1ad66acd563ecd8b454c53863f2cca4f79398/src/Modal.tsx#L422-L430

I've originally looked into using `onEscapeKeyDown` and `onBackdropClick` when implementing this, but I ran into difficulties.

`onEscapeKeyDown` will only trigger if `keyboard` is set to `true` in react-overlays, but this modal animation should run when `keyboard` is set to `false`. 

`onBackdropClick` would not trigger for me, and I found it was due to how the markup and styling was done for the bootstrap modal. The modal class spans the entire screen and covers the backdrop, so clicking would not actually reach the backdrop element.

I wasn't sure if there was a better way to tackle this. Thoughts @taion ?

src/Modal.js
kyletsang: Done

taion:
@kyletsang on the keyboard thing, perhaps it would be better to pass `true` for keyboard in this case? since we can handle the keyboard event here

taion:
that's a bummer about the backdrop thing, but at least we can get out of managing keyboard event listeners?

kyletsang:
> 
> 
> @kyletsang on the keyboard thing, perhaps it would be better to pass `true` for keyboard in this case? since we can handle the keyboard event here

Setting keyboard to `true` would trigger `onHide()` when esc is pressed in react-overlays. 

https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.tsx#L281

taion:
bleh, i wonder if that should check for `event.defaultPrevented` – it seems like it should be possible for the "escape key down" listener to prevent the default "onHide" action there

kyletsang:
That's a good idea. I'll propose that in the react-overlays project.

taion:
released in 3.2.0 on R-O now

kyletsang:
Thanks for pushing new RO so fast. I've removed the keyboard listener here and updated tests.

src/Modal.js
```diff
@@ -320,7 +323,33 @@ const Modal = React.forwardRef(
       waitingForMouseUpRef.current = false;
     };
 
+    const handleStaticModalAnimation = () => {
+      setAnimateStaticModal(true);
+      removeStaticModalAnimationRef.current = transitionEnd(
```
taion, line 328: **suggestion:** we should probably cancel this on unmount, as well

taion:
just some minor comments

src/Modal.js
```diff
@@ -320,7 +323,33 @@ const Modal = React.forwardRef(
...
+        modal.dialog,
+        () => {
+          if (removeStaticModalAnimationRef.current) {
+            removeStaticModalAnimationRef.current();
+          }
```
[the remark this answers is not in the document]
kyletsang: Good catch. I didn't realize the listener was automatically removed until I took a closer look

src/Modal.js
kyletsang: I currently cancel it in the `handleExit` function. Is that good enough?

taion: `onExit` won't be called for the component un-mounting, so the `setState` call at the end will fail

kyletsang: Got it, thanks.

Issue #4140, opened 2019-07-29
Add a modifier to Accordion.Toggle to easily say if Accordion.Collapse is collapsed


Simpleqode:
First of all, thank you so much for such a wonderful library. Makes converting Bootstrap to React a breeze. I have an accordion with _many_ collapsible items in there. I need to style the Accordion.Toggle component depending on if it's Accordion.Collapse pair is collapsed or not. I know I can listen to a click event and update the state (though I don't want to have a state for every Accordion.Toggle & Collapse pair), but how do I know if it was collapsed from the very beginning (i.e. with `defaultActiveKey`)? Thanks in advance!

taion:
How do you envision the API working here for `<Accordion.Toggle>` to inform you of this?

Simpleqode:
Maybe use `aria` attributes as Bootstrap does for it's togglers?

пн, 29 июл. 2019 г., 23:20 Jimmy Jia <notifications@github.com>:

bpas247:
> Maybe use `aria` attributes as Bootstrap does for its togglers? So we could then use `[aria-expanded="true"] { ... }`

I'm not sure if accessing DOM nodes imperatively like that is a good practice to begin with.

> but how do I know if it was collapsed from the very beginning (i.e. with defaultActiveKey)? 

IMO this is something you can infer with the current API. For example, if you don't pass in a `defaultActiveKey` (or `defaultActiveKey` that doesn't match any `Accordion.Toggle`'s `activeKey`s), then the Accordion will start out [fully collapsed](https://react-bootstrap.github.io/components/accordion/#fully-collapsed-state) from the very beginning.

Simpleqode:
Sorry, I don't understand what you mean. I do pass defaultActiveKey to collapase the Accordion.Collapse that I need (I am working on a non-standard navigation and I need to only expand the collapsible menu that I need). Also I am keen to know what's wrong with styling elements using aria attributes. I am 100% confident this is not going to slow down the rendering if I use it here and there.

bpas247:
> Sorry, I don't understand what you mean.

Well I linked to the example docs that shows how to start the Accordion in a fully-collapsed state, but you could also just pass in an `activeKey` that doesn't match any of the togglers to achieve the same effect.

> Also I am keen to know what's wrong with styling elements using aria attributes. 

[...]

PR #5171, opened 2020-05-10
Added Accordion docs example requested in issue #4140

RickeyWard:
based on the idea of diceler found at https://github.com/react-bootstrap/react-bootstrap/issues/4140#issuecomment-604130679. I extended from the existing CustomToggle example making as few changes as possible.

### DIFF
--- a/www/src/components/ReactPlayground.js
+++ b/www/src/components/ReactPlayground.js
@@ -11,6 +11,7 @@ import React, {
   useState,
 } from 'react';
 import * as ReactBootstrap from 'react-bootstrap';
+import AccordionContext from 'react-bootstrap/AccordionContext';
 import ReactDOM from 'react-dom';
 import {
   LiveContext,
@@ -28,7 +29,9 @@ const scope = {
   useEffect,
   useRef,
   useState,
+  useContext,
   ...ReactBootstrap,
+  AccordionContext,
   ReactDOM,
   classNames,
   PropTypes,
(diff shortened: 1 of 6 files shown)


www/src/examples/Accordion/ContexAwareToggle.js
```diff
@@ -0,0 +1,45 @@
+function ContexAwareToggle({ children, eventKey, callback }) {
```
taion, line 1: **nitpick:** should be `ContextAwareToggle`

RickeyWard: good catch, I copied and pasted that around without noticing it. fixed.

www/src/components/ReactPlayground.js
```diff
@@ -11,6 +11,7 @@ import React, {
   useState,
 } from 'react';
 import * as ReactBootstrap from 'react-bootstrap';
+import AccordionContext from 'react-bootstrap/AccordionContext';
```
taion, line 14: **question:** is this not available as `ReactBootstrap.AccordionContext` because it's not exported from the package index? if so, then we should just export it from the package index, instead of doing something special in the examples here

RickeyWard: That is correct. I didn't know if there was a reason that it wasn't exported as part of the index.

taion: yeah, so just add the export then

RickeyWard: Sounds good to me. done.

PR #5177, opened 2020-05-13
Add more variant options for ProgressBar

viniciushiga:


taion:
the variants available here by default are meant to match the ones in bootstrap: https://getbootstrap.com/docs/4.4/components/progress/#backgrounds

PR #5189, opened 2020-05-19
feat(Feedback): Add support for tooltips

### LINKED ISSUE #5188: Add FormControl.Feedback tooltip variant
## Is your feature request related to a problem? Please describe

Noticed the FormControl.Feedback was missing the tooltip variant. See official bootstrap docs:
https://getbootstrap.com/docs/4.5/components/forms/#tooltips

I could create a PR for this. Perhaps adding a `tooltip` boolean prop to Feedback that would set the tooltip class if true and original feedback class if false/undefined?

kyletsang:
Resolves #5188

### DIFF
--- a/src/FormCheck.js
+++ b/src/FormCheck.js
@@ -79,6 +79,9 @@ const propTypes = {
   /** Manually style the input as invalid */
   isInvalid: PropTypes.bool.isRequired,
 
+  /** Display feedback as a tooltip. */
+  tooltip: PropTypes.bool,
+
   /** A message to display when the input is in a validation state */
   feedback: PropTypes.node,
 };
@@ -89,6 +92,7 @@ const defaultProps = {
   disabled: false,
   isValid: false,
   isInvalid: false,
+  tooltip: false,
   title: '',
 };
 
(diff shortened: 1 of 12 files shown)


src/FormCheck.js
```diff
@@ -79,6 +79,9 @@ const propTypes = {
   /** Manually style the input as invalid */
   isInvalid: PropTypes.bool.isRequired,
 
+  /** Display feedback as a tooltip. */
+  tooltip: PropTypes.bool,
```
taion, line 83: **suggestion:** maybe this should be something like `feedbackTooltip`? it'd be more explicit given context

kyletsang: Sounds good. I changed this in both FormCheck and FormFile

jquense:
awesome thank you!

shaungc-su:
Nice work! Wondering would this be included in the next release like `1.1.0`?

kyletsang:
> 
> 
> Nice work, looking forward to using the tooltip variant! Wondering would this be included in the next release like `1.1.0`? Looks like it's not yet included in`v1.1.0-rc.0`.

@taion?

taion:
@kyletsang looks like @jquense is already setting up for 1.1.0 in https://github.com/react-bootstrap/react-bootstrap/pull/5251, so maybe we just wait until then?

PR #5192, opened 2020-05-20
Add first/last column order

abeMedia:
Column order can also be "first" or "last". See https://getbootstrap.com/docs/4.5/layout/grid/#order-classes

### DIFF
--- a/src/Col.js
+++ b/src/Col.js
@@ -37,35 +37,35 @@ const propTypes = {
   /**
    * The number of columns to span on extra small devices (<576px)
    *
-   * @type {(true|"auto"|number|{ span: true|"auto"|number, offset: number, order: number })}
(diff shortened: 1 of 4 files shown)


taion:
Thanks, but could you please update the docstrings in e.g. https://github.com/react-bootstrap/react-bootstrap/blob/46c669838185b9617dc0a289f6c53de86236adb5/src/Col.js#L40 as well?

abeMedia:
Have added them.

kosmiq:
Very solid work, thank you very much!

Would you be willing to add an example to the docs as well?

abeMedia:
@kosmiq sure, why not

abeMedia:
Have added an example to the docs

kosmiq:
Thanks! Looking good.

PR #5193, opened 2020-05-20
test: Added more tests

kyletsang:
Added more tests for #3889

### DIFF
--- a/test/CarouselSpec.js
+++ b/test/CarouselSpec.js
@@ -223,84 +223,6 @@ describe('<Carousel>', () => {
     }, 150);
   });
 
-  it('should call onSlide when slide animation is disabled', () => {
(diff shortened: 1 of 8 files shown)


test/CarouselSpec.js
```diff
@@ -223,6 +223,84 @@ describe('<Carousel>', () => {
     }, 150);
   });
 
+  it('should call onSlide when slide animation is disabled', () => {
+    const clock = sinon.useFakeTimers();
```
taion, line 227: **suggestion:** might be cleaner to wrap these all in a `describe` and set up / tear down sinon in `beforeEach` / `afterEach`

kyletsang: Done!  Thanks

PR #5209, opened 2020-06-01
refactor(Heading+Anchor): Rewrite Heading and Anchor to functional components

moh12594:
This PR is just for rewriting two (Heading and Anchor) components to functional components, if it's good, I'm open to rewrite some other components...

### DIFF
--- a/www/src/components/Anchor.js
+++ b/www/src/components/Anchor.js
@@ -24,23 +24,25 @@ const styles = css`
   }
 `;
 
-class Anchor extends React.Component {
-  static propTypes = {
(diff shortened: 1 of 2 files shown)


taion:
BTW you're aware that those are only used in the docs, right? They're not actually part of our API at all and don't show up in the package.

taion:
Everything that's under `www/` pertains specifically to our docs page.

moh12594:
> BTW you're aware that those are only used in the docs, right? They're not actually part of our API at all and don't show up in the package.

Yes, I know, but as first contribution, I wanted to start with some easy task... 😅

PR #5210, opened 2020-06-01
refactor(Collapse): Rewrite Collapse component as functional component

moh12594:
Hi 👋
Just a little PR in order to rewrite Collapse component into a functional component

### DIFF
--- a/src/Collapse.js
+++ b/src/Collapse.js
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import css from 'dom-helpers/css';
 import transitionEnd from 'dom-helpers/transitionEnd';
 import PropTypes from 'prop-types';
-import React, { useCallback } from 'react';
+import React, { useMemo } from 'react';
 import Transition, {
   ENTERED,
   ENTERING,
@@ -159,62 +159,59 @@ const Collapse = React.forwardRef(
     },
     ref,
   ) => {
+    /* Compute dimension */
+    const computedDimension = getDimension(dimension);
+
     /* -- Expanding -- */
-    const handleEnterAction = useCallback(
-      (elem) => {
-        elem.style[getDimension(dimension)] = '0';
-      },
-      [dimension],
+    const handleEnter = useMemo(
+      () =>
+        createChainedFunction((elem) => {
+          elem.style[computedDimension] = '0';
+        }, onEnter),
+      [computedDimension, onEnter],
     );
 
-    const handleEnteringAction = useCallback(
-      (elem) => {
-        const computedDimension = getDimension(dimension);
-        elem.style[dimension] = collapseHelpers._getScrollDimensionValue(
-          elem,
-          computedDimension,
-        );
-      },
-      [dimension],
+    const handleEntering = useMemo(
+      () =>
+        createChainedFunction((elem) => {
+          elem.style[dimension] = collapseHelpers._getScrollDimensionValue(
+            elem,
+            computedDimension,
+          );
+        }, onEntering),
+      [computedDimension, onEntering, dimension],
     );
 
-    const handleEnteredAction = useCallback(
-      (elem) => {
-        elem.style[getDimension(dimension)] = null;
-      },
-      [dimension],
+    const handleEntered = useMemo(
+      () =>
+        createChainedFunction((elem) => {
+          elem.style[computedDimension] = null;
+        }, onEntered),
+      [computedDimension, onEntered],
     );
 
     /* -- Collapsing -- */
-    const handleExitAction = useCallback(
-      (elem) => {
-        const computedDimension = getDimension(dimension);
-        elem.style[dimension] = `${getDimensionValue(
-          computedDimension,
-          elem,
-        )}px`;
-        triggerBrowserReflow(elem);
-      },
-      [dimension, getDimensionValue],
+    const handleExit = useMemo(
+      () =>
+        createChainedFunction((elem) => {
+          elem.style[dimension] = `${getDimensionValue(
+            computedDimension,
+            elem,
+          )}px`;
+          triggerBrowserReflow(elem);
+        }, onExit),
+      [dimension, onExit, getDimensionValue, computedDimension],
     );
-
-    const handleExitingAction = useCallback(
-      (elem) => {
-        elem.style[getDimension(dimension)] = null;
-      },
-      [dimension],
+    const handleExiting = useMemo(
+      () =>
+        createChainedFunction((elem) => {
+          elem.style[computedDimension] = null;
+        }, onExiting),
+      [computedDimension, onExiting],
     );
 
     delete props.dimension;
     delete props.getDimensionValue;
-    const handleEnter = createChainedFunction(handleEnterAction, onEnter);
-    const handleEntering = createChainedFunction(
-      handleEnteringAction,
-      onEntering,
-    );
-    const handleEntered = createChainedFunction(handleEnteredAction, onEntered);
-    const handleExit = createChainedFunction(handleExitAction, onExit);
-    const handleExiting = createChainedFunction(handleExitingAction, onExiting);
 
     return (
       <Transition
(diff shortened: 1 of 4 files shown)


src/Collapse.js
```diff
@@ -125,78 +139,86 @@ const defaultProps = {
...
-      this.handleEntering,
+      handleEnteringAction,
       onEntering,
     );
-    const handleEntered = createChainedFunction(this.handleEntered, onEntered);
-    const handleExit = createChainedFunction(this.handleExit, onExit);
-    const handleExiting = createChainedFunction(this.handleExiting, onExiting);
+    const handleEntered = createChainedFunction(handleEnteredAction, onEntered);
+    const handleExit = createChainedFunction(handleExitAction, onExit);
+    const handleExiting = createChainedFunction(handleExitingAction, onExiting);
```
taion, line 217: **suggestion:** there's not much point to `useCallback`-ing the wrapped callbacks while not wrapping `createChainedFunction` in anything memoized.

instead of doing this, then, i would wrap the `createChainedFunction` calls in `useMemo`, and not further memoize the individual handlers.

src/Collapse.js
```diff
@@ -213,14 +235,14 @@ class Collapse extends React.Component {
               className,
               children.props.className,
               collapseStyles[state],
-              this.getDimension() === 'width' && 'width',
+              getDimension(dimension) === 'width' && 'width',
```
taion, line 238: **suggestion:** this isn't really relevant as set up any more; just pre-calculate the dimension once and for all up front in the function, and close over it in the callback handlers

test/CollapseSpec.js
```diff
@@ -40,13 +40,16 @@ describe('<Collapse>', () => {
   describe('from collapsed to expanded', () => {
     beforeEach(() => {
       wrapper = mount(<Component>Panel content</Component>);
-
       // since scrollHeight is gonna be 0 detached from the DOM
-      sinon
-        .stub(wrapper.instance().collapse, '_getScrollDimensionValue')
+      getScrollDimensionValue = sinon
+        .stub(collapseHelpers, '_getScrollDimensionValue')
         .returns('15px');
```
taion, line 46: looking at this again – can we just stub out `scrollX` or `scrollY` or whatever on the base node class?

src/Collapse.js
moh12594: Yes, you're right ! I'm fixing it thank you !

src/Collapse.js
moh12594: Yes ! I'm on it :)

test/CollapseSpec.js
moh12594: I'm not sure to understand... I tried to stub the scrollHeight, but I didn't found a way to achieve it 😢

moh12594:
Hi @taion ! Thank you for your review 👋
I pushed some edits according to it

src/Collapse.js
```diff
@@ -29,6 +29,21 @@ function getDimensionValue(dimension, elem) {
...
+
+// for testing
+function _getScrollDimensionValue(elem, computedDimension) {
+  const scroll = `scroll${computedDimension[0].toUpperCase()}${computedDimension.slice(
+    1,
+  )}`;
+  return `${elem[scroll]}px`;
+}
+// For testing purposes we should export _getScrollDimensionValue in order to stub it
+export const collapseHelpers = { _getScrollDimensionValue, getDimension };
```
taion, line 45: **issue:** we generally don't want to have additional exports in modules, because it screws up the "add module exports" logic we need for transparent CJS shimming.

see further comments below

src/Collapse.js
```diff
@@ -125,78 +140,83 @@ const defaultProps = {
...
+          elem.style[computedDimension] = null;
+        }, onExiting),
+      [computedDimension, onExiting],
     );
-    const handleEntered = createChainedFunction(this.handleEntered, onEntered);
-    const handleExit = createChainedFunction(this.handleExit, onExit);
-    const handleExiting = createChainedFunction(this.handleExiting, onExiting);
+
+    delete props.dimension;
+    delete props.getDimensionValue;
```
taion, line 216: **suggestion:** these already get spread out above

```suggestion
```

test/CollapseSpec.js
```diff
@@ -174,7 +172,7 @@ describe('<Collapse>', () => {
     });
 
     it('Defaults to height', () => {
-      assert.equal(wrapper.instance().collapse.getDimension(), 'height');
+      assert.equal(collapseHelpers.getDimension(), 'height');
```
taion, line 175: **suggestion:** instead of doing this (which doesn't do anything to check that the logic in the component itself is correct), assert on the rendered classes from the object, i.e. that `.width` is not present: https://github.com/moh12594/react-bootstrap/blob/2329c05768f09104c3c6453b875abb61902c9db8/src/Collapse.js#L236

test/CollapseSpec.js
```diff
@@ -174,7 +172,7 @@ describe('<Collapse>', () => {
...
     });
 
     it('Uses getCollapsibleDimension if exists', () => {
```
taion, line 178: **nitpick:** this test case could probably use a better name... probably just "uses dimension if exists", given that there's no longer any `getCollapsibleDimension`

test/CollapseSpec.js
```diff
@@ -184,7 +182,10 @@ describe('<Collapse>', () => {
 
       wrapper.setState({ dimension });
 
-      assert.equal(wrapper.instance().collapse.getDimension(), 'whatevs');
+      assert.equal(
+        collapseHelpers.getDimension(wrapper.find(Collapse).props().dimension),
+        'whatevs',
+      );
```
taion, line 188: **suggestion:** per https://github.com/moh12594/react-bootstrap/blob/2329c05768f09104c3c6453b875abb61902c9db8/src/Collapse.js#L236, probably better to just set the dimension to `width`, and assert that the `width` class is present

test/CollapseSpec.js
```diff
@@ -30,21 +30,19 @@ describe('<Collapse>', () => {
...
+    assert.ok(wrapper.find(Collapse).props().in === false);
   });
 
   it('Should have collapse class', () => {
     mount(<Component>Panel content</Component>).assertSingle('.collapse');
   });
 
   describe('from collapsed to expanded', () => {
+    // since scrollHeight is gonna be 0 detached from the DOM
+    sinon.stub(collapseHelpers, '_getScrollDimensionValue').returns('15px');
```
taion, line 42: **issue:** this needs to go into `beforeEach` if it's to do anything? also, don't we need to unwind this stub/mock somewhere?

taion: **suggestion:** instead of mocking this, just update the wrapper itself to set something like `style={{ height: 15 }}` or whatever, such that the logic in question will actually calculate the correct height.

then you can get rid of the helper export

src/Collapse.js
taion: once we remove the exports, we can inline both of these functions, as both will only be used once

src/Collapse.js
```diff
@@ -125,78 +140,83 @@ const defaultProps = {
...
```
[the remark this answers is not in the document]
moh12594: Yes !

taion:
fantastic. thanks!

src/Collapse.js
```diff
@@ -125,78 +125,81 @@ const defaultProps = {
...
+      dimension,
+      getDimensionValue,
       ...props
-    } = this.props;
+    },
+    ref,
+  ) => {
+    /* Compute dimension */
+    const computedDimension =
+      typeof dimension === 'function' ? dimension() : dimension;
```
kyletsang, line 150: We could probably `useMemo` this @taion ?

taion: it's probably not needed... `dimension()` shouldn't be super-expensive, and it's a string, so `===` will pass

kyletsang: Gotcha

PR #5211, opened 2020-06-02
in solidarity

jquense:
https://twitter.com/kwuchu/status/1267436741692133376 and https://github.com/reactjs/reactjs.org

It's literally the least we can do with our small microphone


jquense:
deployed!

PR #5219, opened 2020-06-05
refactor(PageItem): Rewrite createButton return to functional component

moh12594:
Hi 👋
Just a little PR in order to rewrite the createButton returned component into a functional component

### DIFF
--- a/src/PageItem.js
+++ b/src/PageItem.js
@@ -52,14 +52,18 @@ PageItem.displayName = 'PageItem';
 export default PageItem;
 
 function createButton(name, defaultValue, label = name) {
-  return ({ children, ...props }) => {
+  function Button({ children, ...props }) {
     return (
       <PageItem {...props}>
         <span aria-hidden="true">{children || defaultValue}</span>
         <span className="sr-only">{label}</span>
       </PageItem>
     );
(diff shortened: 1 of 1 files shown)


src/PageItem.js
```diff
@@ -52,19 +52,13 @@ PageItem.displayName = 'PageItem';
...
-      );
-    }
+  return ({ children, ...props }) => {
+    return (
+      <PageItem {...props}>
+        <span aria-hidden="true">{children || defaultValue}</span>
+        <span className="sr-only">{label}</span>
+      </PageItem>
+    );
   };
```
taion, line 62: **issue:** for nice-ness in dev tools, we need to assign `displayName`. something like the below (maybe with a better interim function name than `PageItemButton`)

```suggestion
  function PageItemButton ({ children, ...props }) {
    return (
      <PageItem {...props}>
        <span aria-hidden="true">{children || defaultValue}</span>
        <span className="sr-only">{label}</span>
      </PageItem>
    );
  }

  PageItemButton.displayName = name;

  return PageItemButton;
```

moh12594: Yes, changes done ! thanks !

PR #5220, opened 2020-06-05
fix(Toast): Reset autohide timer only if show or autohide changed

### LINKED ISSUE #5173: Toasts - Stacked Toasts with autohide delays does not work correctly in prod build.
I have an array of toast data which I am using to render the stacked toast as mentioned [here](https://react-bootstrap.github.io/components/toasts/#stacking) 

Each of the `<Toast/>` have `autohide=true` with `delay=5000`. 

The problem is when a new `<Toast/>` is added dynamically by pushing a new item in the array, the delay for all the existing `<Toast/>` restarts. 

The expected behavior should be: `<Toast/>` should get hidden after 5 secs of their creation.

Below is my code snippet of my `render()`: 

```jsx
<>  this.toastArray.map((toast) => {
   return <Toast autohide={true} delay={5000} onClose={() => {this.closeToast(toast.id)}}>
                     <Toast.Body>{toast.text}</Toast.Body>
               </Toast>
}) </>
``` 

Interestingly, the issue happens only in prod build, built with`npm run build`. 

It works fine in local dev environment with `npm run start`.

kyletsang:
Fixes #5173.

The autohide timer was being reset on every render, so I changed it only to reset if `show` or `autohide` is changed.  I noticed there was a note about not restarting the timer if `delay` is changed so I left that as is.


src/Toast.js
```diff
@@ -89,7 +89,15 @@ const Toast = React.forwardRef(
...
 
-    autohideTimeout.set(autohideFunc, delayRef.current);
+    useEffect(() => {
+      if (autohide && show) {
+        autohideTimeout.set(autohideFunc, delayRef.current);
+      }
+
+      return () => {
+        autohideTimeout.clear();
+      };
```
taion, line 99:
```suggestion
      autohideTimeout.set(autohideFunc, delayRef.current);
```

this might be sufficient? `autohideFunc` won't do anything if `autohide && show` isn't enabled, and any re-run of the effect will reset the timeout anyway.

kyletsang: This is much cleaner.  Thanks!

src/Toast.js
```diff
@@ -89,7 +89,9 @@ const Toast = React.forwardRef(
       onCloseRef.current();
     }, [autohide, show]);
 
-    autohideTimeout.set(autohideFunc, delayRef.current);
+    useEffect(() => {
+      autohideTimeout.set(autohideFunc, delayRef.current);
+    }, [autohide, show, autohideTimeout, autohideFunc]);
```
taion, line 94:
```suggestion
    }, [autohideTimeout, autohideFunc]);
```

taion: oops, never mind

taion: actually, you probably want a comment noting that `autohide` and `show` are here on purpose.

you may also want to pre-calculate `autohide && show` and use that here

kyletsang: Added changes. Thanks!

PR #5225, opened 2020-06-08
Bump versions

Richard87:
react-bootstrap 0.33 is the only demand for core-js 2 in my legacy project now, and while I work on migrating to bootstrap 4, I would love to have core-js 3, since it would solve alot of dependency issues in unrelated code!

- babel version ^7.0.0 =>  ^7.4.0
- core-js 2 => 3


jquense:
@taion is this safe to do do you think?

Richard87:
Thank you! :)

Issue #5205, opened 2020-05-28
<Col> outputs column classes without sizing if no span is set in an object breaking responsiveness in some cases


kosmiq:
## Describe the bug

Creating an issue to keep track of this. Will look into it whenever I get around to it. It could be by design, if so do please tell.

## To Reproduce

1. Create a `<Col>`
2. Set two breakpoints using the object style, but omit setting a span for one breakpoint: `<Col xs={{ span: "auto", order: 2 }} lg={{ order: 1 }}>`
3. The outputted html becomes `<div class="col-lg col-auto order-lg-2 order-1">` with no sizing set for `lg` breakpoint.
4. This breaks auto sizing that should be inherited from the smaller breakpoint since `col-lg` has CSS without a specified breakpoint.

## Reproducible Example
https://codesandbox.io/s/bold-night-htu76?file=/src/App.js

Not reproducible per se since in my application `col-lg` hits my div (see screenshot) and breaks the auto-sizing of the col that I want to be inherited from xs breakpoint definition.

However you can see that the `col-lg` class is added even though I only really want the order class to be outputted, but the order-lg-* should be added.

## Expected behavior

Sizing should be inherited from the smaller breakpoint as expected and the "empty" `col-lg` class should not be outputted

## Screenshots

![Screenshot 2020-05-28 at 10 55 33](https://user-images.githubusercontent.com/322759/83121039-1be83300-a0d2-11ea-9a1e-73a4105d107b.png)

## Environment (please complete the following information)

- Operating System: MacOS 10.15.5
- Browser, Version Chrome version 81.0.4044.138 (Official Build) (64-bit)
- React-Bootstrap Version Beta 16 and 1.0.1

## Additional context

For some reason the col-lg class exists in the official bootstrap CSS but does not get active in the example on CodeSandbox?

kyletsang:
Looks like span is defaulted to `true` when it's omitted.  Seems like this should be opt-in if you want to output equal width sizing classes?

https://github.com/react-bootstrap/react-bootstrap/blob/dcbed6d9aa039ae65317f1b7d8d5947c259f011e/src/Col.js#L86

kosmiq:
@taion @jquense do you have a suggestion here? For my own use case I would love to be able to omit span for breakpoints and thus more mimic the behavior if I omit the lg tag altogether, and not get col-lg as class. I would be fine with being able to set span: false to stop output, or span: true if I want it and false is the default.

[...]

PR #5232, opened 2020-06-11
feat(Col): Allow false to be used for Col span

kyletsang:
Fixes TS definitions to allow null for span in Col.  Also fix some minor wording in layout docs.

#5205

### DIFF
--- a/src/Col.js
+++ b/src/Col.js
@@ -37,35 +37,35 @@ const propTypes = {
   /**
    * The number of columns to span on extra small devices (<576px)
    *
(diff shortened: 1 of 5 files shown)


kosmiq:
I think we should add false as well. If we allow for passing true you should be able to pass the opposite which is false.

kyletsang:
Sure @kosmiq.  I'll add that in tomorrow.

taion:
could it be better to do `false` instead of `null` as the "don't emit a class" value?

PR #5235, opened 2020-06-12
feat: useMediaBreakpoints hook

tywmick:
Hello!

Not sure if porting any of Bootstrap's Sass mixins is a goal of this project, but in case it is (or can be), I made [a React hook version](https://github.com/tywmick/use-media-breakpoints) of their [media query mixins](https://getbootstrap.com/docs/4.5/layout/overview/#responsive-breakpoints) (`media-breakpoint-up`, `media-breakpoint-down`, `media-breakpoint-only`, `media-breakpoint-between`). Goes a little something like this:

-----

After calling the hook from the top level of your React function...

```js
const breakpoint = useMediaBreakpoints();
// Or specify your own custom breakpoint values:
const breakpoint = useMediaBreakpoints({
  xs: 0,
  sm: 576,
  md: 768,
  lg: 992,
  xl: 1200,
});
```

...it returns an object containing the boolean results of several media queries. For example, if the width of the window is `800px`, the value of `breakpoint` (as defined either way above) will be

```js
{
  xs: false,
  sm: false,
  md: true,
  lg: false,
  xl: false,
  only: {
    xs: false,
    sm: false,
    md: true,
    lg: false,
    xl: false,
  },
  up: {
    xs: true,
    sm: true,
    md: true,
    lg: false,
    xl: false,
  },
  down: {
    xs: false,
    sm: false,
    md: true,
    lg: true,
    xl: true,
  },
  between: {
    xs: {
      sm: false,
      md: true,
      lg: true,
      xl: true,
    },
    sm: {
      md: true,
      lg: true,
      xl: true,
    },
    md: {
      lg: true,
      xl: true,
    },
    lg: {
      xl: false,
    },
  },
}
```

What's that good for? Say you have a React component you only want to display on `md`-sized screens. Thow this into your JSX:

```jsx
{breakpoint.md && <MyComponent />}
{/* OR */}
{breakpoint.only.md && <MyComponent />}
```

Or maybe you want to use one component on larger screens and a different one on smaller screens:

```jsx
{breakpoint.up.lg ? <LargerVersion /> : <SmallerVersion />}
```

Or maybe you want to describe the size of the screen in paragraph form with an odd sort of precision:

```jsx
<p>
  This screen is {breakpoint.between.sm.lg ? '' : 'pretty '}
  {breakpoint.down.sm ? 'small' : breakpoint.up.lg ? 'big' : 'average'}.
</p>
```

-----

Is this hook something that would be useful and/or desired in React-Bootstrap? I'll of course be happy to adapt [my tests](https://github.com/tywmick/use-media-breakpoints/tree/master/test) and add new documentation to this branch if so—I just figured I may as well ask first.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,3 +78,4 @@ export Tooltip from './Tooltip';
 export Toast from './Toast';
 export ToastBody from './ToastBody';
 export ToastHeader from './ToastHeader';
+export useMediaBreakpoints from './useMediaBreakpoints';
--- a/src/useMediaBreakpoints.js
+++ b/src/useMediaBreakpoints.js
@@ -1,0 +1,138 @@
+import { useState, useEffect } from 'react';
+
+const defaultBreakpoints = { xs: 0, sm: 576, md: 768, lg: 992, xl: 1200 };
+
+/**
+ * React hook for using screen width breakpoints.
+ * @param {Object} breakpoints - Set of screen width breakpoints to be used
+ * @param {0} [breakpoints.xs] - Lower bound of extra-small screen widths, in pixels
+ * @param {number} breakpoints.sm - Lower bound of small screen widths, in pixels
+ * @param {number} [breakpoints.md] - Lower bound of medium screen widths, in pixels
+ * @param {number} breakpoints.lg - Lower bound of large screen widths, in pixels
+ * @param {number} [breakpoints.xs] - Lower bound of extra-large screen widths, in pixels
+ * @returns {Object} A full set of breakpoint queries and their boolean values
+ */
+export default function useMediaBreakpoints(breakpoints = defaultBreakpoints) {
+  // Validate breakpoint inputs
+  if (breakpoints.sm === undefined)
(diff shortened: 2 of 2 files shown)


taion:
We've been using something like this in https://github.com/react-restart/hooks/blob/f72b24aef649d3bf1fd05c3c8cb72e97edaa71fc/src/useMediaQuery.tsx

The issue is that, to get the best value out of this, you'd need some way to pull the breakpoint values out of CSS. It's doable with e.g. CSS module exports and webpack, but difficult otherwise.

@jquense WDYT of incorporating something like this into R-B?

jquense:
I think we should probably avoid porting style only sorts of things. It walks to close to our  "don't deal with style things specifically" line i think.

I would offer tho that we also have https://react-restart.github.io/hooks/api/useBreakpoint in restart/hooks that matches the default bootstrap breakpoints already.

tywmick:
Totally makes sense. 😊 Take care!

PR #5263, opened 2020-07-02
[types/helpers] add 'this : void' to SelectCallback

hexagonal-sun:
Add a 'this : void' parameter to the SelectCallback, indicating to TypeScript
that this callback will overwrite the 'this' context. See [1].

[1]: https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters-in-callbacks

### DIFF
--- a/types/components/helpers.d.ts
+++ b/types/components/helpers.d.ts
@@ -39,6 +39,7 @@ export type BsPrefixComponentClass<
 > = React.ComponentClass<ReplaceProps<As, BsPrefixProps<As> & P>>;
 
 export type SelectCallback = (
+  this : void,
   eventKey: string,
   e: React.SyntheticEvent<unknown>,
 ) => void;


types/components/helpers.d.ts
```diff
@@ -39,6 +39,7 @@ export type BsPrefixComponentClass<
 > = React.ComponentClass<ReplaceProps<As, BsPrefixProps<As> & P>>;
 
 export type SelectCallback = (
+  this : void,
```
taion, line 42: **nitpick:** formatting

```suggestion
  this: void,
```

hexagonal-sun:
> should we do this throughout?

There are probably other areas where this should be updated.  I created this PR just to see what people's appetite would be for such a change.

> i don't think this is an especially common pattern here.
> 
> what will this prevent, and is there a way we can opt into this as the "default" behavior?

Adding a `this : void` parameter to a function, says to typescript that this function will be called where `this` will have a different context.  A simple example would consist of:

```typescript
class foo extends React.Component<{}, { value: string }> {
    state: { value: string };
    constructor(props: {}) {
        super(props);
        this.state = { value: 'a' };
    }

    private callback(this: this, e: React.ChangeEvent<HTMLTextAreaElement>)
    {
    }

    render(this : this) {
        return DOM.textarea({
            value: this.state.value,
            onChange: this.callback,
        });
    }
}
```
This code is unsafe as `onChange` will have a different `this` context.  The following error would be raised:
```
No overload matches this call.
  Overload 1 of 2, '(props?: DetailedHTMLProps<TextareaHTMLAttributes<HTMLTextAreaElement>, HTMLTextAreaElement> | null | undefined, ...children: ReactNode[]): DetailedReactHTMLElement<...>', gave the following error.
    Type '(this: this, e: ChangeEvent<HTMLTextAreaElement>) => void' is not assignable to type '(this: void, event: ChangeEvent<HTMLTextAreaElement>) => void'.
      The 'this' types of each signature are incompatible.
        Type 'void' is not assignable to type 'this'.
          'this' could be instantiated with an arbitrary type which could be unrelated to 'void'.
  Overload 2 of 2, '(props?: DetailedHTMLProps<TextareaHTMLAttributes<HTMLTextAreaElement>, HTMLTextAreaElement> | null | undefined, ...children: ReactNode[]): DOMElement<...>', gave the following error.
    Type '(this: this, e: ChangeEvent<HTMLTextAreaElement>) => void' is not assignable to type '(this: void, event: ChangeEvent<HTMLTextAreaElement>) => void'.
```

> TS can't possibly expect that callbacks generally have a `this`, can it?

I guess so.

taion:
@react-bootstrap/collaborators thoughts?

kyletsang:
It seems like this only guards against the case where the user explicitly defines `this: T` in the callback?  If you set up the above example as below, it would pass the TS check, but crash at runtime.

```
class foo extends React.Component<{}, { value: string }> {
    state: { value: string };
    constructor(props: {}) {
        super(props);
        this.state = { value: 'a' };
    }

    private doSomething() {}

    private callback(e: React.ChangeEvent<HTMLTextAreaElement>)
    {
        this.doSomething();
    }

    render(this : this) {
        return DOM.textarea({
            value: this.state.value,
            onChange: this.callback,
        });
    }
}
```

hexagonal-sun:
>  If you set up the above example as below, it would pass the TS check, but crash at runtime.

Agreed. I'm not sure why this behavior is the way it is.  I don't understand why typescript can't imply the `this` type for member functions.

hexagonal-sun:
I've also added a similar PR to the react type definitions.  There's a post there that may help clear things up: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/45834#issuecomment-657706512

taion:
Makes sense. We'll follow the upstream here – will watch the PR there.

Issue #5262, opened 2020-07-01
Error importing and using Feedback component


blms:
## Description

I've been working on a `Form` with react-bootstrap. Using the following syntax, all components are importing correctly except for `Feedback`:
```JSX
import {
  Button, Card, Col, Container, Feedback, Form, Row,
} from 'react-bootstrap';
```

This returns the following error: `Attempted import error: 'Feedback' is not exported from 'react-bootstrap'.`

If I separate the `Feedback` import like so, the import does not fail, but the functionality does not seem to work (at least for a `Form.Check` element).
```JSX
import Feedback from 'react-bootstrap/Feedback';
```
The tooltip feedback for my `Form.Check` element does not change from the default (`Please check this box if you want to proceed.`). Here is the code I wrote to use the `Feedback` component:
```JSX
<Form.Check type="checkbox">
  <Form.Check.Input required type="checkbox" id="tosCheck" />
  <Form.Check.Label for="tosCheck">
    I agree to the ToS.{" "}
  </Form.Check.Label>
  <Feedback type="invalid" for="tosCheck">
    You must accept the ToS to create an account.
  </Feedback>
</Form.Check>
```
I'm not sure if `for="tosCheck"` is required but it seems to fail with or without that. I also tried `Form.Feedback`, `Form.Check.Feedback`, `Form.Control.Feedback` and a `feedback="text"` property in the `Form.Check` element—none of which worked. (It doesn't work if I add a `tooltip` property to the Feedback element either.)

## Reproducible Example

- Example of the import error: https://codesandbox.io/s/elastic-almeida-q4i4d?file=/src/App.js
- Example of the non-functioning Feedback component: https://codesandbox.io/s/sparkling-thunder-ilwz7?file=/src/App.js

Click "Submit" on the latter example without checking the checkbox. The message should read "You must accept the ToS to create an account," but displays the default message instead.

## Environment

- Operating System: macOS 10.15.4 (19E287)
- Browser, Version: Google Chrome 83.0.4103.116 (64-bit)
- React-Bootstrap Version: 1.0.1

## Additional context

The documentation about this seems inconsistent. For example, I couldn't find documentation for the Feedback component though it is used inside the [Form.Check](https://react-bootstrap.netlify.app/components/forms/#form-check-props) and [Form.File](https://react-bootstrap.netlify.app/components/forms/#form-file-props) reference.

blms:
[...]

PR #5264, opened 2020-07-03
docs(Form): Add info about noValidate

kyletsang:
Added some info about `noValidate` in forms

### DIFF
--- a/www/src/pages/components/forms.js
+++ b/www/src/pages/components/forms.js
@@ -410,10 +410,20 @@ export default withLayout(function FormControlsSection({ data }) {
       <p>
         Bootstrap scopes the <code>:valid</code> and <code>:invalid</code>{' '}
         styles to parent <code>.was-validated</code> class, usually applied to
-        the <code>Form</code> (you can use the <code>validated</code> prop as a
-        shortcut). Otherwise, any required field without a value shows up as
-        invalid on page load. This way, you may choose when to activate them
+        the <code>{'<Form>'}</code> (you can use the <code>validated</code> prop
+        as a shortcut). Otherwise, any required field without a value shows up
+        as invalid on page load. This way, you may choose when to activate them
         (typically after form submission is attempted).
+      </p>
+      <p>
+        For custom Bootstrap form validation messages, you’ll need to add the{' '}
+        <code>noValidate</code> boolean attribute to your{' '}
+        <code>{'<Form>'}</code>. This disables the browser default feedback
+        tooltips, but still provides access to the form validation APIs in
+        JavaScript. Try to submit the form below; our JavaScript will intercept
+        the submit button and relay feedback to you. When attempting to submit,
+        you’ll see the <code>:invalid</code> and <code>:valid</code> styles
+        applied to your form controls.
       </p>
       <ReactPlayground codeText={ValidationNative} />
 
--- a/www/src/pages/components/forms.js
+++ b/www/src/pages/components/forms.js
@@ -415,16 +415,12 @@ export default withLayout(function FormControlsSection({ data }) {
         as invalid on page load. This way, you may choose when to activate them
         (typically after form submission is attempted).
       </p>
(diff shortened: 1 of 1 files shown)


www/src/pages/components/forms.js
```diff
@@ -410,11 +410,21 @@ export default withLayout(function FormControlsSection({ data }) {
...
+      <p>
+        For custom Bootstrap form validation messages, you’ll need to add the{' '}
+        <code>noValidate</code> boolean attribute to your{' '}
+        <code>{'<Form>'}</code>. This disables the browser default feedback
+        tooltips, but still provides access to the form validation APIs in
+        JavaScript. Try to submit the form below; our JavaScript will intercept
+        the submit button and relay feedback to you. When attempting to submit,
+        you’ll see the <code>:invalid</code> and <code>:valid</code> styles
+        applied to your form controls.
+      </p>
```
bpas247, line 427: This just looks like a direct copy/paste of the [bootstrap docs](https://getbootstrap.com/docs/4.0/components/forms/#custom-styles)?

I don't think it's a good idea to add this, if anything we can just link to this section of their docs.

kyletsang: Yeah it's mostly a copy/paste with some minor edits for React.  This came up in #5262, so I thought it would help to clear up any confusion for users.  I'm fine with linking to the official docs if that's the way to go.

jquense: I think it's ok to restate docs here if relevant. TBH most of our docs are copies of upstream any :P

bpas247: Idk I still think it’s a hassle to have duplicate info in here, since the upstream docs might change and we’d have to update this once it does 🤷

Maybe just a brief summary followed by an external link to the upstream docs?

kyletsang: I did notice that a lot of info were copies from the official docs with respective React info substituted in (ie. previous paragraph), so I thought I'd follow suit with that.  

I can probably chop the text in half to make it shorter and put in the official link?

jquense: I generally agree with not duplicating info, but this isn't likely to change because it's not bootstrap specific, it's surfacing and noting normal HTML behavior that needs to be considered here. Also people don't really click through to go read the bootstrap docs, or often understand why they would be relevant to RB.  I would just slightly reword the upstream notes a bit.

www/src/pages/components/forms.js
```diff
@@ -410,11 +410,21 @@ export default withLayout(function FormControlsSection({ data }) {
...
-        invalid on page load. This way, you may choose when to activate them
+        the <code>{'<Form>'}</code> (you can use the <code>validated</code> prop
+        as a shortcut). Otherwise, any required field without a value shows up
+        as invalid on page load. This way, you may choose when to activate them
         (typically after form submission is attempted).
       </p>
+      <p>
+        For custom Bootstrap form validation messages, you’ll need to add the{' '}
+        <code>noValidate</code> boolean attribute to your{' '}
+        <code>{'<Form>'}</code>. This disables the browser default feedback
```
jquense, line 421: don't we set noValidate to false automatically in the component? Should we?

www/src/pages/components/forms.js
jquense: I would rework this to something like "Watch out! Browsers, provide their own validation UI by default on `form`s. You can disable the default UI by adding the HTML `noValidate` attribute to your Form/form element."

kyletsang: Thanks for the feedback @bpas247 & @jquense. I've checked in new wording.

www/src/pages/components/forms.js
kyletsang: Do you mean noValidate = true?  It's currently not set by default.  I think it's better to follow the default behavior here by having noValidate as an opt in.

PR #5283, opened 2020-07-09
Fix placements

### LINKED ISSUE #5111: Tooltips with position auto are placed on top of the element

jquense:
closes #5111

### DIFF
--- a/src/Popover.tsx
+++ b/src/Popover.tsx
@@ -94,16 +94,18 @@ const Popover: Popover = (React.forwardRef<HTMLDivElement, PopoverProps>(
     ref,
   ) => {
     const decoratedBsPrefix = useBootstrapPrefix(bsPrefix, 'popover');
+    const [primaryPlacement] = placement?.split('-') || [];
+
     return (
       <div
         ref={ref}
         role="tooltip"
         style={style}
-        x-placement={placement}
+        x-placement={primaryPlacement}
         className={classNames(
           className,
           decoratedBsPrefix,
-          `bs-popover-${placement}`,
+          `bs-popover-${primaryPlacement}`,
         )}
         {...props}
       >
--- a/src/Popover.tsx
+++ b/src/Popover.tsx
@@ -105,7 +105,7 @@ const Popover: Popover = (React.forwardRef<HTMLDivElement, PopoverProps>(
         className={classNames(
           className,
           decoratedBsPrefix,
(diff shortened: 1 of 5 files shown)


src/Tooltip.tsx
```diff
@@ -99,13 +99,19 @@ const Tooltip: Tooltip = React.forwardRef<HTMLDivElement, TooltipProps>(
...
         ref={ref}
         style={style}
         role="tooltip"
-        x-placement={placement}
-        className={classNames(className, bsPrefix, `bs-tooltip-${placement}`)}
+        x-placement={primaryPlacement}
+        className={classNames(
+          className,
+          bsPrefix,
+          `bs-tooltip-${primaryPlacement}`,
```
taion, line 113: **nitpick:** we shouldn't do `bs-tooltip-undefined` or whatever we'd get when `placement` is null, should we?

src/Popover.tsx
```diff
@@ -94,16 +94,18 @@ const Popover: Popover = (React.forwardRef<HTMLDivElement, PopoverProps>(
...
         ref={ref}
         role="tooltip"
         style={style}
-        x-placement={placement}
+        x-placement={primaryPlacement}
         className={classNames(
           className,
           decoratedBsPrefix,
-          `bs-popover-${placement}`,
+          `bs-popover-${primaryPlacement}`,
```
jquense, line 108:
```suggestion
          primaryPlacement  && `bs-popover-${primaryPlacement}`,
```

src/Tooltip.tsx
taion: this should be `primaryPlacement &&` as well

PR #5287, opened 2020-07-10
fix(Popover): arrow offset when scrolling offscreen

jquense:
depends on the react overlays bump. 

this still doesn't cover this problem https://github.com/popperjs/react-popper/issues/375 but at least it won't lock up and crash

### DIFF
--- a/src/usePopperMarginModifiers.tsx
+++ b/src/usePopperMarginModifiers.tsx
@@ -37,7 +37,7 @@ export default function usePopperMarginModifiers(): [
     overlayRef.current = overlay;
   }, []);
 
-  const modifier = useMemo(() => {
+  const offset = useMemo(() => {
     return {
       name: 'offset',
       options: {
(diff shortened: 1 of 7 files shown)


src/usePopperMarginModifiers.tsx
```diff
@@ -61,5 +61,40 @@ export default function usePopperMarginModifiers(): [
...
+      name: 'popoverArrowMargins',
+      enabled: true,
+      phase: 'main',
+      requiresIfExists: ['arrow'],
+      effect({ state }) {
+        if (
+          !overlayRef.current ||
+          !state.elements.arrow ||
+          !hasClass(overlayRef.current, 'popover') ||
+          !state.modifiersData['arrow#persistent']
```
taion, line 77: **question:** where does `#persistent` come from? is this really an `id`?

jquense: its a special key Popper adds, nothing unique about it tho aside from Popper picked it

PR #5296, opened 2020-07-13
Remove Role Attribute From SafeAnchor Component

LeftyDaniels:
Added a type definition to the "isTrivialHref" method's argument.

Removed the use of the "isTrivialHref" in some conditional logic and replaced it with a falsy value detection for the "href" prop.

Removed the assignment of the "role" prop from this conditional logic and simplified the remaining statement to not include a falsy dedection since it will now be redundant.

### DIFF
--- a/src/SafeAnchor.tsx
+++ b/src/SafeAnchor.tsx
@@ -29,7 +29,7 @@ const propTypes = {
   as: PropTypes.elementType,
 };
 
-function isTrivialHref(href) {
+function isTrivialHref(href?: string) {
   return !href || href.trim() === '#';
 }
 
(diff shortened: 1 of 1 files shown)


taion:
wait, this doesn't quite make sense to me. anything that's linking to `"#"` by definition is going to be button-like rather than link-like, no?

LeftyDaniels:
I can see that argument, and if that is the assumption that you want the component to make then that I guess we can disregard this.

MDN does suggest adding a "space" key event listener to go along with the `[role="button]` if you want a link to behave like a button.

image

They also recommend that if you can, use the native HTML button elements where possible.

image

Is there a preference on how to approach this?

taion:
@jquense what's your thought here?

jquense:
We are intentionally covering this case as a button bc it's an old school web dev pattern for a button. The goal was to intercept these usages and handle them correctly as buttons. I'd say if there is a valid use case where this should be a real anchor then let's go for it otherwise id say double down and improve the button emulation

Issue #5304, opened 2020-07-16
Add DropdownMenu responsive alignment


kyletsang:
https://getbootstrap.com/docs/4.5/components/dropdowns/#responsive-alignment

Thinking of using these props:

`alignRight`: boolean | 'sm' | 'md' | 'lg' | 'xl'
`alignLeft`: 'sm' | 'md' | 'lg' | 'xl'

cc @react-bootstrap/collaborators

jquense:
maybe something like the responsive props on Col? that way you can't specify bad combos, e.g. `align={{ 'sm': left', lg: 'right' }}` keep alignRight around for back compat?

kyletsang:
> 
> 
> maybe something like the responsive props on Col? that way you can't specify bad combos, e.g. `align={{ 'sm': left', lg: 'right' }}` keep alignRight around for back compat?

Sounds good to me

kyletsang:
Done in #5307

Issue #5186, opened 2020-05-19
Disabling Popper?


priley86:
Is there any known good way to disable Popper.js with the current Dropdown implementation? I am using the Dropdown from an API only perspective and leveraging an existing Bootstrap CSS implementation at my company. When I upgrade from `1.0.0-beta.14` to latest I see what appears to be some Poppper.js inline injected styles. Is there any way to disable Popper completely if I am not worried about offscreen support and styles conflict w/ my design libraries?

My dropdown looks fairly trivial like this at the moment...

```
import './dropdown.scss';

import classnames from 'classnames';
import React, { SyntheticEvent, useState } from 'react';
import { Dropdown as RBDropdown } from 'react-bootstrap';

import { DropdownItem, DropdownItemProps } from './DropdownItem';
import { DropdownMenu, DropdownMenuProps } from './DropdownMenu';
import { DropdownToggle, DropdownToggleProps } from './DropdownToggle';

export type SelectCallback = (eventKey: string, e: React.SyntheticEvent<unknown, Event>) => void;

export type DropdownMetadata = {
  source: 'select' | 'click' | 'rootClose' | 'keydown';
};

export enum DropdownDirection {
  Right = 'Right',
  Centered = 'Centered',
  Left = 'Left',
}

export interface DropdownProps {
  children: React.ReactNode;
  className?: string;
  direction?: DropdownDirection;
  onToggle?: (isOpen: boolean, event: SyntheticEvent<RBDropdown<'div'>, Event>, metadata: DropdownMetadata) => void;
  onSelect?: SelectCallback;
}

interface DropdownComposition {
  Item: React.FC<DropdownItemProps>;
  Menu: React.FC<DropdownMenuProps>;
  Toggle: React.FC<DropdownToggleProps>;
}

export const InnerDropdown: React.FC<DropdownProps> & DropdownComposition = ({
  children,
  className = '',
  direction = DropdownDirection.Right,
  onToggle = () => {},
  onSelect = () => {},
}) => {
  const [isActive, setIsActive] = useState(false);
  const classes = classnames(className, {
    'dropdown--center': direction === DropdownDirection.Centered,
    'dropdown--left': direction === DropdownDirection.Left,
    'dropdown--right': direction === DropdownDirection.Right,
    active: isActive,
  });
  const onDropdownToggle = (isOpen: boolean, e: SyntheticEvent<RBDropdown<'div'>, Event>, m: DropdownMetadata) => {
[...]

PR #5307, opened 2020-07-16
feat(DropdownMenu): Add responsive menu alignment

kyletsang:
#5304

### DIFF
--- a/src/DropdownMenu.tsx
+++ b/src/DropdownMenu.tsx
@@ -17,10 +17,20 @@ import {
   SelectCallback,
 } from './helpers';
 
+export type AlignDirection = 'left' | 'right';
+
+export interface ResponsiveAlignProp {
+  sm?: AlignDirection;
+  md?: AlignDirection;
+  lg?: AlignDirection;
+  xl?: AlignDirection;
+}
+
 export interface DropdownMenuProps extends BsPrefixPropsWithChildren {
   show?: boolean;
   renderOnMount?: boolean;
   flip?: boolean;
+  align?: ResponsiveAlignProp;
   alignRight?: boolean;
   onSelect?: SelectCallback;
   rootCloseEvent?: 'click' | 'mousedown';
@@ -28,6 +38,22 @@ export interface DropdownMenuProps extends BsPrefixPropsWithChildren {
 }
 
 type DropdownMenu = BsPrefixRefForwardingComponent<'div', DropdownMenuProps>;
+
+const DEVICE_SIZES = [
+  'xl' as const,
+  'lg' as const,
+  'md' as const,
+  'sm' as const,
+];
+
+const alignDirection = PropTypes.oneOf(['left', 'right']);
+
+export const alignPropTypes = PropTypes.shape({
+  sm: alignDirection,
+  md: alignDirection,
+  lg: alignDirection,
+  xl: alignDirection,
+});
 
 const propTypes = {
   /**
@@ -43,6 +69,14 @@ const propTypes = {
 
   /** Have the dropdown switch to it's opposite placement when necessary to stay on screen. */
   flip: PropTypes.bool,
+
+  /**
+   * Aligns the dropdown menu responsively depending on the breakpoint.
+   * The alignment direction will affect the specified breakpoint or larger.
+   *
+   * *Note: Using responsive alignment will disable Popper usage for positioning.*
+   */
+  align: alignPropTypes,
 
   /** Aligns the Dropdown menu to the right of it's container. */
   alignRight: PropTypes.bool,
@@ -86,6 +120,7 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
     {
       bsPrefix,
       className,
+      align,
       alignRight,
       rootCloseEvent,
       flip,
@@ -102,6 +137,16 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
     const prefix = useBootstrapPrefix(bsPrefix, 'dropdown-menu');
     const [popperRef, marginModifiers] = usePopperMarginModifiers();
 
+    const alignClasses: string[] = [];
+    if (align) {
+      DEVICE_SIZES.forEach((brkPoint) => {
+        const direction = align[brkPoint];
+        if (direction) {
+          alignClasses.push(`${prefix}-${brkPoint}-${direction}`);
+        }
+      });
+    }
+
     const {
       hasShown,
       placement,
@@ -114,7 +159,7 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
       rootCloseEvent,
       show: showProps,
       alignEnd: alignRight,
-      usePopper: !isNavbar,
+      usePopper: !isNavbar && alignClasses.length === 0,
       popperConfig: {
         ...popperConfig,
         modifiers: marginModifiers.concat(popperConfig?.modifiers || []),
@@ -137,12 +182,14 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
       (menuProps as any).close = close;
       (menuProps as any).alignRight = alignEnd;
     }
+
     if (placement) {
       // we don't need the default popper style,
       // menus are display: none when not shown.
       (props as any).style = { ...(props as any).style, ...menuProps.style };
       props['x-placement'] = placement;
     }
+
     return (
       <Component
         {...props}
@@ -151,7 +198,8 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
           className,
           prefix,
           show && 'show',
-          alignEnd && `${prefix}-right`,
+          (alignEnd || alignClasses.length) && `${prefix}-right`,
+          ...alignClasses,
         )}
       />
     );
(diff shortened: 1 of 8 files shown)


src/DropdownMenu.tsx
```diff
@@ -29,6 +39,22 @@ export interface DropdownMenuProps extends BsPrefixPropsWithChildren {
 
 type DropdownMenu = BsPrefixRefForwardingComponent<'div', DropdownMenuProps>;
 
+const DEVICE_SIZES = [
+  'xl' as const,
+  'lg' as const,
+  'md' as const,
+  'sm' as const,
+];
```
taion, line 47: **suggestion:** i think you can do:

```suggestion
const DEVICE_SIZES = [
  'xl',
  'lg',
  'md',
  'sm',
] as const;
```

src/DropdownMenu.tsx
```diff
@@ -114,7 +159,7 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
       rootCloseEvent,
       show: showProps,
       alignEnd: alignRight,
-      usePopper: !isNavbar,
+      usePopper: !isNavbar && alignClasses.length === 0,
```
taion, line 162: so we don't want the "flip" behavior when specifying alignment explicitly?

src/DropdownMenu.tsx
```diff
@@ -29,6 +39,22 @@ export interface DropdownMenuProps extends BsPrefixPropsWithChildren {
...
+const DEVICE_SIZES = [
+  'xl' as const,
+  'lg' as const,
+  'md' as const,
+  'sm' as const,
+];
+
+const alignDirection = PropTypes.oneOf(['left', 'right']);
+
+export const alignPropTypes = PropTypes.shape({
```
taion, line 51: `alignPropType` might be a better name?

src/DropdownMenu.tsx
```diff
@@ -44,6 +70,14 @@ const propTypes = {
   /** Have the dropdown switch to it's opposite placement when necessary to stay on screen. */
   flip: PropTypes.bool,
 
+  /**
+   * Aligns the dropdown menu responsively depending on the breakpoint.
+   * The alignment direction will affect the specified breakpoint or larger.
+   *
+   * *Note: Using responsive alignment will disable Popper usage for positioning.*
+   */
+  align: alignPropTypes,
```
taion, line 79: does it make sense to support specifying _just_ `'left'` or `'right'` to have this take effect at all sizes?

taion: e.g. can we merge with and/or deprecate `alignRight`? right now there's e.g. no easy way to force left- (resp. right-) align at all sizes and disable popper

src/DropdownMenu.tsx
kyletsang: Upstream bootstrap uses styles that are incompatible with popper positioning, so they say to disable it.

>If you want to use responsive alignment, disable dynamic positioning by adding the data-display="static" attribute and use the responsive variation classes.

The responsive examples unfortunately do not support flip

src/DropdownMenu.tsx
kyletsang: To clarify, you mean something like this?
```
align?: 'left' | 'right' | ResponsiveAlignProp
```
Where `left` and `right` has the ability to use/not use popper and any responsive align = immediately disable popper?

I guess we can add a `usePopper` prop to the dropdown menu too.  Knock out #5186 at the same time.

taion: yeah... something like that? but maybe we can make `usePopper` just default to `true` unconditionally? or should we default it to false if `align` is specified?

kyletsang: I think we still have to do it conditionally due to responsive alignment classes not working with popper.  `left` and `right` should probably follow the default behavior where popper is enabled?

So `true` for `align={left|right}` and `false` for `align={ResponsiveAlignProp}`?

taion: ah, okay, so you can't enable popper at all in those cases? that sounds reasonable i guess. we'd just need to be careful to document it explicitly so it's not confusing. and then this would be less a matter of `usePopper` so much as "disable popper if this is set"

kyletsang: Thanks for the feedback.  I checked in the changes.

taion:
perhaps we want another type alias for the valid order values?

src/DropdownMenu.tsx
```diff
@@ -102,6 +143,21 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
     const prefix = useBootstrapPrefix(bsPrefix, 'dropdown-menu');
     const [popperRef, marginModifiers] = usePopperMarginModifiers();
 
+    const alignClasses: string[] = [];
+    if (align) {
+      if (typeof align === 'object') {
+        DEVICE_SIZES.forEach((brkPoint) => {
+          const direction = align[brkPoint];
+          if (direction) {
+            alignRight = alignRight || direction === 'left';
```
taion, line 152: **question:** i'm confused by this – we set `alignRight` to `true` if _any_ `direction` is `left`? that seems a little odd to me

src/DropdownMenu.tsx
```diff
@@ -102,6 +143,21 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
     const prefix = useBootstrapPrefix(bsPrefix, 'dropdown-menu');
     const [popperRef, marginModifiers] = usePopperMarginModifiers();
 
+    const alignClasses: string[] = [];
+    if (align) {
+      if (typeof align === 'object') {
+        DEVICE_SIZES.forEach((brkPoint) => {
```
taion, line 149: **suggestion:** it would be cleaner to do `Object.entries(align).forEach(([breakpoint, direction]) => ...)`

the `DEVICE_SIZES` here seems extraneous

src/DropdownMenu.tsx
```diff
@@ -102,6 +143,21 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
...
+          const direction = align[brkPoint];
+          if (direction) {
+            alignRight = alignRight || direction === 'left';
+            alignClasses.push(`${prefix}-${brkPoint}-${direction}`);
+          }
+        });
+      } else if (align === 'right') {
+        alignRight = true;
```
taion, line 157: **suggestion:** if we're deprecating `alignRight`, it might be better to implement `alignRight` in terms of setting `align = 'right'`

src/DropdownMenu.tsx
kyletsang: Sorry should've left a comment here.  In order to responsively align left, the `dropdown-menu-right` class needs to be added in addition to `dropdown-menu-**-left`.  I reused this bool to control toggling of that class.

https://getbootstrap.com/docs/4.5/components/dropdowns/#responsive-alignment

src/DropdownMenu.tsx
kyletsang: Originally I was thinking when we do remove `alignRight` from the API, I would just use this var locally so I can toggle the `dropdown-menu-right` class in the case of responsive left alignment.  See comment above re: responsive left classes requiring this class.

Would this be ok?

src/DropdownMenu.tsx
kyletsang: Thanks.  I used `Object.keys` instead for IE11 compatibility.

taion:
wait a sec... per https://github.com/react-bootstrap/react-bootstrap/issues/5304#issuecomment-659612548 – should we then have some sort of check that only a single key is specified in the object?

i.e. it wouldn't make sense to specify more than one align class, i think?

taion:
e.g. if i'm reading the bootstrap docs correctly, the type should be more like

```ts
AlignDirection
| { sm: AlignDirection }
| { md: AlignDirection }
| { lg: AlignDirection }
| { xl: AlignDirection }
```

taion:
cc @jquense – was that what you meant by "can't specify bad combos"?

src/DropdownMenu.tsx
kyletsang: Added note in a callout:

> Using responsive alignment will disable Popper usage so any dynamic positioning features such as <code>flip</code> will not work.

kyletsang:
Oops, was thinking multiple aligns were allowed.  I'll fix that up

taion:
well, are they? the upstream docs at least suggest that it makes little sense

kyletsang:
Adding the combination of classes do allow you to flip flop on diff screen sizes.  But yeah prob not a practical thing.

kyletsang:
Fixed code to only handle 1 breakpoint in the `align` object and warn if object doesn't contain exactly 1 breakpoint.

jquense:
I think you should be allows to specify combos together no? At least thats supported for other cases of responsive variants.

taion:
@jquense upstream doesn't show it https://getbootstrap.com/docs/4.5/components/dropdowns/#responsive-alignment

would a user really want to flip back and forth on direction? the way the "left" thing is set up doesn't quite enable that...

e.g. if you want it go to "left" at ">= md", we'd need to also add the "align right" as the default... otherwise we'd need a slightly different/clunkier API here? e.g. we'd have to force users to specify `alignEnd` as the "default" behavior?

PR #5309, opened 2020-07-18
Changed the link to point to tsx components

### LINKED ISSUE #5308: LinkToSource on the docs UI is broken after v1.0.0
## Describe the bug
After v1.0.0, we added typescript support. But the LinkToSource still points to .js 

## To Reproduce

1. Go to 'https://react-bootstrap.github.io/' (v.1.2.2) 
2. Click on 'Table' (or any component)
3. Scroll down to 'API' section
4.Click on 'Source Link'

## Expected behavior

The link should point to the github code (.tsx) when version is higher than 1.0.0

Please let me know if a PR can be raised. I would be happy to fix this

tanmaylaud:
This PR is raised to fix issue #5308
It is a simple issue but a silent one as well. I thought I should fix so that proud users like me can navigate on the UI easily.
Please accept.

fixes #5308

### DIFF
--- a/www/src/components/LinkToSource.js
+++ b/www/src/components/LinkToSource.js
@@ -14,7 +14,7 @@ const Link = styled('a')`
 
 export default (props) => {
   const { component } = props;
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

Issue #5313, opened 2020-07-20
The responsive Table section on the docs ui is not responsive (unless window is too small)


tanmaylaud:
## Describe the bug

When we go to [Docs UI Responsive Table](https://react-bootstrap.github.io/components/table/#table-responsive), we see that the Table example is not responsive unless the window is made too small. This may not be apparent for beginners and new users. Hence, new users may not get to know the beauty of responsive tables

## To Reproduce

1. Go to [Docs UI Responsive Table](https://react-bootstrap.github.io/components/table/#table-responsive)
2. Try to scroll horizontal
3. Not scrollable

## Expected behavior

We need to add more columns to the example ( may be 50 ) to show the power of React Boostrap.

It will good to add more columns to the Table (without making the example code lengthy) so that the Table is responsive

kyletsang:
Resolved by #5314

PR #5314, opened 2020-07-20
Example for responsive table changed ( Fixing issue #5313) 

tanmaylaud:
This PR fixes issue #5313
I have added 50 columns so that the responsive behaviour can be clearly displayed ( even on larger windows )
Also, a sentence is added to the doc explaining the example.


taion:
this seems like it makes the example quite unwieldy?

taion:
that's a lot of columns...

tanmaylaud:
> this seems like it makes the example quite unwieldy?

@taion  I have added many columns so that the responsive behaviour can be clearly shown. Even with browsers with large viewport. Regarding the example, I tried to keep it as simple as possible (without hurting the React Playground ). Could you suggest a better way to do it?

taion:
We'd prefer the examples to match the upstream ones as much as possible: https://getbootstrap.com/docs/4.5/content/tables/#responsive-tables

tanmaylaud:
> We'd prefer the examples to match the upstream ones as much as possible: https://getbootstrap.com/docs/4.5/content/tables/#responsive-tables

@taion Thanks for the reference. But, if you see, the upstream UI does not have a runnable playground like ours. So, we need to provide a runnable code. I had thought of directly putting the headers and cells. But then the code would become too lengthy even for 10 columns.

taion:
yeah... the issue here is that the code becomes a little bit awkward/unidiomatic, and it's not quite "proper" in the sense that we don't have `key`s on the elements, &c.

can we just make the lists longer?

taion:
like, if we get up to 12 columns, it should be enough to show the scrolling...

www/src/examples/Table/Responsive.js
```diff
@@ -2,41 +2,37 @@
   <thead>
     <tr>
       <th>#</th>
-      <th>Table heading</th>
-      <th>Table heading</th>
-      <th>Table heading</th>
-      <th>Table heading</th>
-      <th>Table heading</th>
-      <th>Table heading</th>
+      {(function () {
```
mxschmitt, line 5:
```suggestion
      {Array.from({length:5}).map(_ => <th>Table heading</th>)}
```
should work too for both cases.

tanmaylaud:
> like, if we get up to 12 columns, it should be enough to show the scrolling...

Yeah..true. but the example code gets really long. Should I use the snippet suggested by @mxschmitt  instead ?

taion:
in any event, it would be nice if this example set `key`s on the elements, at least

tanmaylaud:
> in any event, it would be nice if this example set `key`s on the elements, at least

@taion I have changed the number of columns to 12, and added a key to each cell

www/src/examples/Table/Responsive.js
tanmaylaud: @mxschmitt  Thanks ! 👍

PR #5316, opened 2020-07-20
feat(OverlayTrigger): allow renderProp pattern

jquense:
Adds 2 features. First, OverlayTrigger render prop for more nuanced positioning situations (added docs example). Second, adds a global shared delay for tooltips in toolbars.


src/useDelayToggleCallback.tsx
```diff
@@ -0,0 +1,90 @@
...
+let showHandle: NodeJS.Timeout;
+let leaveHandle: NodeJS.Timeout;
+const hideCallbacks = [] as Array<(nextValue: boolean) => void>;
+
+function drainHideCallbacks() {
+  hideCallbacks.forEach((h) => h(false));
+  hideCallbacks.length = 0;
+}
+
+function useGlobalTimeout(
```
jquense, line 23: this is extremely inelegant, open to simplifications.

kosmiq:
Super nice features!

I'll leave a proper review soon-ish. :)

src/useDelayToggleCallback.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+type ValueRef = {
+  entered?: boolean;
+  pending?: boolean;
+  hideCallbacks: Array<() => void>;
+};
+
+let entered: boolean;
+let showHandle: NodeJS.Timeout;
+let leaveHandle: NodeJS.Timeout;
```
taion, line 15: **question:** are these going to cause problems w/people not using node types?

src/useDelayToggleCallback.tsx
```diff
@@ -0,0 +1,90 @@
...
+function useGlobalTimeout(
+  delay: Delay,
+  callback: (nextValue: boolean) => void,
+) {
+  const [showDelay, leaveDelay] = delay;
+  return useCallback(
+    (nextValue: boolean) => {
+      clearTimeout(showHandle);
+
+      if (nextValue === true) {
```
taion, line 32: **thought:** seems pointless if we know `nextValue` is a bool?

```suggestion
      if (nextValue) {
```

src/useDelayToggleCallback.tsx
```diff
@@ -0,0 +1,90 @@
...
+function drainHideCallbacks() {
+  hideCallbacks.forEach((h) => h(false));
+  hideCallbacks.length = 0;
+}
+
+function useGlobalTimeout(
+  delay: Delay,
+  callback: (nextValue: boolean) => void,
+) {
+  const [showDelay, leaveDelay] = delay;
```
taion, line 27: **nitpick:** could destructure

```suggestion
  [showDelay, leaveDelay]: Delay,
  callback: (nextValue: boolean) => void,
) {
```

taion:
btw is the global thing present upstream?

jquense:
> if you're in the "show" state, no delay in switching from one to the other

That's the core of it yes, Once a tooltip is shown you want the next one to show immediately until an normal delayed exit occurs. 

I threw this together very quickly b/c it was annoying me and i was hoping i could avoid a more involved API. 

something like: [tippy](https://atomiks.github.io/tippyjs/#singleton) or the reach-ui tooltip

I think it's ok UX to not have more local or nuanced scoping, like a provider, but I agree that as is it's ugly and bad. Like you probably also want to skip the animation for already entered tips but a good way to do that didn't seem obvious. Maybe what you actually want to do is share a single tooltip component and change the content and reference, but that seems hard...

jquense:
Ok I split out the global timer stuff for now to not hold this up

syadav214:
UI is breaking for me on tooltip.

image

Issue #5297, opened 2020-07-14
Netlify claims 1.0.1 is the current version


kosmiq:
## Describe the bug

Visit https://react-bootstrap.netlify.app/. Note that it claims to be version 1.0.1 instead of 1.2.2.

## To Reproduce

1. Go to https://react-bootstrap.netlify.app/
2. Note the version number

## Expected behavior

Netlify should output whatever is the latest in master.

## Screenshots

![Screenshot 2020-07-14 at 12 39 08](https://user-images.githubusercontent.com/322759/87416477-0cb03c80-c5cf-11ea-8557-f1c950448886.png)

bpas247:
I went ahead and triggered a manual build to update the netlify deployment, but I'm still not sure why our builds stopped triggering automatically.

3scava1i3r:
Now we have this rectified so i think this issue should be closed
![image](https://user-images.githubusercontent.com/52003051/88573476-6fd5b080-d05e-11ea-9554-4203ef26497d.png)

PR #5318, opened 2020-07-21
feat(netlify): more resilient docs alert implementation

bpas247:
Continuation of #5078.

Uses the already built-in netlify configuration variables that
will be injected by the netlify build environment, which should
allow use to not have to configure anything extra to get this
feature to work properly.

Also adds some more information about the deployment, specifying
whether or not it's a pull request (and redirects to the right
pr link), or if it's being deployed from a specific branch.


taion:
wow, gatsby is magic?

bpas247:
> wow, gatsby is magic?

more like netlify is magic 😛 there are [build env variables](https://docs.netlify.com/configure-builds/environment-variables/#netlify-configuration-variables) that get injected when the docs are deployed.

I'm not entirely sure why the docs haven't auto-deployed though, I might trigger a manual deployment for this PR to ensure that everything works properly.

taion:
right, but i mean, gatsby would have to be smart enough to inject them into the static build so they're accessible there...

bpas247:
> right, but i mean, gatsby would have to be smart enough to inject them into the static build so they're accessible there...

yeah I'm kinda worried that gatsby might not be smart enough, but once I figure out why it isn't deploying this branch it should be easy to enough to confirm.

jquense:
There are also some Gatsby netlify plugins might be helpful idk

bpas247:
@jquense do you happen to know why deploy previews aren't being built for PRs? seems like something changed after March 29th, since deploy previews for PRs stopped building since then.

bpas247:
after further investigation, it looks like [this deployment](https://app.netlify.com/sites/react-bootstrap/deploys/5ed5ec09cde94a3190719c3f) (which was a manual deployment) somehow caused netlify's CD to stop working *entirely* (related to #5297), as there hasn't been another deployment since then.

I'm not entirely sure why though 🤔 might have to do with netlify's cache being cleared.

bpas247:
Update: I ended up opening a new issue thread on the netlify community forums to track what's going on with our CD and live preview deployment issues.
https://community.netlify.com/t/cd-and-live-previews-has-suddenly-stopped-working/19344

bpas247:
Update: Netlify issues have been resolved and this PR is ready for review 🚢

bpas247:
Should be ready for review now :shipit: ended up just injecting the build-time variables into the static build

www/src/components/MasterDocsAlert.js
```diff
@@ -0,0 +1,38 @@
...
+            href={`https://github.com/react-bootstrap/react-bootstrap/pull/${netlify.reviewId}`}
+          >
+            pull request
+          </Alert.Link>
+        </span>
+      );
+    } else {
+      deploymentDetails = `the ${netlify.branch} branch`;
+    }
+  }
```
taion, line 24: **nitpick:** inline this?

www/src/components/MasterDocsAlert.js
```diff
@@ -0,0 +1,38 @@
...
+        <Alert variant="warning" className="w-100">
+          You are currently viewing the auto-generated docs from{' '}
+          {deploymentDetails}. The docs for the current release are available at{' '}
+          <Alert.Link href="https://react-bootstrap.github.io/">
+            https://react-bootstrap.github.io/
+          </Alert.Link>
+        </Alert>
+      </Row>
+    </Container>
+  ) : null;
```
taion, line 37: **nitpick:** no need for ternary here

```suggestion
  return netlify && (
    <Container fluid>
      <Row>
        <Alert variant="warning" className="w-100">
          You are currently viewing the auto-generated docs from{' '}
          {deploymentDetails}. The docs for the current release are available at{' '}
          <Alert.Link href="https://react-bootstrap.github.io/">
            https://react-bootstrap.github.io/
          </Alert.Link>
        </Alert>
      </Row>
    </Container>
  );
```

www/src/components/MasterDocsAlert.js
taion: (without the netlify condition)

taion:
@bpas247 just to check – is it some transient thing here that we have 2 netlify deploy previews for this build?

bpas247:
> @bpas247 just to check – is it some transient thing here that we have 2 netlify deploy previews for this build?

I think somebody created a new [deployment](https://app.netlify.com/sites/laughing-clarke-0d7e9f/overview) since the old deployment wasn't working.

We should delete this deployment since we don't need it anymore. 

I think @jquense is the only person who has perms to delete deployments on netlify, but I can try to disable it in the meantime.

bpas247:
@taion is it cool if I merge? I can't think of anything else I want to do here haha

PR #5321, opened 2020-07-22
fix(Carousel): Wrap direction calculation in useEffect

### LINKED ISSUE #5319: Carousel is not correctly sliding from first index to last in 'prev' direction

I'm using v1.2.2. Here is an example (not sure if it's the same version) https://codesandbox.io/s/runtime-grass-l2cg4?file=/package.json

related to https://github.com/react-bootstrap/react-bootstrap/issues/3654

kyletsang:
Resolves #5319

Wraps direction calculation in a `useEffect` so it doesn't recalculate the wrong direction on the edges when using `StrictMode`.

Not sure how to add a unit test here.  I tested manually and didn't notice any issues.

### DIFF
--- a/src/Carousel.tsx
+++ b/src/Carousel.tsx
@@ -250,20 +250,24 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
     activeIndex || 0,
   );
 
-  if (!isSliding && activeIndex !== renderedActiveIndex) {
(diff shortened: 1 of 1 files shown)


taion:
i think i have a better fix in https://github.com/react-bootstrap/react-bootstrap/pull/5322

kyletsang:
Close in favor of #5322

PR #5322, opened 2020-07-22
fix: Fix clearing nextDirectionRef in <Carousel>

### LINKED ISSUE #5319: Carousel is not correctly sliding from first index to last in 'prev' direction

taion:
Fixes #5319


src/Carousel.tsx
```diff
@@ -265,6 +267,12 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
     setRenderedActiveIndex(activeIndex || 0);
   }
 
+  useEffect(() => {
+    if (shouldClearNextDirection) {
+      nextDirectionRef.current = null;
+    }
```
kyletsang, line 273: Tested the branch and this statement isn't being triggered for me.  It appears that `shouldClearNextDirection` is getting reset to `false` prior to running `useEffect`

taion: it's closing over the value from the most recent "render" pass. the issue is that strictmode is happier about discarding render results... so this is the side effect we need to put into `useEffect`

you can also repro the bug just by wrapping our example in a `<React.StrictMode>`

taion: and at least on my side, this fixes things

kyletsang: I ran the example without strictmode in my case, so since the directionref isn't being cleared, skipping slides using the indicators on the bottom uses the wrong direction animation.  

Sliding on edges using left/right is fixed tho

taion: I think this is fixed now! Could you take another look?

kyletsang: Using carousel indicators still moves in wrong direction :(

Attaching gif:

carousel

taion: It seems to work for me: https://gfycat.com/annualtautgreatargus

How did you get that repro?

kyletsang: Huh weird.  I just ran the docs locally from this branch.  Ran the uncontrolled example without strictmode

Windows 10
Chrome Version 84.0.4147.105
Firefox 79.0 (64-bit)

taion: I can't reproduce it on either Chrome or Firefox. Do you see it on https://deploy-preview-5322--react-bootstrap.netlify.app/components/carousel/ ?

kyletsang: doh ignore.  my branch wasn't updated..  sorry!

strictmode and non strictmode works

cjmaret:
this issue still exists for me

deumolo:
Was this ever solved ?

PR #5324, opened 2020-07-23
Adding more top space to SidePanel for better visibility

javascript-alex:
**Problem:**
Due to the "Black Lives Matter" banner at the top, the right hand Sidebar panel was partially cut off and not readable or clickable after scrolling down. Especially at the "Forms" menu. See Screenshot as well.
screenshot-sidepanel-before

**Solution:**
Adding a bit more space at the top to the Sidepanel solves the problem. See Screenshot below.
screenshot-sidepanel-after

Thank you very much for reviewing this and I hope this little fix is appreciated.

Cheers,
Alex


taion:
Can you add a comment noting that this extra value is to accommodate the top banner?

kyletsang:
Looks like this was fixed as part of #5316.

I can see the full side nav on latest docs at https://react-bootstrap.github.io.

taion:
ah, nice. guess we don't need this, then. thanks!

javascript-alex:
Great, thanks :)

PR #5328, opened 2020-07-24
v5: Add fullscreen modal support

kyletsang:
Adds prop `fullScreen` which supports `true` or breakpoints `sm` and up.

https://v5.getbootstrap.com/docs/5.0/components/modal/#fullscreen-modal

### DIFF
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -39,7 +39,7 @@ export interface ModalProps
     | 'backdropTransition'
   > {
   size?: 'sm' | 'lg' | 'xl';
-  fullScreen?: true | 'sm' | 'md' | 'lg' | 'xl';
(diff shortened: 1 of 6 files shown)


kyletsang:
Yeah majority is usually "up".  One exception from v4 that comes to mind is the fluid containers.  Those ones are 100% width up until the specified breakpoint.

BestUkpai:
wow so amazing

jquense:
The fluid containers map to the classes tho right? It's `container-sm` not `container-sm-down`, that these classes include the `down` makes wonder if we should be more explicit as well, if only in case they add a `*-up` variants? I don't have a lot of suggestions tho on prop naming :/ `maybe fullscreen='sm-down'` is ok?

kyletsang:
`fullscreen="sm-down"` works and is more clear.  I wonder if bootstrap will rename their container classes in the future to align the naming with behavior.  

I'll fix up this PR.

PR #5334, opened 2020-07-27
v5: Remove CardDeck component

kyletsang:
Removed CardDeck and added example where you can achieve the equivalent using grid system.

https://v5.getbootstrap.com/docs/5.0/migration/#cards
> Removed card decks in favor of the grid which adds more flexibility over responsive behavior.

https://v5.getbootstrap.com/docs/5.0/components/card/#grid-cards


kyletsang:
I think having our own is pretty handy.  It's nice to reference them from the react point of view.

PR #5337, opened 2020-07-28
v5: Combine native and custom checkboxes/radios

kyletsang:
V5 combines native and custom checkboxes, so we no longer need to specify `custom` prop.

This PR contains following changes:

- Remove `custom` and `bsCustomPrefix` props from `FormCheck`
- Remove `bsCustomPrefix` prop from FormCheckInput
- Remove `isStatic` prop from FormCheckInput.  `position-static` is no longer needed when rendering checks with no labels
- Do not output `bsPrefix` if no label is specified. [Upstream docs](https://v5.getbootstrap.com/docs/5.0/forms/checks/#without-labels) say to omit `.form-check` if there is no label


src/FormCheck.tsx
```diff
@@ -176,9 +148,9 @@ const FormCheck: FormCheck = (React.forwardRef(
           style={style}
           className={classNames(
             className,
-            bsPrefix,
-            custom && `custom-${type}`,
+            label && bsPrefix,
             inline && `${bsPrefix}-inline`,
+            type === 'switch' && 'form-switch',
```
kyletsang, line 153: Should we add a bsPrefix override for `form-switch`?

taion: maybe we should treat `<Switch>` as a separate component that passes down its own `bsPrefix` that defaults to `form-switch`?

kyletsang: The switch still needs the `.form-check` class alongside `.form-switch`

ie.  `<div class="form-check form-switch">`

I guess users would need ability to control both.

taion: ah, so like `switchBsPrefix` then?

kyletsang: Yeah.  Wanted to double check before I added something like that :)

Issue #5339, opened 2020-07-29
NavDropdown alignRight failing on typescript check


xiaohong3:
## Describe the bug

I updated react-bootstrap from 1.0.1 to 1.3.0, after that my component is failing typescript (3.9.7) checking.
I have a component:
```
<NavDropdown
  title="title"
  id="basic-nav-dropdown"
  alignRight={true}
>
...
</NavDropdown>
```
And it complains on the alignRight property:
`Property 'alignRight' does not exist on type 'IntrinsicAttributes & Pick<Pick<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>, "slot" | ... 253 more ... | "is"> & { ...; }, "ref" | ... 250 more ... | "is"> & BsPrefixProps<...> & NavDropdownProps & { ...; }'`

I checked types of 1.0.1, the NavDropdown is defined differently (extends from Dropdown component which has alignRight prop).

## Expected behavior

Run-time seems to be working just fine (when you have alignRight the dropdown menu will align correctly).
However, react-scripts build will fail because typescript check failed.
I think the NavDropdown should extends Dropdown's properties in this case?

## Environment (please complete the following information)

- Operating System: Linux mint 20
- Browser, Version Chromium Version 84.0.4147.89
- React-Bootstrap Version 1.3.0

taion:
Makes sense to me. Want to send in a PR?

xiaohong3:
I'd love to.
I checked the [NavDropdown](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/NavDropdown.tsx) component, and find that it extends `BsPrefixPropsWithChildren` but not `Dropdown` (although Dropdown extends BsPrefixPropsWithChildren too).
I am not very familiar with the code, so I need to ask: Is NavDropdown a full extension of Dropdown? Like it should contain all of Dropdown's props and some more

xiaohong3:
Oh sorry, I closed the issue accidentally. Not very familiar with the github community, only use it to see if any of my bugs are reported, this is my first time reporting anything.
I'd love to give it a try.

taion:
It can/should extend all of those interfaces. See e.g. the very similar https://github.com/react-bootstrap/react-bootstrap/blob/521fd9276c17532d382bc4dedd87844894da36dd/src/DropdownButton.tsx#L8-L17

PR #5341, opened 2020-07-29
Fix NavDropdown alignRight failing typescript check.

xiaohong3:
Fixes issue #5339 
I changed the `NavDropdownProps` to extends from `DropdownProps` instead of `BsPrefixPropsWithChildren`. This way it will include all of Dropdown's props.

### DIFF
--- a/src/NavDropdown.tsx
+++ b/src/NavDropdown.tsx
@@ -5,12 +5,10 @@ import Dropdown, { DropdownProps } from './Dropdown';
 import NavItem from './NavItem';
 import NavLink from './NavLink';
 import { BsPrefixRefForwardingComponent } from './helpers';
-import { PropsFromToggle } from './DropdownToggle';
 
 export interface NavDropdownProps
   extends DropdownProps,
-    Omit<React.HTMLAttributes<HTMLElement>, 'onSelect' | 'title'>,
-    React.PropsWithChildren<PropsFromToggle> {
+    Omit<React.HTMLAttributes<HTMLElement>, 'onSelect' | 'title'> {
   id: string;
   title: React.ReactNode;
   disabled?: boolean;
@@ -76,9 +74,6 @@ const NavDropdown: NavDropdown = (React.forwardRef(
       disabled,
       active,
       renderMenuOnMount,
-      href,
-      size,
-      variant,
       ...props
     }: NavDropdownProps,
     ref,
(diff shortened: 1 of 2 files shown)


src/NavDropdown.tsx
```diff
@@ -1,15 +1,12 @@
...
 import NavItem from './NavItem';
 import NavLink from './NavLink';
-import {
-  BsPrefixPropsWithChildren,
-  BsPrefixRefForwardingComponent,
-} from './helpers';
+import { BsPrefixRefForwardingComponent } from './helpers';
 
-export interface NavDropdownProps extends BsPrefixPropsWithChildren {
+export interface NavDropdownProps extends DropdownProps {
```
kyletsang, line 9: We should omit `as` from `DropdownProps` because `NavDropdown` doesn't use it.

taion:
do we need any of these here? https://github.com/react-bootstrap/react-bootstrap/blob/521fd9276c17532d382bc4dedd87844894da36dd/src/DropdownButton.tsx#L10-L11

src/NavDropdown.tsx
kyletsang: Hmm actually this might be difficult with `BsPrefixRefForwardingComponent`

kyletsang:
> 
> 
> do we need any of these here?
> 

Looks like we needed the `React.HTMLAttributes` to get stuff like `id` to work.  

`React.PropsWithChildren<PropsFromToggle>` to get `'href' | 'size' | 'variant' | 'disabled';`

taion:
okay, let's add that then? @xiaohong3

src/NavDropdown.tsx
```diff
@@ -1,15 +1,16 @@
...
-  BsPrefixRefForwardingComponent,
-} from './helpers';
+import { BsPrefixRefForwardingComponent } from './helpers';
+import { PropsFromToggle } from './DropdownToggle';
 
-export interface NavDropdownProps extends BsPrefixPropsWithChildren {
+export interface NavDropdownProps
+  extends DropdownProps,
+    Omit<React.HTMLAttributes<HTMLElement>, 'onSelect' | 'title'>,
+    React.PropsWithChildren<PropsFromToggle> {
```
kyletsang, line 13: We can drop `React.PropsWithChildren<PropsFromToggle>` actually. I looked thru the code and these props are not needed. Other than that, looks good!

Issue #3029, opened 2018-03-05
v4: Layout component?


jquense:
**Read first!** [Contributing](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md) | [Structuring a PR](https://github.com/react-bootstrap/react-bootstrap/blob/v4-support/V4-Contributing.md ) | [Bootstrap docs](https://getbootstrap.com/docs/4.0/layout/utilities-for-layout/)

v4 contains a ton more helper and utility classes for layout, flexbox, margins, padding, etc. It might be a good idea to try and wrap some of that up into a single abstract `Layout` or `Flex` component as a library supported way to use some of these new utilities. It's unclear to me whether using it internally would be a good idea but I think it'd be a useful value-add component

highskillzz:
Could you please explain more what do you mean?

chenesan:
Hi @jquense I'd like to take this one 😄

jquense:
@chenesan go for it! Setting expectations a bit first though so we are on the same page. I'm not sure what this component should/would look like, so there will probably be a lot of back and forth on the api and structure. It's not entirely clear yet whether adding it is even a good call, but we'd love to see some exploration!

chenesan:
@jquense Thanks! I'll read through the utility api docs and try to write some possible usage examples in the next few days.

chenesan:
Hi @jquense 

I think we can have a `wrapLayoutUtil` hoc to handle all the utility className for internal use:

```javascript

// hoc.js

const utilityClassNameMap = {
  dFlex: "d-flex",
  // and other layout utility...
}

const wrapLayoutUtil = (Component) => {
  return (props) => {
    const utilityClassNames = Object.keys(props)
    .map(propName => utilityClassNameMap[propName])
    .filter(className => !!className)
    return <Component
      {...this.props}
      className={classnames(utilityClassNames, this.props.className)}>
    </Component>
  }
}

export default wrapLayoutUtil


// Button.js (or any react-bootstrap component)

import wrapLayoutUtil from "./hoc"

class Button extends React.Component {
  // Implementation of Button component.
  // Must has a `className` props to receive passed custom class name
}

export default wrapLayoutUtil(Button)

// App.js (client code)

// will output <button className="btn d-flex"></button>
const App = () => <Button dFlex></Button>

```

With this way, It's easy to map all the utility class onto all library components. Also it's less error prone than using className directly.

[...]

PR #5343, opened 2020-07-30
feat(v5): drop form row and form inline support

bpas247:
Drops support for the form-row and form-inline classes, and
removes any doc references associated with either of them.
https://v5.getbootstrap.com/docs/5.0/migration/#forms


kyletsang:
I don't see the removal of the `inline` prop in Form.tsx

bpas247:
> I don't see the removal of the `inline` prop in Form.tsx

yeah and I also noticed I didn't clean up all of the references of both features, so I ended up going through everything and cleaned up everything I found.

The new diffs were much larger so I ended up rebasing and splitting it across two different commits.

jquense:
nice work folks, if it's not too much trouble for these PR's could we drop a reference link to the upstream changelog or src that inspired the change, should make it a bit easier for reviewers

src/Form.tsx
```diff
@@ -91,7 +75,7 @@ const FormImpl: Form = (React.forwardRef(
         className={classNames(
           className,
           validated && 'was-validated',
-          inline && `${bsPrefix}-inline`,
+          bsPrefix,
```
kyletsang, line 78: I think we can drop `bsPrefix` here?  I looked at the v5 docs and there is no `.form` class usage.

bpas247: Hmm, why was the `form` class dropped? Also, if it was dropped, should we just drop our Form component entirely?

I have noticed that a lot of other component libraries (e.g. MUI and Chakrai-UI) don't have an overall Form component, so I'm curious if we should carry that trend over to our library as well.

kyletsang: Took a closer look at bootstrap docs and it looks like there wasn't a `.form` class in 4.0 either.  I've seen usage of `.form-inline` and `.was-validated`.  

https://getbootstrap.com/docs/4.0/components/forms/

So in v5 with inline being dropped, this just leaves `.was-invalidated` to handle.  I guess it's nice to have Form around even if it's to handle 1 class... unsure :)

bpas247: Either way the diffs for dropping support for the Form component would warrant it to be in a separate PR altogether, so we should open a new issue and discuss dropping support there.

In the meantime I'll drop `bsPrefix` and get this merged 👍

jquense:
Given the heavier reliance on utility classes in v5 we should probably come up with an answer for #3029 and friends, since more and more of the "public" api of these components is shifting to these utils

PR #5350, opened 2020-07-31
fix(Form): Fix types

kyletsang:
For v5

Fixes a type error I saw in TS tests

### DIFF
--- a/src/Form.tsx
+++ b/src/Form.tsx
@@ -71,7 +71,7 @@ const FormImpl: Form = (React.forwardRef(
 ) as unknown) as Form;
 
 FormImpl.displayName = 'Form';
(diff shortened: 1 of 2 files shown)


src/Form.tsx
```diff
@@ -72,7 +71,7 @@ const FormImpl: Form = (React.forwardRef(
 ) as unknown) as Form;
 
 FormImpl.displayName = 'Form';
-FormImpl.propTypes = propTypes;
+FormImpl.propTypes = propTypes as any;
```
taion, line 74: i wonder if we should do this more broadly. inferred props from `propTypes` don't play super well with explicit TS types

kyletsang: Yeah this is so finicky :/

PR #5351, opened 2020-07-31
feat(v5): Drop InputGroupPrepend and InputGroupAppend

kyletsang:
https://v5.getbootstrap.com/docs/5.0/forms/input-group/

Note: 
I did run into an issue with the dropdowns that might need to be solved in another PR.  The new markup for dropdowns has the dropdown toggle and menu directly under the `InputGroup`.  

In v4 this was solved with `<DropdownButton as={InputGroup.Prepend}>`, but there's no parent element within `InputGroup` this time.

I don't think there's a way to solve this with the existing components... anybody have any suggestions?

Changes:
- Drop `InputGroupPrepend` and `InputGroupAppend`
- Replace input in `InputGroupCheckbox` with `FormCheckInput`
- Fix tests and examples

### DIFF
--- a/src/InputGroup.tsx
+++ b/src/InputGroup.tsx
@@ -5,14 +5,11 @@ import React from 'react';
 
 import createWithBsPrefix from './createWithBsPrefix';
 import { useBootstrapPrefix } from './ThemeProvider';
+import FormCheckInput from './FormCheckInput';
 import {
   BsPrefixPropsWithChildren,
   BsPrefixRefForwardingComponent,
 } from './helpers';
-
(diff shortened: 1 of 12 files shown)


jquense:
> I did run into an issue with the dropdowns that might need to be solved in another PR. The new markup for dropdowns has the dropdown toggle and menu directly under the InputGroup.

https://github.com/react-bootstrap/react-overlays/pull/843 may unlock some API improvements there, since we are trying to remove the need for Dropdown to render a wrapping element

kyletsang:
Thanks @jquense.  I'll merge this one first.

Issue #3765, opened 2019-05-09
Sidebar scroll not working


BeastMaster07:
## Describe the bug

Sidebar scroll is not working properly

## To Reproduce

1. Go to react-bootstrap docs
2. View any component 
3. Scroll the particular component and then scroll the sidebar .

## Expected behavior

Scroll any component and then scroll the sidebar , it is not working most of the times. It works only on clicking the sidebar and not by hover ( which is ideal, I guess).
If the sidebar scroll works, just double click on the main component screen and now try to scroll the sidebar, it won't work.

## Environment (please complete the following information)

- Operating System: Windows 10
- Browser, Version: Chrome 74
- React-Bootstrap Version: 1.0.0-beta.6

taion:
Ouch. It's worth noting that the same thing happens on upstream docs as well, though.

jquense:
what the heck? this seems like a change in how the browser determines the "active" scroll area, needing to "activate" it is new and weird

taion:
I think it might have to do with the double-sidebar layout. I don't see this on e.g. our single sidebar page.

jquense:
I can't find an upstream issue maybe they are also unaware

BeastMaster07:
Any updates?

peterkmurphy:
This issue has been around for more than a year. I observe it both on Firefox 74.0 and Google Chrome 83.0 (on Linux Mint 18.3).

@taion : any idea what can be done and when it can be solved? Because this is a serious **_disincentive_** for using react-bootstrap. I don't want to be nasty, but a product in such as state that it destroys its own documentation....

jquense:
> any idea what can be done and when it can be solved?

fastest way to get something fixed is to send a PR. 

> but a product in such as state

This isn't a product, it's an open source project built by people in their free time with free labor.

>  I don't want to be nasty

Then don't be. be productive and jump in and help

peterkmurphy:
@jquense: I can do my little bit by observing that on the[ Components page](https://react-bootstrap.github.io/components/alerts/), I see:

```
@media (min-width:768px) {
 .SideNav-OverflowWrapper-module--cls2--2_4Kd {
  overflow:hidden;
  display:block!important;
  height:100%!important
 }
}
```
Turn that overflow to "auto", and your problem is solved. But would this one-liner fix causing problems elsewhere? 

[...]

PR #5358, opened 2020-08-04
Solving sidebar scroll not working #3765

peterkmurphy:
Hopefully this minor change is all that is necessary.


taion:
@jquense this isn't _quite_ right layout-wise because of the negative margin business, but it's an improvement... should we go ahead with this?

jquense:
ya seems ok to me

PR #5359, opened 2020-08-04
feat(carousel): add individual item intervals

### LINKED ISSUE #5305: Add individual CarouselItem intervals
https://getbootstrap.com/docs/4.5/components/carousel/#individual-carousel-item-interval

bpas247:
Adds the ability to specify intervals per-item, which should
hopefully match the functionality of the upstream feature.

TODOs:
- [x] Fix false positive issues from the feature test
- [x] Add prop types for the `interval` prop for `CarouselItem`

fixes #5305

### DIFF
--- a/src/Carousel.tsx
+++ b/src/Carousel.tsx
@@ -505,7 +505,7 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
 
     intervalHandleRef.current = window.setInterval(
       document.visibilityState ? nextWhenVisible : next,
-      activeChildIntervalRef.current || interval || undefined,
+      activeChildIntervalRef.current ?? interval ?? undefined,
     );
 
     return () => {
--- a/src/Carousel.tsx
+++ b/src/Carousel.tsx
@@ -251,12 +251,6 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
     activeIndex || 0,
   );
 
-  const intervals = useRef<number[]>([]);
-
-  const intervalRef = useCommittedRef(
-    intervals.current[activeIndex as number] ?? interval,
-  );
-
   if (!isSliding && activeIndex !== renderedActiveIndex) {
     if (nextDirectionRef.current) {
       setDirection(nextDirectionRef.current);
@@ -278,12 +272,18 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
   });
 
   let numChildren = 0;
+  let activeChildInterval: number | undefined;
+
   // Iterate to grab all of the children's interval values
   // (and count them, too)
   forEach(children, (child, index) => {
-    intervals.current[index] = child.props.interval as number;
-    numChildren++;
+    ++numChildren;
+    if (index === activeIndex) {
+      activeChildInterval = child.props.interval as number | undefined;
+    }
   });
+
+  const activeChildIntervalRef = useCommittedRef(activeChildInterval);
 
   const prev = useCallback(
     (event) => {
@@ -505,7 +505,7 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
 
     intervalHandleRef.current = window.setInterval(
       document.visibilityState ? nextWhenVisible : next,
-      intervalRef.current,
+      activeChildIntervalRef.current || interval || undefined,
     );
 
     return () => {
@@ -513,7 +513,7 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
         clearInterval(intervalHandleRef.current);
       }
     };
-  }, [shouldPlay, next, intervalRef, interval, nextWhenVisible]);
+  }, [shouldPlay, next, activeChildIntervalRef, interval, nextWhenVisible]);
 
   const indicatorOnClicks = useMemo(
     () =>
--- a/src/Carousel.tsx
+++ b/src/Carousel.tsx
@@ -250,6 +250,8 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
     activeIndex || 0,
   );
 
+  const intervals = useRef<number[]>([]);
+
   if (!isSliding && activeIndex !== renderedActiveIndex) {
     if (nextDirectionRef.current) {
       setDirection(nextDirectionRef.current);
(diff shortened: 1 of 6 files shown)


test/CarouselSpec.js
```diff
@@ -315,6 +315,43 @@ describe('<Carousel>', () => {
...
+      expect(onSelectSpy).to.have.been.calledOnce;
+      expect(onSelectSpy.firstCall).to.have.been.calledWith(1);
+
+      clock.tick(intervals[1] * 2);
+      expect(onSelectSpy).to.have.been.calledTwice;
+      expect(onSelectSpy.secondCall).to.have.been.calledWith(2);
+
+      clock.tick(intervals[2]);
+      expect(onSelectSpy).to.have.been.calledThrice;
+      expect(onSelectSpy.thirdCall).to.have.been.calledWith(0);
```
bpas247, line 352: Ticking the clock like this seems to lead to pretty inconsistent behavior... not sure if just ticking it once and make the assertions I want to make would lead to more consistent results.

test/CarouselSpec.js
```diff
@@ -315,6 +315,39 @@ describe('<Carousel>', () => {
...
+
+      const onSelectSpy = sinon.spy();
+      mount(
+        <Carousel interval={defaultInterval} onSelect={onSelectSpy}>
+          {itemsWithIntervals}
+        </Carousel>,
+      );
+
+      const total = intervals.reduce((sum, current) => sum + current, 0);
+      clock.tick(total * 1.1);
```
bpas247, line 343: Yep, this is still causing consistency issues 😆 anyone gotta better idea for a more consistent approach to testing this feature?

src/Carousel.tsx
```diff
@@ -487,17 +489,23 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
...
+      chosenInterval,
     );
 
     return () => {
       if (intervalHandleRef.current !== null) {
         clearInterval(intervalHandleRef.current);
       }
     };
-  }, [shouldPlay, next, interval, nextWhenVisible]);
+  }, [shouldPlay, next, activeIndex, interval, nextWhenVisible]);
```
taion, line 508: i think we just want:

```js
const intervalRef = useCommittedRef(
  intervals.current[activeIndex] ?? interval,
)
```

using non-committed refs is unsafe (as the other carousel issue observes), and we don't want the effect hook to unnecessarily invalidate, either

src/Carousel.tsx
```diff
@@ -542,6 +550,7 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
       <div className={`${prefix}-inner`}>
         {map(children, (child, index) => {
           const isActive = index === renderedActiveIndex;
+          intervals.current[index] = child.props.interval as number;
```
taion, line 553: put this in the `map` clause where we count `numChildren`

bpas247: https://github.com/react-bootstrap/react-bootstrap/blob/3b0091002defe502c69a7c5a7ab09e75c090e13c/src/Carousel.tsx#L270-L272
Hmm, I'm not exactly sure where the `map` clause would be here... I'm probably missing something obvious 😆

src/Carousel.tsx
```diff
@@ -1,5 +1,6 @@
 import useEventCallback from '@restart/hooks/useEventCallback';
 import useUpdateEffect from '@restart/hooks/useUpdateEffect';
+import useComittedRef from '@restart/hooks/useCommittedRef';
```
taion, line 3:
```suggestion
import useCommittedRef from '@restart/hooks/useCommittedRef';
```

src/Carousel.tsx
taion: in https://github.com/react-bootstrap/react-bootstrap/blob/ba7d0f3a8e300b8b09d37f575fed51fac2b6f725/src/Carousel.tsx#L275-L277 – change that to a `forEach` or something instead of a `.length`, and extract both the current interval _and_ the number of children

src/Carousel.tsx
bpas247: 😂 default exports are *so* great

src/Carousel.tsx
```diff
@@ -250,6 +251,12 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
     activeIndex || 0,
   );
 
+  const intervals = useRef<number[]>([]);
+
+  const intervalRef = useCommittedRef(
+    intervals.current[activeIndex as number] ?? interval,
+  );
+
```
taion, line 259: use the intervals below.. .see

```suggestion
```

src/Carousel.tsx
```diff
@@ -265,9 +272,13 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
...
-  const numChildren = React.Children.toArray(children).filter(
-    React.isValidElement,
-  ).length;
+  let numChildren = 0;
+  // Iterate to grab all of the children's interval values
+  // (and count them, too)
+  forEach(children, (child, index) => {
+    intervals.current[index] = child.props.interval as number;
+    numChildren++;
+  });
```
taion, line 281: let's try

```suggestion
  let numChildren = 0;
  let activeChildInterval = null;

  // Iterate to grab all of the children's interval values
  // (and count them, too)
  forEach(children, (child, index) => {
    ++numChildren;
    if (index === activeIndex) {
      activeChildInterval = child.props.interval as (number | undefined);
    }
  });

  const activeChildIntervalRef = useCommittedRef(activeChildInterval);
```

bpas247:
> we should also update typings on `Carousel.Item` to include this new `interval` prop

yep I have that in my TODOs 👍 main issue with that is that we won't be able to use the `createWithBsPrefix` HOC anymore, since it doesn't support allowing to pass `propTypes` to the underlying component

src/Carousel.tsx
```diff
@@ -494,15 +505,15 @@ function CarouselFunc(uncontrolledProps: CarouselProps, ref) {
 
     intervalHandleRef.current = window.setInterval(
       document.visibilityState ? nextWhenVisible : next,
-      interval || undefined,
+      activeChildIntervalRef.current || interval || undefined,
```
taion, line 508: this will allow `interval`s to be set to 0 (which is useless, but not meaningless)

```suggestion
      activeChildIntervalRef.current ?? interval ?? undefined,
```

test/CarouselSpec.js
taion: not really... and this isn't directly asserting that item intervals individually get respected.

maybe just set something up where the carousel has an interval of `1000`, the first item has an interval of `100`, and make sure that we're on the 2nd item after `200` units have lapsed?

Issue #3459, opened 2019-02-09
How to override modal-content class in Modal?


zverexe:
Hi, I'm using Modal component and cannot override modal-content class or add some custom class to that div. I'm using css modules. I cannot find in the docs appropriate way of adding custom components to the Modal except modalClassName

taion:
set the `bsPrefix` on the `<ModalDialog>` to something other than `modal`; the class is calculated from there

alternatively, use a nested selector on whatever your modal class is, and the global `.modal-content`

zverexe:
I use react bootstrap react-bootstrap": "^0.32.1" and I think I cannot use bsPrefix there

zverexe:
@taion ?

hazemyth:
Could we re-open this issue?

> set the `bsPrefix` on the `<ModalDialog>` to something other than `modal`; the class is calculated from there
> 
> alternatively, use a nested selector on whatever your modal class is, and the global `.modal-content`

The recommended route would also alter the class of the Modal.Dialog component. As far as I can see, there isn't any way to modify the class of the model-content element specifically.

Most of the components of the  modal can be targeted in precise fashion, either directly with props such as 'as' or 'className' or indirectly with specialized props such as 'dialogAs' or 'backdropClassName', when the component/element comes packaged inside another.

However, the model-content element comes packaged inside the Modal.Dialog component, without any independent means of modification that I can identify.

taion:
sure, that's a slightly different question – if you need to modify that class name for some reason, like you're using CSS modules and you want to compose stuff in...

We'd be happy to take a PR that adds a `contentClassName` prop or something.

akzarma:
any update on this?

taion:
As above, we'd be happy to take a PR to add a `contentClassName` prop.

Gantushig77:
Same here.

PR #5367, opened 2020-08-06
create contentClassName prop

IHIutch:
Adds contentClassName prop for modal, see https://github.com/react-bootstrap/react-bootstrap/issues/3459

### DIFF
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -484,10 +484,9 @@ const Modal: Modal = (React.forwardRef(
           role="document"
           onMouseDown={handleDialogMouseDown}
           className={dialogClassName}
+          contentClassName={contentClassName}
         >
-          <div className={classNames(`${bsPrefix}-content`, contentClassName)}>
-            {children}
-          </div>
+          {children}
         </Dialog>
       </div>
     );
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -44,6 +44,7 @@ export interface ModalProps
   backdropClassName?: string;
   animation?: boolean;
   dialogClassName?: string;
+  contentClassName?: string;
   dialogAs?: React.ElementType;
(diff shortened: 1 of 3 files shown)


src/Modal.tsx
```diff
@@ -478,7 +485,9 @@ const Modal: Modal = (React.forwardRef(
           onMouseDown={handleDialogMouseDown}
           className={dialogClassName}
         >
-          {children}
+          <div className={classNames(`${bsPrefix}-content`, contentClassName)}>
```
taion, line 488: **issue:** why move this here instead of just passing `contentClassName` down to `<ModalDialog>`? this is technically a breaking change in the API

IHIutch: Honestly, because Im new to this and am not really familiar! Just want to see this made, what do I need to do?

IHIutch: I also thought it made sense to go here since backdropClassName and dialogClassName both exist on Modal

taion: pass down `dialogClassName` from the modal component to the dialog component, and keep rendering that `<div>` inside the dialog component

also, it'd be nice to see a test for this

IHIutch:
Like so?

src/Modal.tsx
IHIutch: Updated.

> it'd be nice to see a test for this

If you have an example of what you're looking for, I'll get you what you need.

taion: yup, take a look at https://github.com/react-bootstrap/react-bootstrap/blob/f846dfbe9b8d85fa9ec8cd3d7e50d1310fd12220/test/ModalSpec.js#L192-L199

IHIutch: Thanks, added

IHIutch:
Thanks for the feedback, cheers

PR #5368, opened 2020-08-06
Bring back noCaret

IHIutch:
Forgive my ignorance, but would something like this work to hide the caret on dropdowns? This is one of the first issues I ran into when I first started using this library. This also would keep bsPrefix intact,

### DIFF
--- a/src/DropdownToggle.tsx
+++ b/src/DropdownToggle.tsx
@@ -18,6 +18,7 @@ export interface DropdownToggleProps
   split?: boolean;
   childBsPrefix?: string;
   eventKey?: any; // TODO: fix this type
+  noCaret?: boolean;
 }
 
 type DropdownToggle = BsPrefixRefForwardingComponent<
@@ -47,6 +48,11 @@ const propTypes = {
   as: PropTypes.elementType,
 
   /**
+   * Removes default .dropdown-toggle.
+   */
+  noCaret: PropTypes.bool,
+
+  /**
    * to passthrough to the underlying button or whatever from DropdownButton
    * @private
    */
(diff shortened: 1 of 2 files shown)


taion:
Can't users just do `bsPrefix=""` for cases like this?

IHIutch:
It doesnt work with an empty string

taion:
That's probably a bug in `useBootstrapPrefix`, I think? It's using `||` when it should be using `??`. I don't have a strong view on whether we should add `noCaret`. @react-bootstrap/collaborators ?

kyletsang:
The `.dropdown-toggle` class is also responsible for the border radius on the right side of the button in some scenarios like when it's nested under a button group.  

I think it's better for users to just create a new class to remove it... something like this?

```
.nocaret::before {
  content: none !important;
}
.nocaret::after {
  content: none !important;
}
```

IHIutch:
Do you have an example of that? I looked through the CSS and don't recall seeing anything to that effect

kyletsang:
Check the example with the different color variants under the `.btn-group` here:
https://getbootstrap.com/docs/4.5/components/dropdowns/

Use dev tools to delete the `.dropdown-toggle` class.

You'll see something like this:

image

IHIutch:
Ahhh, yes I see, thanks

IHIutch:
Does that also happen when modifying bsPrefix though?

IHIutch:
Also, interestingly, that appears to not be the case the react-bootstrap docs 
image

https://react-bootstrap.netlify.app/components/button-group/#nesting

kyletsang:
I think we can close this?  There's no good way to remove the caret with the toggling of bootstrap classes.

jefelewis:
This worked for me!

```css
.dropdown-toggle::after{
    display: none;
}
```

mednche:
The ```.dropdown-toggle::after{
    display: none;
}``` trick used to work but no longer works in production environment since upgrading to React 18. Is there any patch for this since React 18?

PR #5371, opened 2020-08-07
feat(ToggleButtonGroup): Add vertical option

eisnstein:
I am using react-bootstrap in a React/TypeScript project and after upgrading from 1.0.1 to 1.3.0 TypeScript tells me, that "Property 'vertical' does not exist on..." on the `ToggleButtonGroup` component.

So here I added the `vertical` property to the `ToggleButtonGroup` component.

(right now I am making heavy use of `// @ts-ignore` before every ToggleButtonGroup and it works as expected)


eisnstein:
Added change as proposed.

PR #5373, opened 2020-08-09
feat(v5): add layout

### LINKED ISSUE #5363: v5: Layout component!
# The "What"

From #3029:
> v4 contains a ton more helper and utility classes for layout, flexbox, margins, padding, etc. It might be a good idea to try and wrap some of that up into a single abstract `Layout` or `Flex` component as a library supported way to use some of these new utilities. 

# The "Why"

From #5343:
> Given the heavier reliance on utility classes in v5 we should probably come up with an answer for #3029 and friends, since more and more of the "public" api of these components is shifting to these utils

From #5362:
> yep, having a Layout component would be great since we could handle three major use cases:
> 
> * provide a generic React component that our end users can directly use (instead of just using host elements)
> * unify our `as` prop APIs everywhere - since any component that uses the `as` prop can just use the Layout component internally and have that handle the functionality for that
> * improve the TypeScript DX for our end-users that want to utilize any Bootstrap utility classes
> 
> Not to mention, this generic layout component API is not anything new in the component library scene (e.g. the [Box](https://material-ui.com/components/box/#api) component from MUI and the [Box](https://chakra-ui.com/box) component from Chakra-UI), so end-users who are switching from other component libraries to ours will already be familiar in some form to the idea behind it.

# The "How"

The scope of a component like this will be *huge*, but it does seem feasible enough to get this done before we officially release v5 support (since v5 isn't even officially released yet, lol)

We can use the fantastic work by @chenesan in #3089 as a starting point for the work that would need to be done here.

If we all think this is a good idea, I wouldn't mind taking point on this 😃 

## Things to Consider: Breaking Changes

Obviously the biggest roadblock for implementing this for v5 at this point in time is that it's still in alpha release, so chances of there being breaking changes for the utility classes is pretty high.

However I do believe that if we can come up with a flexible enough component implementation, these breaking changes shouldn't be too terrible to manage.

cc @jquense @taion @kyletsang

bpas247:
(special thanks to @chenesan for giving us a good base to start out with 😄)

Note: This PR is very much **WIP**, mainly opening this now to start gathering feedback on the API design.

TODOs:
- [ ] more complete documentation
- [ ] refactor existing primitive components to be composed with `Box`
- [ ] add the remaining flex APIs
- [ ] add borders API
- [ ] add colors API
- [ ] add float API
- [ ] add interaction API 
- [ ] add overflow API 
- [ ] add position API
- [ ] add shadow API 
- [ ] add sizing API 
- [ ] add spacing API 
- [ ] add text API 
- [ ] add vertical alignment API 

### Why `Box` over `Layout`?

The main reason why I prefer `Box` over `Layout` is because it's a fairly common name for a low-level utility component, so it would be an easier transition for end-users coming from other component libraries/frameworks.

And because it's shorter 😉 

### Why is the API so verbose?

Thinking back on [Jason's comment](https://github.com/react-bootstrap/react-bootstrap/pull/3089#issuecomment-375990461) in the original PR, it makes sense for our API to be explicit over terse. Keeping our API as flat as possible also has its advantages, especially when it comes to the complexity of our prop and typescript type implementations.

This also gives us the opportunity to add shorthand aliases to these prop names, mirroring the api design for [styled-systems](https://styled-system.com/api/).

Fixes #5363


jquense:
This is very cool, glad you are taking it up 💪 

I don't have much to say about the code atm, but worth hashing through the API a bit yeah? I do generally think explicit naming is better, if only b/c upstream is fairly explicit (vs something like tailwindcss). I think the tension and balance we need to strike comes between clear and explicit, and two other things.

### intuitive mapping to the upstream utils

It's important from an expectation and DX perspective that users have a good sense of how to consistently map their knowledge of bootstrap util classes to this component. The trouble tho there is we are more bound to how TWBS names things, even if they are inconsistent, eg. `d-inline-block` vs `px-1`.

### A Prop  API surface area that plays well with any components

This is, I think, the hardest part to get right, and most likely to cause frustration (at least from my experience with similar components). Because `Box` is so low level, users will _constantly_ use it with `as` to add layout to higher level components. OR if not using `as`, Box at least will underpin most components  so it's important that being able to easily spread through layout props through components works well. I think this is in direct tension with many flat props because there are potentially 10's of "extra" props that need to be passed through. We need to make sure that it's really easy to _type_ components that spread through Layout props as well as  do our best to keep those props from conflicting with other common names (explicit naming helps there!).

TBH I do think that a lot of those problems could be offset with discipline and good practice on users, but we would need to model that behavior at least to help folks down the right path (e.g. in general _don't_ spread through all layout props, only expose the ones you actually want to be configurable, _don't_ go crazy with `as` layers, etc)

taion:
Are you thinking a functional child thing?

Or maybe just helpers and/or hooks to calculate the merged class name from some sort of config object?

bpas247:
@jquense First of all, thanks for the amazing advice, I really appreciate the input as always! 🙏 

Secondly, going off of your second point, would it make more sense to have all of the utility APIs nested at a second level rather than all just at the surface? e.g. instead of doing
```jsx
<Box 
  flexDirectionRow="sm"
  alignItemsCenter="md"
/>
```
we could do something like
```jsx
<Box
  bs={{
    flexDirectionRow: "sm",
    alignItemsCenter: "md",
  }}
/>
```
the name of the surface prop can definitely be changed 😂 but this could help resolve any conflicts that we could potentially have with the former approach, while also still being relatively shallow as to avoid potential complexity issues with the API being too nested.

One downside to the latter approach is that our `react-docgen` integration has a *very* tough time parsing the potential values you can pass to props that are nested, so we'd have to tweak the configurations a bit to dig into the `Box` API a little bit more to generate the appropriate API docs.

If we still decided to go with the former approach, we could also abstract the logic in the `Box` component that's being used to calculate the appropriate classnames into its own utility function. That way our end-users could still use the utility logic of the `Box` component for use cases where using the `Box` component directly would be too complex (or annoying).

jquense:
Hey @bpas247 sorry for the silence, a bit short on time but not forgotten about this!

PR #5376, opened 2020-08-11
feat(v5): Set ColOrder types to max 5

kyletsang:
https://v5.getbootstrap.com/docs/5.0/migration/#utilities

> Decreased the number of responsive order utilities per breakpoint. The highest order utility with a number now is .order-5 instead of .order-12.

This PR just changes the TS types to restrict string numbers to max 5

### DIFF
--- a/src/Col.tsx
+++ b/src/Col.tsx
@@ -23,7 +23,8 @@ type NumberAttr =
   | '11'
   | '12';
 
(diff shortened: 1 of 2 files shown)


taion:
perhaps we want another type alias for the valid order values?

kyletsang:
> 
> 
> perhaps we want another type alias for the valid order values?

Not sure I understand.  Something in addition to `ColOrder`?

taion:
i mean something like `NumberAttr`

PR #5387, opened 2020-08-18
docs: Fix small typo in Dropdown documentation

mflores88:
Changed  'a' to 'of'. It reads a bit nicier and I think it's gramatically correct.

Before:
` but it's very specific to a certain kind a menu. <ARIA /> menus, must`

After:
` but it's very specific to a certain kind of menu. <ARIA /> menus, must`


jquense:
thanks!

PR #5345, opened 2020-07-30
v5: Update Badge classes

thakkaryash94:
[ x ] Change .badge-variant class to .bg-variant
[ x ] Change .badge-pill class to .rounded-pill

Link: https://github.com/react-bootstrap/react-bootstrap/projects/2#card-42008293

PR #5389, opened 2020-08-19
v5: Update Badge classes

thakkaryash94:
[ x ] Change .badge-variant class to .bg-variant
[ x ] Change .badge-pill class to .rounded-pill

Link: https://github.com/react-bootstrap/react-bootstrap/projects/2#card-42008293

New request from: https://github.com/react-bootstrap/react-bootstrap/pull/5345

### DIFF
--- a/src/Badge.tsx
+++ b/src/Badge.tsx
@@ -4,11 +4,12 @@ import React from 'react';
 
 import { useBootstrapPrefix } from './ThemeProvider';
 import { BsPrefixProps, BsPrefixRefForwardingComponent } from './helpers';
-import { Variant } from './types';
+import { Color, Variant } from './types';
 
 export interface BadgeProps extends BsPrefixProps {
-  variant?: Variant;
+  bg?: Variant;
   pill?: boolean;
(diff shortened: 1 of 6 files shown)


kyletsang:
Since all of these badge changes are now using util classes (border, bg color, text color), I'm thinking we should address this after #5373 is done.

@bpas247 what do you think?

bpas247:
> Since all of these badge changes are now using util classes (border, bg color, text color), I'm thinking we should address this after #5373 is done.

Hmm, that's a tough call since that PR isn't going to be done for awhile 🤔 not sure if I want this to be blocked for that long.

Maybe we could merge this (pending approval) and we'll just rebase in #5373 accordingly?

Worst-case scenario some (or all) of the work that's done here will get deprecated later on 😛

jquense:
ya lets merge this in, if we have larger changes that affect this later no big deal 👍

thakkaryash94:
Thank you everyone for your support

bpas247:
Absolutely 😄 thank you for contributing!

PR #5392, opened 2020-08-20
pass {props.children} via ...props in DropdownToggle.tsx

artcg:
it is better practice to include props.children directly to the component here (?) I can't think of a reason why it is passed separately like this...


artcg:
(Note this style matches the style in ./DropdownMenu.tsx so they should at least be consistent regardless of which style is preferred)

PR #5402, opened 2020-08-24
Add Yarn command to introduction page.

Zandor300:
For convenience sake.


jquense:
I believe we've generally left off these sort of duplications. It can be confusing for folks who don't know much about JS package management and just use what comes with node. On the other hand for folks that do use yarn, they likely know how to translate `npm` commands having decided to use the "not default"

kyletsang:
Closing as per above comment

PR #5416, opened 2020-09-02
Fix popover arrow misaligned when class is prefixed

juangl:
I'm facing a bug where the popover arrow isn't properly aligned with the overlay target when using prefixed classes. I'm solving the problem in user-space. Just took a moment to create a PR in case that this feels like a proper fix.
<div align="center">Screen Shot 2020-09-01 at 22 01 00</div>


jquense:
nice thanks!

kyletsang:
Any clean way we can extend this fix for those that pass in `bsPrefix` directly @jquense?

jquense:
Ah, hmm I don't think so. Not neatly anyway.

kyletsang (approved):

PR #5422, opened 2020-09-04
Attempt at fixing the onChange problem in the documentation

### LINKED ISSUE #5421: I think the docs are incomplete about FormFile.Input and FormFile
## Describe the bug

`FormFile` and `FormFile.Input` both accept an `onChange` but are referenced in the docs.

## To Reproduce

1. Go to http://localhost:8000/components/forms/#form-file-input-props
2. Notice there is no onChange documented.

1. Go to http://localhost:8000/components/forms/#form-file-props
2. Notice there is no onChange documented.

## Reproducible Example

Not Needed

## Expected behavior

All the props to be documented (I think they are pulled from the propTypes so this hints as a propTypes issue as well, but I'm not 100% sure.

## Screenshots

The one screenshot that fits on one screen....
image

## Environment

- Operating System: Not important
- Browser, Version Not important
- React-Bootstrap Version   "1.3.0"

## Additional context

I think the error stems from here (but not 100% certain) :
```
export interface FormFileProps
  extends BsPrefixPropsWithChildren,
    BsCustomPrefixProps,
    Pick<React.HTMLAttributes<HTMLElement>, 'style'> {
  inputAs?: React.ElementType;
  id?: string;
  disabled?: boolean;
  label?: React.ReactNode;
  custom?: boolean;
  isValid?: boolean;
  isInvalid?: boolean;
  feedback?: React.ReactNode;
  feedbackTooltip?: boolean;
  lang?: string;
  'data-browse'?: any; // ???
}
```

boatcoder:
fixes #5421 

Found the path being used to generate the docs and took a stab at fixing it.  It is probably incomplete but didn't want to invest the effort if the solution isn't acceptable to this part.

`yarn start` in the www folder does not auto-reload on changes though:

```
info changed file at /private/tmp/react-bootstrap/src/FormFileInput.tsx
success onPreExtractQueries - 0.032s
success extract queries from components - 0.181s
success write out requires - 0.002s
success Re-building development bundle - 2.683s
warn Warning: Event "xstate.after(1000)#waitingMachine.batchingNodeMutations" was sent to stopped service "waitingMachine". This service has already reached its final state, and will not transition.
Event: {"type":"xstate.after(1000)#waitingMachine.batchingNodeMutations"}
```

### DIFF
--- a/src/FormFileInput.tsx
+++ b/src/FormFileInput.tsx
@@ -15,6 +15,7 @@ export interface FormFileInputProps extends BsPrefixProps, BsCustomPrefixProps {
   isValid?: boolean;
   isInvalid?: boolean;
   lang?: string;
+  onChange?: (event: any) => void;
 }
 type FormFileInput = BsPrefixRefForwardingComponent<
   'input',
@@ -52,6 +53,10 @@ const propTypes = {
 
   /** The language for the button when using custom file input and SCSS based strings */
   lang: PropTypes.string,
+
+  /** Called when the value of the input changes with an `event`.
+      `event.target` is the `<input type='file'>` component with the changed value, `event.target.files` is the current list of files */
+  onChange: PropTypes.func,
 };
 
 const FormFileInput: FormFileInput = React.forwardRef(
--- a/www/yarn.lock
+++ b/www/yarn.lock
@@ -45,12 +45,21 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/generator@^7.10.3", "@babel/generator@^7.11.0", "@babel/generator@^7.11.4", "@babel/generator@^7.3.4", "@babel/generator@^7.9.6":
+"@babel/generator@^7.10.3", "@babel/generator@^7.11.0", "@babel/generator@^7.3.4", "@babel/generator@^7.9.6":
   version "7.11.4"
   resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.11.4.tgz#1ec7eec00defba5d6f83e50e3ee72ae2fee482be"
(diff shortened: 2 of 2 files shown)


www/yarn.lock
```diff
@@ -45,7 +45,7 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
-"@babel/generator@^7.10.3", "@babel/generator@^7.11.0", "@babel/generator@^7.11.4", "@babel/generator@^7.3.4", "@babel/generator@^7.9.6":
```
kyletsang, line 48: Yarn lock shouldn't need to change here

src/FormFileInput.tsx
```diff
@@ -15,6 +15,7 @@ export interface FormFileInputProps extends BsPrefixProps, BsCustomPrefixProps {
   isValid?: boolean;
   isInvalid?: boolean;
   lang?: string;
+  onChange?: (event: any) => void;
```
kyletsang, line 18: Better to extend from `InputHTMLAttributes` so we don't have to manually type onChange.

src/FormFileInput.tsx
```diff
@@ -52,6 +53,10 @@ const propTypes = {
 
   /** The language for the button when using custom file input and SCSS based strings */
   lang: PropTypes.string,
+
+  /** Called when the value of the input changes with an `event`.
+      `event.target` is the `<input type='file'>` component with the changed value, `event.target.files` is the current list of files */
```
kyletsang, line 58: Maybe this can be simplified to just `A callback fired when the file changes`

boatcoder:
This is why working on Opensource is so unrewarding.  No issues 2 years ago, bug is still probably unfixed....

Issue #5151, opened 2020-04-28
Switch issues without a label or ID


igorsantos07:
## Describe the bug
The Switch component behaves a little different from the Check in regards to the `label` and `id` attributes. It doesn't render without the label, and can't be toggled without the id. Both cases work fine with the standard checkbox.

## Reproducible Example

[Minimal example](https://codesandbox.io/s/brave-firefly-g2zzn?file=/src/App.js). I even included aria-label to see if that helped, but no. Passing an empty label is enough, but it should behave like the plain checkbox.

This is also noticeable on the doc live blocks.

taion:
That's just how it works upstream.

Do we just need to fix the types?

jquense:
It probably worth noting or warning about in the component? The underlying issue is that unless to link the link the label to input clicking the label has no effect on the underlying (hidden) checkbox. That's just how HTML works, but we could make that clear somewhere

igorsantos07:
I didn't understand the "upstream" until I went to the Boostrap website to look for the said switches. There's so much time I don't go there, I thought the switch was an addition by this library! :scream: 

It does work that way upstream, but it's a tad easier to understand because all HTML elements are well separated... The switch is appended to the label and defined in the wrapper div, not in the input, so if there's no label, there's no button to render. Given we don't need to provide a label element nor define the wrapper div (and at least I don't usually inspect the HTML generated by React), a notice about this known issue on the doc would be helpful, as @jquense mentioned.

Nonetheless, I guess a PropType setting label as required would be helpful? Not sure if React-Bootstrap could go a step further and "fix" it by defaulting the label to an empty string - if it encapsulates the label element, which is required, it could have a better default as well.

jquense:
Lets make the situation better then! Want to craft a PR @igorsantos07 ?

igorsantos07:
I could! Maybe in the next few weeks when I get rid of a second, time-consuming project :) Might be an interesting adventure into TS.

Nonetheless, I would like your guidance on what to do. Not sure about what's the "react-bootstrap way" on this matter - is it opinionated enough to create an empty label or such, for instance?

PR #5427, opened 2020-09-09
feat(FormCheck): Allow custom controls to render without a label

### LINKED ISSUE #4767: onChange not being called on Switch components (Form.Check type=switch)
## Describe the bug
This issue has closed but the problem persists. 
https://github.com/react-bootstrap/react-bootstrap/issues/4478

When using a `Form.Check` component, changing `type="checkbox"` to `type="switch"`, the onChange function no longer triggers on click. This is also true when using `Form.Switch`.

## To Reproduce

Create a  `Form.Check` with `type="switch"` prop and `onChange` function prop.
Click switch and notice function not being triggered.

kyletsang:
This PR improves the docs in the FormCheck component.

Fixes #4767 and #5151

- Mentions that `label` and `id` are required props for a switch and fixes proptypes to check this
- Fill in missing docs for inline, disabled, title, label props
- Fixes switch tests so it doesn't fail with new proptype check
- Fix lint warning in helpers.ts
- Change types to extend HTMLAttributes - I notice that when using inline event handlers, the vars are implicitly set as `any`.  This resolves that to the proper type.

### DIFF
--- a/src/FormCheck.tsx
+++ b/src/FormCheck.tsx
@@ -65,8 +65,10 @@ const propTypes = {
 
   /**
    * A HTML id attribute, necessary for proper form accessibility.
+   * An id is recommended for allowing label clicks to toggle the check control.
    *
-   * This is **required** when `type="switch"`.
+   * This is **required** for custom check controls or when `type="switch"` due to
+   * how they are rendered.
    */
   id: PropTypes.string,
 
@@ -100,13 +102,13 @@ const propTypes = {
 
   /**
    * Label for the control.
-   *
-   * This is **required** when `type="switch"`.
    */
   label: PropTypes.node,
 
   /** Use Bootstrap's custom form elements to replace the browser defaults */
-  custom: PropTypes.bool,
+  custom: all(PropTypes.bool, ({ custom, id }) =>
+    custom && !id ? Error('Custom check controls require an id to work') : null,
+  ),
 
   /**
    * The type of checkable.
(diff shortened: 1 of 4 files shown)


src/FormCheck.tsx
```diff
@@ -65,7 +63,11 @@ const propTypes = {
    */
   as: PropTypes.elementType,
 
-  /** A HTML id attribute, necessary for proper form accessibility. */
+  /**
+   * A HTML id attribute, necessary for proper form accessibility.
+   *
+   * This is **required** when `type="switch"`.
+   */
```
jquense, line 70: The ID also ensures that clicking on the label works for _any_ check, in the Switch case, the whole component is built around that behavior

src/FormCheck.tsx
```diff
@@ -81,9 +83,26 @@ const propTypes = {
...
+   * `title` attribute for the underlying `FormCheckLabel`.
+   */
   title: PropTypes.string,
+
+  /**
+   * Label for the control.
+   *
+   * This is **required** when `type="switch"`.
+   */
   label: PropTypes.node,
```
jquense, line 106: A label shouldn't actually be required to use the switch. The issue is we don't render a `label` element at all when it's not provided. More correctly we should render an empty label for the switch case I think.

src/FormCheck.tsx
kyletsang: I thought that might be implied with the accessibility bit.  Did you want to make this more explicit in the docs?

jquense: It is an a11y thing, i think it's also just a behavior that people expect and tend to forget why it works. So maybe worth mentioning explicitly just to be extra helpful?

kyletsang:
Thanks for the review @jquense.  I've modified this to render an empty label for both custom radio/check and switch cases since all custom controls won't function without the label.

jquense (approved):

PR #5430, opened 2020-09-12
fix(docs): roll-back resolution versions

bpas247:
Rolls back the resolution versions (borked by #5385) due to a known [issue](https://github.com/benjamn/ast-types/issues/383) with the `ast-types` dependency.


jquense:
Hopefully this should be addressed soon, looks like the fix upstream was merged

jquense (approved):

PR #5432, opened 2020-09-14
Fixing small type error in docs for form controls

lx4r:
When using React-Boostrap with TypeScript `rows="3"` will cause a type error: `Type 'string' is not assignable to type 'number | undefined'.`  
Thanks for maintaining React-Bootstrap!


kyletsang (approved):

kyletsang:
Thanks!

lx4r:
@kyletsang Thank you for the fast approval!

PR #5456, opened 2020-10-01
feat: add default variant to Alert and Badge

davidjb:
This sets a default variant of "primary" for Alert and Badge, allowing them to be used without having to set a variant.  Bootstrap's docs on Alerts/Badges require the use of a contextual class (https://getbootstrap.com/docs/4.5/components/alerts/) in order to have proper styling.

This aligns Alert and Badge variant functionality with that of Button and Navbar, where default variants are provided and can be used like `<Button>Hello</Button>`.

This could be viewed as a breaking change (e.g. for someone using Alert without a variant as it now adds one).  However, it could be viewed as a feature as it is adding & declaring functionality in the public API where behaviour wasn't previously explicitly defined in examples or docs.

### DIFF
--- a/src/Alert.tsx
+++ b/src/Alert.tsx
@@ -87,6 +87,7 @@ const propTypes = {
 };
 
 const defaultProps = {
+  variant: 'primary',
   show: true,
   transition: Fade,
   closeLabel: 'Close alert',
--- a/src/Badge.tsx
+++ b/src/Badge.tsx
@@ -43,6 +43,7 @@ const propTypes = {
 };
 
 const defaultProps = {
+  bg: 'primary',
   pill: false,
 };
 
--- a/test/AlertSpec.js
+++ b/test/AlertSpec.js
@@ -31,10 +31,19 @@ describe('<Alert>', () => {
       .simulate('click');
   });
 
-  it('Should have use variant class', () => {
+  it('Should default to variant="primary"', () => {
+    mount(<Alert>Message</Alert>).assertSingle(`.alert-primary`);
+  });
+
+  it('Should use variant class', () => {
     mount(<Alert variant="danger">Message</Alert>).assertSingle(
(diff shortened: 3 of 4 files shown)


src/Button.tsx
```diff
@@ -106,7 +106,7 @@ const Button: Button = React.forwardRef(
       className,
       prefix,
       active && 'active',
-      `${prefix}-${variant}`,
+      variant && `${prefix}-${variant}`,
```
jquense, line 109: this is definitely a bug fix, i would split it into another PR with just this change so that it doesn't get locked up in the discussion around the other feature

jquense (commented):
thoughts folks? I think this is probably reasonable consistency change, however i do think it'd have to be considered a breaking change so maybe (if accepted) should go in the v5 branch instead?

src/Button.tsx
davidjb: For sure, will do - thanks for the feedback.  Here we go: #5458.  Will remove this change from this PR shortly and force push.

kyletsang:
Sounds good to me.  Let's push it to v5

kdgyimah:
> thoughts folks? I think this is probably reasonable consistency change, however i do think it'd have to be considered a breaking change so maybe (if accepted) should go in the v5 branch instead?

I would agree on that as well.
I usually set up my Alerts to render primary as the default variant anyways 😄.

kyletsang:
@davidjb, sorry for the delay on this.  Can you rebase this and change `variant` in Badge to `bg`?  Things have changed in v2.  We should be good to merge after this.

Thanks!

davidjb:
@kyletsang Done!  That should be it - rebased, updated the Badge and its tests, plus updated commit message.

kyletsang (approved):

kyletsang:
Thanks!

PR #5458, opened 2020-10-01
fix(Button): display variant only when set

davidjb:
This bugfix aligns the `<button>` variant class implementation with Alert/Badge/Navbar -- only outputing a `prefix-variant` formatted class when a variant is specified and avoiding the situation of outputting `.btn-null` or `.btn-` when the variant is null or an empty string respectively.

This adds several tests to confirm that the default variant gets removed but also the invalid variants don't get outputted.

Split from #5456

### DIFF
--- a/src/Button.tsx
+++ b/src/Button.tsx
@@ -106,7 +106,7 @@ const Button: Button = React.forwardRef(
       className,
       prefix,
       active && 'active',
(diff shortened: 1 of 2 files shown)


jquense (approved):

jquense:
thanks!

PR #5462, opened 2020-10-04
Clarified the BrowserGlobalsCodeBlock in the GettingStardedPage with …

Sporego:
### Code block was confusing and could be more informative
*Used React.js Documentation comment style
*Clarified React.js Dependency `<script/>` tag
*Changed last script tag to `your code here` style instead of being ambiguous

### DIFF
--- a/www/src/components/BrowserGlobalsCodeBlock.js
+++ b/www/src/components/BrowserGlobalsCodeBlock.js
@@ -8,17 +8,23 @@ function BrowserGlobalsCodeBlock() {
     <CodeBlock
       mode="html"
       codeText={`
+/* This is React.js a React-Bootstrap depenancy... duhh*/ 
+/*Automatically added, if you are using a react framework (e.g., next.js, create-react-app, and gastby).*/
+
 <script src="https://unpkg.com/react/umd/react.production.min.js" crossorigin></script>
 
 <script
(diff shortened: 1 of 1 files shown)


jquense:
Thanks, I think we'll keep it simple and not have extra comments.

Sporego:
@jquense It is not simple and unclear because: 
1. If you follow the default, main, or recommended React tutorials @ [reactjs.org](https://reactjs.org/docs/create-a-new-react app.html#recommended-toolchains) you would have used React via create-react-app or some other recommended toolchain (what Facebook calls react framework).
    * Any unexperienced or new react dev would probably just copy and paste the whole code block... find that things are working and have unnecessary code in their project or worse get confused
1. `<script>var Alert = ReactBootstrap.Alert;</script>` can be `/*Your other script tags*/`. This is obviously clearer and more direct then the current example which for whatever reason references ReactBootsrap...

    Honestly, I'm a little butthurt because I was super excited to fork>commit>make a pull request. However, I understand why you may think my method of clarifying (e.g., too wordy, untasteful, "insert dissenting opinion 3" ) the documentation issues stated above isn't simple. I'm totally cool with that. 

My hope is you understand the issues stated above and solve it in whatever way you please. May your solution be ever clearer and simpler! 

Thanks and Best Wishes,

PR #5463, opened 2020-10-05
Improved Browser Globals Section of Getting Started Webpage

Sporego:
### Code block was confusing and could be more informative
*Used React.js Documentation comment style
*Clarified React.js Dependency `<script/>` tag
*Changed last script tag to `your code here` style instead of being ambiguous


Sporego:
Duplicate

PR #5466, opened 2020-10-06
fix(DropdownMenu): Fix base styles being overridden by popper styles

kyletsang:
Passing `style` to `DropdownMenu` currently gets overwritten with the style from popper instead of being merged.  This was a regression from the typescript merge.

### DIFF
--- a/src/DropdownMenu.tsx
+++ b/src/DropdownMenu.tsx
@@ -204,10 +204,11 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
       (menuProps as any).alignRight = alignEnd;
     }
 
+    let style = (props as any).style;
     if (placement) {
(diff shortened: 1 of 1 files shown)


src/DropdownMenu.tsx
```diff
@@ -204,17 +204,19 @@ const DropdownMenu: DropdownMenu = React.forwardRef(
       (menuProps as any).alignRight = alignEnd;
     }
 
+    let style = (props as any).style;
```
Garz4, line 207: Using `let` is usually for variables, but this is a constant. I would recommend you to use `const`.

kyletsang: This is the correct usage. I am reassigning the var in a block below.

jquense (approved):

PR #5471, opened 2020-10-09
feat(v5): Update dropdown

kyletsang:
https://v5.getbootstrap.com/docs/5.0/components/dropdowns/

Added the dark dropdown variant initially.

Changes:
- Set default element of DropdownDivider to hr
- Add variant prop for DropdownMenu
- Add menuVariant prop for DropdownButton and NavDropdown

I noticed that Bootstrap changed their dropdown menu/item markup back to using ul/li in https://github.com/twbs/bootstrap/pull/28591 . We should probably do the same here.  I wanted to get some feedback before adding those changes though. In `DropdownItem`, I could make it so the props control the `li` element, then apply the anchor specific props/classes to the inner anchor... but then this probably need additional props to control rendering of the anchor.  Alternatively, add a bare `li` wrapping element and keep the current implementation as is?

### DIFF
--- a/src/Dropdown.tsx
+++ b/src/Dropdown.tsx
@@ -20,6 +20,7 @@ const DropdownHeader = createWithBsPrefix('dropdown-header', {
   defaultProps: { role: 'heading' },
 });
 const DropdownDivider = createWithBsPrefix('dropdown-divider', {
+  Component: 'hr',
   defaultProps: { role: 'separator' },
 });
 const DropdownItemText = createWithBsPrefix('dropdown-item-text', {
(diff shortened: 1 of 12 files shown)


jquense (approved):
looks great.

re: 'li' tags, i think it'd be hard to default to them given how flexible menus are intended to be. I'd be worried that folks would accidentally great more bad markup. We should probably do the right thing in doc examples tho via `as` maybe?

kyletsang:
Won't be able to use `as` here since we still need the anchor to render within the `li`.  I think I vaguely recall that the class in BS3 used to be on the `li` element, but looks like they opted to place it in the children instead for BS5.
```
<li><a class="dropdown-item" href="#">Action</a></li>
```

Would it make sense to add `as="li"` as a special case for DropdownItem, and have that render the above code?

kyletsang:
Going to kick the `li` can down the road since the styling works with the current markup.

Issue #5468, opened 2020-10-08
Add docs page for CloseButton component


davidjb:
## Is your feature request related to a problem? Please describe

As per #2623, `CloseButton` is exported publicly but isn't documented in the components list.

## Describe the solution you'd like

A page added to the docs. The component is simple but publishing its API would avoid needing to look up the source code. I'm happy to make this as a PR, just seeking feedback before starting + checking it'll be accepted.

## Describe alternatives you've considered

Reading the source code at https://github.com/react-bootstrap/react-bootstrap/blob/master/src/CloseButton.tsx - not hard, but involves remembering where it is.

## Additional context

N/A

kyletsang:
I think this is a good idea.  It was on my todo list for v5 docs since there are new improvements there.

davidjb:
@kyletsang Thanks 👍  Did you want me to look at a PR or did you have this in hand, given the other improvements happening?

kyletsang:
Sure go for it @davidjb.  Thanks :)

davidjb:
Had a crack at it @kyletsang - I found your already-open PR for adding the changes to the docs for v5 so opened a PR against that branch.  Cherry-pick / merge accordingly :+1: 

Wasn't quite sure how the white variants should be conveyed in an example (e.g. I could have used `exampleClassName` on the Playground component and added `bg-dark` there) but I felt this was clearer.

kyletsang:
Fixed in v2 alpha

PR #5474, opened 2020-10-12
docs(CloseButton): add component page

davidjb:
See #5468. Some adjustment of the TS API may be required, see below where the `variant`'s type is the string literal of `'white'` (`PropTypes.oneOf[null, 'white']`?):

image

### DIFF
--- a/www/src/examples/CloseButton/Variants.js
+++ b/www/src/examples/CloseButton/Variants.js
@@ -1,0 +1,4 @@
+<div className="bg-dark p-3">
+  <CloseButton variant="white" />
+  <CloseButton variant="white" disabled />
+</div>;
--- a/www/src/pages/components/close-button.mdx
+++ b/www/src/pages/components/close-button.mdx
@@ -1,0 +1,51 @@
+import { graphql } from 'gatsby';
+
+import ComponentApi from '../../components/ComponentApi';
+import ReactPlayground from '../../components/ReactPlayground';
+import CloseButtonBasic from '../../examples/CloseButton/Basic';
+import CloseButtonDisabled from '../../examples/CloseButton/Disabled';
+import CloseButtonVariants from '../../examples/CloseButton/Variants';
+import CloseButtonLabelled from '../../examples/CloseButton/Labelled';
+
+
+# Close Button
+
+A generic close button for dismissing content such as modals and alerts.
+
+<ReactPlayground codeText={CloseButtonBasic} />
+
+## Disabled state
+
+Bootstrap adds relevant styling to a disabled close button to prevent user
+interactions.
+
+<ReactPlayground codeText={CloseButtonDisabled} />
+
+## Variants
+
+Change the default dark color to white using `variant="white"`.
+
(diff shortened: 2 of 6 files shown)


www/src/examples/CloseButton/Variants.js
```diff
@@ -0,0 +1,4 @@
+<div className="bg-dark">
```
kyletsang, line 1:
```suggestion
<div className="bg-dark p-3">
```
I think some extra padding would look nicer so the buttons aren't at the edges

src/CloseButton.tsx
```diff
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 
-export type CloseButtonVariant = 'white';
+export type CloseButtonVariant = string;
```
kyletsang, line 5: I checked over the bootstrap sass files and it looks like they don't offer any additional customizing here like they did with button colors.  We can leave this as `white`.

www/src/pages/components/close-button.mdx
```diff
@@ -0,0 +1,53 @@
...
+Bootstrap adds relevant styling to a disabled close button to prevent user
+interactions.
+
+<ReactPlayground codeText={CloseButtonDisabled} />
+
+## Variants
+
+All standard visual variants are available for both animation styles by
+setting the `variant` property. Alternatively spinners can be custom
+sized with the `style` property, or custom CSS classes.
```
kyletsang, line 28:
```suggestion
Change the default dark color to white using `variant="white"`.
```
Just added a random blurb to replace the spinner text.  You can add more if needed.

www/src/pages/components/close-button.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+<ReactPlayground codeText={CloseButtonVariants} />
+
+## Accessibility
+
+To ensure the maximum accessibility for Close Button components, it is
+recommended that you provide relevant text for screen readers.  The example
+below provides an example of accessible usage of this component by way of the
+ARIA `label` property.
```
kyletsang, line 37:
```suggestion
`aria-label` property.
```

kyletsang (commented):
This looks good!  Left a few minor adjustments in the comments.

www/src/examples/CloseButton/Variants.js
davidjb: Sounds good -- I debated doing this but didn't want to add too much to the example other than the CloseButton itself

src/CloseButton.tsx
davidjb: Oops, I didn't meant to leave this change in the code; I was just testing how the rendered docs would look and this must have been swept up in my change.  Will clear it out.

davidjb:
Thanks for the review @kyletsang - all changes made and pushed into this PR

kyletsang (approved):

kyletsang:
Thanks!

PR #5497, opened 2020-10-28
fix(types): Specify FormCheck's onChange type exactly

akx:
This should make e.g. `onChange={e => e.target.checked}` work out of the box.

### DIFF
--- a/src/FormCheck.tsx
+++ b/src/FormCheck.tsx
@@ -16,7 +16,7 @@ export type FormCheckType = 'checkbox' | 'radio' | 'switch';
 
 export interface FormCheckProps
   extends BsPrefixPropsWithChildren,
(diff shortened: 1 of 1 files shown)


kyletsang (commented):
Might be better to extend from `React.InputHTMLAttributes` instead of `React.HTMLAttributes` for the props interface?

akx:
> Might be better to extend from `React.InputHTMLAttributes` instead of `React.HTMLAttributes` for the props interface?

Sure thing. I thought the original implementation would've been possibly less disruptive, but reworked :)

kyletsang (approved):

kyletsang:
Nice thanks.  I figured extending would work better in case there's other weird things being inferred :P

PR #5508, opened 2020-11-07
Pass through the menuAlign prop for NavDropdown to the embedded Dropdown.Menu.

Zandor300:
There is no way to align the `Dropdown.Menu` inside the `NavDropdown` to the left. This implements it by adding the prop `menuAlign` to `NavDropdown` and passing it's value to the `align` prop of `Dropdown.Menu`.

### DIFF
--- a/src/NavDropdown.tsx
+++ b/src/NavDropdown.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import Dropdown, { DropdownProps } from './Dropdown';
+import { alignPropType as DropdownMenuAlignPropType } from './DropdownMenu';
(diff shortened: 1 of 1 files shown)


jquense:
The align prop is passed to the Dropdown component which passes through context to the Menu

Zandor300:
> The align prop is passed to the Dropdown component which passes through context to the Menu

@jquense That isn't working for me. With `<NavDropdown align="right">`, the dropdown menu still aligns left.

Additionally, with `<NavDropdown align="right">`, when the parent Navbar is collapsed, the actual item is aligned to the right which isn't what I want.

jquense:
we should fix that then before adding another prop here

PR #5517, opened 2020-11-12
feat(v5): Update Accordion

kyletsang:
https://v5.getbootstrap.com/docs/5.0/components/accordion/

Accordion is now its own dedicated component in v5.

Note: I left the examples for the custom-toggle card-based components just because it's still possible to create them.  Might be useful down the line?

Changes:
- Rewrite component to match new upstream Accordion component
- Combine `AccordionContext` and `SelectableContext` together to simplify and remove null `SelectableContext` context fix
- Update docs
- Update tests


jquense:
awesome, gonna try and get some time to take a good look at this. Ping me if i forget!

kyletsang:
Hey @jquense, are you avail to review some of these existing v5 PRs?  I got some free time next week, so I can finish up adding the new beta changes once the existing PRs are merged in.  

Thanks!

src/AccordionToggle.tsx
```diff
@@ -1,17 +1,19 @@
...
 
-export interface AccordionToggleProps extends BsPrefixPropsWithChildren {
+export interface AccordionToggleProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement>,
+    BsPrefixPropsWithChildren {
   eventKey: string;
-  onClick?: EventHandler;
 }
 
 type AccordionToggle = BsPrefixRefForwardingComponent<
```
jquense, line 19: should we call this `AccordionButton` now?

www/src/examples/Accordion/AllCollapse.js
```diff
@@ -1,22 +1,34 @@
...
-        Click me!
-      </Accordion.Toggle>
-    </Card.Header>
+  <Accordion.Item>
+    <Accordion.Header>
+      <Accordion.Toggle eventKey="0">Accordion Item #1</Accordion.Toggle>
+    </Accordion.Header>
     <Accordion.Collapse eventKey="0">
-      <Card.Body>Hello! I'm the body</Card.Body>
+      <Accordion.Body>
```
jquense, line 7: The verbose component API starts to feel unnecessary now that "accordion" is standalone component. E.g. you are less likely to mix and match the markup, making the need to provide both the body and the Collapse feel a bit excessive?

Perhaps we should combine the collapse and body components? Still expose Collapse as its own component for customization but have the Body component handle it automatically. Worst case someone can drop the single body class on a div if they need to for custom cases

jquense (commented):
looks great, small nit and API thoughts

www/src/examples/Accordion/AllCollapse.js
kyletsang: Looks like we can do the same for the header too so we can potentially reduce the markup to something like this:

```
<Accordion.Item>
  <Accordion.Header eventKey="0">
    Accordion Item #1
  </Accordion.Header>  
  <Accordion.Body eventKey="0">
    Body text
  </Accordion.Body>
</Accordion.Item>
```

kyletsang:
New changes:
- Renamed `AccordionToggle` to `AccordionButton`
- Reduced the markup required for the accordion by combining `AccordionHeader` with `AccordionButton` and `AccordionBody` with `AccordionCollapse`
- Removed the need to specify matching eventKeys in the header and body.  Now it's passed through `AccordionItem` using a context, so you only need to declare it once.

```
  <Accordion.Item eventKey="0">
    <Accordion.Header>Accordion Item #1</Accordion.Header>
    <Accordion.Body>Body</Accordion.Body>
  </Accordion.Item>
  <Accordion.Item eventKey="1">
    <Accordion.Header>Accordion Item #2</Accordion.Header>
    <Accordion.Body>Body</Accordion.Body>
  </Accordion.Item>
</Accordion>
```

jquense (approved):

jquense:
awesome. We've also come full circle to rebuilding the 5 year old Accordion component API from like bootstrap v3

PR #5531, opened 2020-11-19
Fix autocomplete for button and progress bar types

norman-ags:
The autocompletion for variants is missing.

This PR is base on:
https://github.com/microsoft/TypeScript/issues/29729

Here's a sample of the autocomplete: 
image

### DIFF
--- a/src/Button.tsx
+++ b/src/Button.tsx
@@ -10,7 +10,7 @@ import {
 } from './helpers';
 import { ButtonVariant } from './types';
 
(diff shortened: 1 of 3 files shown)


kyletsang:
I'm a bit wary of a workaround like this because there could potentially be other side effects.

norman-ags:
> I'm a bit wary of a workaround like this because there could potentially be other side effects.

Got it, I just thought it would help a lot of people if we have that kind of autocompletion. And thanks for reviewing my first PR! 😄

PR #5542, opened 2020-11-26
fix(types): Set @types/react as optional peer dependency

### LINKED ISSUE #5541: Typescript compile break in react 17 project because of conflicting react types
## Describe the bug
Currently, several `@types` packages are listed as dependency in `package.json`.

This causes a lot of the following "duplicate identifier" typescript compile errors in our case as our project is using `react` 17 and we have `@types/react:17.0.0` installed in our project:
```
Error: node_modules/@types/react/index.d.ts(2981,14): error TS2300: Duplicate identifier 'LibraryManagedAttributes'.
Error: node_modules/react-bootstrap/node_modules/@types/react/index.d.ts(2982,14): error TS2300: Duplicate identifier 'LibraryManagedAttributes'.
Error: node_modules/react-bootstrap/node_modules/@types/react/index.d.ts(2995,13): error TS2717: Subsequent property declarations must have the same type.  Property 'a' must be of type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>', but here has type 'DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>'.
Error: node_modules/react-bootstrap/node_modules/@types/react/index.d.ts(2996,13): error TS2717: Subsequent property declarations must have the same type.  Property 'abbr' must be of type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>', but here has type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>'.
Error: node_modules/react-bootstrap/node_modules/@types/react/index.d.ts(2997,13): error TS2717: Subsequent property declarations must have the same type.  Property 'address' must be of type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>', but here has type 'DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>'.
Error: node_modules/react-bootstrap/node_modules/@types/react/index.d.ts(2998,13): error TS2717: Subsequent property declarations must have the same type.  Property 'area' must be of type 'DetailedHTMLProps<AreaHTMLAttributes<HTMLAreaElement>, HTMLAreaElement>', but here has type 'DetailedHTMLProps<AreaHTMLAttributes<HTMLAreaElement>, HTMLAreaElement>'.
```

The reason seems to be that the `@types` packages are dependencies in `react-bootstrap` and not "only" devDependencies and therefore the `@types/react:16.9.35` gets installed "by" `react-bootstrap` (in `node_modules/react-bootstrap/node_modules/@types/react`).
Since the typescript compile picks up these types then on top of the locally installed v17.0.0 types (installed in `node_modules/@types/react`), type definitions collide.

## To Reproduce

1. Create a Typescript project with `react` ^17.0.0 and `@types/react` ^17.0.0
2. `npm i react-bootstrap`
3. Typescript compile

## Reproducible Example

Not possible in Code Sandbox.

## Expected behavior

I would expect @types/react to only be a devDependency in `react-bootstrap` because of the following reasons:
Having it as devDependency would mean that `npm i react-bootstrap` in consuming projects would not install `@types/react` because of `react-boostrap`. **This is good because**:
1. Any typescript react project should be able to choose it's own version of `@types/react` by installing `@types/react` in the corresponding version according to the installed react version used in the project without suddenly getting type conflicts because `react-bootstrap` pulls in another version of `@types/react` that conflicts.
2. Vanilla Javascript react projects do not need typedefs at all, hence keeping `@types/react` as dependency in `react-boostrap` just clutters the dependency tree for those projects
(Additionally, `react` itself is only a devDependency (& peer-Dependency) and therefore does not get installed with `react-bootstrap` - so why should the respective types should be a dependency and not a devDependency?)

## Screenshots

N/A

- Operating System: all
- Browser, Version: Not relevant
- React-Bootstrap Version: 1.4.0

## Additional context

This probably applies to other `@types/` packages installed as dependencies in `react-bootstrap` as well but I am not sure about them. We hit the issue currently only with `@types/react` but my feeling is that most of the other @types also should move to devDependencies.

kyletsang:
Fixes #5541

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "@types/classnames": "^2.2.10",
     "@types/invariant": "^2.2.33",
     "@types/prop-types": "^15.7.3",
-    "@types/react": "^16.9.35",
+    "@types/react": ">=16.9.35",
     "@types/react-transition-group": "^4.4.0",
     "@types/warning": "^3.0.0",
     "classnames": "^2.2.6",
--- a/yarn.lock
+++ b/yarn.lock
@@ -83,14 +83,10 @@
   dependencies:
     "@babel/highlight" "^7.10.4"
 
-"@babel/compat-data@^7.8.6", "@babel/compat-data@^7.9.0":
-  version "7.9.0"
-  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.9.0.tgz#04815556fc90b0c174abd2c0c1bb966faa036a6c"
-  integrity sha512-zeFQrr+284Ekvd9e7KAX954LkapWiOmQtsfHirhxqfdlX6MEC32iRE+pqUGlYIBchdevaCwvzxWGSy/YBNI85g==
-  dependencies:
-    browserslist "^4.9.1"
-    invariant "^2.2.4"
-    semver "^5.5.0"
+"@babel/compat-data@^7.12.5", "@babel/compat-data@^7.12.7":
+  version "7.12.7"
+  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.12.7.tgz#9329b4782a7d6bbd7eef57e11addf91ee3ef1e41"
+  integrity sha512-YaxPMGs/XIWtYqrdEOZOCPsVWfEoriXopnsz3/i7apYPXQ3698UFhS6dVT1KN5qOsWmVgw/FOrmQgpRaZayGsw==
 
 "@babel/core@^7.12.9", "@babel/core@^7.6.4", "@babel/core@^7.7.5":
   version "7.12.9"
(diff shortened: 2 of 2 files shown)


jquense:
I'm not sure that peer deps are the correct place for the types? only this package needs this range of types. the duplication in the issue is coming from npm/yarn not being able to reconcile and dedupe a common version. It does seem to break which isn't great but alternatively if it's a peer dep the user would still have an invalid setup trying to use the v17 types, AND npm users would get warnings saying they are missing a peer dep because peer deps meta isn't supported in npm (maybe in v7?).

not sure what the correct thing to do is...

kosmiq:
About npm v7: It does installs peer deps by default. Npm v6 does not however.

kyletsang:
Peer deps meta should be supported as of 6.11.0:
https://github.com/npm/cli/releases/tag/v6.11.0

This release is a lil over a year old now, so most people should be good with the peer deps meta support.  Warnings might be ok otherwise, since they'll likely be seeing warnings about the jquery peer deps in bootstrap already :P

I did check the build against the react 17 types and didn't see any issues.  I think it should be ok for the types to support >=16.9.35?

ceisele-r:
Thanks @kyletsang for taking care of this. It would be super helpful to get a release with this fix once you decided on the final solution as this breaks compilation for projects trying to use react 17 types.

jquense:
@ceisele-r it will continue to be broken with react-17 types until RB supports that version...moving it to a peer dep will only move the conflict

package.json
```diff
@@ -139,9 +139,15 @@
     "webpack": "^4.44.2"
   },
   "peerDependencies": {
+    "@types/react": ">=16.9.35",
```
jquense, line 142: I think the initial step is just to change the normal dep to `>=` range to allow the 17 types and enable deduping. That is if we are actually compatible with those types

kyletsang: Is there a way for us to test this locally to make sure dedupe works as expected?

jquense (approved):

package.json
jquense: no, but it should, has up until now at least

Issue #5544, opened 2020-11-29
Alert does not rest props if `transition={false}`


thepuzzlemaster:
## Describe the bug
If trying to pass in some HTMLDivElement props to the Alert, but setting `transition={false}` prevents the props from making it to the rendered div.

## To Reproduce
1. Create the following: `<Alert children='Alert me' transition={false} onMouseOver={() => {console.log('onMouseOver')}  />
2. Mouse over the Alert. Notice no console output
3. Delete `transition={false}` from the code
4. Mouse over the Alert. Notice that sweet, sweet output
Offending line is here: https://github.com/react-bootstrap/react-bootstrap/blob/v1.4.0/src/Alert.tsx#L121

## Reproducible Example
Minimal example [CodeSandbox](https://codesandbox.io/s/brave-sun-7mugf).

## Expected behavior

I can't think of any reason props wouldn't be rested to the div

## Environment (please complete the following information)

- Operating System: [macOSX 10.14.6]
- Browser, Version [87.0.4280.67]
- React-Bootstrap Version [1.4.0]

PR #5545, opened 2020-11-29
Properly rest html props for `Alert` when `transition={false}`

thepuzzlemaster:
I believe this is the fix for the following issue I opened: https://github.com/react-bootstrap/react-bootstrap/issues/5544

It looked like the ternary was just backwards, where if a Transition was being used, it would add the props both to the Alert, and the Transition, where it should be the other way around, and only add the props to the Alert, if NO Transition is being used.

The test I wrote fails without this change.

Manual testing and tests seem to verify this conclusion.

### DIFF
--- a/test/AlertSpec.js
+++ b/test/AlertSpec.js
@@ -48,6 +48,26 @@ describe('<Alert>', () => {
     expect(wrapper.find('.fade').length).to.equal(0);
   });
 
+  it('should spread props to alert when transition=false', () => {
+    const alertId = 'alert-id';
+    const wrapper = mount(
+      <Alert transition={false} id={alertId}>
+        Message
+      </Alert>,
+    );
+    expect(wrapper.getDOMNode().getAttribute('id')).to.equal(alertId);
+  });
+
+  it('should spread props to alert when transition=true', () => {
+    const alertId = 'alert-id';
+    const wrapper = mount(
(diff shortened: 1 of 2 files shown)


test/AlertSpec.js
```diff
@@ -48,6 +48,26 @@ describe('<Alert>', () => {
     expect(wrapper.find('.fade').length).to.equal(0);
   });
 
+  it('should rest props when transition=false', () => {
```
kyletsang, line 51:
```suggestion
  it('should spread props to alert when transition=false', () => {
```

test/AlertSpec.js
```diff
@@ -48,6 +48,26 @@ describe('<Alert>', () => {
...
+    const alertId = 'alert-id';
+    const wrapper = mount(
+      <Alert transition={false} id={alertId}>
+        Message
+      </Alert>,
+    );
+    expect(wrapper.getDOMNode().getAttribute('id')).to.equal(alertId);
+  });
+
+  it('should rest props when transition=true', () => {
```
kyletsang, line 61:
```suggestion
  it('should spread props to alert when transition=true', () => {
```

kyletsang (commented):
Minor nitpicks, but the rest looks good to me

kyletsang (approved):

kyletsang (approved):

PR #5567, opened 2020-12-16
feat(v5): Add floating labels

kyletsang:
https://getbootstrap.com/docs/5.0/forms/floating-labels/

These are initial thoughts on the floating label API.  

There's 2 approaches to creating floating labels:
1. Using a `FloatingLabel` and wrapping an input element.  This approach takes care of rendering the label and passing the control id down to the input/label.
2. Using `FormFloating` and wrapping an input element and a label.  This is to allow customization in case the above doesn't work out. 

TODO:
- Tests

### DIFF
--- a/src/FloatingLabel.tsx
+++ b/src/FloatingLabel.tsx
@@ -1,0 +1,56 @@
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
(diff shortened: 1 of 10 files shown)


jquense (approved):
this looks great

jquense:
maybe worth adding a few conveniences to some of the form controls to have typing but I think this is a good base API at least

Issue #5552, opened 2020-12-05
NavDropdown should support 'as' prop


mperkh:
It would be nice, if NavDropdown supported the 'as' prop, to use it like:

`<NavDropdown title="Dropdown" id="basic-nav-dropdown" as='li'>`

Currently it always results in a `<div>` without an escape hatch. This makes it impossible to create an menu using an unordered list and recreate the menu show here: https://getbootstrap.com/docs/4.5/components/navbar/#supported-content

Garz4:
I would like to work on this issue!

PR #5576, opened 2020-12-21
Support to 'as' as prop in NavDropdown

Garz4:
**What does this PR do?**
Supports `as` as prop in NavDropdown.

**Description of tasks completed**
- Added support to prop `as`: `PropTypes.elementType` to `NavDropdown` component.
- Created a test for this feature.

**What is the related issue?**
#5552


Garz4:
@kyletsang Should I include tests that checks every single `as` prop?

kyletsang:
> 
> 
> @kyletsang Should I include tests that checks every single `as` prop?

A single `as` test would be good here

src/NavDropdown.tsx
```diff
@@ -78,7 +80,7 @@ const NavDropdown: NavDropdown = (React.forwardRef(
     }: NavDropdownProps,
     ref,
   ) => (
-    <Dropdown ref={ref} {...props} as={NavItem}>
+    <Dropdown ref={ref} {...props} as={props.as ? props.as : NavItem}>
```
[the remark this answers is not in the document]
Garz4: Thank you! I also added an assert to check the existance of a single `li.nav-item`.

src/NavDropdown.tsx
```diff
@@ -77,28 +82,37 @@ const NavDropdown: NavDropdown = (React.forwardRef(
...
-        renderOnMount={renderMenuOnMount}
-        rootCloseEvent={rootCloseEvent}
+  ) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'nav-item');
+
+    return (
+      <Dropdown
+        ref={ref}
+        {...props}
+        as={props.as ? props.as : NavItem}
```
kyletsang, line 92: You can remove the `as={...}` line.  Since you are adding in the `.nav-item` class already, we don't need to use `NavItem` anymore.

test/NavDropdownSpec.js
```diff
@@ -59,4 +59,15 @@ describe('<NavDropdown>', () => {
...
   });
+
+  it('should support as as prop', () => {
+    const wrapper = mount(
+      <NavDropdown as="li" id="test-id" title="title">
+        <DropdownItem eventKey="1">Item 1</DropdownItem>
+      </NavDropdown>,
+    );
+
+    wrapper.assertSingle('li');
```
kyletsang, line 70:
```suggestion
```

The second assert takes care of the first assert.

src/NavDropdown.tsx
```diff
@@ -77,28 +82,37 @@ const NavDropdown: NavDropdown = (React.forwardRef(
...
-      >
-        {title}
-      </Dropdown.Toggle>
-
-      <Dropdown.Menu
-        role={menuRole}
-        renderOnMount={renderMenuOnMount}
-        rootCloseEvent={rootCloseEvent}
+  ) => {
+    bsPrefix = useBootstrapPrefix(bsPrefix, 'nav-item');
```
kyletsang, line 86: The `bsPrefix` prop is normally passed down to the dropdown toggle.  You'll need to create a separate var for the dropdown to use internally.

src/NavDropdown.tsx
Garz4: You are right! Thanks, Kyle!

src/NavDropdown.tsx
Garz4: Done!

test/NavDropdownSpec.js
Garz4: Didn't notice that haha thanks!

src/NavDropdown.tsx
```diff
@@ -77,28 +79,36 @@ const NavDropdown: NavDropdown = (React.forwardRef(
...
-    <Dropdown ref={ref} {...props} as={NavItem}>
-      <Dropdown.Toggle
-        id={id}
-        eventKey={null}
-        active={active}
-        disabled={disabled}
-        childBsPrefix={bsPrefix}
-        as={NavLink}
+  ) => {
+    const toBsPrefix = useBootstrapPrefix(bsPrefix, 'nav-item');
```
kyletsang, line 83:
```suggestion
    const navItemPrefix = useBootstrapPrefix(undefined, 'nav-item');
```

kyletsang:
Thanks looks good to me.  @jquense, double checking this with you :)

jquense (commented):
seems ok...this works because NavItem has no logic in it right? it's purely presentational? Maybe we should make a little note about that

kyletsang:
Yeah NavItem is purely presentational.  Could probably shrink that to `createWithBsPrefix` another day.

Garz4:
So, should I comment line 83 with something like `/* The following works because it's purely presentational. to-do: shrink to createWithBsPrefix */` ?

kyletsang:
> 
> 
> So, should I comment line 83 with something like `/* The following works because it's purely presentational. to-do: shrink to createWithBsPrefix */` ?

Maybe something like `NavItem has no additional logic. Can set nav item class here to support "as"`?

You can leave a TODO note in NavItem if you want.  I was thinking of reducing some code in v5 branch later on.

jquense (approved):

kyletsang:
Thanks!

PR #5583, opened 2020-12-27
Add aria-label to carousel prev/next btn

herol3oy:
Based on this issue on lighthouse I created this pull:
image

### DIFF
--- a/src/Carousel.tsx
+++ b/src/Carousel.tsx
@@ -180,10 +180,10 @@ const defaultProps = {
   wrap: true,
   touch: true,
 
-  prevIcon: <span aria-hidden="true" className="carousel-control-prev-icon" />,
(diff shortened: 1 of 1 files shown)


src/Carousel.tsx
```diff
@@ -180,10 +180,10 @@ const defaultProps = {
   wrap: true,
   touch: true,
 
-  prevIcon: <span aria-hidden="true" className="carousel-control-prev-icon" />,
+  prevIcon: <span aria-label="previous slide" aria-hidden="true" className="carousel-control-prev-icon" />,
```
kyletsang, line 183: The carousel also renders `<span class="sr-only">Previous</span>` like in the upstream Bootstrap examples.  Are these not working properly for screen readers?

herol3oy: Seems nothing wrong with that! However, adding `aria-label` was necessary because I've got error on one project in lighthouse test.

PR #5577, opened 2020-12-21
chore(ci): Fix brew cask install

kyletsang:
Fixes the following error

> Error: Calling brew cask install is disabled! Use brew install [--cask] instead.

PR #5586, opened 2020-12-28
Update ToggleButtonProps.inputRef typedef

thepuzzlemaster:
I'm adding some <ToggleButton>'s to my project and I can't quite find a way to currently set up a11y. I want to add an `aria-label` to the `<input>` elements that are rendered by the `ToggleButton` so that when using a screen reader, they announce their labels correctly. However, I noticed that the `inputRef` prop is typed to a `React.LegacyRef<'input'>, which doesn't allow me to properly use `React.useRef<HTMLInputElement>` in my functional component.

I admit, there may be some reason why this is typed the way it is, without me being aware of it, or why this might be a bad idea, but I thought the easiest way to ask this question was to just create this PR.

I verified that copying the built `ToggleButton.js` and its accompanying typedef file over to my project fixes the issue for me.

I'm then using a `useEffect` to get the reference to the inputs and set the `aria-label` attributes on them.

Alternatively, we could pass some props directly to the `input` rather than applying them to the `label`, but that gets messy, and makes some assumptions that we probably don't want to make. 

Thanks for taking a look.


thepuzzlemaster:
I took a look at the CI failure, but it appears there is already a PR out to fix it: https://github.com/react-bootstrap/react-bootstrap/pull/5577

src/ToggleButton.tsx
```diff
@@ -17,7 +17,7 @@ export interface ToggleButtonProps
   disabled?: boolean;
   onChange?: React.ChangeEventHandler<HTMLInputElement>;
   value: unknown;
-  inputRef?: React.LegacyRef<'input'>;
+  inputRef?: React.LegacyRef<'input'> | React.RefObject<HTMLInputElement>;
```
kyletsang, line 20:
```suggestion
  inputRef?: React.LegacyRef<HTMLInputElement>;
```
I think this was just an oversight and the original typing is wrong.  The `value: unknown` right above it should be narrowed to `value: string | ReadonlyArray<string> | number;` to match that in the React types.

thepuzzlemaster: Thanks. I made the update. I also realized I forgot to update the propTypes. Please make sure I did those correctly, as I've not used those before.

Also, worth noting that changing the `value` typedef from `unknown` will be a compile-time error for anyone not using a `string | number`. I had actually been using `boolean` for my value, as in my case I was toggling between 2 states (on/off), and that was the cleanest way to track the value for me. Easy enough to update to string, and map back to boolean for the consumers of my widget, but I thought I would mention it, as I'm not sure how you handle changes like that in this library.

src/ToggleButton.tsx
```diff
@@ -56,13 +55,21 @@ const propTypes = {
...
   /**
    * A ref attached to the `<input>` element
    * @type {ReactRef}
    */
-  inputRef: PropTypes.any,
+  inputRef: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.func,
+    PropTypes.shape({ current: PropTypes.instanceOf(HTMLInputElement) }),
+  ]),
```
kyletsang, line 72:
```suggestion
  inputRef: PropTypes.oneOfType([
    PropTypes.func,
    PropTypes.object,
  ]),
```

Although the types do allow string refs, we should discourage it as it's deprecated.  Also, for ref objects, HTMLInputElement isn't avail for SSR environments.  For that, I think we can change that bit to just `PropTypes.object` for simplicity.

kyletsang (approved):

PR #5588, opened 2020-12-30
Fix Formik validation example bug in docs

amonae:
Added `terms: false` within the `initialValues` prop. This fixed a bug where when trying to check the checkbox, an error appeared that said ``terms must be a `boolean` type, but the final value was: `[ "\"on\"" ]`.``


kyletsang (commented):
There's another example with Formik but uses the tooltip feedbacks, that needs this fix too.

Also, I think we should display the invalid feedback when the checkbox is not checked.  Currently there's no feedback displayed in either checked/unchecked state.

kyletsang:
Fixed in #5592

PR #5592, opened 2021-01-04
Fix some bugs in example code in docs.

### LINKED ISSUE #5574: Form.Check Formik example return "terms must be a `boolean` type, but the final value was: `[ "\"on\"" ]`."
## Describe the bug

Both checkbox validations on the Formik examples seem to be borked. 

    terms must be a `boolean` type, but the final value was: `[ "\"on\"" ]`.

is the message it prints out to the user immediately when clicking it. 

It's an array type because switching the yup definition to yup.array().required() makes it go away BUT that creates a different UI issue such that once you check the box, then uncheck it, you can still submit (aka, submits to console.log) Or maybe that issue was always there, hard to tell when it won't validate in either state.

## To Reproduce

 Pretty simple really, click on the "Agree to terms and conditions" check box. 
 
## Expected behavior

A user friendly message only when NOT clicking the checkbox.
 
- Operating System: Windows 
- Browser, Version Firefox, Chrome, Edge, take your pick.
- React-Bootstrap Version 1.4.0
 Also, on Android Firefox (wasn't sure it was just Windows).

### LINKED ISSUE #5585: Form Text Inline documentation is incorrect
## Describe the bug

The Form.Text id and the Form.Control aria-describedby should match, and in the example they do not.

Form.Text id should be changed to `passwordHelpInline`

## Additional context

https://github.com/react-bootstrap/react-bootstrap/blob/332a4a035614a8493ff527bdbf82e90be4546037/www/src/examples/Form/FormTextInline.js

JingWangTW:
- Fix #5585 
    - Props `id` in `Form.Text` changed to `passwordHelpInline`.
- Fix #5574
    - Added `terms: false` within the `initialValues` prop and set the schema to restrict the checkbox must be checked.

### DIFF
--- a/www/src/examples/Form/FormTextInline.js
+++ b/www/src/examples/Form/FormTextInline.js
@@ -7,7 +7,7 @@
       id="inputPassword6"
       aria-describedby="passwordHelpInline"
     />
-    <Form.Text id="passwordHelpBlock" muted>
+    <Form.Text id="passwordHelpInline" muted>
       Must be 8-20 characters long.
     </Form.Text>
   </Form.Group>
--- a/www/src/examples/Form/ValidationFormik.js
+++ b/www/src/examples/Form/ValidationFormik.js
@@ -1,13 +1,13 @@
 const { Formik } = formik;
 
-const schema = yup.object({
+const schema = yup.object().shape({
   firstName: yup.string().required(),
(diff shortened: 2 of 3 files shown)


kyletsang (approved):

kyletsang:
Thanks

PR #5593, opened 2021-01-05
fix: Prevent focus on disabled nav items or list group items only

### LINKED ISSUE #4737: Custom tabs not focusable if none are active
## Describe the bug

Tabs cannot be selected using keyboard navigation if none of the tab items are active. Issue caused by setting `tabindex="-1"` on inactive tabs, even if none are active.

## To Reproduce

1. Create a tabbed navigation without an `activeKey` or `defaultActiveKey` prop set
2. Press tab key to focus the navigation

## Reproducible Example

[Example](https://codesandbox.io/s/code-sandbox-examples-2nthn)

## Expected behavior

Should focus first tab within the tabbed navigation

## Environment

- MacOS 10.14
- Latest Chrome and Safari
- React-bootstrap 1.0.0-beta.12

kyletsang:
Fixes a couple of accessibility issues:

- Prevents disabled list group items from being focused when tabbing.
- Allow inactive tab items to be focused when tabbing

Fixes #4737

### DIFF
--- a/src/AbstractNavItem.tsx
+++ b/src/AbstractNavItem.tsx
@@ -93,8 +93,11 @@ const AbstractNavItem: AbstractNavItem = React.forwardRef(
     }
 
     if (props.role === 'tab') {
-      props.tabIndex = isActive ? props.tabIndex : -1;
       props['aria-selected'] = isActive;
+    }
+
+    if (props.disabled) {
+      props.tabIndex = -1;
     }
 
     const handleOnclick = useEventCallback((e) => {
--- a/src/AbstractNavItem.tsx
+++ b/src/AbstractNavItem.tsx
@@ -34,7 +34,7 @@ const propTypes = {
   role: PropTypes.string,
 
   href: PropTypes.string,
-  tabIndex: PropTypes.string,
+  tabIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   eventKey: PropTypes.any,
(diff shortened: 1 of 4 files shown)


src/AbstractNavItem.tsx
```diff
@@ -93,10 +93,13 @@ const AbstractNavItem: AbstractNavItem = React.forwardRef(
     }
 
     if (props.role === 'tab') {
-      props.tabIndex = isActive ? props.tabIndex : -1;
       props['aria-selected'] = isActive;
     }
 
+    if (props.disabled) {
+      props.tabIndex = -1;
+    }
```
jquense, line 101: i'm not sure this is right. I think we only want to insist on this in the tab case. OItherwise it's necessarily true that items styled as disabled should not be tabbable, especially if a tabindex was provided by the user explicitly

jquense: and we should probably add aria-disabled in the case where it is disabled

kyletsang: > i'm not sure this is right. I think we only want to insist on this in the tab case. OItherwise it's necessarily true that items styled as disabled should not be tabbable, especially if a tabindex was provided by the user explicitly

Changed to allow user overrides for tabIndex in ListGroupItem but not in tabs.  Also added aria-disabled.

jquense (approved):

Issue #5519, opened 2020-11-13
OverlayTrigger - findDOMNode is deprecated in StrictMode


brvaland:
Hi

I am using Tooltip overlay from react-bootstrap but getting error `findDOMNode is deprecated in StrictMode`. As per [documentation](https://react-bootstrap.github.io/components/overlays/) using the function form of overlaytrigger avoids a React.findDOMNode call, for those trying to be strict mode compliant but I am still getting the error as below.

```
index.js:1 Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
    in div (created by Tooltip)
    in Tooltip (at NavMenu.tsx:31)
    in Transition (created by Fade)
    in Fade (created by Overlay)
    in Overlay (created by Overlay)
    in Overlay (created by OverlayTrigger)
    in OverlayTrigger (at NavMenu.tsx:28)
    in div (created by ForwardRef)
    in ForwardRef (created by Nav)
    in Nav (at NavMenu.tsx:27)
    in div (created by Context.Consumer)
    in Transition (created by ForwardRef)
    in ForwardRef (created by Context.Consumer)
    in NavbarCollapse (at NavMenu.tsx:23)
    in nav (created by Navbar)
    in Navbar (at NavMenu.tsx:18)
    in StrictMode (at src/index.tsx:13)
```

**Code for OverLayTrigger**
```
<OverLayTrigger 
     key="left" 
     placement="left" 
     overlay={<Tooltip id="tooltip-initials">{`Logged in as ${this.context.user?.name}`}</Tooltip>}>

{
    ({ref, ...triggerHandler}) => (
        <div className="avatar-circle" {...triggerHandler} >
            <span className="initials" id="initials" ref={ref}>
                {`${this.context.user?.given_name.substring(0,1)}${this.context.user?.family_name.substring(0,1)}`}
            </span>
        </div>
    )
}
</OverLayTrigger>
```

Can you please advise what I am missing above ?

douglasduartetb:
When you start a new react project, by default now came with strict mode. Im still waiting a solution for that to, is ugly the errors on console 

`
<React.StrictMode>
        <App />
</React.StrictMode>
`

kyletsang:
[...]

PR #5605, opened 2021-01-11
fix: remove call to obsolete findDOMNode

mmomtchev:
ReactDom.findDOMNode is now obsolete
#5519

### DIFF
--- a/src/OverlayTrigger.tsx
+++ b/src/OverlayTrigger.tsx
@@ -2,7 +2,6 @@ import contains from 'dom-helpers/contains';
 import PropTypes from 'prop-types';
 import React, { cloneElement, useCallback, useRef } from 'react';
 import useTimeout from '@restart/hooks/useTimeout';
-import safeFindDOMNode from 'react-overlays/safeFindDOMNode';
 import warning from 'warning';
 import { useUncontrolledProp } from 'uncontrollable';
 import Overlay, { OverlayChildren, OverlayProps } from './Overlay';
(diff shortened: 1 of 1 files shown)


jquense:
Sorry we can't remove this right now, as it would be a breaking API change. Luckily tho there is already a workaround to avoid the case where findDOMNode is used, only use the function child of Overlay, and pass the ref to something that resolves to a DOM node. If you are getting the warning that findDOMNode is being used, then it means this change would actually _break_ your code because it's following a path that _requires_ findDOMNode.

jquense:
Looking at #5519 it seems like you are still getting the warning? Can you put together a repro and we can figure out why and how to fix it in a backwards compatible way 👍

mmomtchev:
@jquense , in the case where `children` is a DOM element my PR actually works 😄 
But you are right that there is also the case where `children` is a React element and in this case `findDOMNode` cannot be avoided

I will convert my `OverlayTrigger`s to function calls, this is a valid workaround

jquense:
We will definitely do this in the next major version tho!

mmomtchev:
@jquense - the problem was on my side - the warning was not triggered the Overlaytrigger I had in mind but by another one - I realized this when I created the Codepen

Zeratoxx:
> 
> 
> We will definitely do this in the next major version tho!

Is there a known date for?

kyletsang:
@Zeratoxx, we'll aim to have this in the bootstrap 5 release.

CalvinJamesHeath:
Im having the same error on my React app, im using a popover with 	
<OverlayTrigger trigger='click' placement='bottom' overlay={popover} rootClose>
<TetxField />
<OverlayTrigger/>
import OverlayTrigger from 'react-bootstrap/OverlayTrigger';

And my Error:  findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of RefHolder which is inside StrictMode. Instead, add a ref directly to the element you want to reference.

Any help or solutions would be appreciated. Thanks!

PR #5618, opened 2021-01-20
feat(InputGroup): Add hasValidation prop

kyletsang:
> To detect what elements need rounded corners inside an input group with validation, an input group requires an additional .has-validation class.

Ref: https://getbootstrap.com/docs/4.6/components/forms/#input-group-validation


jquense (approved):
Bleh that's an annoying quirk, too bad it couldn't be handled via CSS (upstream)

Issue #5621, opened 2021-01-24
transition={false} prevents Tooltip from showing


mmomtchev:
Here is my custom tooltip component:

```js
import { OverlayTrigger, Tooltip as ReactTooltip } from 'react-bootstrap';

export default class Tooltip extends React.Component {
    render() {
        return (
            <OverlayTrigger placement="auto" transition={false}
                delay={{ show: 250, hide: 4000 }}
                overlay={(props) => (
                    <ReactTooltip {...props}>{this.props.tooltip}</ReactTooltip>
                )}>
                {({ ref, ...triggerHandler }) => (
                    <div {...triggerHandler} ref={ref}>
                        {this.props.children}
                    </div>
                )}
            </OverlayTrigger>
        );
    }
};
```

The `transition={false}` prevents the tooltip from showing - as its opacity stays at 0.

Workaround is

```css
.tooltip {
  opacity: unset !important;
}
```

kyletsang:
`show` class isn't being applied when children is a function in `Overlay`.

A workaround in your case would be to use `overlay={<ReactTooltip>{this.props.tooltip}</ReactTooltip>}` in your `OverlayTrigger` props.

mmomtchev:
I see only one case where there might be some breakage: `transition={true}` and a function child that deconstructs the props *after* a `className` property

```js
(props) => <Div className="test" {...props} />
```

Do you see another one?

PR #5631, opened 2021-01-29
show tooltips for func overlay without transition

mmomtchev:
#5621 

How about this? The user will have to add the classes himself when he is using a function `overlay`

### DIFF
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -205,6 +205,7 @@ function Overlay({
             ...overlayProps,
             placement,
             show,
+            className: !transition && show && 'show',
             popper,
             arrowProps,
           });
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -205,7 +205,7 @@ function Overlay({
             ...overlayProps,
             placement,
             show,
-            className: !transition && show && 'show',
+            ...(!transition && show && { className: 'show' }),
             popper,
             arrowProps,
           });
--- a/test/OverlayTriggerSpec.js
+++ b/test/OverlayTriggerSpec.js
@@ -39,6 +39,27 @@ describe('<OverlayTrigger>', () => {
     wrapper.assertSingle('div.test');
   });
 
+  it('Should show the tooltip when transitions are disabled', () => {
+    const overlay = ({ className, ...props }) => (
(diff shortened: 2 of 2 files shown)


kyletsang:
I think this will need to target the bs5 branch since it's breaking behavior.

jquense:
I'm tempted to call this a bug fix...do you think anyone would be depending on the tooltip not working without a transition? I guess someone might be using it to toggle visibility?

kyletsang:
I was initially thinking people were using the `show` prop that's passed along to toggle visibility manually, but maybe a lot of people just use transitions by default.  This sounds pretty low risk, so I'm down for merging this.

jquense:
We can always revert if it causes a problem

jquense (approved):

src/Overlay.tsx
```diff
@@ -205,6 +205,7 @@ function Overlay({
             ...overlayProps,
             placement,
             show,
+            className: !transition && show && 'show',
```
kyletsang, line 208:
```suggestion
            ...(!transition && show && { className: 'show' }),
```

This change will avoid overwriting `className` when transition is true and spreading props after a className is declared.

mmomtchev:
@jquense Not that many in fact :smile: 
The main problem is of course that recent versions of `create-react-app` have strict mode enabled - but no matter one's goodwill - and I tried hard - you end up disabling it for any serious project

src/Overlay.tsx
mmomtchev: `overlayProps` won't contain the `className` if `overlay` is a function? The `className` will probably be hard-coded in the function.

kyletsang: There's no `className` in the `overlayProps`.  This is what's passed back from react-overlays:

https://github.com/react-bootstrap/react-overlays/blob/f9f1a14752a8d8ff9aa7c7d207aa0c413ca70660/src/Overlay.tsx#L117-L130

The above suggestion will prevent the below `className="abc"` from being overwritten when `transition=true`.
```
(props) => <Tooltip className="abc" {...props}>my tooltip</Tooltip
```

kyletsang:
Thanks!

PR #5663, opened 2021-02-19
fix(Popover): use Bootstrap default offsets

kyletsang:
For bootstrap 5

Popovers are not aligned due to a missing default offset found in upstream bootstrap:
https://github.com/twbs/bootstrap/blob/5c32767e0e0dbac2d934bcdee03719a65d3f1187/js/src/popover.js#L28

The approach here uses a `useOverlayOffset` hook to add the proper offset if it's a popover, otherwise default [0, 0].

This is the last usage of `usePopperMarginModifiers`, so we can remove that in this PR after #5662 is merged.


src/useOverlayOffset.tsx
```diff
@@ -0,0 +1,34 @@
...
+      name: 'offset',
+      options: {
+        offset: () => {
+          if (
+            overlayRef.current &&
+            hasClass(overlayRef.current, popoverClass)
+          ) {
+            // Default popover offset.
+            // https://github.com/twbs/bootstrap/blob/5c32767e0e0dbac2d934bcdee03719a65d3f1187/js/src/popover.js#L28
+            return [0, 8];
```
jquense, line 24: I think we should expose this via a module constant, maybe in Popover, just to ensure users have access to these hard coded values in case they change

src/Overlay.tsx
```diff
@@ -166,7 +166,7 @@ function Overlay({
   ...outerProps
 }: OverlayProps) {
   const popperRef = useRef({});
-  const [ref, marginModifiers] = usePopperMarginModifiers();
+  const [ref, modifiers] = useOverlayOffset();
```
jquense, line 169: hmm this API is unfortunate...ideally you'd want the the Popover component itself to specify this, might be feasible with context, but i'd be afraid of timing issues. Lets make sure we aren't calling this hook "public" in case we want to remove it later

src/useOverlayOffset.tsx
kyletsang: Done.  Had to move `ArrowProps` and `Placement` types from `Overlay` over to `types.tsx` to fix a circular dependency warning.

jquense (approved):

PR #5665, opened 2021-02-19
feat(v5): replace screen reader classes

ggtaylor:
Replace sr-only with visually-hidden 
Rename srOnly prop to visuallyHidden


kyletsang (approved):

kyletsang:
Thanks!

PR #5676, opened 2021-02-24
refactor: simplify code

kyletsang:
Use optional chaining with function calls
Avoid destructuring params if not needed
Shorten TabContent, NavItem to use createWithBsPrefix


src/Image.tsx
```diff
@@ -54,19 +54,19 @@ const Image = React.forwardRef<HTMLImageElement, ImageProps>(
...
         {...props}
-        className={classNames(className, classes)}
+        className={classNames(
+          className,
+          classNames(
+            fluid && `${bsPrefix}-fluid`,
+            rounded && `rounded`,
+            roundedCircle && `rounded-circle`,
+            thumbnail && `${bsPrefix}-thumbnail`,
+          ),
```
jquense, line 68: this could be unnested i think

jquense (approved):

src/Image.tsx
kyletsang: Good catch!   Fixing...

PR #5687, opened 2021-03-01
feat!: add StrictMode support

### LINKED ISSUE #5075: Warning: findDOMNode is deprecated in StrictMode.
Error generated when using <OverlayTrigger >
Warning: findDOMNode is deprecated in StrictMode.

Tested with latest 1.0.0 release of recat-bootstrap 

From what I can see this is a known issue with react-bootstrap but I am not clear on when it will be fixed or any workarounds in the meantime. Console errors are not helpful when developing. Turning off strict mode compromises other development.

see: https://reactjs.org/docs/strict-mode.html

### LINKED ISSUE #5519: OverlayTrigger - findDOMNode is deprecated in StrictMode

kyletsang:
An attempt at fixing strict mode issues for transitions, carousel and overlay trigger. 

Fixes #5519
Fixes #5075


src/TransitionWrapper.tsx
```diff
@@ -0,0 +1,93 @@
...
+  children:
+    | React.ReactElement
+    | ((
+        status: TransitionStatus,
+        props: Record<string, unknown>,
+      ) => React.ReactNode);
+};
+
+// Normalizes Transition callbacks when nodeRef is used.
+const TransitionWrapper = React.forwardRef<
```
jquense, line 20: I can never remember what RTG is doing now...do we need this wrapper? If it's in v5 can we just break the API and require node ref compatible components?

src/Collapse.tsx
```diff
@@ -229,19 +230,20 @@ const Collapse = React.forwardRef(
         onEntered={handleEntered}
         onExit={handleExit}
         onExiting={handleExiting}
+        childRef={(children as any).ref}
```
jquense, line 233: this will always be falsey, children, if it's an element would have ref on `children.props` not on the element itself

src/Fade.tsx
```diff
@@ -92,19 +94,20 @@ const Fade = React.forwardRef<Transition<any>, FadeProps>(
...
     );
 
     return (
-      <Transition
+      <TransitionWrapper
         ref={ref}
         addEndListener={transitionEndListener}
         {...props}
         onEnter={handleEnter}
+        childRef={(children as any).ref}
```
jquense, line 108: ditto

src/OverlayTrigger.tsx
```diff
@@ -186,6 +181,10 @@ function OverlayTrigger({
   ...props
 }: OverlayTriggerProps) {
   const triggerNodeRef = useRef(null);
+  const mergedRef = useMergedRefs<unknown>(
+    triggerNodeRef,
+    (children as any).ref,
```
jquense, line 186: ditto

src/TransitionWrapper.tsx
kyletsang: RTG has this quirk where the node is excluded in the callback signature when using nodeRef.

ie.  `(node, isAppearing) => ...` becomes `(isAppearing) => ...`

This wrapper abstracts all that so callers don't have to worry about it (like they can just use it like it was without using nodeRef).  I kinda figured this would be a cleaner approach.

src/Collapse.tsx
kyletsang: Hmm `children.ref` is actually defined when I was testing this.  I've tried `children.props.ref` but I get a warning:

Warning: div: `ref` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.

kyletsang: tbh, children.ref did seem kinda hacky (esp. with the casting).  I'm not 100% sure on what the correct approach is :P

jquense: o hmm i must have it backwards, been doing this for too long. I'm suprised the types don't correctly cover this tho

src/TransitionWrapper.tsx
jquense: ah, that seems like a bad API choice i probably ok'd oops

src/Collapse.tsx
kyletsang: I found this on the react github project that seems to confirm the `children.ref` usage:

https://github.com/facebook/react/issues/8873#issuecomment-275423780

jquense (approved):

FunctionDJ:
This will only be in v5, not v4, correct?

kyletsang:
> 
> 
> This will only be in v5, not v4, correct?

Correct

jgomez-nydig:
> > This will only be in v5, not v4, correct?
> 
> Correct

Looking at the tags I see 1.5.2 and 2.x, what are v5 and v4?

kyletsang:
> > > This will only be in v5, not v4, correct?
> > 
> > 
> > Correct
> 
> Looking at the tags I see 1.5.2 and 2.x, what are v5 and v4?

V4 and v5 corresponds to the bootstrap versions, so in terms of react bootstrap this would be 1.x and 2.x respectively.

augustjanse:
I was under the impression that `react-bootstrap@next` would fix these warnings, but I'm still getting them with `OverlayTrigger` and `Tooltip`. I pushed an example [here](https://github.com/augustjanse/react-bootstrap-example/blob/main/src/App.js). Am I mistaken?

kyletsang:
@augustjanse, try passing a function as a child to OverlayTrigger.

See:  https://react-bootstrap.netlify.app/components/overlays/#customizing-trigger-behavior

augustjanse:
Oh! As you might have guessed, I didn't realize that the other way of doing it would still trigger the warnings. This worked fine, thanks!

PR #5695, opened 2021-03-03
refactor!: Migrate to new JSX transform

### LINKED ISSUE #5293: What changed React-bootstrap 1.0.1 to 1.2.2
## Describe the bug
I am using WebStorm and IDE shows to me available components in the React bootstrap. But now react-bootstrap 1.2.2, my IDE does not show to me any suggestion, autocompletion. 

If I use `yarn add react-bootstrap@1.0.1` IDE completion still works. My problem is updated version.

L4hotAC9ZE
nupyqnwsa0

### LINKED ISSUE #5354: After migrating to 1.3.0 I can't use Tab
## After migrating to 1.3.0 I can't use Tab

My code used to work fine with react-bootstrap 1.0.1. But after migrating to 1.3.0, I get this error when I try to use <Tab />

TS2786: 'Tab' cannot be used as a JSX component.
  Its instance type 'Tab' is not a valid JSX element.
    Type 'Tab' is missing the following properties from type 'ElementClass': context, setState, forceUpdate, props, and 2 more.

And I also get this error:
TS2607: JSX element class does not support attributes because it does not have a 'props' property.

Both are Typescript compilation errors

## This could be related to my Typescript configuration but not sure.  
Here a copy of my tsconfig.json

```
{
  "compilerOptions": {
    "sourceMap": true,
    "jsx": "react",
    "target": "es2015",
    "module": "commonjs",
    "moduleResolution": "node",
    "noImplicitAny": true,
    "skipLibCheck": true,
    "typeRoots": [
      "./node_modules/@types"
    ]
  },
  "include": [
    "./src/**/*",
    "./test/**/*"
  ]
}
```

- Operating System: Windows 10
- React-Bootstrap Version 1.3.0

## Additional context

This could be an issue in my Typescript configuration, but it is strange that it used to work just fine in react-bootstrap 1.0.1

### LINKED ISSUE #5420: Cannot find module 'react-bootstrap' or its corresponding type declarations. ts(2307)
## Describe the bug

Error message in TypeScript: Cannot find module 'react-bootstrap' or its corresponding type declarations. ts(2307)

## To Reproduce

1. Add `react-bootstrap` to a project containing TypeScript code.
2. Import something from `react-bootstrap` in a .ts file, such as: `import { Container, Row, Col, Button, Image } from "react-bootstrap"`
3. The TypeScript compiler gives this error message: Cannot find module 'react-bootstrap' or its corresponding type declarations. ts(2307)

## Expected behavior

The types should be imported normally with no error. The library contains type definitions, but there is something wrong with how they're being exported.

## Environment

- Operating System: Ubuntu 20
- Browser not applicable (this is a build error).
- React-Bootstrap Version: 1.3.0

kyletsang:
Should be a good time to do this?

Ref: https://github.com/facebook/react/pull/18102

Fixes #5420 (maybe?)
Fixes #5354
Fixes #5293


jquense:
I wonder if it might be better to upgrade to the new JSX babel transform which will automatically add the `as * ` import? A potential "issue" i've seen with other projects trying to do this early is that the type defs don't actually enforce that everything is a named export (because it's not actually React is only CommonJs), so TS tooling and editors will auto important as `import React from 'react'` and there isn't much that can be done to enforce the style we want aside from manual intervention

jquense:
also i've never understood why typescript picks one format of a commonJS export as "more correct" if anything the `import React from 'react'` is more correct, given how modules in NodeJs actually work

kyletsang:
Hmm I'm not so well-versed in this area.  I think the current TS issues above is more of the import syntax in the `d.ts` files forcing a particular tsconfig setup?  Would it make sense to revisit the babel transform when react releases their esm build?

jquense:
we can upgrade to the newer babel transform now i think, which involves some config (and exploration) for the react preset. Worth investigating probably as it would save us manual headaches https://babeljs.io/blog/2020/03/16/7.9.0#a-new-jsx-transform-11154httpsgithubcombabelbabelpull11154 I guess it might require react 17, but honestly that's not a big deal for v5 here

jquense (approved):
sweeet

kyletsang:
Hmm tests are failing all of a sudden, but I don't think it's related to this change...  The proptypes console error args aren't being formatted in the error message so it causes the tests to fail.

Will fix this in a follow-up

PR #5697, opened 2021-03-05
feat(docs): add premium themes page

EINazare:
Hello guys,

We were building UI Dashboards on top of React-Bootstrap since 2018 and more than 100.000 developers used them so far, with a rating of 4.9/5 ( example: https://www.creative-tim.com/product/light-bootstrap-dashboard-react/reviews). 
I think this will be a good addition to the community that wants to speed up their work. 

Thank you,
Manu

### DIFF
--- a/www/src/components/NavMain.js
+++ b/www/src/components/NavMain.js
@@ -90,6 +90,10 @@ const NAV_LINKS = [
     link: '/components/alerts',
     title: 'Components',
   },
+  {
+    link: '/premium-themes/',
+    title: 'Premium Themes',
+  },
 ];
 
 const propTypes = {
--- a/www/src/pages/premium-themes.js
+++ b/www/src/pages/premium-themes.js
@@ -28,12 +28,12 @@ const Content = styled('div')`
   display: flex;
   flex-direction: column;
   justify-content: center;
-  min-height: 450px;
+  min-height: 28rem;
   margin: 0 auto;
-  max-width: 800px;
+  max-width: 50rem;
 
   @media (max-width: 800px) {
-    padding: 0 40px;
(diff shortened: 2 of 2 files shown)


sohamsshah (changes requested):
Good work @EINazare ! Thanks for this PR.

I have a suggestion for your PR:
* There are various places in the code where px, rem and % values are used for styling. As an enhancement, can you avoid using pixel values? IMO rem and % are good choices; so curious if there is a particular reason for using pixel for css styling.

Thank you!

EINazare:
Hello there, @sohamsshah ,

I've changed all usages of `px` to `rem`, but I strongly think this is bad.
If you check the `www/src/pages/index.js` you will see that on some components, such as the `Content`, there was used `px` for heights and widths. Is it ok to have that in `px` and on this page to have `rem`?

Best,
Manu

sohamsshah:
> Hello there, @sohamsshah ,
> 
> I've changed all usages of `px` to `rem`, but I strongly think this is bad.
> If you check the `www/src/pages/index.js` you will see that on some components, such as the `Content`, there was used `px` for heights and widths. Is it ok to have that in `px` and on this page to have `rem`?
> 
> Best,
> Manu

Makes sense. Code Consistency is also something can be considered as well. Would like if the Members/Maintainers can review this and suggest best solution :)

jquense:
Hey @EINazare these themes look great, however, i don't think it's a good idea for us to showcase paid third party themes like these as it gives the impression that the RB team, maintains or supports them. We don't, ourselves, use these themes, and while they look great, i don't feel comfortable "sponsoring" them like this. Thanks again!

PR #5714, opened 2021-03-10
feat(v5): update ResponsiveEmbed to Ratio

kyletsang:
https://getbootstrap.com/docs/5.0/helpers/ratio/

ResponsiveEmbed has been changed to Ratio as per upstream. Fixed impl, tests, docs.


src/Ratio.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * This component requires a single child element
+   */
+  children: PropTypes.element.isRequired,
+
+  /**
+   * Set the aspect ration of the embed
+   */
+  aspectRatio: PropTypes.oneOf<AspectRatio>(['1x1', '4x3', '16x9', '21x9']),
```
jquense, line 31: should handle the custom ratio case automatically? could accept a number, so someone could do 3/4 and we set 
```js
style={{'--bs-aspect-ratio': `${num * 100}%`}}
```

maybe with some guard rails around if the number is 0 -1 or 0 1-100 handle it gracefully?

kyletsang: Hmm we might be able to take any custom "AxB" string, split it, then do
```
style={
  '--bs-aspect-ratio': 'calc(${B} / ${A} * 100%)'
}
```

jquense: That would work too, I'm not sure if it's too clever tho? A user might want to specify a percent directly instead of thinking of a spectic ratio to produce that percent. Idk. Could support everything it's a simple enough component I guess otherwise

kyletsang: Ended up supporting just the defaults and a fraction/percentage number.  I'm thinking this is probably the best flexibility to allow users to create custom ratio classes and make them responsive in css.

jquense (approved):

PR #5720, opened 2021-03-15
fix(Nav): add .card-header-* only if nested in card header

### LINKED ISSUE #4771: <Nav> inside <Card> has "card-header-*" class even outside <Card.Header>
## Describe the bug

If you put a `<Nav>` inside a `<Card>`, the resulting `.nav` element is always given one of the `card-header-*` classes (`card-header-tabs` or `card-header-pills`, depending on the `<Nav>`'s `variant` prop).

This happens even if the `<Nav>` is not inside a `<Card.Header>`, eg if it's in the card's body; and in those situations, the unwanted class messes up the `<Nav>`'s margins.

## To Reproduce

1. Create a `<Card>`
2. Put a `<Nav>` inside it, that isn't inside a `<Card.Header>`
3. Give the `<Nav>` a `variant` (either will do), and at least one complete `<Nav.Item>`
3. Add another element after the `<Nav>`
4. Observe that the `<Nav>` overlaps the element after it

## Reproducible Example

https://codesandbox.io/s/romantic-leakey-9ob7c

## Expected behavior

The `card-header-*` class should only be added to the `.nav` if the `<Nav>` is inside a `<Card.Header>`.

## Screenshots

image

- Operating System: n/a
- Browser, Version: n/a
- React-Bootstrap Version: v1.0.0-beta.14

## Additional context

I originally discovered this while trying to put a `<Tabs>` inside a `<Card.Body>`, but from looking in the source, I found that the underlying problem was caused by the `<Nav>` code that `<Tabs>` uses internally. In particular, it's [this part of Nav.js](https://github.com/react-bootstrap/react-bootstrap/blob/0a8a20c8ff800042f0314e34cd3815a2b3600cab/src/Nav.js#L109-L135):

```javascript
let … cardHeaderBsPrefix;
…
const cardContext = useContext(CardContext);
…
if (navbarContext) {
  …
} else if (cardContext) {
  ({ cardHeaderBsPrefix } = cardContext);
}

return (
  <AbstractNav
    …
    className={classNames(className, {
      …
      [`${cardHeaderBsPrefix}-${variant}`]: !!cardHeaderBsPrefix,
```

Perhaps  `<Card.Header>` should provide its own context type, so the `<Nav>` code could look for that one instead of `CardContext`?

kyletsang:
Fixes #4771 for bs5 branch

Rename CardContext to CardHeaderContext since it's only used for CardHeaders.  Also added missing components into docs


jquense (approved):
bootstrap is getting a little out of hand with these combinations

PR #5727, opened 2021-03-18
fix(types): fix types for event key

kyletsang:
This fixes a number of type inconsistencies surrounding event keys.  The types were changed to allow string and number only.


jquense (approved):
nice

PR #5736, opened 2021-03-23
refactor(Modal): forward ref to react-overlays Modal

kyletsang:
I think this is a good time to remove `ref._modal`? 

Should we merge/pass the ref along to `BaseModal`?


jquense (commented):
if BaseModal accepts a ref, then yeah lets pass it through

kyletsang:
Done

jquense (approved):

PR #5738, opened 2021-03-24
feat(v5): add Offcanvas component

kyletsang:
https://getbootstrap.com/docs/5.0/components/offcanvas/

Offcanvas is a simpler version of Modal, so ended up using Modal as a base and removed the unneeded pieces.  I also moved code from `ModalHeader` into `AbstractModalHeader` so both `ModalHeader` and `OffcanvasHeader` can share it.  They only differ by a `bsPrefix`.

I used a root close handler in here to handle the case where there isn't a backdrop.

Preview: https://deploy-preview-5738--react-bootstrap-v5.netlify.app/components/offcanvas/

### DIFF
--- a/src/AbstractModalHeader.tsx
+++ b/src/AbstractModalHeader.tsx
@@ -1,0 +1,81 @@
+import PropTypes from 'prop-types';
+import * as React from 'react';
+import { useContext } from 'react';
+import useEventCallback from '@restart/hooks/useEventCallback';
+import CloseButton, { CloseButtonVariant } from './CloseButton';
+import ModalContext from './ModalContext';
+
+export interface AbstractModalHeaderProps
+  extends React.HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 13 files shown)


jquense (approved):
this is awesome, nice work

NickCoooke:
Hey!  Newbie here to react and js. Could you go into a little more detail describing the mechanics of the position attribute? I'm struggling to understand your example, and why the following doesn't work:

<OffCanvasExample position='end' />

Thanks for your time and for making such an awesome component!

kyletsang:
@NickCoooke, have a look at the docs for Offcanvas here:
https://react-bootstrap-v5.netlify.app/components/offcanvas/

The prop you're looking for is actually called `placement`

If you're missing any styles, make sure you install Bootstrap and set up the css.  See this link for more details:
https://react-bootstrap-v5.netlify.app/getting-started/introduction/

PR #5741, opened 2021-03-27
reversable card

ahmedatallah00:
**Reversible Card**

This is a reversible and stylish card created from scratch. The card is extremely responsive. The component could be reusable and integrated within any website that uses bootstrap.


jquense:
Hey this is cool. Unfortunately, It's not the right fit for react bootstrap, which doesn't add new components over plain bootstrap. Thanks for show us though

Issue #5751, opened 2021-04-01
v5: Drop `noGutters` from Row


kyletsang:
`.no-gutters` has been removed in v5.

> Removed .no-gutters in favor of new gutter spacing classes. Use .g-0 instead. 

ref: https://getbootstrap.com/docs/5.0/migration/#grid-and-layout

Tasks
- [ ] Remove `noGutters` from Row
- [ ] Remove associated docs
- [ ] Remove associated tests

mdctleo:
Do you mind if I take this one?

kyletsang:
> 
> 
> Do you mind if I take this one?

Sure go for it.  Thanks!

christo9090:
It appears this was never completed? Just upgraded to bootstrap 5 and react bootstrap 2.2.1 and looks like noGutters was removed by g-0 has no effect on a Row component? Am I missing something or is this still open?

PR #5761, opened 2021-04-07
changed .no-gutter to g-0

mdctleo:
#5751 @kyletsang can't seem to add reviewers, can you let me know if I understood the specs correctly? Thank you!

### DIFF
--- a/src/Row.tsx
+++ b/src/Row.tsx
@@ -150,7 +150,7 @@ const Row: BsPrefixRefForwardingComponent<'div', RowProps> = React.forwardRef<
         className={classNames(
           className,
           decoratedBsPrefix,
-          noGutters && 'no-gutters',
+          noGutters && 'g-0',
           ...classes,
         )}
       />
--- a/src/Row.tsx
+++ b/src/Row.tsx
@@ -51,8 +51,6 @@ const propTypes = {
    */
   bsPrefix: PropTypes.string,
 
-  /** Removes the gutter spacing between `Col`s as well as any added negative margins. */
-  noGutters: PropTypes.bool.isRequired,
   as: PropTypes.elementType,
 
   /**
@@ -150,7 +148,6 @@ const Row: BsPrefixRefForwardingComponent<'div', RowProps> = React.forwardRef<
         className={classNames(
           className,
           decoratedBsPrefix,
-          noGutters && 'g-0',
           ...classes,
         )}
       />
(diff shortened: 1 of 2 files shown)


kyletsang:
I was thinking of removing the `noGutters` prop entirely, so that way we can add a better prop in the future that allows us to set all the different gutter sizes (not sure what that looks like yet though).

kyletsang:
Don't worry about merging master into your branch.  I need to merge this manually into bs5-dev once I finish some other minor things.

src/Row.tsx
```diff
@@ -150,7 +148,6 @@ const Row: BsPrefixRefForwardingComponent<'div', RowProps> = React.forwardRef<
         className={classNames(
           className,
           decoratedBsPrefix,
-          noGutters && 'no-gutters',
```
kyletsang, line 153: Need to remove `noGutters` from the TS interface above as well.  Also don't forget to fix the warnings after you're done.

mdctleo: Done! And sorry, I merged master to see if it would fix CI fails

kyletsang (changes requested):
There's some warnings that need to be resolved:

>  src/Row.tsx#L116
>
> 'noGutters' is defined but never used. Allowed unused args must match /^_/u                                                                

You can remove the `noGutters` here

>  src/Row.tsx#L147
> 
> Replace `⏎··········className,⏎··········decoratedBsPrefix,⏎··········...classes,⏎········` with `className,·decoratedBsPrefix,·...classes`

This can be fixed by running Prettier and that will format the code for you.

kyletsang:
Hey @mdctleo, still interested in finishing this?  It’s pretty close.

src/Row.tsx
```diff
@@ -101,9 +101,7 @@ const propTypes = {
   xxl: rowColumns,
 };
 
-const defaultProps = {
-  noGutters: false,
-};
+const defaultProps = {};
```
kyletsang, line 104: We can drop defaultProps now since there's nothing to set in it

kyletsang (changes requested):
Can you also rebase your branch so it just includes your changes?  The `bs5-dev` branch has been updated with all the changes in master already. Thanks!

kyletsang:
Looks good.  Once you rebase the branch we'll be good to go.

mdctleo:
@kyletsang  hopefully I rebased correctly, thank you for your help and guidance!

kyletsang (approved):

kyletsang:
Nice job, thanks!

PR #5787, opened 2021-05-02
docs(contributing): expand Documentation section and add links

kelley-sharp:
Hello! First time contributor here. As I was reading the contributor docs, I had questions about where the Component API docs were being auto generated and with what technologies. So I thought I would provide more information in that section for other newcomers. Let me know if I can do anything to improve these changes or if you need to reject my request for any reason. Thanks!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,15 +46,16 @@ desired change easier.
 
 ## Documentation
 
-Please update the docs with any API changes, the code and docs should always be
-in sync.
+Please update the docs with any API changes, the code and docs should always be in sync.
 
-Component prop documentation is generated automatically from the React components
-and their leading comments. Please make sure to provide comments for any `propTypes` you add
-or change in a Component.
+The main documentation lives in the `www` folder which is a Gatsby project that uses [MDX](https://www.gatsbyjs.com/docs/how-to/routing/mdx/). The long-form documentation for components, including interactive examples and guides, is found within the `pages/components` directory.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -46,15 +46,16 @@ desired change easier.
...
-Please update the docs with any API changes, the code and docs should always be
-in sync.
+Please update the docs with any API changes, the code and docs should always be in sync.
 
-Component prop documentation is generated automatically from the React components
-and their leading comments. Please make sure to provide comments for any `propTypes` you add
-or change in a Component.
+The main documentation lives in the `www` folder which is a Gatsby project that uses [MDX](https://www.gatsbyjs.com/docs/how-to/routing/mdx/). The long-form documentation for components, including interactive examples and guides, is found within the `pages/components` directory.
+
+Separately, component prop API documentation is generated automatically from the React components in the `src` directory and their leading comments. This is the documentation that shows up in the tables at the bottom of component doc pages, e.g. [here](https://react-bootstrap.github.io/components/accordion/#api). Please make sure to provide [JSDOC-style](https://jsdoc.app/about-getting-started.html) comments* for any `propTypes` you add or change in a component.
```
jquense, line 53: I guess since we moved to typescript the comment style is closer to (if not exactly) https://tsdoc.org/ which is jsdoc but with most of the type info being optional since typescript already has type info.

jquense (commented):
this is a great refresh thanks!

CONTRIBUTING.md
kelley-sharp: Gotcha. I updated the link. Does that look good?

jquense (approved):

PR #5795, opened 2021-05-05
Removed unnecessary dependency "@types/classnames@2.3.1".

Popescu-PfeifferMarc:
The dependency "@types/classnames@2.3.1" is a stub type definition, because classnames provides its own type definitions.


kyletsang:
`classnames` should also be updated to the latest version which bundles the types.

PR #5807, opened 2021-05-10
Add custom variant support for Dropdown.Menu Component.

ubheamar:
In the earlier version, only dark variant was available by adding type string developer will have more control of dropdown variant.


kyletsang (approved):

kyletsang:
Thanks!

PR #5816, opened 2021-05-16
fix: make a11y id prop consistent

kyletsang:
Noticed that `id` is a bit inconsistent in several components.  Docs mention it should be required, but TS types sometimes require it/sometimes not. 

I changed it so it's consistent everywhere and removed ability to set a number, just so it matches base TS types for `id` which is a string.

### DIFF
--- a/src/DropdownButton.tsx
+++ b/src/DropdownButton.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
-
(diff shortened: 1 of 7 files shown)


kyletsang:
@jquense not entirely sure if this approach is correct, but we should try to make it somewhat consistent

jquense:
the restart/ui changes would actually make this obsolete, they use auto generated ID's when not provided

kyletsang:
> 
> 
> the restart/ui changes would actually make this obsolete, they use auto generated ID's when not provided

Awesome!

Issue #5797, opened 2021-05-07
Border of Get Started Button has margin greater in Mobile mode (320 px) 


Anmolreshi:
I could Modify this ! Shall I send a PR

kyletsang:
Seems fine to me.  Is there an issue this is addressing?

Anmolreshi:
The Button is not fitting the container in 320px . Visit This Link where I have shown the issue of button overflow.
Link : https://drive.google.com/file/d/1gw5R3RA8Qfh1i64qYVEm6YutIIcDiUoH/view?usp=sharing

PR #5827, opened 2021-05-20
Fixed homepage appearance in mobile view

nancy2681:
Fixed issue #5797 

Mobile View

Before :

image

After:

image

Web View

Before:

image

After:

image

### DIFF
--- a/www/src/pages/index.js
+++ b/www/src/pages/index.js
@@ -84,7 +84,7 @@ const ButtonToolbar = styled('div')`
     margin: -1rem;
 
     & > * {
-      width: 90%;
+      width: calc(100% - 1.85rem);
       max-width: 300px;
       margin: 1rem;
     }
--- a/www/src/pages/index.js
+++ b/www/src/pages/index.js
@@ -86,7 +86,7 @@ const ButtonToolbar = styled('div')`
     & > * {
       width: 100%;
       max-width: 300px;
-      margin: 1rem;
+      margin: 0.1rem;
     }
   }
 `;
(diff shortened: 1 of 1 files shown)


nancy2681:
@kyletsang Can you please review!?

www/src/pages/index.js
```diff
@@ -84,7 +84,7 @@ const ButtonToolbar = styled('div')`
     margin: -1rem;
 
     & > * {
-      width: 100%;
+      width: 90%;
```
kyletsang, line 87: I think it's better to use `calc()` here because the width does not account for the margins hence the overflow

thelokeshgoel00:
The Website doesn't looks good after the update on desktop, this reduced height isn't better. Maybe, separate code base for height on mobile and desktop

nancy2681:
> The Website doesn't looks good after the update on desktop, this reduced height isn't better. Maybe, separate code base for height on mobile and desktop

There is no significant visible difference I can see with this code change. If you can send screenshot we will get to know more, otherwise i think everything is working fine

nancy2681:
After using calc():

image

nancy2681:
@kyletsang I have made required changes. Please review

www/src/pages/index.js
```diff
@@ -84,7 +84,7 @@ const ButtonToolbar = styled('div')`
     margin: -1rem;
 
     & > * {
-      width: 100%;
+      width: calc(100% - 1.85rem);
```
kyletsang, line 87: Should actually be `calc(100% - 2rem)` since the margins added on each side sum up to 2.

nancy2681:
@kyletsang Made the required changes. Can you please review!

jquense (approved):

kyletsang (approved):

Issue #5788, opened 2021-05-03
Cannot adjust modal transition duration


sebammon:
## Describe the bug

I want to adjust the duration of the modal's TRANSITION_DURATION and BACKDROP_TRANSITION_DURATION but this is not possible. I've had a look at the source code and it seems that there two properties on the Modal: `Modal.TRANSITION_DURATION` and `Modal.BACKDROP_TRANSITION_DURATION` but changing them produces no results.

It becomes apparent when you adjust the CSS animation for the `.fade` class to something greater than 300ms, since this is the default value for the `TRANSITION_DURATION`.

## To Reproduce

Have a look at this minimal example [CodeSandbox](https://codesandbox.io/s/pedantic-torvalds-5ller). You will notice that when the Modal opens, the CSS animations work perfectly. However, when the modal is hidden, the Modal unmounts before the animation is complete.

## Expected behavior
I would expect that it is possible to at least change the duration of the TRANSITION_DURATION and BACKDROP_TRANSITION_DURATION property.

## Possible solution

I would be happy to help on this issue. If I'm not mistaken, it may just be a matter of passing the two value to the functions `DialogTransition(...)` and `BackdropTransition(...)`.

kyletsang:
I wonder if it's better to omit the timeout here in the transition since we already use the `addEndListener` handler.  This way, the duration can be controlled directly in the css.

sebammon:
@kyletsang Are you talking about the [animation events](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event)? Is it supported enough across browsers?

kyletsang:
> 
> 
> @kyletsang Are you talking about the [animation events](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event)? Is it supported enough across browsers?

[Transition end](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event) events.  We already hooked up a handler here:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Fade.tsx#L103

`Fade` has a default `timeout` value which I believe is causing the animation to cut out early.  I think we can override the default by passing a null or 0 to `Fade` inside the `DialogTransition`/`BackdropTransition`.

sebammon:
[...]

PR #5837, opened 2021-05-25
Update Modal's dialog and backdrop transition timeout

sebammon:
This PR addresses issue: #5788 

When setting a transition that is longer than the default `300ms` timeout as set by the `Fade` component's default props, the Modal is unmounted on exit before completing the animation. By setting the `timeout=null` for the Fade component, we let the Transition complete fully through the `transitionEndListener`.


kyletsang (approved):

kyletsang:
Thanks!

Issue #5753, opened 2021-04-01
v5: RTL support


kyletsang:
Ref: https://getbootstrap.com/docs/5.0/migration/#rtl
BS5 Cheatsheet: https://getbootstrap.com/docs/5.0/examples/cheatsheet-rtl/

CSS support should be finished, so all we need now is the JS support.

Here are some components that need updating from a quick glance at bootstrap's source:
- [ ] Dropdown
- [x] Carousel
- [ ] Modal
- [ ] Tooltip

Maybe we should add a setting to `ThemeProvider` for RTL?

Up for discussions here.

golota60:
Hi,

I wanted to pick up `Modal` changes and I've noticed that there *might* be a few more things to change for RTL.
- `dismissible` prop for `Alert` should have close icon aligned to the left
- `Offcanvas` - also close icon should be aligned to the left
- `Toast` - also close icon should be aligned to the left

All of above boil down to `CloseIcon` to be aligned to the left when RTL is true

- `Accordion` - should have the `^` icon aligned fully to the left

Please confirm those, as I've got no experience with RTL 😄

kyletsang:
Those are actually handled by the css, so if you import the `bootstrap.rtl.min.css` instead of the original, you'll see the changes. I think I've gotten all the css bits working already.  Only thing missing is anything JS related.

For the remaining RTL tasks, I think we would need to integrate `@restart/ui` prior to finishing them.  These are the things left to do:

- Modal - Add padding/margin to left side instead of the right when handling scrollbars, and some navbar-related items.  I did the scrollbar handling in the base ModalManager in RUI already.
- Dropdown - Right-align menus by default.  RUI has a new placement API that makes this a bit easier.
- Tooltip - Flip start/end placements.  Actually I think this might not need RUI, but I kinda left it since the other 2 above did.

PR #5855, opened 2021-06-02
feat(Carousel): add RTL keyboard support

kyletsang:
Part of #5753

### DIFF
--- a/src/ThemeProvider.tsx
+++ b/src/ThemeProvider.tsx
@@ -2,29 +2,49 @@ import PropTypes from 'prop-types';
 import * as React from 'react';
 import { useContext, useMemo } from 'react';
 
-export interface ThemeProviderProps {
-  prefixes: Record<string, unknown>;
+export interface ThemeContextValue {
+  prefixes: Record<string, string>;
+  dir?: string;
 }
 
-const ThemeContext = React.createContext({});
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface ThemeProviderProps extends Partial<ThemeContextValue> {}
+
+const ThemeContext = React.createContext<ThemeContextValue>({ prefixes: {} });
 const { Consumer, Provider } = ThemeContext;
 
-function ThemeProvider({ prefixes, children }) {
-  const copiedPrefixes = useMemo(() => ({ ...prefixes }), [prefixes]);
+const ThemeProvider: React.FC<Partial<ThemeContextValue>> = ({
+  prefixes = {},
+  dir,
+  children,
+}) => {
+  const contextValue = useMemo(
+    () => ({
+      prefixes: { ...prefixes },
+      dir,
+    }),
+    [prefixes, dir],
+  );
 
-  return <Provider value={copiedPrefixes}>{children}</Provider>;
-}
+  return <Provider value={contextValue}>{children}</Provider>;
+};
 
 ThemeProvider.propTypes = {
-  prefixes: PropTypes.object.isRequired,
-};
+  prefixes: PropTypes.object,
+  dir: PropTypes.string,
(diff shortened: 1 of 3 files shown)


src/ThemeProvider.tsx
```diff
@@ -2,31 +2,51 @@ import PropTypes from 'prop-types';
...
-const ThemeContext = React.createContext({});
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface ThemeProviderProps extends Partial<ThemeContextValue> {}
+
+const ThemeContext = React.createContext<ThemeContextValue>({ prefixes: {} });
 const { Consumer, Provider } = ThemeContext;
 
-function ThemeProvider({ prefixes, children }) {
-  const copiedPrefixes = useMemo(() => ({ ...prefixes }), [prefixes]);
+const ThemeProvider: React.FC<Partial<ThemeContextValue>> = ({
```
jquense, line 16: what's the reason btw for switching to an arrow function here? I find the typing slightly harder to understand vs a plain function

src/ThemeProvider.tsx
```diff
@@ -2,31 +2,51 @@ import PropTypes from 'prop-types';
...
 export function useBootstrapPrefix(
   prefix: string | undefined,
   defaultPrefix: string,
 ): string {
-  const prefixes = useContext(ThemeContext);
+  const { prefixes } = useContext(ThemeContext);
   return prefix || prefixes[defaultPrefix] || defaultPrefix;
 }
 
+export function useRTL() {
```
jquense, line 45:
```suggestion
export function useIsRTL() {
```
Maybe? just a thought

src/ThemeProvider.tsx
```diff
@@ -2,31 +2,51 @@ import PropTypes from 'prop-types';
...
+  const { dir } = useContext(ThemeContext);
+  return dir === 'rtl';
```
jquense, line 47: What alternatives to storing this in context have we considered? This adds a new RB specific api required in addition to the upstream one. What about stealing their JS helper? https://github.com/twbs/bootstrap/blob/main/js/src/util/index.js#L230 

If we want a reactive hook still could combine with a mutationObserver to trigger updates, though i'm not sure how common dynamically changing the read direction is aside from demonstrations in docs

src/ThemeProvider.tsx
kyletsang: Yeah `useIsRTL` is more clear.  I flip flopped between the 2 when working on it 😛

src/ThemeProvider.tsx
kyletsang: I used the context to support SSR since `document` wasn't available.  Wasn't sure if there was another way?

jquense: hmm you are right. what if we made this a function? presumably this is actually only relevant on the client, but we'd need to defer accesing the document til after render...or maybe set `dir` state in a useEffect? Do we think there is a reason why we'd need this during a SSR run, i suspect most usage is in event handlers.

kyletsang: From my list before, the components that need RTL changes are dropdown, modal, carousel, tooltip.  I think you're right that these aren't really needed during SSR - will have to double check.

kyletsang: @jquense looking at upstream, this is where RTL is used:
- Modal - setting padding for scrollbar when modal is shown
- Dropdown - menu placement
- Tooltip/Popover - tooltip/popover placement
- Carousel - keyboard control

Seems like we would need SSR support for dropdowns with renderOnMount or tooltips/popovers with show=true?

jquense (approved):

PR #5863, opened 2021-06-04
fix(TabPane): support mountOnEnter/unmountOnExit when not using a transition

### LINKED ISSUE #5048: Support for mountOnEnter on TabContainer with transition={false}
## Is your feature request related to a problem? Please describe
Currently when TabContainer has transition set to false, mountOnEnter does nothing. Only after removing transition, mountOnEnter starts to work. It would be nice, if both props could be used simultaneously

Add support for mountOnEnter even if transition={false}

Or if there is another solution to this issue, please tell me, maybe I'm missing something. Thanks

akx:
Since #4312, marking a TabPane as `unmountOnExit` will have had the same effect as `mountOnEnter`, i.e. that the component is not there at all when the tab is not active. 

This adds full support mount-on-enter and unmount-on-exit also when not using a Transition.

Follows up on #3497, #4312 (cc @fungjj92)
Fixes #5048

### DIFF
--- a/src/TabPane.tsx
+++ b/src/TabPane.tsx
@@ -149,8 +149,7 @@ const TabPane: BsPrefixRefForwardingComponent<'div', TabPaneProps> =
     } = useTabContext(props);
 
     const prefix = useBootstrapPrefix(bsPrefix, 'tab-pane');
-
-    if (!active && !Transition && unmountOnExit) return null;
+    const hasMountedRef = React.useRef(false);
 
     let pane = (
       <Component
@@ -162,7 +161,9 @@ const TabPane: BsPrefixRefForwardingComponent<'div', TabPaneProps> =
       />
     );
 
-    if (Transition)
+    if (Transition) {
+      // If we're using transitions, the Transition will deal
+      // with mountOnEnter/unmountOnExit behavior.
       pane = (
         <Transition
           in={active}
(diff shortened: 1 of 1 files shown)


src/TabPane.tsx
```diff
@@ -150,7 +150,7 @@ const TabPane: BsPrefixRefForwardingComponent<'div', TabPaneProps> =
 
     const prefix = useBootstrapPrefix(bsPrefix, 'tab-pane');
 
-    if (!active && !Transition && unmountOnExit) return null;
+    if (!active && !Transition && (mountOnEnter || unmountOnExit)) return null;
```
kyletsang, line 153: We'll need to continue rendering the tab pane when it's not active after it has been shown once, similar to the behavior in react-transition-group.  I think we can track that in a ref here?

akx: Should the pane just be `display: none` in that case?

kyletsang: Yeah, the `.active` class controls the `display: block` for the current visible pane, and the rest are set to `display: none` by default.

akx: Reworked the logic. I'd appreciate extra eyes on it – so many booleans 😂

kyletsang: Just tested this and having just `unmountOnExit` is not working.  I think we might be able to get away with adding the extra case of "not active and mountOnEnter and tab was never shown" to the original condition to return null

Thinking like this:

```
    const hasMountedRef = React.useRef(false);

    if (
      !active &&
      !Transition &&
      (unmountOnExit || (mountOnEnter && !hasMountedRef.current))
    ) {
      return null;
    }

    ...

    hasMountedRef.current = true;

    return ...
```

We'll definitely need some tests here since there's several booleans like you mentioned 😛

akx:
Guess this is outdated. Closing to clean up my PR list ✌️

PR #5873, opened 2021-06-09
Added autoClose behaviour for dropdowns

jotoh98:
More about the autoClose behaviour [here](https://getbootstrap.com/docs/5.0/components/dropdowns/#auto-close-behavior).

Additionally to the implementation, I've added

- a playground example
- a documentation section
- unit-tests

Preview: https://deploy-preview-5873--react-bootstrap.netlify.app/

### DIFF
--- a/src/Dropdown.tsx
+++ b/src/Dropdown.tsx
@@ -47,7 +47,7 @@ export interface DropdownProps
   focusFirstItemOnShow?: boolean | 'keyboard';
   onSelect?: SelectCallback;
   navbar?: boolean;
-  autoClose: 'true' | 'outside' | 'inside' | 'false';
+  autoClose: boolean | 'outside' | 'inside';
 }
 
 const propTypes = {
@@ -129,13 +129,13 @@ const propTypes = {
    * Controls the auto close behaviour of the dropdown when clicking outside of
    * the button or the list.
    */
-  autoClose: PropTypes.oneOf(['true', 'outside', 'inside', 'false']),
+  autoClose: PropTypes.oneOf([true, 'outside', 'inside', false]),
 };
 
 const defaultProps: Partial<DropdownProps> = {
   navbar: false,
   align: 'start',
-  autoClose: 'true',
+  autoClose: true,
 };
 
 const Dropdown: BsPrefixRefForwardingComponent<'div', DropdownProps> =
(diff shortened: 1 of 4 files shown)


src/Dropdown.tsx
```diff
@@ -156,8 +165,20 @@ const Dropdown: BsPrefixRefForwardingComponent<'div', DropdownProps> =
         if (
           event.currentTarget === document &&
           (source !== 'keydown' || event.key === 'Escape')
-        )
+        ) {
           source = 'rootClose';
+
+          const noOuterCloseModes =
```
kyletsang, line 171: Might be worth leaving comments about click what elements triggers the closing of the dropdown so it's more clear

test/DropdownSpec.js
```diff
@@ -333,4 +333,141 @@ describe('<Dropdown>', () => {
       ).assertSingle('div.dropdown-menu');
     });
   });
+
+  describe('autoClose behaviour', () => {
+    const createDocumentListenersMock = () => {
```
kyletsang, line 338: Do we need this?  We have other methods in other tests to trigger click events.

kyletsang (commented):
Overall, looks good.  Thanks for taking this on.  Left a few comments.

src/Dropdown.tsx
```diff
@@ -47,6 +47,7 @@ export interface DropdownProps
   focusFirstItemOnShow?: boolean | 'keyboard';
   onSelect?: SelectCallback;
   navbar?: boolean;
+  autoClose: 'true' | 'outside' | 'inside' | 'false';
```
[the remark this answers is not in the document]
jotoh98: Yes, I thought of that as well. Wasn't sure about the consistency but I will change that.

test/DropdownSpec.js
jotoh98: I didn't notice any method concerned with simulating clicks outside of the component. Correct me if I'm wrong. Should we keep the outside click methods?

kyletsang: We did use `simulant.fire(document.body, 'click');` inside an Offcanvas test.  Can we try that and see if it works for your test?

www/src/examples/Dropdown/AutoClose.js
```diff
@@ -0,0 +1,15 @@
+<>
+  {[true, 'outside', 'inside', false].map((autoClose, index) => (
```
kyletsang, line 2: Wonder if we should rearrange/reword the examples so it mirrors upstream Bootstrap examples:
https://getbootstrap.com/docs/5.0/components/dropdowns/#auto-close-behavior

I got thrown off with the autoclose behavior for "inside" and "outside" differing from Bootstrap's docs, but turns out it was an issue of ordering/wording.

kyletsang (approved):

kyletsang:
Looks good to me - left 1 comment regarding docs demo.

@jquense, would be good to get your eyes on this.

jquense (approved):
LGTM, happy with how this was relatively straightforward to implement outside of react-overlays this is

kyletsang:
Thanks!

Issue #5644, opened 2021-02-08
<Tab/> component to receive custom HTML Attribute


Bizzle-Dapp:
## Is your feature request related to a problem? Please describe:

Currently i'm trying to get my Cypress tests to target the <Tab/> element by an "Automation-Id" as given by our acceptance tests. From my digging through the API and from further digging online. It appears this isn't possible.

## Describe the solution you'd like

Ideally, it would be nice to pass custom attributes to a <Tab/> element as such:

`<Tab tabProps={{"automation-id":"myAutomationTestTargetAttribute", "someOtherAttribute":"Blurgh"}} />`

These should then be reflected inside the `<a/>` attribute within the containing `<div/>`. 

## Describe alternatives you've considered

I've searched online, however have had to access the existing `id` attribute on the <a/> element provided by the Tab and augmented with the Title field.

## Additional context

N/A

PR #5879, opened 2021-06-12
feat(Tabs): Added possibility of passing custom attributes to Tab com…

golota60:
…ponent

Resolves:  #5644 

`<Tab />` can now receive `tabAttrs` object which will be passed to underlying navlink component as attributes where `key` is the name of attribute and `value` is it's value


rsaid99:
How do we get an approver for this?  This fix will save me the work of building my own tab control (so I can instrument the control with a Cypress tag for automated testing)

kyletsang (approved):

kyletsang:
Looks alright to me.  @jquense did you want to double check?

Issue #5864, opened 2021-06-04
New FormSelect component in v2 beta not setting the "id" attribute.


kb3eua:
## Describe the bug

The new `FormSelect` component for Bootstrap v5 is not applying the `controlId` of its parent `FormGroup` component for its `id` attribute.  This prevents the `<label>` from being related to the input as its `htmlFor` property points to a non-existent `id`.

## To Reproduce

1. define a `<Form.Group>` element with the `controlId` set
2. nest a `<Form.Label>` element under the `<Form.Group>`
3. nest a `<Form.Select>` element also under the `<Form.Group>`

## Reproducible Example

```tsx
<Form.Group controlId="test-id">
  <Form.Label>My Label</Form.Label>
  <Form.Select>
    <option>1</option>
    <option>2</option>
  </Form.Select>
</Form.Group>
```

## Expected behavior

HTML with an `id` attribute on the `<select>` element:
```html
<div>
  <label class="form-label" for="test-id">My Label</label>
  <select class="form-select" id="test-id">
    <option>1</option>
    <option>2</option>
  </select>
</div>
```

## Screenshots

![image](https://user-images.githubusercontent.com/4258026/120808660-4657a200-c517-11eb-858d-c224f91cc46a.png)

## Environment (please complete the following information)

- Operating System: Windows 10
- Browser, Version: all
- React-Bootstrap Version: 2.0.0-beta.0

## Additional context

Should behave similarly to how the `FormControl` component handles `id`/`controlId`

kyletsang:
Good catch.  Same needs to be done for `FormRange`

PR #5880, opened 2021-06-12
fix(form): fixed controlId override behavior on FormRange and FormSelect

golota60:
Resolves: #5864 

Fixes the bug described in the ticket, and applies the same `controlId` behavior to `<FormRange />` as specified in comment by @kyletsang


jquense (approved):

kyletsang:
Thanks!

Issue #5856, opened 2021-06-02
"modal-open" is not defined 


broofa:
## Describe the bug

**Note:**  This appears to be an issue with the latest react & react-bootstrap buildes.  Specifically:
* bootstrap@5.0.1
* react-bootstrap@2.0.0-alpha.2

When rendering a `Modal`, the body element is assigned the "modal-open" CSS class, however this class is not defined by `bootstrap`.

![image](https://user-images.githubusercontent.com/164050/120491133-b2d67380-c354-11eb-992a-e5112fd15844.png)

(FWIW, this was picked up by the [`checkcss`](https://www.npmjs.com/package/checkcss) module I wrote that monitors the DOM for references to undefined CSS classes.  'May be of interest to... someone?)

## Reproducible Example

https://codesandbox.io/s/stupefied-taussig-s0p72?file=/src/App.js:0-193

## Expected behavior

'Shouldn't be using CSS classes that aren't defined anywhere

## Environment (please complete the following information)

  System:
    OS: macOS 11.4
    CPU: (8) x64 Apple M1
    Memory: 15.25 MB / 16.00 GB
    Shell: 5.1.0 - /usr/local/bin/bash
  Binaries:
    Node: 14.16.1 - ~/.nvm/versions/node/v14.16.1/bin/node
    npm: 7.15.0 - ~/.nvm/versions/node/v14.16.1/bin/npm
  Browsers:
    Chrome: 91.0.4472.77
    Firefox: 88.0.1
    Safari: 14.1.1
  npmPackages:
    @parcel/babel-plugin-transform-runtime: 2.0.0-nightly.2312 => 2.0.0-nightly.2312
    @parcel/packager-raw-url: 2.0.0-nightly.2312 => 2.0.0-nightly.2312
    @parcel/transformer-sass: 2.0.0-nightly.690 => 2.0.0-nightly.690
    @parcel/transformer-svg-react: 2.0.0-nightly.2312 => 2.0.0-nightly.2312
    @parcel/transformer-webmanifest: 2.0.0-nightly.2312 => 2.0.0-nightly.2312
    @types/react-dom: 17.0.3 => 17.0.3
    @types/react-router-dom: 5.1.7 => 5.1.7
    @typescript-eslint/eslint-plugin: 4.20.0 => 4.20.0
    @typescript-eslint/parser: 4.20.0 => 4.20.0
    bootstrap: 5.0.1 => 5.0.1
[...]

PR #5882, opened 2021-06-13
fix(Modal): Remove unnecessary modal-open bootstrap class

golota60:
Resolves: #5856 

Removes `modal-open` as it was removed upstream in https://github.com/twbs/bootstrap/pull/33551

### DIFF
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -496,7 +496,6 @@ const Modal: BsPrefixRefForwardingComponent<'div', ModalProps> =
             onExiting={onExiting}
             onExited={handleExited}
             manager={getModalManager()}
-            containerClassName={`${bsPrefix}-open`}
             transition={animation ? DialogTransition : undefined}
             backdropTransition={animation ? BackdropTransition : undefined}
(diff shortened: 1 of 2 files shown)


golota60:
I compared the behavior to https://getbootstrap.com/docs/5.0/components/modal/ and it seems to behave exactly the same

jquense:
it looks like upstream still leaves the class to avoid a breaking change. Even if the class doesn't do anything. We can afford the breaking change but i'm not sure we should deviate from the upstream behavior more than we need to?

golota60:
I didn't think of that before - but yeah, I think you're right, that would add additional unnecessary complexity when migrating. 

It's just an empty class after all - it's not really "worth" a breaking change imo

golota60:
Feel free to close this when you make your mind up, i won't be too upset 😸

jquense:
@kyletsang  your call :P

kyletsang:
Ah didn't realize upstream kept the class around in their JS.  In that case, let's keep the behavior consistent.  Thanks for looking into this @golota60 👍

PR #5886, opened 2021-06-18
Fixed autoClose being mandatory

### LINKED ISSUE #5885: autoClose for dropdowns is mandatory
## Describe the bug

The autoClose property in the interface `DropdownProps` is mandatory though the specification states an implicit standard value. Upon working with `NavDropdown`, I stumbled across this.

## Expected behavior

The autoClose property should not be mandatory. If not defined, the standard value `true` should be used.

jotoh98:
Closes #5885.

Made autoClose optional, refactored code slightly and improved the naming in the doc to be like https://getbootstrap.com/docs/5.0/components/dropdowns/#auto-close-behavior


kyletsang (approved):

kyletsang:
Thanks, totally missed that

jotoh98:
@kyletsang Yeah, if I hadn't used it myself, I also would've missed it. This was the time working with `propTypes` together with a ts prop type declaration. Naming a `defaultProp` and leaving the `oneOf()` without `.isRequired` doesn't seem to satisfy typescript.

Issue #5732, opened 2021-03-22
missing `backdrop` attribute on `ModalProps` when using typescript


achmadk:
## Describe the bug

I am using react-bootstrap with typescript. I want to prevent modal component from being clicked outside its own element, so I just set `backdrop` props value into `static`. Unfortunately, `backdrop` wasn't show when using vscode intellisense. 

## To Reproduce

1. Import `ModalProps` from `react-bootstrap`
2. create dummy type variable called `test`
3. assign `test` variable into `ModalProps['backdrop']`
4. When `test` variable is hovered, its type is `boolean | "static"`, but `any` is shown.

## Reproducible Example

https://codesandbox.io/s/busy-mirzakhani-1ly1w?file=/src/App.tsx.

## Expected behavior

backdrop props type is boolean | 'static' | undefined
```ts
  backdrop?: boolean | 'static'
```

## Screenshots

![Screenshot from 2021-03-22 17-29-49](https://user-images.githubusercontent.com/8057028/111977856-d18ae380-8b35-11eb-94ed-f78aaa40094d.png)

## Environment (please complete the following information)

- Operating System: ubuntu
- Browser, Version: Chrome 89
- React-Bootstrap Version: 1.5.2

kyletsang:
Looks like `Omit` doesn't play nice with index signatures.  We could use a `Pick` here and redefine the index signature in RB `ModalProps`?  Unsure if there's a cleaner way.

golota60:
@kyletsang  I got around to fixing it by using `Pick`, but an idea came to my mind, but I'd like to get it confirmed before implementing, cause it would require changes in `react-overlays`. 

So, in `react-overlays`, instead of having `ModalProps` we could split them like so:
```js

interface IndexSignature { [other: string]: any};

export interface ModalPropsWithoutIndexSignature {
    children?: React.ReactElement;
    role?: string;
    style?: React.CSSProperties;
    className?: string;
    show?: boolean;
    container?: DOMContainer;
    onShow?: () => void;
    onHide?: () => void;
    manager?: ModalManager;
    backdrop?: true | false | 'static';
    renderDialog?: (props: RenderModalDialogProps) => React.ReactNode;
    renderBackdrop?: (props: RenderModalBackdropProps) => React.ReactNode;
    onEscapeKeyDown?: (e: KeyboardEvent) => void;
    onBackdropClick?: (e: React.SyntheticEvent) => void;
    containerClassName?: string;
[...]

PR #5894, opened 2021-06-27
fix(Modal): Fix wrong types on ModalProps

golota60:
Resolves: #5732

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "invariant": "^2.2.4",
     "prop-types": "^15.7.2",
     "prop-types-extra": "^1.1.0",
-    "react-overlays": "^5.1.0",
+    "react-overlays": "^5.1.1",
     "react-transition-group": "^4.4.1",
     "uncontrollable": "^7.2.1",
     "warning": "^4.0.3"
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "invariant": "^2.2.4",
     "prop-types": "^15.7.2",
     "prop-types-extra": "^1.1.0",
-    "react-overlays": "^5.0.1",
+    "react-overlays": "^5.1.0",
     "react-transition-group": "^4.4.1",
     "uncontrollable": "^7.2.1",
     "warning": "^4.0.3"
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -412,8 +412,8 @@ const Modal: BsPrefixRefForwardingComponent<'div', ModalProps> =
         onExit?.(node);
       };
 
-      const handleEntering = (node) => {
-        onEntering?.(node);
+      const handleEntering = (node, ...args) => {
+        onEntering?.(node, ...args);
 
         // FIXME: This should work even when animation is disabled.
         addEventListener(window as any, 'resize', handleWindowResize);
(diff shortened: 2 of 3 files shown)


golota60:
Sorry for waiting such a long time on this, had a busy week.

Anyways, this change probably picked up(and allowed me to fix) a bug where `onExited` wasn't passing a node(see screenshot)
image

I've removed passing of `...args` in `handleEnter`, `handleExit`, `handleExited`,`handleExited` - I've checked and it doesn't break anything, and I'm pretty sure they weren't supposed to be there in the first place(I've traced the code and there is always only one `node` argument passed).

I've tried to trace a reason as to why/when `..args` were added(and if they're passed for a reason) and found https://github.com/react-bootstrap/react-bootstrap/pull/3187 , which leads me to believe they are not supposed to be there

kyletsang:
> 
> 
> Sorry for waiting such a long time on this, had a busy week.
> 
> Anyways, this change probably picked up(and allowed me to fix) a bug where `onExited` wasn't passing a node(see screenshot)
> 
> I've removed passing of `...args` in `handleEnter`, `handleExit`, `handleExiting`,`handleExited` - I've checked and it doesn't break anything, and I'm pretty sure they weren't supposed to be there in the first place(I've traced the code and there is always only one `node` argument passed).
> 
> I've tried to trace a reason as to why/when `..args` were added(and if they're passed for a reason) and found #3187 , which leads me to believe they are not supposed to be there

Nice catch on the missing node arg in `onExited`.  

We should keep the args, because all the "enter" callbacks do have an additional `isAppearing` arg in them: (http://reactcommunity.org/react-transition-group/transition/#Transition-prop-onEnter).  There's still an issue in the `TransitionCallbacks` type in `helpers.ts` though, because it only specify a node arg.

golota60:
Thanks for looking into this!

It seems like `...args` should be only in `handleEnter`, cause according to http://reactcommunity.org/react-transition-group/transition/#Transition-prop-onExit, `onExit`,`onExiting` and `onExited` are passing only `node`.

Commited the change, but it still needs https://github.com/react-bootstrap/react-overlays/pull/962.

kyletsang:
> 
> 
> Thanks for looking into this!
> 
> It seems like `...args` should be only in `handleEnter`, cause according to http://reactcommunity.org/react-transition-group/transition/#Transition-prop-onExit, `onExit`,`onExiting` and `onExited` are passing only `node`.
> 
> Commited the change, but it still needs [react-bootstrap/react-overlays#962](https://github.com/react-bootstrap/react-overlays/pull/962).

`handleEntering` needs it as well.  Hmm seems like CI is failing, but shouldn't be related to your changes...

golota60:
@kyletsang  I wasn't able to pin down why `...args` wasn't working(TS just gave me `A spread argument must either have a tuple type or be passed to a rest parameter.` which wasnt really helpful). But since we only will ever pass `isAppearing`, I just decided to replace `...args` with it.

golota60:
@kyletsang I wasn't able to pin down why ...args wasn't working(TS just gave me A spread argument must either have a tuple type or be passed to a rest parameter. which wasnt really helpful). But since we only will ever pass isAppearing, I just decided to replace ...args with it.

But in order for this to pass CI, we need https://github.com/react-bootstrap/react-overlays/pull/962 to be merged & released.

kyletsang:
@golota60, sorry missed that PR.  It's merged and released in 5.1.1 now

golota60:
@kyletsang everything should be ready to merge 😄

kyletsang (approved):

kyletsang:
Thanks!

PR #5907, opened 2021-07-05
replace deprecated lifecycle method in Dropdown.js

Rogach:
Fixes the warning in React.StrictMode.

### DIFF
--- a/src/Dropdown.js
+++ b/src/Dropdown.js
@@ -132,13 +132,14 @@ class Dropdown extends React.Component {
     this.focusNextOnOpen();
   }
 
-  UNSAFE_componentWillUpdate(nextProps) { // eslint-disable-line
-    if (!nextProps.open && this.props.open) {
+  getSnapshotBeforeUpdate(prevProps) {
+    if (!this.props.open && prevProps.open) {
       this._focusInDropdown = contains(
         ReactDOM.findDOMNode(this.menu),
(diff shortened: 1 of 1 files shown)


Rogach:
Hi! Can somebody help me with understanding the checks failure?
Running `cd www && yarn && yarn build` completes successfully for me on my local machine.
Could the error be somehow related to corejs version update in #5225 ?

Rogach:
Apologies, seems I mixed up some branches. `yarn build` in `www` directory does seem to fail with the same errors locally. Reverting #5225 fixes the build, I'll look into it.

jquense:
Hi @Rogach thanks for taking the time to send over a PR. While, we definitely appreciate the work, I don't believe we are accepting any more patches to the bs3 line, now that that v5 is out. It's too much of a maintenance burden to support 3 major versions of bootstrap

jquense:
cc @kyletsang

Rogach:
@jquense I understand that core react-bootstrap developers are not going to support the v3, and that's okay. But my idea was that the community, like me in this example :) can provide some small ongoing patches to keep v3 alive.

A lot of people are still using bootstrap v3 and not planning to migrate for various reasons (e.g. not happy with the new look or too hard to migrate).

Rogach:
What do you think (about community-supported v3 branch)?

PR #5911, opened 2021-07-06
feat(types): increase perf of dynamic types by loosening the type slightly

jquense:
I stole this from styled-components DT type def. Basically the large union is a lot of the per cost to these types. By having it be just `string` most of the way down saves a lot of time. The downside is that you don't get strict validation of the string types in `as` but I think that's a reasonable trade off.


kyletsang:
Looks like we need to update `as` inside `AbstractNavProps` and `AbstractNavItemProps` to match

Issue #5914, opened 2021-07-08
Feature to copy code at the click of the button in documentation.


VibhorJaiswal:
As of currently there is no easy way to copy code from the documentation without manually selecting and copy the code.

I would like to contribute this feature by which code can be copied by a single click of a button.

kyletsang:
Sure go for it, thanks!

VibhorJaiswal:
ok! I already started working on it!

VibhorJaiswal:
How do I run it on my local machine? It is giving me this error. 
![image](https://user-images.githubusercontent.com/69683278/125103881-65d78100-e0fa-11eb-84ca-5e14c45ca69b.png)

kyletsang:
> 
> 
> How do I run it on my local machine? It is giving me this error.

https://github.com/react-bootstrap/react-bootstrap#local-setup

`yarn run bootstrap` will install all the deps for the lib + docs site

VibhorJaiswal:
Hey! The code is ready but, Can you please tell where should be the related css?

VibhorJaiswal:
@kyletsang reminder that I have opened a pull request with all the features discussed in this issue. It has passed all the checks and is completely bug free so please review the code.

kyletsang:
@VibhorJaiswal  Sorry a bit tied up with other tasks at the moment.  Will review when I get a chance.

PR #5919, opened 2021-07-11
feat(copy button): Added button to copy example code.

VibhorJaiswal:
Added a button to copy example code without copying it manually. As discussed in #5914 .
Here are the related screenshots.

image
------------------------------------------------------------------------------------------------
image
------------------------------------------------------------------------------------------------
image


eirikb (approved):
LGTM

kyletsang (changes requested):

VibhorJaiswal:
1) react-bootstrap `Tooltip` does not scale according to its contents, so when the content inside the `Tooltip` changes it results like this 
image

also using two different `Tooltips ` will decrease code readability significantly. Although I did use the react-bootstrap `Button`.

2) 

> It's better to declare this in a function like `handleBlur` for readability

Done ✅


3) 

> Haven't looked too deeply into react-live, but does the editor expose a way for you to set code as a value so we can use it as a controlled editor?

As far as I know there is a  `onChange` event but when using it, the `react-live` stops showing errors on erroneous code.
Also when I tested, it was working totally fine when using `onBlur` event(also when the code is changed).

www/src/components/ReactPlayground.js
```diff
@@ -219,18 +223,22 @@ function Editor() {
...
   const showMessage = keyboardFocused || (focused && !ignoreTab);
 
+   
+
   return (
     <div className="position-relative">
       <StyledEditor
         onFocus={handleFocus}
-        onBlur={handleBlur}
+        onBlur={(e) => {handleBlur(), handleCodeChange(e)}}
```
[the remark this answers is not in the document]
VibhorJaiswal: 1) react-bootstrap `Tooltip` does not scale according to its contents, so when the content inside the `Tooltip` changes it results like this 
image

also using two different `Tooltips ` will decrease code readability significantly. Although I did use the react-bootstrap `Button`.

2) 

> It's better to declare this in a function like `handleBlur` for readability

Done ✅


3) 

> Haven't looked too deeply into react-live, but does the editor expose a way for you to set code as a value so we can use it as a controlled editor?

As far as I know there is a  `onChange` event but when using it, the `react-live` stops showing errors on erroneous code.
Also when I tested, it was working totally fine when using `onBlur` event(also when the code is changed).

VibhorJaiswal:
@kyletsang 
1) react-bootstrap `Tooltip` does not scale according to its contents, so when the content inside the `Tooltip` changes it results like this 
image

also using two different `Tooltips ` will decrease code readability significantly. Although I did use the react-bootstrap `Button`.

2) 

> It's better to declare this in a function like `handleBlur` for readability

Done ✅


3) 

> Haven't looked too deeply into react-live, but does the editor expose a way for you to set code as a value so we can use it as a controlled editor?

As far as I know there is a  `onChange` event but when using it, the `react-live` stops showing errors on erroneous code.
Also when I tested, it was working totally fine when using `onBlur` event(also when the code is changed).

kyletsang:
> react-bootstrap `Tooltip` does not scale according to its contents, so when the content inside the `Tooltip` changes it results like this
>

Might be able to use this for the dynamic content:
https://react-bootstrap.github.io/components/overlays/#overlays-dynamic-updates

VibhorJaiswal:
@kyletsang 
1) > Can we use existing react-bootstrap components here (Button and Tooltip)? That way, you wouldn't have to define new styles for the button/tooltip. ✅

2) > It's better to declare this in a function like `handleBlur`for readability. ✅

www/src/components/ReactPlayground.js
```diff
@@ -256,11 +267,39 @@ const propTypes = {
...
+  const handleCodeChange = (e) => { setcodeToCopy(e.target.value) }
+
+  const handleCopy = () => {
+    navigator.clipboard.writeText(codeToCopy) //copies code to clipboard
+      .then(setCopy('Copied!'))
+
+  }
+
+
+  const UpdatingPopover = React.forwardRef(
```
kyletsang, line 287: Components shouldn't be nested within another component.  This should be declared outside.

www/src/components/ReactPlayground.js
```diff
@@ -269,8 +308,28 @@ function Playground({ codeText, exampleClassName, showCode = true }) {
...
+
+          <>
+            <div onMouseOut={resetCopyStatus}>
+              <OverlayTrigger
+                trigger={['hover', 'focus']}
+                overlay={
+                  <UpdatingPopover id="popover-contained">{copyStatus}</UpdatingPopover>
+                }
+              >
+                <Button onClick={handleCopy} className={styles.styledCopyButton} variant='dark' value={codeText} >
```
kyletsang, line 325: Is `value` needed here?

kyletsang (commented):
Looking good.  Couple more comments.  Can you also run prettier on the files you've changed?  I'm noticing there's some style inconsistencies.

VibhorJaiswal:
@kyletsang 
> Can you also run prettier on the files you've changed? I'm noticing there's some style inconsistencies.

I have prettified according to my prettier, if this is not what you want then can you please help me in this case.

SaltyBuckets:
> Can you also run prettier on the files you've changed? 

@kyletsang  tbh you should add a bot to do that to decrease the load on new contributers

kyletsang (approved):

kyletsang:
I think this is a good first pass.  @jquense, want to review?

kyletsang:
Sorry for delay on this.  Thanks for PR!

VibhorJaiswal:
@kyletsang no problem - thanks

PR #5920, opened 2021-07-11
Add Bootstrap v4.6.0 to the dependencies

e0ne:
React-Bootstrap can't work without Bootstrap so this patch adds it
to the dependencies.


kyletsang:
Thanks for the PR.  Bootstrap isn't included as a dependency because you can choose not to use the installed dependency.  For example, you can import it via CDN or compile/host your own elsewhere etc.

PR #5932, opened 2021-07-16
feat!: integrate @restart/ui

kyletsang:
Integration of @restart/ui. 

Items integrated:
- Button
- Modal
- Overlay
- Dropdown
- Nav/NavItem
- Tabs

Preview: https://deploy-preview-5932--react-bootstrap.netlify.app/


jquense:
🙌

src/Button.tsx
```diff
@@ -79,37 +72,23 @@ const defaultProps = {
...
+    ({ bsPrefix, variant, size, active, className, ...props }, ref) => {
       const prefix = useBootstrapPrefix(bsPrefix, 'btn');
 
-      const classes = classNames(
-        className,
-        prefix,
-        active && 'active',
-        variant && `${prefix}-${variant}`,
-        size && `${prefix}-${size}`,
+      return (
```
jquense, line 78: might be worth using the hook directly? avoids an extra component for a little more logic

test/NavSpec.js
```diff
@@ -151,139 +151,139 @@ describe('<Nav>', () => {
...
-          .simulate('keydown', {
-            key,
-          });
-
-        expect(document.activeElement).to.equal(anchors.at(0).getDOMNode());
-        sinon.assert.calledOnce(keyDownSpy);
-      });
-    });
-  });
+  // describe('keyboard navigation', () => {
```
kyletsang, line 154: Currently this doesn't work because restart ui checks for existence of a tab context before handling keys.  I guess this might be ok?  If users want keyboard control, they can use `Tabs` instead.

kyletsang:
@jquense I think I finally got everything 😄 

Ready for your review

test/NavSpec.js
jquense: I think this is correct yeah. At least i can't imagine why you'd want this keyboard behavior outside of a tablist

src/SelectableContext.tsx
```diff
@@ -1,14 +0,0 @@
...
-
-export const makeEventKey = (
-  eventKey?: string | number | null,
-  href: string | null = null,
-): string | null => {
-  if (eventKey != null) return String(eventKey);
-  return href || null;
-};
-
-export default SelectableContext;
```
jquense, line 14: should these be re-exported? IDK if users are using this directly, probably they shouldn't

jquense (approved):
LGTM i'm sure i missed something but lets go for it

src/SelectableContext.tsx
kyletsang: Let's leave it out for now and see if we get any feedback about it being missing

src/BootstrapModalManager.tsx
```diff
@@ -35,11 +36,14 @@ class BootstrapModalManager extends ModalManager {
...
-    super.setContainerStyle(containerState, container);
+  setContainerStyle(containerState: ContainerState) {
+    super.setContainerStyle(containerState);
 
-    if (!containerState.overflowing) return;
-    const size = getScrollbarSize();
+    const container = this.getElement();
+    addClass(container, 'modal-open');
+
+    const size = this.getScrollbarWidth();
```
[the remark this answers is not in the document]
kyletsang: Fixed

PR #5983, opened 2021-08-19
Fix `Form.Check.Label` spacing

### LINKED ISSUE #5938: `<Form.Check>` and `<Form.Check.Label>` should add `form-check` and `form-label` class automatically.
`<Form.Check>` and `<Form.Check.Label>` should add `form-check` and `form-label` class automatically.

https://react-bootstrap-v5.netlify.app/components/forms/#forms-check-api

TyMick:
Fixes #5938. Currently, when using a `Form.Check.Label` component to customize `Form.Check` rendering, there will be no space between the checkbox and the label. This is because `Form.Check` is currently relying on the presence of a `label` prop to apply the `form-check` class name to the wrapping `<div>`, because checkboxes without labels [don't need the wrapping element to have the `form-check` class name](https://github.com/react-bootstrap/react-bootstrap/issues/5938#issuecomment-883704209).

This PR adds a utility to check whether an element has a child of a certain type. It then uses that utility to check if a `Form.Check` element has a `Form.Check.Label` child and takes that into account when determining whether the checkbox has a label.

Adding a special property (currently called `typeName`, but that can certainly change) to components for this utility is necessary because React minifies the `displayName` property in production.

Here's a side-by-side screenshot of the current bug as shown in the documentation site, compared to the fixed version when I run the docs locally after this fix:

Screen Shot 2021-08-19 at 7 58 58 AM

-----

**_EDIT:_** Only the [changes from the first two commits](https://github.com/react-bootstrap/react-bootstrap/pull/5983/files/f1305813394de810eef1b5eb1c2fe90af42c74ed) are required to fix this spacing. The [third commit](https://github.com/react-bootstrap/react-bootstrap/commit/5822d94f6c3d19d2f66e5d84291cf3e668690d2a) adds a [related feature](https://github.com/react-bootstrap/react-bootstrap/pull/5983#issuecomment-902024263) that I maybe should've saved for a separate PR. 🙈


TyMick:
Another thing I just thought of is that `Form.Check` currently doesn't allow users to use a `label` prop and a custom `Form.Check.Input` at the same time...

https://github.com/react-bootstrap/react-bootstrap/blob/3ff951b0e711b020bdec202e7465fc8d34f126da/src/FormCheck.tsx#L181-L186

...since the presence of _any_ children will prevent a prop-generated label from rendering. Is that something we want to support?

kyletsang:
Wonder if we should allow bools for `label` here to control the `.form-check` class in the custom rendering scenario. 

@jquense this might be a cleaner approach?

TyMick:
Do you mean requiring the user to pass `label={true}` to `Form.Check` when using a custom `Form.Check.Label`? Just checking if I'm interpreting correctly. 👍🏼

TyMick:
Also I just thought of one way to allow users to use a `label` prop and a custom `Form.Check.Input` at the same time (or a custom `Form.Check.Label` without having to include a custom `Form.Check.Input`): 5822d94. I can revert if it's too much, though.

kyletsang:
> 
> 
> Do you mean requiring the user to pass `label={true}` to `Form.Check` when using a custom `Form.Check.Label`? Just checking if I'm interpreting correctly. 👍🏼

Yeah, so when using custom rendering, setting `label={true}` would toggle the class

```
<FormCheck label={true}>
  <FormCheckLabel>Custom</FormCheckLabel>
  ...
</FormCheck>
```

TyMick:
Right on. As a user, I'd prefer `FormCheck` to be able to detect a `FormCheckLabel` child element, but if there's a consensus that that feature isn't worth the overhead of adding a `typeName` property and a `getChildOfType` (or `includesType`) function, I can get behind that.

kyletsang:
@jquense your thoughts on this?

TyMick:
And FWIW, only the [changes from the first two commits](https://github.com/react-bootstrap/react-bootstrap/pull/5983/files/f1305813394de810eef1b5eb1c2fe90af42c74ed) are required to fix the `Form.Check.Label` spacing. The [third commit](https://github.com/react-bootstrap/react-bootstrap/commit/5822d94f6c3d19d2f66e5d84291cf3e668690d2a) adds a related feature that I maybe should've saved for a separate PR. 🙈

TyMick:
Thanks for the great feedback, @kyletsang and @jquense! How do things look after this latest commit? I've stopped trying to force the custom render components into a certain order, and now the only thing this PR adds to `FormCheck` is it checks to see if a direct child is a `FormCheckLabel` and takes that into account when applying the correct Bootstrap class. Then if someone decides to wrap `FormCheckLabel`, the official solution can be passing `label={true}` to `FormCheck`? I think checking for `FormCheckLabel` will help people out in the majority of cases, though.

src/FormCheck.tsx
```diff
@@ -152,7 +153,9 @@ const FormCheck: BsPrefixRefForwardingComponent<'input', FormCheckProps> =
         [controlId, id],
       );
 
-      const hasLabel = label != null && label !== false && !children;
+      const hasLabel =
+        (!children && label != null && label !== false) ||
+        hasChildOfType(children, FormCheckInput);
```
kyletsang, line 158: Should this be `FormCheckLabel` instead of `FormCheckInput`?

kyletsang (commented):
I think this should be ok since it covers the most general case? 

Your call @jquense 😄

src/FormCheck.tsx
TyMick: Whoops, good catch!

TyMick: [`f0d8e9f`](https://github.com/react-bootstrap/react-bootstrap/pull/5983/commits/f0d8e9f04b70d248a17c0f76d381ea6bded8c9c5)

kyletsang:
@TyMick can you remove the changes in both the package.json files?  We don't need those.

I'll merge after

jquense:
sorry i've been so absent. seems ok with me, certainly don't block on me

TyMick:
@kyletsang Sounds good! I found them useful for testing the production build, but they certainly aren't necessary. 👍🏼

kyletsang (approved):

kyletsang:
Thanks!  And sorry for the delays 😅

TyMick:
No worries; plenty of those delays were on my end anyway. 😊

Issue #5977, opened 2021-08-16
Collapse - Horionztal animation


willd-i2o:
As of v5.1 Bootstrap now supports horizontal collapse animations with the `.collapse-horizontal` class https://getbootstrap.com/docs/5.1/components/collapse/#horizontal

I'm guessing the Collapse component will need to be updated to use this when the `dimension` prop === `width`? (`<Collapse dimension="width"></Collapse>` )

kyletsang:
Sounds reasonable.  Currently, when computedDimension, it adds a `.width` class though, but I don't see that in the bootstrap docs.  Maybe we can replace that class?  

cc @jquense 

We'd be happy to take a PR to add this feature.

jquense:
yeah that was the placeholder from v4. we should update it to whatever is actually supported

nsi319:
Hi @kyletsang 
Can I work on adding horizontal collapse to Collapse component?

kyletsang:
@nsi319 feel free to take on any of the issues here.

nsi319:
@kyletsang Should there be a new Component which does horizontal animation (similar to `Fade`) or should it come in Collapse when `dimension` prop === width?

nsi319:
Hi @kyletsang. When the `dimension` prop === `height`, the existing Collapse component is sufficient. When the value is `width`, we need to make the updates. Could you guide me on how to introduce horizontal animations to it?

kyletsang:
@nsi319 

We can replace the `.width` class that's added here with `.collapse-horizontal`:
https://github.com/react-bootstrap/react-bootstrap/blob/74894de1b50e005280d0270719a4a36e2bf4026c/src/Collapse.tsx#L248

ie.

```
computedDimension === 'width' && 'collapse-horizontal',
```

PR #5992, opened 2021-08-23
Fix Collapse - Horionztal animation #5977

Daim-Nickel-Penny:
Fixed the Collapse - Horionztal animation #5977


kyletsang:
Thanks for the PR.  Could you add a test for this along with an example for the docs?  Thanks.

PR #6002, opened 2021-08-27
fix(ListGroupItem): fix .active not being set with activeKey

### LINKED ISSUE #6001: ListGroup.Item no longer getting the "active" class when active in 2.0.0-beta.5
## Describe the bug

Between versions 2.0.0-beta.4 and 2.0.0-beta.5, the behavior of defaultActiveId seems broken.  When used correctly, the attribute should render an "active" class on the active ListGroup.Item.  This no longer appears to be the case.

## To Reproduce

In version versions 2.0.0-beta.4:
```
<ListGroup defaultActiveKey="#link1">
  <ListGroup.Item action href="#link1">
    Link 1
  </ListGroup.Item>
  <ListGroup.Item action href="#link2" disabled>
    Link 2
  </ListGroup.Item>
  <ListGroup.Item action>This one is a button</ListGroup.Item>
</ListGroup>
```

will render the following html:
image

Now switch to version 2.0.0-beta.5 (or 2.0.0-beta.6):

The same code renders the following html:
image

The "active" class is missing from the rendered html and as a result the styles are not correct.  

## Reproducible Example
versions 2.0.0-beta.4
https://codesandbox.io/s/shy-wildflower-x84k4?file=/src/App.js

Version 2.0.0-beta.5 (or 2.0.0-beta.6)
https://codesandbox.io/s/young-tree-8rxh0?file=/src/App.js

kyletsang:
Fixes #6001

### DIFF
--- a/src/ListGroupItem.tsx
+++ b/src/ListGroupItem.tsx
@@ -1,10 +1,12 @@
 import classNames from 'classnames';
 import * as React from 'react';
-import { useCallback } from 'react';
 import PropTypes from 'prop-types';
-import BaseNavItem, {
+import useEventCallback from '@restart/hooks/useEventCallback';
+import {
+  useNavItem,
   NavItemProps as BaseNavItemProps,
 } from '@restart/ui/NavItem';
+import { makeEventKey } from '@restart/ui/SelectableContext';
 import { useBootstrapPrefix } from './ThemeProvider';
 import { BsPrefixProps, BsPrefixRefForwardingComponent } from './helpers';
 import { Variant } from './types';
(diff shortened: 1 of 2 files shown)


kyletsang:
Hey @ishields, sorry for the delay.  I think we should be good to merge this in.  I just rebased and will re-test this later in the day.

Will include this as part of RC release later today.

ishields:
Awesome, thank you!  👍

jquense (approved):
SORRY FOR THE DELAY!!

kyletsang:
Thanks for the review @jquense 😄

Issue #6017, opened 2021-09-04
Tooltip gives me console errors


multiwebinc:
I'm using react-bootstrap 1.6.1 with the Tooltip component. It is functional, however, I get the following error in the console every time the OverlayTrigger is moused over:

> Popper: modifier "popoverArrowMargins" provided an invalid "fn" property, expected "function" but got "undefined" 
    at https://lacolhost.com:3000/static/js/vendors~main.chunk.js:136717:20
    at Overlay (https://lacolhost.com:3000/static/js/vendors~main.chunk.js:131862:22)
    at OverlayTrigger (https://lacolhost.com:3000/static/js/vendors~main.chunk.js:132001:22)
    at PromoDetailsModal (https://lacolhost.com:3000/static/js/main.chunk.js:10961:3)
    at div
    ...

The following code produces the error:

```jsx
import { OverlayTrigger, Tooltip } from 'react-bootstrap';

const PromoDetailsModal = () => {
  const renderTooltip = props => (
    <Tooltip id="button-tooltip" {...props}>
      Price details
    </Tooltip>
  );

  return (
    <OverlayTrigger overlay={renderTooltip}>
      <h1>Hover for price details</h1>
    </OverlayTrigger>
  );
};

export default PromoDetailsModal;
```

Here is my package.json dependencies in case that is relevant:

```json
{
  "dependencies": {
    "@babel/runtime": "^7.12.5",
    "@fortawesome/fontawesome-svg-core": "^1.2.34",
    "@fortawesome/free-brands-svg-icons": "^5.15.2",
    "@fortawesome/free-regular-svg-icons": "^5.15.2",
    "@fortawesome/free-solid-svg-icons": "^5.15.2",
    "@fortawesome/react-fontawesome": "^0.1.14",
    "@paypal/react-paypal-js": "^7.1.0",
    "@reduxjs/toolkit": "^1.5.0",
    "@styled-icons/crypto": "^10.25.0",
    "@testing-library/jest-dom": "^5.11.9",
    "@testing-library/react": "^12.0.0",
    "@testing-library/user-event": "^13.0.6",
    "@vgs/collect-js": "^0.4.0",
    "bootstrap": "^5.0.2",
    "currency-format": "^1.0.13",
    "formik": "^2.2.6",
    "history": "^5.0.0",
    "i18next": "^20.1.0",
    "i18next-browser-languagedetector": "^6.0.1",
    "i18next-http-backend": "^1.0.21",
[...]

PR #6021, opened 2021-09-06
fix: avoid @popperjs/core console warning

val1984:
`fn` is mandatory for custom modifiers but it may return `undefined` as stated in [popper.js docs](https://popper.js.org/docs/v2/modifiers/#example-modifier).
fixes "Popper: modifier "popoverArrowMargins" provided an invalid "fn" property, expected "function" but got "undefined"" #6017


jquense (approved):

kyletsang (approved):

kyletsang:
Thanks!

PR #6068, opened 2021-10-01
fix(Toast): add optional event arg to Toast onClose

### LINKED ISSUE #5726: Toast.onClose: Type omits argument
Using a `Toast` like this...

```tsx
<Toast onClose={((e: React.MouseEvent<HTMLElement>) => {
    e.stopPropagation();
    doSomething();
}) as () => void} onClose={doSomethingElse}>
```

works absolutely fine, i.e. clicking the “×” triggers `onClose`, but not `onClick`, while clicking somewhere else on the toast only triggers `onClick`. For this behavior, the call `e.stopPropagation()` is needed.
However, this use requires the type assertion `as () => void` in there to type correctly, as the type of `onClose` omits the argument: `ToastProps.onClose?: (() => void) | undefined`
Unless I'm doing something wrong here, it looks like the type of `onClose` should be amended.

I'm using version 1.5.2 of `react-bootstrap`.

(I've skipped the remainder of the bug template as this is not a behavioral bug.)

BerniWittmann:
correct the type signator for the Toast component onClose property

fixes #5726

### DIFF
--- a/src/Toast.tsx
+++ b/src/Toast.tsx
@@ -19,7 +19,7 @@ export interface ToastProps
   animation?: boolean;
   autohide?: boolean;
   delay?: number;
-  onClose?: () => void;
+  onClose?: (e?: React.MouseEvent | React.KeyboardEvent) => void;
   show?: boolean;
   transition?: TransitionComponent;
   bg?: Variant;


kyletsang (approved):

kyletsang:
Thanks!

PR #6078, opened 2021-10-06
Code clean up - Remove the noop function

lcheunglci:
While I was running the tests using `yarn run tdd`, I saw that there's a `noop` function test missing in `ToggleButton.tsx` in the test coverage and while trying to write a test for it, I realized it cannot be called and this function is not needed, so I removed it.


lcheunglci:
Looks like the CI does run different tests that uses the noop function, so this removal is not required.

PR #6090, opened 2021-10-12
ci(workflow): add cache to workflows using actions/setup-node

oscard0m:
## Description

Add `cache` to workflows using `actions/setup-node`

## Context

`setup-node` GitHub Action just released a new option to add cache to steps using it.

You can find the details here: https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/

---

🤖 This PR has been generated automatically by [this octoherd script](https://github.com/oscard0m/octoherd-script-add-cache-to-node-github-action), feel free to run it in your GitHub user/org repositories! 💪🏾

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,16 +35,6 @@ jobs:
       - name: Install Firefox (macOS)
         if: matrix.os == 'macOS-latest' && matrix.browser == 'FirefoxHeadless'
         run: brew install --cask firefox
-      - name: Cache dependencies
-        id: cache-dependencies
-        uses: actions/cache@v2
-        with:
-          path: |
-            .cache/yarn
-            node_modules
-            www/node_modules
-          key: react-bootstrap-${{ hashFiles('yarn.lock') }}
-          restore-keys: react-bootstrap
       - name: Install Dependencies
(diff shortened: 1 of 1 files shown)


kyletsang:
We already have caching in place.  Is there an improvement/something different this provides?

oscard0m:
Applied a couple of fixes from initial PR:
- Using `yarn` cache since you use yarn
- Removed unnecessary step of `actions/cache` since `setup/node -> cache` is equivalent

> We already have caching in place. Is there an improvement/something different this provides?

- In terms of performance it should be equivalent since `setup/node`-> `cache` option relies on `actions/cache`:  [You can read more about it here](https://github.com/actions/setup-node/issues/271)

The goals for this option are the following ones:
- ~**Decrease the entry threshold for using the cache for Node.js dependencies and simplify initial configuration**~ -> It does not apply in your case because you already did the "effort" of integrating `actions/cache`
- **Simplify YAML pipelines because no need additional steps to enable caching** -> ✅ 
- ~**More users will use cache for Node.js so more customers will have fast builds!**~ -> I would say this is a goal from `actions/setup-node` team, so not relevant for `react-bootstrap`

PR #6091, opened 2021-10-12
feat(Accordion): add `alwaysOpen` prop

### LINKED ISSUE #6063: Option for accordion items to stay open

I think it would be handy to have a prop that turns off the accordion's auto-close behavior so that multiple items can stay open at the same time.

kyletsang:
Ref: https://getbootstrap.com/docs/5.1/components/accordion/#always-open

Closes #6063


kyletsang:
Hey @jquense, whenever you're avail, could you review this?  Looking to use this in a project.

jquense (approved):

PR #6129, opened 2021-11-18
Update ButtonSpec.js

marcelolima-hotmart:


kyletsang:
Thanks for the PR.  You just need to add `disabled`, not `disabled="true"`:  https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

PR #6146, opened 2021-12-09
Update colSpan in Table documentation

tayabsoomro:
This PR updates the `colSpan` attribute in [Table](https://react-bootstrap.netlify.app/components/table/#tables) in the React-Bootstrap documentation. This attribute must be an integer and not a string (which it currently is). Having it as a string throws an error with ReactJS with Typescript enabled.


kyletsang (approved):

kyletsang:
Thanks!

Issue #6153, opened 2021-12-15
Race Condition: modal disappears but is not removed from the DOM


nigelgutzmann:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

I have a fairly heavy-weight frontend application, and some of our users reported that after they try to close a modal, the whole page becomes unresponsive.

Digging in a little, I realized that there is a race condition in the way that modals with animations are closed. The process is:

1. user clicks close
2. react-transition-group's `Transition` enters "exiting" status, and the `Fade` component removes the `show` class from the DOM: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Fade.tsx#L130
3. the listener is set up to listen for the end of the transition in the DOM: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/transitionEndListener.ts
4. the `transitionend` listener fires, triggering react-transition-group's `Transition` to enter `exited` status
5. `react-overlay`'s base `Modal` sets it's internal `exited` state to true: https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.tsx#L301
6. The modal is removed from the DOM: https://github.com/react-bootstrap/react-overlays/blob/master/src/Modal.tsx#L307

Our users were getting into a state where the first part of this process was working as expected, but steps 4-6 were occasionally not completed. See the screenshot, where the modal is still in the DOM but the "show" class is missing.

<img width="1792" alt="Screen Shot 2021-12-15 at 2 40 33 PM" src="https://user-images.githubusercontent.com/5893167/146269631-d35a7ebc-2b18-4246-b65e-ba88954a9f92.png">

The problem with the modal remaining in the DOM is that it still covers the whole screen, but has the following css in bootstrap:

```
.fade:not(.show) {
    opacity: 0;
}
```

So to the end user, it just seems like they can't interact with the page at all (all clicks and interactions hit the invisible modal rather than what the user sees).

Note that this would only happen SOMETIMES, because it is a race condition.

[...]

PR #6154, opened 2021-12-15
fix(Modal): option to fix race condition when closing modals

nigelgutzmann:
Fixes https://github.com/react-bootstrap/react-bootstrap/issues/6153

I opted to keep the current behaviour for backwards compatibility, but this adds the option to provide a timeout as a fallback for consumers of react-bootstrap.

### DIFF
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -232,11 +232,11 @@ const defaultProps = {
 
 /* eslint-disable no-use-before-define, react/no-multi-comp */
 function DialogTransition(props) {
-  return <Fade {...props} timeout={null} />;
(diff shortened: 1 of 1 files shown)


src/Modal.tsx
```diff
@@ -515,6 +515,9 @@ export default Object.assign(Modal, {
   Title: ModalTitle,
   Footer: ModalFooter,
   Dialog: ModalDialog,
-  TRANSITION_DURATION: 300,
-  BACKDROP_TRANSITION_DURATION: 150,
```
nigelgutzmann, line 519: these properties are unused and should be removed for clarity-sake

nigelgutzmann:
This isn't the fix that I thought it was, will put more information in the associated issue

PR #6175, opened 2022-01-06
fix: prioritize poppers position prop in overlaytrigger

### LINKED ISSUE #6173: Arrow placement in overlay not working with `flip` modifier
### Prerequisites

### Describe the bug

Classname does not change when overlay position changes if flip is enabled. And also on older versions of bootstrap, react-bootstrap everything works perfectly. I hope you fix this, thanks!

### Expected behavior

It was expected that when I passed prop flip = {true} and position = {'top'} my arrow in overlay would change its position depending on how much I scrolled.

### To Reproduce

_No response_

### Reproducible Example

This example reflects the error, as well as the solution that I use on my project.
https://codesandbox.io/s/festive-clarke-s9pud?file=/src/App.js

### Screenshots

_No response_

Windows

Chrome

2.1.0

5.1.3

### Additional context

_No response_

golota60:
Fixes #6173. What was happening is that `placement` prop was being passed to `BaseOverlay` and then passed  to the `overlay` prop, ignoring popper object, which in turn resulted in hard-coded the `placement` property

### DIFF
--- a/src/Overlay.tsx
+++ b/src/Overlay.tsx
@@ -181,7 +181,6 @@ const Overlay = React.forwardRef<HTMLElement, OverlayProps>(
       >
         {(overlayProps, { arrowProps, popper: popperObj, show }) => {
           wrapRefs(overlayProps, arrowProps);
-          // Need to get placement from popper object first, handling case when overlay is flipped using 'flip' prop
           const updatedPlacement = popperObj?.placement;
           const popper = Object.assign(popperRef.current, {
             state: popperObj?.state,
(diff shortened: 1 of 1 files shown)


src/Overlay.tsx
```diff
@@ -181,18 +181,20 @@ const Overlay = React.forwardRef<HTMLElement, OverlayProps>(
       >
         {(overlayProps, { arrowProps, placement, popper: popperObj, show }) => {
           wrapRefs(overlayProps, arrowProps);
+          // Need to get placement from popper object first, handling case when overlay is flipped using 'flip' prop
+          const updatedPlacement = popperObj?.placement || placement;
```
kyletsang, line 185: Hmm I wonder if we need to fallback to `placement` here.

I think(?) popper's placement should always be defined.

I think this might be enough `const updatedPlacement = popperObj?.placement;`

golota60: i think you're right, will change

kyletsang (approved):

Issue #6156, opened 2021-12-17
Switch tests to @testing-library


kyletsang:
To future proof our tests for new React releases, we should probably switch away from enzyme.  Need to convert all tests over to RTL and typescript.

Help would be appreciated here.

If you take on one of the test files, please leave a comment here so there's no overlap.

- [x] AccordionButton - @VibhorJaiswal 
- [x] Accordion
- [x] Alert - @Sanchitverma78
- [x] Badge
- [x] BootstrapModalManager
- [x] BreadcrumItem
- [x] Breadcrumb
- [x] ButtonGroup
- [x] Button
- [x] ButtonToolbar
- [x] CardImg
- [x] Card - @jayeclark
- [x] CarouselCaption
- [x] Carousel
- [x] CloseButton
- [x] Collapse
- [x] Col
- [x] Container
- [x] createChainedFunction
- [x] createUtilityClasses
- [x] DropdownButton
- [x] DropdownItem
- [x] DropdownMenu
- [x] Dropdown
- [x] DropdownToggle
- [x] Fade
- [x] Feedback
- [x] FigureCaption
- [x] Figure
- [x] FloatingLabel
- [x] FormCheck
- [x] FormControlFeedback
- [x] FormControl
- [x] FormGroup
- [x] FormLabel
- [x] FormRange
- [x] FormSelect
- [x] Form
- [x] FormText
- [x] helpers
- [x] Image
- [x] InputGroup
- [x] ListGroupItem
- [x] ListGroup
- [x] ModalBody
- [x] ModalFooter
- [x] ModalHeader
- [x] Modal
- [x] ModalTitle
- [x] NavbarBrand
- [x] NavbarOffcanvas 
- [x] Navbar
- [x] NavbarToggle
- [x] NavDropdown
- [x] NavItem
- [x] NavLink
- [x] Nav
- [x] Offcanvas
- [x] Overlay
- [x] OverlayTrigger
- [x] PageItem
- [x] Pagination
- [x] PlaceholderButton
- [x] Placeholder
- [x] Popover
- [x] ProgressBar
- [x] Ratio
- [x] Row
- [x] Spinner
- [x] SplitButton
- [x] Stack
- [x] TabContent
- [x] Table
- [x] Tabs
- [x] ThemeProvider
- [x] ToastBody
- [x] ToastContainer
- [x] ToastHeader
- [x] Toast
- [x] ToggleButtonGroup
- [x] Tooltip
[...]

PR #6178, opened 2022-01-08
test: Move Tabs and ThemeProvider specs to rtl

golota60:
Part of #6156 

Futher changes, but there's one big caveat here - I've had to remove one test, cause it seems like `chai`'s `expect` is unable to catch a error thrown by `<Tab />` when it's rendered by itself. I've tried wrapping it in error boundaries and trying to workaround this issue in different ways, but was unable to port the test.

To be sure that this is the problem with `chai` and not my approach I've implemented a simple `jest` setup under [a new branch](https://github.com/golota60/react-bootstrap/tree/jest-test), which shows that there are no issues when using `jest`. 

If you want to see it simply checkout the branch, install deps and run `yarn test:jest` which runs the said test.

There's always a possibity that I'm doing something wrong, feel free to point it out or have a stab at porting the test yourself

### DIFF
--- a/test/TabsSpec.tsx
+++ b/test/TabsSpec.tsx
@@ -1,5 +1,4 @@
-import { render } from '@testing-library/react';
-import { shallow } from 'enzyme';
+import { fireEvent, render } from '@testing-library/react';
 import sinon from 'sinon';
 import React from 'react';
 
@@ -97,7 +96,7 @@ describe('<Tabs>', () => {
       </Tabs>,
     );
 
-    getByText('Tab 2').click();
+    fireEvent.click(getByText('Tab 2'));
     onSelectSpy.should.have.been.called;
   });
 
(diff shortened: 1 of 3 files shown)


test/TabsSpec.tsx
```diff
@@ -0,0 +1,193 @@
+import { render } from '@testing-library/react';
+import { shallow } from 'enzyme';
```
kyletsang, line 2:
```suggestion
```

kyletsang (approved):
Hmm, I guess it's ok to leave that test out.  We can revisit later

test/TabsSpec.tsx
```diff
@@ -0,0 +1,193 @@
...
+        <Tab title="Tab 1" eventKey="1">
+          Tab 1 content
+        </Tab>
+        <Tab title="Tab 2" eventKey="2">
+          Tab 2 content
+        </Tab>
+      </Tabs>,
+    );
+
+    getByText('Tab 2').click();
```
kyletsang, line 100: Can we use `fireEvent` here?

PR #6183, opened 2022-01-10
test: port OverlayTriggerSpec.tsx to rtl

golota60:
Part of #6156 

the biggest spec yet!

since it's so big, i think it's best to have it in a separate pr


test/OverlayTriggerSpec.tsx
```diff
@@ -0,0 +1,507 @@
...
+
+    // aria-describedby gets assigned after a slight delay
+    setTimeout(() => {
+      buttonElem.getAttribute('aria-describedby')!.should.equal('test-tooltip');
+      done();
+    });
+  });
+
+  describe('trigger handlers', () => {
+    let mountPoint;
```
kyletsang, line 223: Do we need this?  Doesn't seem to be used

kyletsang (approved):

PR #6187, opened 2022-01-12
test: migrate Nav, NavItem, NavLink, NavDropdown to rtl

golota60:
Part of #6156


test/NavDropdownSpec.tsx
```diff
@@ -7,34 +7,46 @@ import NavDropdown from '../src/NavDropdown';
...
       </NavDropdown>,
     );
+    const navDropdownElem = getByTestId('test');
+    navDropdownElem.classList.contains('dropdown').should.be.true;
+    navDropdownElem.classList.contains('test-class').should.be.true;
 
-    wrapper.assertSingle('div.dropdown.test-class');
-
-    wrapper.assertSingle('a.nav-link').text().should.equal('Title');
+    navDropdownElem.firstElementChild!.classList.contains('nav-link');
```
kyletsang, line 26: Need an assertion here `should.be.true`

test/NavDropdownSpec.tsx
```diff
@@ -44,56 +56,48 @@ describe('<NavDropdown>', () => {
       </Nav>,
     );
 
-    wrapper
-      .assertSingle('a.active')
-      .text()
-      .should.equal('DropdownItem 2 content');
+    document
```
kyletsang, line 59: I think we can put a `data-testid` on DropdownItem 2, then check for the active class and text content instead of using querySelector

test/NavSpec.tsx
```diff
@@ -0,0 +1,173 @@
...
+    const onSelectSpy = sinon.spy();
+
+    render(
+      <Nav onSelect={onSelectSpy}>
+        <NavDropdown title="Dropdown" id="nav-dropdown-test" renderMenuOnMount>
+          <NavDropdown.Item eventKey={1}>Dropdown item</NavDropdown.Item>
+        </NavDropdown>
+      </Nav>,
+    );
+    const dropdownItem = document.querySelector('a.dropdown-item');
```
kyletsang, line 132: We can put a `data-testid` on the dropdown item and use `getByTestId`

kyletsang (approved):

PR #6195, opened 2022-01-17
docs: add Carbon ads

kyletsang:
Adds the ad into a couple places:
- Main page under the Get Started buttons
- Bottom of the sidenav where the components are listed

Preview:
https://deploy-preview-6195--react-bootstrap.netlify.app/

### DIFF
--- a/www/src/components/CarbonAds.js
+++ b/www/src/components/CarbonAds.js
@@ -1,4 +1,4 @@
-import { useEffect, useRef } from 'react';
+import React, { useEffect, useRef } from 'react';
 
 const CarbonAds = () => {
   const ref = useRef();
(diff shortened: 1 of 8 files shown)


www/src/components/CarbonAds.js
```diff
@@ -0,0 +1,22 @@
...
+  useEffect(() => {
+    if (typeof document !== 'undefined') {
+      const script = document.createElement('script');
+      script.async = true;
+      script.id = '_carbonads_js';
+      script.type = 'text/javascript';
+      script.src =
+        '//cdn.carbonads.com/carbon.js?serve=CE7IP2QY&placement=react-bootstrapgithubio';
+
+      ref.current.appendChild(script);
```
jquense, line 15: Does this need any clean up logic?

jquense (approved):

www/src/components/CarbonAds.js
kyletsang: Don't think we need any.  I think the node unmounting should be enough

kyletsang: Added one just in case 😄

PR #6197, opened 2022-01-20
fix(TabPane): add role="tabpanel"

marcalexiei:
Adds role attribute on `TabPane`.

I added a test inside `<Tabs>` suite since it is quite simple.

Let me know if I need to change / integrate something.

Thanks


kyletsang:
Thanks for the PR.  I believe this change should be done inside `@restart/ui`'s `useTabPanel` hook.  See https://github.com/react-restart/ui/blob/main/src/TabPanel.tsx

marcalexiei:
Considering that Pr on `@restart/ui` is merged I consider to use this PR to:

1. Change version of `@restart/ui` dependencies to `^1.0.x` after new release
2. I'll add the test anyway, better to have it than to not.

marcalexiei:
I changed this PR to add test support for role="tabpanel" on `TabPane` component

kyletsang (approved):

kyletsang:
Thanks

PR #6202, opened 2022-01-23
test(Breadcrumb): migrate to rtl

marcalexiei:
Part of #6156

### ~~Additional change~~
- ~~add `esModuleInterop` option  on test’s tsconfig to allow type checking on sinon~~


test/BreadcrumbItemSpec.tsx
```diff
@@ -0,0 +1,165 @@
...
+    container.querySelectorAll('a').length.should.equal(0);
+    container.querySelectorAll('button').length.should.equal(1);
+  });
+
+  it('Should be able to pass props to the customized inner link element', () => {
+    const { getByRole } = render(
+      <Breadcrumb.Item linkAs={Button} linkProps={{ type: 'submit' }} />,
+    );
+
+    (getByRole('button').getAttribute('type') as string).should.equal('submit');
```
kyletsang, line 147: Is this cast needed?  `getAttribute` should already be a string?

test/BreadcrumbItemSpec.tsx
```diff
@@ -0,0 +1,165 @@
...
+  });
+
+  it('Should be able to pass attributes to the link element', () => {
+    const { getByRole } = render(
+      <Breadcrumb.Item linkProps={{ foo: 'bar' }}>Crumb</Breadcrumb.Item>,
+    );
+
+    (getByRole('button').getAttribute('foo') as string).should.equal('bar');
```
kyletsang, line 155: Same as above

test/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     "jsx": "react-jsx",
     "types": ["mocha", "chai", "sinon", "sinon-chai"],
     "rootDir": "..",
+    "esModuleInterop": true
```
kyletsang, line 7: This isn't needed because the base config already has this set.

marcalexiei: I was getting an error on VSCode that suggest me to add this option in the config. 
I tried today without this option and everything is working 😑

test/BreadcrumbItemSpec.tsx
marcalexiei: The signature of `getAttribute` returns a `string | null` so I added a casting to avoid `Object is possibly 'null'.ts(2531)` error.

kyletsang: Another thing you could do is use a non-null assertion:

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

`expect` works just as well 😄

kyletsang (approved):

kyletsang:
Thanks!

PR #6206, opened 2022-01-25
test(Carousel): migrate CarouselSpec and CarouselCaptionSpec to rtl

marcalexiei:
Part of #6156


test/CarouselSpec.tsx
```diff
@@ -0,0 +1,998 @@
...
+
+    function generateTouchEvents(params: {
+      target: HTMLElement;
+      startX: number;
+      endX: number;
+    }): void {
+      const { target, startX, endX } = params;
+
+      fireEvent.touchStart(target, {
+        touches: [new Touch({ identifier: 1, target, clientX: startX })],
```
marcalexiei, line 763: In order to have touch event working on Chrome during local build I have added `new Touch(...)` when firing touch events.

Without `new Touch(...)` I get the following error:
```
Failed to construct 'TouchEvent': Failed to read the 'touches' property from 'TouchEventInit': Failed to convert value to 'Touch'.
```

However this seems to be not supported on headless browser. Any idea on how to fix this?

test/CarouselSpec.tsx
```diff
@@ -0,0 +1,1025 @@
...
+
+      // fireEvent.touchStart(target, {
+      //   touches: [new Touch({ identifier: 1, target, clientX: startX })],
+      // });
+      // fireEvent.touchMove(target, {
+      //   touches: [new Touch({ identifier: 1, target, clientX: endX })],
+      // });
+      // fireEvent.touchEnd(target);
+
+      Simulate.touchStart(target, {
```
marcalexiei, line 778: On FireFox there are errors due to missing Touch class. (See [build log for details](https://github.com/react-bootstrap/react-bootstrap/runs/4940744877?check_suite_focus=true)).

In order to fix the issue I dig up how enzyme works and I see that they use `Simulate` from `react-dom/test-utils`.
Using this utility seems to resolve the issue ([Build log for details](https://github.com/react-bootstrap/react-bootstrap/runs/4943579729?check_suite_focus=true), Chrome builds stopped before running test so I think that everything should be fine).

Let me know if this solution is feasible for you.

Thanks!

marcalexiei: Re-execute of the action after force push the same code produces no error. 
Ready for review / merge.

kyletsang (approved):
Thanks for taking this on.  This is one of the more difficult ones to handle.  I think it's a good first pass since a majority of the tests have been converted.  We can investigate using RTL instead of react-dom utils for touch events later.

PR #6210, opened 2022-01-27
docs: fix Offcanvas typo

malmgrenola:
Noticed that Offcanvas examples was not totally aligned with Api's.


kyletsang:
Thanks for the PR, but the syntax in the example is actually correct.  Offcanvas exposes its subcomponents for convenience here:

https://github.com/react-bootstrap/react-bootstrap/blob/35463a133388b826000190c60d9ca2760b32e8be/src/Offcanvas.tsx#L324-L328

You're free to use whichever method you prefer when creating your UI.

malmgrenola:
So it does! 
I must have messed up my named imports in the project Im working on.
Thank you so much for reviewing.

I'm closing the PR

PR #6212, opened 2022-01-28
fix: do not use `defaultPrefix` when `bsPrefix` prop is an empty string

rtabulov:


rtabulov:
when passing a `bsPrefix` props as empty string at the moment `defaultPrefix` is used
my current workaround for this is `bsprefix=" "` (a whitespace string)

i think this prop should behave like in prop-types package - eg. empty string should be considered a valid prop, thus not triggering default prop

kyletsang:
What's the use case for setting a component's bsPrefix to an empty string?

PR #6225, opened 2022-02-01
test(Accordion): migrate to rtl

marcalexiei:
Part of #6156

### DIFF
--- a/test/AccordionSpec.tsx
+++ b/test/AccordionSpec.tsx
@@ -1,5 +1,6 @@
-import { mount } from 'enzyme';
-import { fireEvent, render } from '@testing-library/react';
+import { fireEvent, render, waitFor } from '@testing-library/react';
+import sinon from 'sinon';
+import { expect } from 'chai';
 
 import Accordion from '../src/Accordion';
 import AccordionCollapse from '../src/AccordionCollapse';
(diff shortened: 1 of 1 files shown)


test/AccordionSpec.tsx
```diff
@@ -9,183 +10,181 @@ import Nav from '../src/Nav';
...
           <Accordion.Body>body text</Accordion.Body>
         </Accordion.Item>
       </Accordion>,
     );
-    const collapses = wrapper.find('AccordionCollapse');
 
-    collapses.at(0).getDOMNode().className.should.include('show');
-    collapses.at(1).getDOMNode().className.should.include('collapse');
+    expect(getByTestId('item-0').querySelector('.show')).to.exist;
+    expect(getByTestId('item-1').querySelector('.collapse')).to.exist;
```
golota60, line 54: It would look a little bit cleaner if we stick to convention and check `classList`. It would also better mimic the behavior of previous test. Something similar to this
```
    getByTestId('item-0').classList.contains('show').should.be.true;
    getByTestId('item-1').classList.contains('collapse').should.be.true;
```
might differ a bit from this of course

test/AccordionSpec.tsx
```diff
@@ -9,183 +10,181 @@ import Nav from '../src/Nav';
...
 
-    // Enzyme doesn't really provide support for async utilities
-    // on components, but in an ideal setup we should be testing for
-    // this className to be `show` after the collapsing animation is done
-    // (which is possible in `@testing-library` via `waitForElement`).
-    // https://testing-library.com/docs/dom-testing-library/api-async#waitforelement
-    collapses.at(1).getDOMNode().className.should.include('collapsing');
+    await waitFor(() => expect(item1.querySelector('.show')).to.exist, {
+      container: item1,
+    });
```
golota60, line 86: same as above

test/AccordionSpec.tsx
```diff
@@ -9,183 +10,181 @@ import Nav from '../src/Nav';
...
-    // on components, but in an ideal setup we should be testing for
-    // this className to be `show` after the collapsing animation is done
-    // (which is possible in `@testing-library` via `waitForElement`).
-    // https://testing-library.com/docs/dom-testing-library/api-async#waitforelement
-    collapses.at(0).getDOMNode().className.should.include('collapsing');
+    const item0 = getByTestId('item-0');
+    expect(item0.querySelector('.collapsing')).to.exist;
+    await waitFor(() => expect(item0.querySelector('.show')).to.not.exist, {
+      container: item0,
+    });
```
golota60, line 117: same as above

test/AccordionSpec.tsx
```diff
@@ -9,183 +10,181 @@ import Nav from '../src/Nav';
...
-    wrapper.find('a.dropdown-item').simulate('click');
+    fireEvent.click(getByText('Dropdown Action'));
 
-    wrapper
-      .find('AccordionCollapse')
-      .at(0)
-      .getDOMNode()
-      .className.should.include('show');
+    expect(getByTestId('item-0').querySelector('.accordion-collapse.show')).to
+      .exist;
```
golota60, line 143: can we use `getByTestId` instead of `querySelector`?

test/AccordionSpec.tsx
```diff
@@ -9,183 +10,181 @@ import Nav from '../src/Nav';
...
-    wrapper.find('ListGroupItem').simulate('click');
+    fireEvent.click(getByText('List Group Item 1'));
 
-    wrapper
-      .find('AccordionCollapse')
-      .at(0)
-      .getDOMNode()
-      .className.should.include('show');
+    expect(getByTestId('item-0').querySelector('.accordion-collapse.show')).to
+      .exist;
```
golota60, line 165: can we use `getByTestId` instead of `querySelector`?

test/AccordionSpec.tsx
```diff
@@ -9,183 +10,181 @@ import Nav from '../src/Nav';
...
-    wrapper.find('NavLink').simulate('click');
+    fireEvent.click(getByText('Nav Link Item 0'));
 
-    wrapper
-      .find('AccordionCollapse')
-      .at(0)
-      .getDOMNode()
-      .className.should.include('show');
+    expect(getByTestId('item-0').querySelector('.accordion-collapse.show')).to
+      .exist;
```
golota60, line 187: can we use `getByTestId` instead of `querySelector`?

golota60 (commented):
looks great! commented some minor nitpicks, nothing I feel strongly about, if you had a reason why you implemented it a certain way, please share!

cc @kyletsang if you could double-check my comments if they're okay, I'd be grateful 🙏

test/AccordionSpec.tsx
golota60: love the usage of `waitFor` by the way! ❤️

test/AccordionSpec.tsx
marcalexiei: `show` and `collapse` classes are set on `accordion-collapse` element (`AccordationCollapse` component) which is not the were data-testid is set. So that's why I'm using `querySelector`.

----
Snippet from test:
image

To add more context I can add `.accordion-collapse` class to the selector

golota60: oh, I didn't realize that, all is good then 👍

golota60 (approved):
👍

kyletsang (approved):

kyletsang:
Thanks

PR #6233, opened 2022-02-05
Fix(Tabs): adds transition=true as a default property (resolves #6232)

### LINKED ISSUE #6232: Tabs component has no Fade transition by default
### Prerequisites

### Describe the bug

The `Tabs` component has no `Fade` transition by default as described in the documentation.

### Expected behavior

_No response_

### To Reproduce

_No response_

### Reproducible Example

You can see it on the official website:
https://react-bootstrap.github.io/components/tabs/

### Screenshots

_No response_

macOS

Chrome

2.1.2

5.1

### Additional context

I think the problem might come from this function:

https://github.com/react-bootstrap/react-bootstrap/blob/5141a1e1b7f2d189fd80df40cb2e4f558792e719/src/getTabTransitionComponent.ts#L5-L13

As you can see, `Fade` is used as a fallback only if `transition` is a boolean and not be `false`.
But the `transition` only accept `false` or `elementType`:

https://github.com/react-bootstrap/react-bootstrap/blob/5141a1e1b7f2d189fd80df40cb2e4f558792e719/src/Tabs.tsx#L49-L52

watinha:
Resolves #6232

### DIFF
--- a/src/Tabs.tsx
+++ b/src/Tabs.tsx
@@ -47,7 +47,7 @@ const propTypes = {
    * @default {Fade}
    */
   transition: PropTypes.oneOfType([
-    PropTypes.oneOf([false]),
+    PropTypes.oneOf([true, false]),
     PropTypes.elementType,
   ]),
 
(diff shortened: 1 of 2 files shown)


kyletsang:
Thanks for the PR.  I'm closing this in favor of #6235 because the root cause of the bug was the default TabPane transition was incorrectly set to a noop transition.

PR #6239, opened 2022-02-10
Replace Badge variant with bg

IonicaBizau:
The `bg` prop seems to be the right one to set the badge style/variant as per docs.


kyletsang (approved):

kyletsang:
Thanks

PR #6253, opened 2022-02-22
feat: support custom breakpoints

### LINKED ISSUE #6247: How to use custom breakpoints in components and types?
I have customised the breakpoints available in the Bootstrap, and would like components such as `<Col />` and `<Container />` to accept the new breakpoint values.

The breakpoints can be customised in Bootstrap SASS via the `$grid-breakpoints` SASS variable. For example:

```
$grid-breakpoints: (
  xs: 0,
  sm: 576px,
  md: 768px,
  lg: 992px,
  xl: 1200px,
  xxl: 1400px,
  // -- New breakpoint!
  xxxl: 1600px,
);
```

Result:

CleanShot 2022-02-15 at 10 48 12

kyletsang:
Closes #6247

Loosens the types for any breakpoint-related prop so it takes in any string.  For the components that use a hard coded breakpoint list, I've allowed the user to specify their own list of breakpoints in the ThemeProvider so these components can take advantage of it.

Haven't added docs yet.  Wanted to make sure the approach looks good before doing so.

### DIFF
--- a/src/Col.tsx
+++ b/src/Col.tsx
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import * as React from 'react';
 import PropTypes from 'prop-types';
 
-import { useBootstrapPrefix } from './ThemeProvider';
+import { useBootstrapPrefix, useBootstrapBreakpoints } from './ThemeProvider';
 import { BsPrefixProps, BsPrefixRefForwardingComponent } from './helpers';
 
 type NumberAttr =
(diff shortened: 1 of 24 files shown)


golota60 (approved):
👍 lgtm

src/Col.tsx
```diff
@@ -36,9 +36,9 @@ export interface ColProps
   lg?: ColSpec;
   xl?: ColSpec;
   xxl?: ColSpec;
+  [key: string]: any;
```
lindapaiste, line 39: This is rough -- `[key: string]: any` allows pretty much any prop.  But I don't think there's any good way to avoid it.  These are top-level properties so `[key: string]: ColSpec` would produce an incompatible index signature error.

PR #6261, opened 2022-03-06
Add `onHidden` prop to Toast

spigelli:


kyletsang:
Thanks for opening this PR.  The transition props should follow the same conventions as the other components.  See `Modal` or `TabPane` as an example

Issue #6250, opened 2022-02-17
OverlayTrigger offset property has no effect


avlisesac:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

I am trying to use OverlayTrigger and Popover to create a popover that has a 2px gap between the popover and the overlay element. Bootstrap uses a property called offset to provide this:
![image](https://user-images.githubusercontent.com/18335571/154521780-7ddf0ab3-6359-484d-8e7a-bec6ef8edc9e.png)

The Popover component in react-bootstrap does not have this property, but the OverlayTrigger does. The offset property accepts an array of two values that can be a number, null, or undefined:
![image](https://user-images.githubusercontent.com/18335571/154521994-f5abedbf-54a5-478c-8885-708cc6f2b466.png)
![image](https://user-images.githubusercontent.com/18335571/154522084-e106be9c-3bdc-448e-a14e-72ead05ba7da.png)

I pass in an array with two numbers to test this, but there is no change to the popover.
![image](https://user-images.githubusercontent.com/18335571/154522204-339fde43-985c-473a-bd71-d695da53526d.png)

### Expected behavior

Passing values to the offset property should provide a gap/offset between the target element and the popover.

### Reproducible Example

Here is a codesandbox example of the issue
https://codesandbox.io/s/charming-hofstadter-udoyku?file=/src/App.js

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Chrome

### What version of React-Bootstrap are you using?

1.6.3

### What version of Bootstrap are you using?

4.6.1

golota60:
Good catch! This *seems* to be an upstream issue in `@restart/ui`. As a workaround, try setting the offset this way
```ts
      popperConfig={{
        modifiers: {
          name: "offset",
          options: {
            offset: [200, 200]
          }
        }
      }}
```
Here's a sandbox if you want to see it working: https://codesandbox.io/s/tender-dust-nibeg6?file=/src/App.js:321-472

avlisesac:
[...]

PR #6264, opened 2022-03-09
fix(useOverlayOffset): use offset prop in useOverlayOffset

golota60:
Fixes: https://github.com/react-bootstrap/react-bootstrap/issues/6250

`useOverlayOffset` was overriding user-defined `offset` prop. This passes the value explicitly to the helper hook


kyletsang:
Can we add a test in `useOverlayOffsetSpec`?

golota60:
getting some spooky errors - will fix in a sec

golota60:
fixed, i thought i could arrange a scenario more similar to what's in the docs but i don't think that's possible

kyletsang (approved):

Issue #6263, opened 2022-03-08
DefaultActiveIndex not being applied after change in component state.


laurencefass:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

I rasied an issue about activeIndex prop freezing the carousel and was advised to use defaultActiveIndex which has enabled navigation but for reasons I dont understand I cant set the defaultActiveIndex from a useEffect->setState() only when I initialise the component. 

The Carousel is not responding to the value change on line 31 in the codesandbox detailed below. This may have something to do with controlled/uncontrolled components but I cant find anything in the docs relating to handling  of controlled/uncontrolled Carousels.

Thanks

### Expected behavior

I expected the component to update whenever I change the defaultActiveIndex value prop but it only updates on component initialisation.

### To Reproduce

Run sandbox.

### Reproducible Example

https://codesandbox.io/s/react-bootstrap-carousel-defautlativeindex-f3x54f

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Chrome

### What version of React-Bootstrap are you using?

2.1.0

### What version of Bootstrap are you using?

5.1.3

kyletsang:
`defaultActiveIndex` is only used when you first render the carousel.  

Since you're looking to update this index from a useEffect or useState, this means you want to "control" the component.  You would need to use `activeIndex` instead and implement the carousel callbacks such as `onSelect` to update the `activeIndex`.

See this example on a controlled carousel:
https://react-bootstrap.github.io/components/carousel/#controlled

laurencefass:
The problem I face here is that I dont want to update the index from an internal carousel state change rather I want to instigate the change from an external event. Is that possible?

```
function ControlledCarousel() {
  const [index, setIndex] = useState(0);

  const handleSelect = (selectedIndex, e) => {
    setIndex(selectedIndex);
  };

  return (
    <Carousel activeIndex={index} onSelect={handleSelect}>
     ...
```

[...]

PR #6265, opened 2022-03-09
fix(Carousel): fix issue where changing `activeIndex` won't work

kyletsang:
Issue was brought up in #6263

If you set the slides and then change the `activeIndex`, it will result in the carousel locking up because `isSliding` is set to true and RTG's `onEntered` isn't triggered to reset this.  

Minimal repro:
https://codesandbox.io/s/react-bootstrap-carousel-active-index-forked-xht0f1?file=/src/App.js

Issue appears to be fixed by moving some of the code that handles the rendered active index into a `useEffect` hook.  
Note, I had to tweak the delays of 1 test because the timers seem to be triggering too fast when testing.

### DIFF
--- a/src/Carousel.tsx
+++ b/src/Carousel.tsx
@@ -263,21 +263,23 @@ const Carousel: BsPrefixRefForwardingComponent<'div', CarouselProps> =
       activeIndex || 0,
     );
 
-    if (!isSliding && activeIndex !== renderedActiveIndex) {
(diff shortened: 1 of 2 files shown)


golota60 (approved):
lgtm 👍  does this work when `slide={false}`?

kyletsang:
> lgtm 👍 does this work when `slide={false}`?

Yup slide=false works 👍

Issue #5953, opened 2021-07-29
ARIA role="dialog" must have an accessible label


dualcyclone:
(moved from https://github.com/react-bootstrap/react-overlays/issues/965 & https://github.com/react-bootstrap/react-overlays/issues/964)

## Describe the bug

Related to https://github.com/react-bootstrap/react-overlays/issues/964

Adding any `aria` attributes to the `Modal` adds these to the child DOM element, which means the main parent element still does not have an accessible label.

## To Reproduce

Create a `Modal` and add `aria-label="test"` to it:

```
import Modal from "react-bootstrap/Modal";

...
<Modal
  ...
  aria-label="test"
>
...
```

Observe that the `aria-label` added does not get added to the element with `role="dialog"`, but instead onto its first child.

```
<div role="dialog" aria-modal="true">
    <div aria-label="test">
    ....
    </div>
</div>

```

## Reproducible Example

Minimal example by using [CodeSandbox](https://codesandbox.io/s/old-wave-1xc5v?file=/src/App.js).

## Expected behavior

The main element with `role="dialog"` should have the `aria` attributes added to it.

## Additional context

[MDN - ARIA: dialog role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role)

> However, adding role="dialog" alone is not sufficient to make a dialog accessible. Additionally, the following needs to be done:
>
> The dialog must be properly labeled

devceline:
Put out a PR to hopefully fix this issue [here](https://github.com/react-bootstrap/react-bootstrap/pull/6269)

PR #6269, opened 2022-03-12
cs/move aria label

devceline:
*Issue*: https://github.com/react-bootstrap/react-bootstrap/issues/5953

### Effects
- Aria-label now forwarded to the modal instead of its child

### Verify
- Edit the example modal to have an aria property like `aria-label="foo"`. Previously the prop would be passed to the first child instead of the parent

### Quick note:
This is my first time contributing to a big library like this, please let me know if anything should be changed in PR/code formatting. Thank you.

### DIFF
--- a/src/Modal.tsx
+++ b/src/Modal.tsx
@@ -290,6 +290,16 @@ const Modal: BsPrefixRefForwardingComponent<'div', ModalProps> =
       const handleHide = useEventCallback(onHide);
       const isRTL = useIsRTL();
 
+      const ariaProps = Object.keys(props)
+        .filter((prop) => prop.indexOf('aria') !== -1)
+        .reduce<{ [x: string]: any }>(
+          (ariaPropsObject, ariaProp) => ({
+            ...ariaPropsObject,
+            [ariaProp]: props[ariaProp],
+          }),
+          {},
+        );
+
       bsPrefix = useBootstrapPrefix(bsPrefix, 'modal');
 
       const modalContext = useMemo(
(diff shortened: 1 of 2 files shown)


kyletsang:
I think we should just forward `aria-describedby` and `aria-label` to the dialog.  Don't need to map all aria properties.

devceline:
> I think we should just forward `aria-describedby` and `aria-label` to the dialog. Don't need to map all aria properties.

But isn't it more intuitive that the props you pass to a component go to what it is rendering directly, and also more accessibility friendly? 

```tsx
const Component = (props) => {
  return (
    <div {...props}> {/*It makes sense that props would be forwarded here*/}
      <span {...props}> </span> {/* instead of here */}
    </div>
  )
```

May I ask why single out only 2 aria props?

kyletsang:
According to the accessibility docs on dialogs, you would only really need the two aria properties anyways (aria-label might be debatable because a modal should have a visible title?).

devceline:
> According to the accessibility docs on dialogs, you would only really need the two aria properties anyways (aria-label might be debatable because a modal should have a visible title?).

Oh alright, do you recommend I close this PR or should I just update it to accept `aria-label` optionally?

kyletsang:
Feel free to amend this PR

kyletsang:
Looks good.  Could you add a test please?

devceline:
> Looks good. Could you add a test please?

Alrighty, added one :) Tried to follow the same test for the labelledby

kyletsang (approved):

kyletsang:
Looks good, thanks!

PR #6276, opened 2022-03-17
fix(Tabs): fix transition functions

golota60:
While poking around in `Tab` component, I've noticed that none of the react-transition-group functions work (`onEnter, onExit etc.`). This was because in `useTabPanel` hook we expose them in the first argument, not the second one (https://github.com/react-restart/ui/blob/c6f9e22dd0cfcef412d3d00618aa5ec453f63bb6/src/TabPanel.tsx#L88).

 This PR moves all of these functions to the correct position. Also added a test for it

### DIFF
--- a/src/TabPane.tsx
+++ b/src/TabPane.tsx
@@ -106,16 +106,16 @@ const TabPane: BsPrefixRefForwardingComponent<'div', TabPaneProps> =
           className,
           // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
           as: Component = 'div',
-          ...rest
-        },
-        {
-          isActive,
           onEnter,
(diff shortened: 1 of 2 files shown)


kyletsang:
Nice catch.  We should actually fix this in `@restart/ui`.  I meant to put the transition callback props into the hook's metadata return, not the first.

golota60:
alright! will create a PR in restart/ui shortly

golota60:
done - do we want to keep a test in here too or will the one in `restart/ui` be enough?

kyletsang:
> done - do we want to keep a test in here too or will the one in `restart/ui` be enough?

I think the tests in restart/ui is good enough 👍 

Thanks for making those changes!

golota60:
no problem! glad i can help

Issue #5102, opened 2020-04-07
Modal Animation wont allow input to auto focus


sirajtahrawashmen:
## Describe the bug

If you have any input with autoFocus property inside a Modal, the input wont be auto focused when the Modal is set to be shown, unless you set the Modal property `animation={false}` 

## Reproducible Example

```    
<Modal
      ref={ModalRef}
      show={isSearchModalOpen}
      onHide={handleClose}
      centered
      animation={false}
    >
      <Modal.Header closeButton>
        <Modal.Title>Quick Search</Modal.Title>
      </Modal.Header>
      <Modal.Body>
        <Form>
          <Form.Group controlId="searchValue">
            <Form.Control
              type="text"
              placeholder="Search..."
              ref={inputRef}
              autoFocus
            />
          </Form.Group>
          <Button variant="primary" type="submit">
            <i className="fa fa-search"></i>
          </Button>
        </Form>
      </Modal.Body>
</Modal>
```

## Expected behavior
Autofocus should work on Inputs even if the animation is set to true on the Modal component

## Environment 
- Operating System: macOS
- Browser, Version Chrome 74
- React-Bootstrap Version 1.0.0

## Additional context

you can manually focus the input using the `onEntered` prop, but there shouldn't be a need to

taion:
Right... looks like this broke at some point. Not sure when. It'd be a bug in https://github.com/react-bootstrap/react-overlays/blob/d0cdd9d5204ef0eaadaca8d616fa2ebd5449f4c1/src/Modal.js#L381-L390, which is supposed to handle this.

shaungc-su:
Experiencing this issue as well, I see people suggesting using `useRef` to grab the input element and call `focus()` on it in `onEnter` prop callback, but feel it bit of an overhead doing that just to achieve auto focus.

> This also works correctly with any Modal children that have the `autoFocus` prop

Looking at the comment of Modal component, looks like `autoFocus` prop on Modal should work well with `autoFocus` on children input components.

I'm guessing perhaps after `show` changed and `this.onShow()` called, at the time of Modal children (and inputs) get mounted, the animation transition hasn't finished yet, so even if autoFocus worked on inputs when they get mounted on Modal, the completion of animation somehow override the focus. But I have to say I'm not familiar with how `react-transition-group` interact with `Modal` here, so I'm not sure if it's an ordering issue.

omgoshjosh:
[...]

PR #6278, opened 2022-03-20
Fix modal focus

tkrotoff:
Fix for #5102


src/Modal.tsx
```diff
@@ -445,10 +444,9 @@ const Modal: BsPrefixRefForwardingComponent<'div', ModalProps> =
 
       const baseModalStyle = { ...style, ...modalStyle };
 
-      // Sets `display` always block when `animation` is false
-      if (!animation) {
```
tkrotoff, line 449: Could not spot the difference in term of animation. I don't know why display was not always set to block.

You can slow down the animation with this (3s instead of 150ms):
```
.fade {
    transition: opacity 3s linear;
}
```

www/src/examples/Modal/Focus.js
```diff
@@ -0,0 +1,49 @@
...
+          <Modal.Title>Modal heading</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <Form>
+            <Form.Group className="mb-3" controlId="exampleForm.ControlInput1">
+              <Form.Label>Email address</Form.Label>
+              <Form.Control
+                type="email"
+                placeholder="name@example.com"
+                autoFocus
```
tkrotoff, line 24: `autoFocus` attribute on an input to demo the behavior

tkrotoff:
btw, each time I modify Modal.tsx, the documentation refreshes (good) but it freezes my Chrome tab: I have to open a new tab each time 😱

src/Modal.tsx
kyletsang: Looks like the logic was supposed to follow upstream bootstrap, although theirs appear to set display block after the animation finishes.  I think the issue may be that we're running the `autoFocus` code in `@restart/ui` too early?  Maybe we should be running it during one of the transition callbacks when animation is enabled.

cc @jquense

kyletsang: Looks like the logic was supposed to follow upstream bootstrap, although theirs appear to set display block after the animation finishes.  I think the issue may be that we're running the `autoFocus` code in `@restart/ui` too early?  Maybe we should be running it during one of the transition callbacks when animation is enabled.

cc @jquense

tkrotoff: Here the algorithm they are using:

```TypeScript
// https://github.com/twbs/bootstrap/blob/v5.0.0/js/src/modal.js#L125-L129
scrollBarHide();
document.body.classList.add('modal-open');
///

// https://github.com/twbs/bootstrap/blob/v5.0.0/js/src/modal.js#L144
// https://github.com/twbs/bootstrap/blob/v5.0.0/js/src/util/backdrop.js#L46-L54
reflow(modalBackdrop);
modalBackdrop.classList.add('show');
await emulateAnimation(modalBackdrop);
///

// https://github.com/twbs/bootstrap/blob/v5.0.0/js/src/modal.js#L144
// https://github.com/twbs/bootstrap/blob/v5.0.0/js/src/modal.js#L243-L261
==> modal.style.display = 'block';
reflow(modal);
==> modal.classList.add('show');
enforceFocus();
///
```

I think line `modal.style.display = 'block'` does not change anything because of

```CSS
// https://github.com/twbs/bootstrap/blob/v5.1.3/scss/_transitions.scss#L4-L6
.fade:not(.show) {
    opacity: 0;
}
```

Line `modal.classList.add('show')` cancels `opacity: 0`.


I think they use `display: none` because 
`<div class="modal fade"><div class="modal-content">...</div></div>` is already inside the DOM even if the modal is not displayed.
In React, `<div class="modal fade">...</div>` is added to the DOM only when the modal is displayed with prop `show={true}` via `createPortal()`.

Check https://getbootstrap.com/docs/5.1/components/modal/ and remove `display: none` from `.modal { ...; display: none; ...; }`, you will see that you are unable to click anywhere on the page.

tkrotoff: Reading https://developer.mozilla.org/en-US/docs/Web/CSS/display:

> display: none
>
> Turns off the display of an element so that **it has no effect on layout (the document is rendered as though the element did not exist)**. All descendant elements also have their display turned off. To have an element take up the space that it would normally take, but without actually rendering anything, use the [visibility](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility) property instead.

That's why `display: none` is not compatible with `autoFocus` on an input inside children.

So *the issue may be that we're running the autoFocus code in @restart/ui too early?* in my opinion won't fix this issue.

Check https://bugzilla.mozilla.org/show_bug.cgi?id=569955: *elements with display: none; are not focusable.* (they even have added a test for this :-) [*Add a test to make sure that autofocus on input elements does not work if they don't have a frame*](https://hg.mozilla.org/mozilla-central/rev/db50c47cb0bd))

tkrotoff: Related Bootstrap issues: https://github.com/twbs/bootstrap/issues/12525, https://github.com/twbs/bootstrap/issues/22402

www/src/examples/Modal/Focus.js
kyletsang: Sorry, I originally thought that `autoFocus` in `@restart/ui` handled inputs explicitly, but that's not the case.  

I think this should be ok.  @jquense , @golota60 did you want to take a look?

jquense: seems ok. honestly i don't remember why we needed this style override originally.

golota60: looks alright to me

kyletsang (approved):

kyletsang:
Thanks for making this fix @tkrotoff

PR #6282, opened 2022-03-23
fix(Modal): render aria-describedby on the top level modal element

kyletsang:
`aria-describedby` should be rendered on the element with the `role="dialog"`


golota60 (approved):
👍 lgtm

PR #6290, opened 2022-03-31
no need to specify version in npm install command installation docs, react-bootstrap's latest tag is 5.1.3

tejasc0:
npm package bootstrap has 5 tags
latest
latest-5
latest-4
latest-3
latest
The latest tag points to 5.1.3
so there is no need to specify bootstrap's version in the installation command

Expected command: `npm install react-bootstrap bootstrap`


kyletsang (approved):

kyletsang:
Thanks

PR #6300, opened 2022-04-08
chore(deps): move all the types dependencies to dev dependencies

### LINKED ISSUE #6299: Remove `@types/*` from dependencies
### Prerequisites

### Describe the bug

`@types/*` being added in `dependencies` instead of `devDependencies`.

### Expected behavior

`@types/*` should be added in `devDependencies`

### To Reproduce

_No response_

### Reproducible Example

There are no steps to reproduce it.

### Screenshots

_No response_

_No response_

_No response_

2.2.2

5.1.3

### Additional context

_No response_

gmukul01:
Moving all types dependencies to dev dependencies.

This fixes #6299


kyletsang:
Thanks for opening a PR, but the types actually do belong in the dependencies.  See @jquense's response in #6299

> Types are supposed to be in dependencies because they need to be installed when some one installs the package. Other wise typescript will not be able to correctly typecheck when you use react-bootstap

gmukul01:
@kyletsang  At least we can move it to `peerDependency`. Keeping it in dependency is failing our project because it is installing latest version on react types(v18) which is not compatible as of now in our project. And since it is in `dependency` we can not control the version.

PR #6303, opened 2022-04-11
fix(deps): clean up @types deps list

kyletsang:
Moves `@types/react` into peer deps.  This should be safe for npm 7 users when resolving peer dependencies because of the optional field in `peerDependenciesMeta`.  All users using TS would've installed a version of `@types/react` anyway, so we shouldn't have problems here.

For `invariant`, `prop-types`, and `warning`, those can be safely moved into devDeps because their types aren't exposed on the public d.ts files.  They are just used internally within the implementation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,11 +62,7 @@
     "@babel/runtime": "^7.17.2",
     "@restart/hooks": "^0.4.6",
     "@restart/ui": "^1.2.0",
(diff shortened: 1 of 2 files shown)


golota60 (approved):
lgtm! yeah, _I think_ it shouldn't break anything for anyone

gmukul01:
When are we planning to merge this ?

PR #6305, opened 2022-04-14
fix(TabPane): Omit Transition when prop is false

headdu:
While using tabs I hit a small issue:

When passing the prop `transition={false}` to Tabs/TabPanes the `fade` class was always applied thus there's always a fade animation when we change Tabs.

I was able to reproduce this using the latest version of the lib and made a [codesandbox](https://codesandbox.io/s/quirky-torvalds-21wjbz)

### DIFF
--- a/www/src/examples/Tabs/NoAnimation.js
+++ b/www/src/examples/Tabs/NoAnimation.js
@@ -4,10 +4,10 @@
   id="noanim-tab-example"
   className="mb-3"
 >
-  <Tab eventKey="home" title="Home">
+  <Tab eventKey="home" title="Home" transition={false}>
     <Sonnet />
   </Tab>
-  <Tab eventKey="profile" title="Profile">
+  <Tab eventKey="profile" title="Profile" transition={false}>
     <Sonnet />
   </Tab>
   <Tab eventKey="contact" title="Contact" disabled>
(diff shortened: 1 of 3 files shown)


kyletsang:
Thanks for the PR.  I believe the issue here is that `getTabTransitionComponent` should actually return a different transition component rather than `undefined` because this ends up defaulting to `Fade` in the `TabPane`:
https://github.com/react-bootstrap/react-bootstrap/blob/e32c53d9001e07f5685c701a41f0c0b7d5a39cca/src/TabPane.tsx#L121

The change should be made here:
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/getTabTransitionComponent.ts

Add

```
import NoopTransition from '@restart/ui/NoopTransition';
```

then replace line 9 with

```
return transition ? Fade : NoopTransition;
```

See if this works out

headdu:
Thank you for the pointers.

I did the changes you suggested and that seems to have done it as well 👍

www/src/examples/Tabs/NoAnimation.js
```diff
@@ -4,10 +4,10 @@
   id="noanim-tab-example"
   className="mb-3"
 >
-  <Tab eventKey="home" title="Home">
+  <Tab eventKey="home" title="Home" transition={false}>
```
kyletsang, line 7: The changes in this file can be reverted.  The tab panes will inherit the transition prop from the parent

headdu: Done!

kyletsang (approved):

kyletsang:
Thanks!

PR #6310, opened 2022-04-25
fix(NavbarOffcanvas): fix rendering of nav items when expand is used

### LINKED ISSUE #6309: Navbar Offcanvas not showing nav items with `expand` prop
### Prerequisites

### Describe the bug

The offcanvas navbar component found here https://react-bootstrap.github.io/components/navbar/#offcanvas is very similar to the one found in regular bootstrap - and I've tried to make this one function like the one in regular bootstrap, so once your window is at a certain size (expand="lg"), so 992px, the navbar will then collapse into its offcanvas format, rather than appearing as a normal navbar - but instead the normal navbar doesn't appear, instead disappearing and leaving me with just a gray bar.

### Expected behavior

I expected the navbar to function like so:

https://getbootstrap.com/docs/5.1/examples/offcanvas-navbar/#

### To Reproduce

1. Go to https://react-bootstrap.github.io/components/navbar/#offcanvas
2. Replicate the code
3. Change "expand={false}" to expand="lg"
4. Note that the navbar doesn't function as original bootstrap intended.

### Reproducible Example

https://react-bootstrap.github.io/components/navbar/#offcanvas

### Screenshots

Collapsed

Non Collapsed

Note that until it is collapsed, nothing appears on the navbar besides the brand, which is outside of the offcanvas.

Windows

Chrome

2.3.0

5.1.3

### Additional context

_No response_

kyletsang:
Fixes #6309

When the navbar hits the expand breakpoint, it will render the children into a div instead of using the Offcanvas component to do it.  We don't need the functionality when the nav items are shown this way anyway.  One thing that needs to be done is to add a default breakpoint size for xxl in @restart/hooks.

I think this approach should be ok?  Will add tests later after review.

### DIFF
--- a/src/NavbarOffcanvas.tsx
+++ b/src/NavbarOffcanvas.tsx
@@ -1,13 +1,80 @@
 import * as React from 'react';
 import { useContext } from 'react';
+import useBreakpoint from '@restart/hooks/useBreakpoint';
+import classNames from 'classnames';
+import { useBootstrapPrefix } from './ThemeProvider';
 import Offcanvas, { OffcanvasProps } from './Offcanvas';
 import NavbarContext from './NavbarContext';
 
-const NavbarOffcanvas = React.forwardRef<HTMLDivElement, OffcanvasProps>(
-  (props, ref) => {
-    const context = useContext(NavbarContext);
+export type NavbarOffcanvasProps = Omit<OffcanvasProps, 'show'>;
 
-    return <Offcanvas ref={ref} show={!!context?.expanded} {...props} />;
+const NavbarOffcanvas = React.forwardRef<HTMLDivElement, NavbarOffcanvasProps>(
+  (
+    {
+      className,
+      bsPrefix,
+      backdrop,
+      backdropClassName,
+      keyboard,
+      scroll,
+      placement,
+      autoFocus,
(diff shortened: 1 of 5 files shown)


src/NavbarOffcanvas.tsx
```diff
@@ -1,13 +1,80 @@
...
+        restoreFocusOptions={restoreFocusOptions}
+        onShow={onShow}
+        onHide={onHide}
+        onEscapeKeyDown={onEscapeKeyDown}
+        onEnter={onEnter}
+        onEntering={onEntering}
+        onEntered={onEntered}
+        onExit={onExit}
+        onExiting={onExiting}
+        onExited={onExited}
```
golota60, line 74: why change this from `{...props}`? They're being forwarded anyway

golota60 (approved):
lgtm, left one nitpick/question

src/NavbarOffcanvas.tsx
golota60: i mean _most_ of these

kyletsang: lol yeah it's a bit ugly.  I pulled it out because I also spread the remainder of the props into the div above for the navbar expand case.

PR #6311, opened 2022-04-25
docs: fix examples

kyletsang:
Not sure when these broke, but some of the examples on the site don't work.


kyletsang:
Nice catch @jirihofman, thanks!

PR #6316, opened 2022-05-01
Add containerPosition prop for ToastContainer

### LINKED ISSUE #6315: Do not add `position-absolute` if `className` has `position-*`
### Is your feature request related to a problem? Please describe

ToastContainer hard-codes `position-absolute` utility class, making it impossible to replace it with `position-fixed`:

```jsx
    return (
      <Component
        ref={ref}
        {...props}
        className={classNames(
          bsPrefix,
          position && `position-absolute ${positionClasses[position]}`,
          className,
        )}
      />
    );
```

Code: https://github.com/react-bootstrap/react-bootstrap/blob/master/src/ToastContainer.tsx#L81

https://github.com/react-bootstrap/react-bootstrap/issues/5799 is relevant; I'd like to know if there is a preferred solution for this.

### Describe the solution you'd like

Check the `className` to see if there are any `position-*` classes:

```jsx
      <Component
        ref={ref}
        {...props}
        className={classNames(
          bsPrefix,
          position && `${classNameHasPosition ? '' : 'position-absolute'} ${positionClasses[position]}`,
          className,
        )}
      />
```

This seems to be the most backward-compatible approach. The most difficult part would be to define `classNameHasPosition`, because `className` has a very flexible type (can be strings, array, etc.).

#### Add a new prop that defaults to `absolute` if `position` is provided

```jsx
<ToastContainer position="bottom-center" anchoring="fixed">
```

This change should be backward-compatible and relatively easier to implement than the solution above. However, I am not sure about the new property name 🙃 

#### Extend the possible values for `position`

```jsx
<ToastContainer position="absolute-bottom-center">
<ToastContainer position="fixed-bottom-center">
```

While this change should be backward-compatible, the enumeration will increase the component size. 

#### Do nothing; let users add `position-fixed` class

```jsx
<ToastContainer position="bottom-center" className="position-fixed">
```

This is suggested in https://github.com/react-bootstrap/react-bootstrap/issues/5799.

I tested and it seems to work on Chrome; however, I am not sure if this is correct because [class order in HTML does not seem to be relevant](https://stackoverflow.com/questions/15670631/does-the-order-of-classes-listed-on-an-item-affect-the-css).

#### Remove `position` prop entirely, force users to use the Bootstrap utility classes

```jsx
<ToastContainer className="position-fixed bottom-0 start-50 translate-middle-x">
```

This breaks existing `ToastContainer` usage. However, this will make the component simpler and removes dependency to Bootstrap utility class.

### Additional context

The current workaround is to not use `position` prop and directly uses the relevant Bootstrap utility classes:

```jsx
<ToastContainer className="position-fixed bottom-0 start-50 translate-middle-x">
```

If I have the solution, I think I can submit a PR for this.

tkesgar:
Fixes #6315.

I am unsure whether we should strictly limit the containerPosition values to [position-* utility classes](https://getbootstrap.com/docs/5.1/utilities/position/) or allow any string, because it is possible to customize the utility classes.

### DIFF
--- a/src/ToastContainer.tsx
+++ b/src/ToastContainer.tsx
@@ -19,6 +19,7 @@ export interface ToastContainerProps
   extends BsPrefixProps,
     React.HTMLAttributes<HTMLElement> {
   position?: ToastPosition;
+  containerPosition?: string;
 }
 
 const propTypes = {
@@ -41,6 +42,13 @@ const propTypes = {
     'bottom-center',
     'bottom-end',
   ]),
+
+  /**
+   * By default the container is rendered with `position-absolute` utility class. Provide a string to use other `position-*` utility classes, or `false` to remove it.
+   *
+   * @default 'absolute'
+   */
+  containerPosition: PropTypes.string,
 };
 
 const positionClasses = {
@@ -63,6 +71,7 @@ const ToastContainer: BsPrefixRefForwardingComponent<
     {
       bsPrefix,
       position,
+      containerPosition = 'absolute',
       className,
       // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
       as: Component = 'div',
@@ -78,7 +87,8 @@ const ToastContainer: BsPrefixRefForwardingComponent<
         {...props}
         className={classNames(
           bsPrefix,
-          position && `position-absolute ${positionClasses[position]}`,
+          position &&
+            `position-${containerPosition} ${positionClasses[position]}`,
           className,
         )}
       />
--- a/src/ToastContainer.tsx
+++ b/src/ToastContainer.tsx
@@ -44,7 +44,7 @@ const propTypes = {
   ]),
 
   /**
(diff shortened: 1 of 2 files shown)


src/ToastContainer.tsx
```diff
@@ -41,6 +42,13 @@ const propTypes = {
     'bottom-center',
     'bottom-end',
   ]),
+
+  /**
+   * By default the container is rendered with `position-absolute` utility class. Provide a string to use other `position-*` utility classes, or `false` to remove it.
```
golota60, line 47: >or `false` to remove it.

Doesn't this accept only strings? 

I think providing an empty string would achieve what you want here 😄 (also it could be a good idea to add a test for that)

golota60 (commented):
lgtm, only one small issue/nitpick

src/ToastContainer.tsx
tkesgar: Oh forget to update this 🙈 I initially allow `false` to remove the `position-*` class, but I feel it is simpler to just use `"static"` or whatever utility class. Using an empty string should work as well, I'll update it.

golota60 (approved):
lgtm, thanks! 😄

PR #6317, opened 2022-05-01
Update toasts.mdx

clementbowe14:
Fixed small grammatical errors.


kyletsang:
Can we remove this line?

> But be aware that autohide only triggers when the `onClose` function is called

clementbowe14:
Sure thing!

PR #6344, opened 2022-05-19
feat(Table): added support for striped columns

andydetorres:
Added support for Striped Columns as per v5.2 

https://getbootstrap.com/docs/5.2/content/tables/#striped-columns


kyletsang:
Thanks!

PR #6345, opened 2022-05-19
Remove extra `onChange` check

RyanCavanaugh:
This const variable is already checked on line 95, so must be truthy here as well.

This bug was accidently detected by TypeScript 4.7-RC


kyletsang (approved):

kyletsang:
Thanks!

PR #6353, opened 2022-05-25
feat(FormCheck): add support for reverse checks and radios

andydetorres:
Added support for reverse prop in Checks and Radios as per v5.2

https://getbootstrap.com/docs/5.2/forms/checks-radios/#reverse

image


kyletsang (approved):

kyletsang:
Thanks!

Issue #4952, opened 2020-01-27
Update Documentation to include imports & constants in code examples


joeldom:
## Describe the bug

When attemtping to add features, copying code shown, you then have to see if you missed and element because a list isn't shown of what is imported

## To Reproduce

1. copy code example
2. paste in app
3. run
4. error: `Line ?: 'Component' is not defined react/jsx-no-undef`

## Reproducible Example

 [Navbars](https://react-bootstrap.github.io/components/navbar/).

## Expected behavior

I'd like to be able to make a new feature component to quickly import to test it within the context of the app with the imports for the component example shown, as well as any state/functions needed to quickly add features.\

## Documentation Example
[Atlaskit ~ Dropdown Item](https://atlassian.design/components/dropdown-menu/dropdown-item/examples)

## Environment (please complete the following information)

- Operating System: [macOS]
- Browser, Version [FF Dev]
- React-Bootstrap Version ["react-bootstrap": "^1.0.0-beta.16"]

nbriannl:
If I understand correctly, for a given code-block sample, you would like to have all the necessary import statements of all the components used in that code-block copied as well?

lslee714:
Please on this issue, it drives me nuts every time I go in the docs to figure out how to use a component

yelkar-snehal:
Hi, I'm new to open source. Is this issue still open for grabs?

VibhorJaiswal:
I would like to work on it.

Aaron-Ennis:
This issue doesn't look like it's seen any movement for a while. I started picking at it for a school assignment. To make it happen with `react-live` it looks like `<LiveProvider>` in the `ReactPlayground` component needs the `transformCode` prop added and defined as a function that uses some regex to strip the imports and exports. I got it to work using a class component or a function component in the example/live code area, but might have to do some additional transformation to make it work with arrow functions. In any case, is this what we're looking for here? 

![Screen Shot 2022-05-06 at 18 26 00](https://user-images.githubusercontent.com/34149819/167232578-37bf0f95-0cd4-4baf-8de3-caa9da584470.png)

kyletsang:
@Aaron-Ennis looks good to me.  Feel free to open a PR for review.  You may want to remove the `import React from 'react'` bit as that's not really needed nowadays.

PR #6368, opened 2022-06-04
docs(all): Add imports to playground examples (#4952)

Aaron-Ennis:
This PR addresses issue #4952 by adding the imports for components used in the code examples in the user documentation. `react-live` provides a way to do this using a transform function passed as a prop to `LIveProvider`. In this case, the simple transform function added in `ReactPlayground.js` uses some regex to strip the import and export statements added to the example code before sending it to be rendered in the Live Preview. I only added the imports and exports to code examples that have the `Copy` button.

### DIFF
--- a/www/src/components/ReactPlayground.js
+++ b/www/src/components/ReactPlayground.js
@@ -25,8 +25,8 @@ import * as yup from 'yup';
 import useEventCallback from '@restart/hooks/useEventCallback';
 import useIsomorphicEffect from '@restart/hooks/useIsomorphicEffect';
 import useMutationObserver from '@restart/hooks/useMutationObserver';
-import PlaceholderImage from './PlaceholderImage';
-import Sonnet from './Sonnet';
+import { PlaceholderImage } from './PlaceholderImage';
+import { Sonnet } from './Sonnet';
 
 const scope = {
   useEffect,
@@ -312,6 +312,8 @@ function Editor() {
 const PRETTIER_IGNORE_REGEX =
   /({\s*\/\*\s+prettier-ignore\s+\*\/\s*})|(\/\/\s+prettier-ignore)/gim;
 
+const IGNORE_IMPORTS_EXPORTS_REGEX = /^.*\b(import|export)\b.*$/gim;
+
 const propTypes = {
   codeText: PropTypes.string.isRequired,
 };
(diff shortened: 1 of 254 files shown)


www/src/components/ReactPlayground.js
```diff
@@ -25,8 +25,8 @@ import * as yup from 'yup';
 import useEventCallback from '@restart/hooks/useEventCallback';
 import useIsomorphicEffect from '@restart/hooks/useIsomorphicEffect';
 import useMutationObserver from '@restart/hooks/useMutationObserver';
-import PlaceholderImage from './PlaceholderImage';
```
kyletsang, line 28: Didn't see these files changed.  Do these imports need to be changed?

kyletsang:
Nice job on this!

www/src/components/ReactPlayground.js
Aaron-Ennis: I was going to ask about this in a comment on the PR, but it slipped my mind. When I tried to create a commit, eslint was throwing the `no-named-as-default` rule violation on these imports for some reason.

kyletsang: Might be a false positive.  Could you revert this change?  This import wouldn't work

Aaron-Ennis: Yes. There are also a couple of the example files that use these imports. I will change those to use the default import rather than the named import, as well.

Aaron-Ennis: @kyletsang I'd kind of like to iron out this commit error issue so I can more easily contribute in the future. Would you mind taking a look at this error output (see attached file), and see if you can tell what's going on? I have not changed any config or rc files in my forked version of the repo, so not sure why it's just my local environment causing this. Much appreciated.
[commit_error_output.txt](https://github.com/react-bootstrap/react-bootstrap/files/8865380/commit_error_output.txt)

kyletsang: Just checked my end and looks like we have an issue with our config. We haven't updated the doc components in a while, so this probably broke sometime long ago.

You can skip the precommit step for now.  I'll take a look at fixing this issue when I get a chance.

kyletsang (approved):

kyletsang:
Thanks!  Great first contribution 👍

PR #6371, opened 2022-06-07
Bug/breakpoints smaller than xs

### LINKED ISSUE #6370: Bug: Custom Breakpoints Do Not Work On Col Elements
### Prerequisites

### Describe the bug

- Custom breakpoints do not render properly on `Col` elements.
- Changing an existing breakpoint works fine and uses the correct width that I changed, this is only an issue with new breakpoints

### Expected behavior

- Using new breakpoints will render the breakpoint as a new prop like `<div class="col" xxs="4">` rather than `<div class="col-4">`
- I determined that this is not an issue with bootstrap itself, because when I manually changed the `class` attribute to what it should be in my browser's inspector it worked perfectly as it should.

### To Reproduce

- Adding a new `xxs` breakpoint and using `<Col xxs={6}>` will render as `<div class="col" xxs="6">`. It should be `<div class="col-6">`
- Adding a new `xxs` breakpoint and using `<Col xxs={6} xs={8}>` will render as `<div class="col-8" xxs="6">`. It should be `<div class="col-xs-8 col-6">`
- I also tried a new breakpoint somewhere between the existing breakpoints to see if it was just an issue with breakpoints smaller than any of the defaults and the same thing happens. Adding a new `xmd` breakpoint (between `md` and `lg`) and using `<Col md={4} xmd={5} lg={6}>` will render as `<div class="col-lg-6 col-md-4" xmd="5">`. It should be `<div class="col-lg-6 col-xmd-5 col-md-4">`

### Reproducible Example

- custom.scss:
```
@import 'bootstrap/scss/_functions.scss';
@import 'bootstrap/scss/_variables.scss';

$grid-breakpoints: (
  xxs: 0,
  xs: 392px,
  sm: 576px,
  md: 768px,
  xmd: 888px,
  lg: 992px,
  xl: 1200px,
  xxl: 1400px
);

$container-max-widths: (
  xs: 368px,
  sm: 540px,
  md: 720px,
  xmd: 852px,
  lg: 960px,
  xl: 1140px,
  xxl: 1320px
);

@import 'bootstrap/scss/bootstrap.scss';
```

- example.tsx:
```
import Col from "react-bootstrap/Col";
import Row from "react-bootstrap/Row";

export default function Example() {
  return (
    <Row>
      <Col xxs={12} xs={11} sm={10} md={9} xmd={8} lg={7} xl={6} xxl={5}>
        <p>xxs and xmd do not work properly here</p>
      </Col>
    </Row>
  );
}
```

### Screenshots

Inspector Elements:
image

Inspector Styles:
image

Windows

Chrome, Safari, Firefox, Microsoft Edge, Opera

2.4.0

5.1.3

### Additional context

_No response_

Jordan9232:
Closes #6370 

Adds a `minBreakpoint` prop to ThemeProvider which defaults to "xs". This fixes all occurrences of logic like `brkPoint !== 'xs'` to instead use value from new `minBreakpoint` prop.

### DIFF
--- a/src/ThemeProvider.tsx
+++ b/src/ThemeProvider.tsx
@@ -3,10 +3,12 @@ import * as React from 'react';
 import { useContext, useMemo } from 'react';
 
 export const DEFAULT_BREAKPOINTS = ['xxl', 'xl', 'lg', 'md', 'sm', 'xs'];
+export const DEFAULT_MIN_BREAKPOINT = 'xs';
 
 export interface ThemeContextValue {
   prefixes: Record<string, string>;
   breakpoints: string[];
+  minBreakpoint?: string;
   dir?: string;
 }
 
@@ -17,12 +19,14 @@ export interface ThemeProviderProps extends Partial<ThemeContextValue> {
 const ThemeContext = React.createContext<ThemeContextValue>({
   prefixes: {},
   breakpoints: DEFAULT_BREAKPOINTS,
+  minBreakpoint: DEFAULT_MIN_BREAKPOINT,
 });
 const { Consumer, Provider } = ThemeContext;
 
 function ThemeProvider({
   prefixes = {},
   breakpoints = DEFAULT_BREAKPOINTS,
+  minBreakpoint = DEFAULT_MIN_BREAKPOINT,
   dir,
   children,
 }: ThemeProviderProps) {
@@ -30,9 +34,10 @@ function ThemeProvider({
     () => ({
       prefixes: { ...prefixes },
       breakpoints,
+      minBreakpoint,
       dir,
     }),
-    [prefixes, breakpoints, dir],
+    [prefixes, breakpoints, minBreakpoint, dir],
   );
 
   return <Provider value={contextValue}>{children}</Provider>;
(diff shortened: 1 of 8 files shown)


src/ThemeProvider.tsx
```diff
@@ -77,6 +88,11 @@ export function useBootstrapBreakpoints() {
   return breakpoints;
 }
 
+export function useBootstrapMinBreakpoint() {
+  const { minBreakpoint } = useContext(ThemeContext);
+  return minBreakpoint;
+}
+
 export function useIsRTL() {
```
Jordan9232, line 96: Should we maybe throw an exception somewhere in here if the `minBreakpoint` does not exist in the `breakpoints` array?

kyletsang: We could probably enforce the ordering of the breakpoints array and have the code check for `breakpoints[0]`.  We could avoid adding the extra prop.   Thoughts?

Jordan9232: That was actually my first thought too but I didn't want to causing breaking changes between updates for some people. I do like it better though, if you think that's not much of a concern I can change it.

kyletsang: Was originally thinking it might not be an issue, but noticed it's been ~3 months since support was released.  I guess this is ok for now, and we'll clean this up in the next major release.  Thanks!

kyletsang (approved):

Issue #5230, opened 2020-06-10
Modal fade animation gets interrupted by .btn transition


omermecitoglu:
There is no problem when you click on the close button (X) or backdrop. But this is happening when you click on a button in the footer. Just click fast enough to see the difference on the fade animation.

https://react-bootstrap.github.io/components/modal/#modals-live

I think it's caused by transition event in bootstrap's .btn class, as @bbtfr mentioned [here](https://github.com/react-bootstrap/react-bootstrap/issues/2151#issuecomment-244724366).

```
.btn {
	transition: color 0.15s ease-in-out, background-color 0.15s ease-in-out, border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
}
```

uuykay:
A hacky solution, but I have found that if you add a small timeout, (smallest being 200 milliseconds) you can see the closing animation.

Note I am using Redux in this example, but you can just as easily use useState and the function that sets the boolean. 
```
      <Button
        onClick={() => {
          setTimeout(function() {
            dispatch(toggleUIModal("hide"))
          }, 200)
        }}
      >
        Try answering again
      </Button>
```

kyletsang:
I guess we could use something like this @taion ?  In my testing, I get about 5 `transitionend` events from the button before getting 1 from the modal.

```
function elementTransitionEnd(node, done) {
  const remove = listen(node, 'transitionend', (e) => {
    if (e.target === node) {
      remove();
      done();
    }
  });
}
```

taion:
It seems reasonable in this case; I think there are other cases where the transition may not be on the element itself. Could we make an upstream fix in dom-helpers? cc @jquense – would this be correct?

kyletsang:
Yeah I think a fix in dom-helpers is the best place for it.  Could add an option like `exactElement: boolean` or something to `transitionEnd`

MSchiavi:
Not sure if this is still ongoing but this problem also occurs with Toasts as well.

PR #6377, opened 2022-06-13
When listening for transitionEnd, look for event.currentTarget instead of event.target

kubabogaczewicz:


kubabogaczewicz (commented):
I've found the issue when using modals with animation turned on.
It's hard to reproduce (in my case it's a small react app mounted inside larger angularjs one with some reproduction steps), but sometimes transitionEnd event can be fired not from the whole modal, but instead for an element inside it, modal then ignores it and stays behind blocking the rest of application.
I don't think this change can be dangerous as it's very seldom that you actually want to check e.target instead of e.currentTarget.

kyletsang:
Unfortunately this will cause #5230 to reappear.  transitionend events can be fired from child nodes, but the one we really want to listen to is the modal.

kubabogaczewicz:
You're right, that "fix" would cause problems. I do see strange behavior where sometimes transitionend event is fired by child element instead of modal itself, but that looks very strange and I've figured out a workaround.
Thank you for taking time to answer the PR.

Issue #6288, opened 2022-03-29
`OverlayTrigger` should support more of the `Overlay` interface, and also `delay` is not implemented (or not documented)


zachsiegel-capsida:
### Is your feature request related to a problem? Please describe

The `OverlayTrigger` interface is by design limited, to make it easier to use than `Overlay`. However, I doubt I'm the only developer who wishes the interface was defined as "all the features of `Overlay` without any of the parameters related to identifying the element to which the overlay points". I'm grateful you made this convenient wrapper for `Overlay`, but I don't see why `OverlayTrigger` can't support, say, `rootClose` and `rootCloseEvent`. I can't see how this change would be breaking for anyone, and I would appreciate it myself.

There is at least [one other open issue](https://github.com/react-bootstrap/react-bootstrap/issues/5175) regarding an `Overlay` parameter that is unavailable on `OverlayTrigger` (also `rootClose`).

Also - I noticed the `delay` arguments only seem to apply to `hover` trigger events. I don't see why there can't be an automatic close when there is a `close` field in the `delay` argument even when the trigger is `click`. Either the documentation should clarify that `delay` is only for `hover` (and possibly `focus`) events, or it should apply to `click` trigger events.

### Describe the solution you'd like

Implement the `rootClose` parameter on `OverlayTrigger` (probably just pass it to the `Overlay`), update the docs.

Also, either `delay` should apply to `click` events or the documentation should express that it doesn't.

### Describe alternatives you've considered

I will not use `OverlayTrigger`, but will use `Overlay` directly. This is a definite downgrade in conciseness.

zachsiegel-capsida:
After testing it out (and spending my and your time with this issue), I realize that `rootClose` seems to work just fine on `OverlayTrigger`. Perhaps it could be indicated in the documentation that some (all?) `Overlay` props work just fine on `OverlayTrigger`.

Thanks so much for developing this package.

JustinTime42:
[...]

PR #6389, opened 2022-06-29
Added a note in the documentation for OverlayTrigger

JustinTime42:
Regarding this issue: https://github.com/react-bootstrap/react-bootstrap/issues/6288

Added a note to the documentation for `OverlayTrigger` that the `click` trigger ignores the `delay` to avoid this confusion.

### DIFF
--- a/src/OverlayTrigger.tsx
+++ b/src/OverlayTrigger.tsx
@@ -74,8 +74,6 @@ const propTypes = {
   /**
    * Specify which action or actions trigger Overlay visibility
    *
-   * The `click` trigger ignores the configured `delay`.
-   *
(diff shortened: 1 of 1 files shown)


kyletsang:
Thanks for the PR, but I think we'd like to keep the click behavior as it is for now.  We could add documentation stating that delay doesn't work on click though

kyletsang:
I believe you reverted too many commits?  I'm seeing no files changed here

JustinTime42:
@kyletsang Sorry about that - it's fixed now

kyletsang (approved):

kyletsang:
Thanks!

Issue #6321, opened 2022-05-05
Using Escape to Close Modal Window that has autoFocus causes underlying page to scroll away from current position


ivaluedev01:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

Open a Modal window with a Form inside Modal.Body and autoFocus on first Element.

When Modal is closed using Escape Key the underlying Page scrolls away from its starting position. To top or bottom.

E.g., Demo version on site [Here](https://react-bootstrap.github.io/components/modal/#focus-on-specific-element)

### Expected behavior

The underlying page should remain where it is. AutoFocus should not fire on closing Modal

### To Reproduce

Go to [Demo](https://react-bootstrap.github.io/components/modal/#focus-on-specific-element)

Click Launch Demo Modal

Click Escape after Modal Appears and Focus is in First Element

Page will Scroll to Top

### Reproducible Example

`
<Modal show={true}>
  <Modal.Body>
    <Form>
      <Form.Group>
          <Form.Label>Segment Name</Form.Label>
          <Form.Control autoFocus/>
        </Form.Group>
      </Form>
    </Modal.Body>
</Modal>
`

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Safari

### What version of React-Bootstrap are you using?

2.3.1

### What version of Bootstrap are you using?

5.1.3

kyletsang:
Looks like a Safari-specific issue.  Some browsers set focus to the button element when triggering the modal, so that when the modal is closed, focus is restored to the button.  

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus

jeraymond:
I found [this workaround](https://github.com/reach/reach-ui/issues/754#issuecomment-823507275) that prevents the issue on Safari. Before closing the dialog (setting `show` to `false`) do this:

```javascript
if (document.activeElement instanceof HTMLElement) {
      document.activeElement.blur();
}
```

PR #6392, opened 2022-07-05
fix: #6321

koshpendi:
Regarding this issue: https://github.com/react-bootstrap/react-bootstrap/issues/6321


kyletsang:
Hi, not sure how this addresses the focus issue?  I believe the fix is better left to the user since we have no context of how the modal was triggered

Issue #4461, opened 2019-09-13
Error when trying to create a customized variant for Button


longnguyennn:
## Describe the bug

I'm getting error from Typescript when trying to create a customized variant for Button, following the official guide: https://react-bootstrap.github.io/getting-started/theming/

## To Reproduce

Use Button with any variant other than the provided variants and you will get the error from Typescript.

## Screenshots

<img width="1081" alt="Screen Shot 2019-09-12 at 10 07 04 PM" src="https://user-images.githubusercontent.com/24534566/64832918-e209e780-d5a9-11e9-82dc-4dd9cacf8a14.png">

taion:
You'll need to override the `ButtonProps` interface to add your new variant.

MichaelSchwabDev:
@taion I'm not sure if this is possible.  Overriding of interfaces isn't possible unless I'm really missing something: https://stackoverflow.com/questions/41285211/overriding-interface-property-type-defined-in-typescript-d-ts-file

And in their documentation it states:
> Non-function members of the interfaces should be unique. If they are not unique, they must be of the same type. The compiler will issue an error if the interfaces both declare a non-function member of the same name, but of different types.

https://www.typescriptlang.org/docs/handbook/declaration-merging.html

My attempts to override ButtonProps with a new or modified version have so far been unsuccessful.  I also tried the method augmentation method at the end of the above article to no avail.  Either the declaration is not picked up by the compiler or it says I cannot overwrite "variant".

liorbentov:
@MichaelSchwabDev , did you manage solving it?

AndreyKuleb:
Try <Button variant={''something" as any} />

ericorruption:
> You'll need to override the `ButtonProps` interface to add your new variant.

Do you have an example of how to do this?

kyletsang:
@taion Should we relax the type strictness for variant?   Like `variant?: string`?  This would reduce the friction in usage when creating custom classes or using sass `$theme-colors`.

kosmiq:
This has been brought up in various ways before. @taion gave a reply here: https://github.com/react-bootstrap/react-bootstrap/issues/5047 when I suggested a way to do it.

taion:
ugh, i have no idea. it's nice to get type checking here that works nicely with bootstrap default classes out of the box.

is there really no good way to override these?

kosmiq:
[...]

PR #6398, opened 2022-07-12
Ensure Consistent API for custom variants

ryanolee:
# What
This PR follows on from https://github.com/react-bootstrap/react-bootstrap/issues/4461 and aims to ensure consistency when using custom variants across react bootstrap. Currently a number of components do not properly support custom variants in the way the issue was resolved as not everything uses Variant as a base. (Such as Navbar). 

# Why
Given this looks to be the agreed upon approach for custom variants all `variant` attributes might need a type of "| string" adding

# How
By adding "| string" to the remaining typedefs where variant is still using a Union of hardcoded values. This should not cause any bc breaks as it is extending hardcoded unions of strings.

Thanks for looking at this!


kyletsang (approved):

kyletsang:
Thanks!

PR #6405, opened 2022-07-20
docs: Emphasis on bootstrap peer-dependency

emranbm:
For new projects and new joiners, it's almost always a waste of time
guessing why the components of this cool library don't work, before
realizing that `bootstrap` is not installed and imported automatically.
i.e. `react-bootstrap`, by its name, misleads the users to leave the
installation of `bootstrap` itself to `react-bootstrap`.
This commit adds an extra emphasis on this point through the `Usage`
section.


kyletsang:
It's probably better to link to the docs instead since it covers this + more:
https://react-bootstrap.github.io/getting-started/introduction

PR #6407, opened 2022-07-22
Accordion additional event oncomplete and additional properties for onselect

jeevasusej:
Accordion additional event oncomplete and additional properties for onselect

- OnComplete will be fired after collapse or expand completes
- OnSelect will have additional two properties which is currentkey and IsExpanded. currentkey is to know which key has been clicked and IsExpanded is to know the current operation without doing any additional operation

### DIFF
--- a/src/Accordion.tsx
+++ b/src/Accordion.tsx
@@ -10,6 +10,7 @@ import AccordionCollapse from './AccordionCollapse';
 import AccordionContext, {
   AccordionSelectCallback,
   AccordionEventKey,
+  AccordionCompleteCallback,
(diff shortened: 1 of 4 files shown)


atomosdovini (approved):

jeevasusej:
Do we have any idea to merge this PR?

kyletsang:
We should expose the existing `AccordionCollapse` transition callbacks on `Accordion` instead (ie. `onEnter`, `onExit` etc). As for the open/closed state in the callback, you could use your `activeKey` state and calculate based on that

PR #6434, opened 2022-08-26
Update README.md

SuchitraSwain:
I have done this in npm package-lock.json and it worked.


kyletsang:
Thanks for the PR, but we'll pass on this.  While npm does work, we're already using yarn (and checked in our yarn lock).  Adding another package manager to the mix will end up with mismatched package resolution.

PR #6453, opened 2022-09-29
Replace direction:String prop with horizontal:Bool

pretendprogrammer:
This makes it easier to set a stack to horizontal.

If a Stack defaults to vertical anyways, why require a specific string to be provided instead of a boolean? There are only two directions, unless there are plans for a depth-stack, using z-index.

I removed references to a prop called 'direction', and replaced it with a prop called horizontal.

### DIFF
--- a/src/Stack.tsx
+++ b/src/Stack.tsx
@@ -32,6 +32,11 @@ const propTypes = {
   bsPrefix: PropTypes.string,
 
   /**
+   * Arranges the elements of the stack horizontally.
+   */
+  horizontal: PropTypes.bool,
+
+  /**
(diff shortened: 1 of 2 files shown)


kyletsang:
Thanks for opening the PR, but we won't be considering any breaking changes to Stack at this time.

pretendprogrammer:
@kyletsang May I open a new PR if I re-work it to remove breaking changes?

kyletsang:
@pretendprogrammer , we're going to stick with the current API.  FWIW, the API is similar to that of MUI's Stack component and bootstrap may implement similar capabilities such as reverse ordering.  The current API would make these trivial to implement.

PR #6458, opened 2022-10-04
docs: add docs to close button

lcheunglci:
I noticed that the api docs had the onclick and aria-label fields missing, so I added it in.


kyletsang (approved):

kyletsang:
Thanks

PR #6460, opened 2022-10-05
docs: update yarn-install link

SebastiaanNijland:
The yarn-install link in the Readme is pointing to the old, classic version of yarn, instead of the modern version. This changes the url to use the new version.


kyletsang:
Thanks for the PR.  We are actually still using yarn classic and haven't done the migration to later versions yet.

PR #6461, opened 2022-10-06
docs: add onclick and href docs to listgroupitem

lcheunglci:
I noticed that the onClick and href is missing for the ListGroupItem api, so I added it in this PR.

### DIFF
--- a/src/ListGroupItem.tsx
+++ b/src/ListGroupItem.tsx
@@ -50,7 +50,7 @@ const propTypes = {
   /** A callback function for when this component is clicked.  */
   onClick: PropTypes.func,
 
-  /** Providing a `href` will render the ListGroup.Item as an `<a>` element (unless `as` is provided). */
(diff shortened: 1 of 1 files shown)


src/ListGroupItem.tsx
```diff
@@ -36,19 +36,21 @@ const propTypes = {
...
+   * Sets list item state as disabled.
    */
   disabled: PropTypes.bool,
 
   eventKey: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 
+  /** A callback function for when this component is clicked.  */
   onClick: PropTypes.func,
 
+  /** Providing a `href` will render the ListGroup.Item as an `<a>` element (unless `as` is provided). */
```
kyletsang, line 53: The `action` prop is also required along with `href` to render an anchor.

lcheunglci: oh ok, thanks! I'll update it

kyletsang (approved):

kyletsang:
Thanks!

PR #6462, opened 2022-10-07
fix(ListGroupItem): prevent a div with a href

lcheunglci:
While adding the docs in #6461 , I realize that if `action=false` and `href` is set, a `<div href="...">` is generated.  I think the `href` attribute should only be for the `<a>` tag or custom component if the `as` is specified.


kyletsang:
Thinking on this... I think we should show a warning instead because the combination of `action=false` with a `href` is invalid.  Use the `warning` function for this.  There are examples in the codebase on how to do it.

kyletsang (approved):

kyletsang:
Thanks!

PR #6465, opened 2022-10-11
fix(types): Fix type of 'safeFindDOMNode' function

RyanCavanaugh:
Based on its body, `safeFindDOMNode` appears to have an incorrect function signature (and TypeScript 4.9 complains about it as a result). The declared type of `React.ComponentClass` refers to *the thing you `new`*, whereas `React.Component` refers to the instance of the class. `setState` is an instance method, not a static method, so will (presumably) never be `in` the former.


kyletsang (approved):

kyletsang:
Thanks!

PR #6468, opened 2022-10-17
make the spinner animation default to border

factoidforrest:
There's no good reason to always type out the animation for the spinner. There are only two choices, and the "border" option is the obvious choice as a default.

This should be a non-breaking change.


kyletsang:
Sounds like a good idea.  For your code, you could add a default parameter to animation and keep the existing bsPrefix bit the same.

factoidforrest:
@kyletsang you're right, that's cleaner, I changed it. How does that look?

kyletsang:
Thanks!

PR #6478, opened 2022-11-04
feat(AccordionBody): add transition callback props

kyletsang:
Replaces #6407

Forwards transition callbacks on `AccordionBody` to the underlying `AccordionCollapse`


kyletsang:
@jquense can I get your review on this?

PR #6490, opened 2022-11-13
feat(Dropdown): add support for centered dropdown menus

### LINKED ISSUE #6477: Support `dropdown-center`
### Is your feature request related to a problem? Please describe

With Bootstrap 5.2, the option to center a dropdown was introduced, but React-Bootstrap does not support it.

### Describe the solution you'd like

Support for `dropdown-center`.

### Describe alternatives you've considered

/

### Additional context

https://getbootstrap.com/docs/5.2/components/dropdowns/#centered

kyletsang:
Closes #6477

I thought about putting a new option `center` in the `align` prop as that seems more intuitive, but it might get confusing because upstream bootstrap doesn't have support for responsive menu centering.  These classes are also placed on the `Dropdown` component instead of the `DropdownMenu` where the `align` prop is shared...

### DIFF
--- a/src/Dropdown.tsx
+++ b/src/Dropdown.tsx
@@ -46,7 +46,14 @@ const propTypes = {
   /**
    * Determines the direction and location of the Menu in relation to it's Toggle.
    */
-  drop: PropTypes.oneOf(['up', 'start', 'end', 'down']),
+  drop: PropTypes.oneOf<DropDirection>([
+    'up',
+    'up-centered',
+    'start',
+    'end',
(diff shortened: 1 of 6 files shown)


jquense (approved):
LGTM, I think it's a bit weird upstream doesn't just use the popper placements for this sort of thing but maybe it's related to additional styling?

golota60 (approved):
lgtm - this makes the `align` prop a little bit awkward to use but i can't see how to address that

Issue #6497, opened 2022-11-16
indeterminate checkbox support


nicko-winner:
### Is your feature request related to a problem? Please describe

Checkbox does not seem to support intermediate states documents in bootstrap.
<img width="843" alt="2022-11-16_12-38-13" src="https://user-images.githubusercontent.com/35456401/202253288-fc2e8938-f7f9-451a-84e9-465c166cf8a6.png">

related documentation:
https://getbootstrap.com/docs/5.0/forms/checks-radios/#indeterminate

Example of how it could be implemented in React:
https://www.robinwieruch.de/react-checkbox-indeterminate/

### Describe the solution you'd like

support for intermediate checkboxes

### Describe alternatives you've considered

Building a customer wrapper around the bootstrap component to support this using the input ref.

jquense:
This is more of a React issue than a RB one. You can set `indeterminate` on the FormCheck, but there isn't an prop because React doesn't support a declarative version of this: https://github.com/facebook/react/issues/1798

Issue #6496, opened 2022-11-16
Feature


jkinley:
### Is your feature request related to a problem? Please describe

There is no way to use Bootstrap's indeterminate class on Checkboxes. This seems like an odd omission.

### Describe the solution you'd like

Have the ability to pass a prop?

### Describe alternatives you've considered

Is it possible to make a custom checkbox for this? I suppose that would work, but seems silly this is not built in.

PR #6498, opened 2022-11-16
feat: support indeterminate checkbox state

golota60:
#6496 #6497 Adds `indeterminate` checkbox state support. 

Full disclosure, I've implemented this before I've seen https://github.com/react-bootstrap/react-bootstrap/issues/6497#issuecomment-1317600419 this comment, so I fully understand if we don't wanna merge this(plus it's doing a lot of unelegant logic which also may be troublesome)

### DIFF
--- a/src/FormCheck.tsx
+++ b/src/FormCheck.tsx
@@ -26,6 +26,7 @@ export interface FormCheckProps
   feedback?: React.ReactNode;
   feedbackType?: FeedbackType;
   bsSwitchPrefix?: string;
+  indeterminate?: boolean;
 }
 
 const propTypes = {
@@ -145,6 +146,7 @@ const FormCheck: BsPrefixRefForwardingComponent<'input', FormCheckProps> =
         children,
         // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
         as = 'input',
+        indeterminate = false,
         ...props
       },
       ref,
@@ -173,6 +175,7 @@ const FormCheck: BsPrefixRefForwardingComponent<'input', FormCheckProps> =
           isInvalid={isInvalid}
           disabled={disabled}
           as={as}
+          indeterminate={indeterminate}
         />
       );
 
--- a/src/FormCheckInput.tsx
+++ b/src/FormCheckInput.tsx
@@ -2,9 +2,14 @@ import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import * as React from 'react';
 import { useContext } from 'react';
+import { isForwardRef } from 'react-is';
 import FormContext from './FormContext';
 import { useBootstrapPrefix } from './ThemeProvider';
-import { BsPrefixProps, BsPrefixRefForwardingComponent } from './helpers';
+import {
+  BsPrefixProps,
+  BsPrefixRefForwardingComponent,
+  isDOMElement,
+} from './helpers';
 
 type FormCheckInputType = 'checkbox' | 'radio';
 
@@ -69,27 +74,24 @@ const FormCheckInput: BsPrefixRefForwardingComponent<
     bsPrefix = useBootstrapPrefix(bsPrefix, 'form-check-input');
 
     const inputRef = React.useRef<HTMLInputElement>();
-    React.useImperativeHandle(
-      ref,
-      () =>
-        ({
-          current: {
-            ...inputRef.current,
-            indeterminate: !!indeterminate,
-          },
-        } as any),
-    );
-
+    const existingRef =
+      isForwardRef(<Component />) || isDOMElement(<Component />)
+        ? ref || inputRef
+        : undefined;
     React.useEffect(() => {
-      if (inputRef.current && type === 'checkbox' && indeterminate) {
-        inputRef.current.indeterminate = true;
+      if (
+        (existingRef as any)?.current &&
+        type === 'checkbox' &&
+        indeterminate
+      ) {
+        (existingRef as any).current.indeterminate = true;
       }
-    }, [indeterminate, ref, type]);
+    }, [existingRef, indeterminate, ref, type]);
 
     return (
       <Component
         {...props}
-        ref={inputRef}
+        ref={existingRef}
         type={type}
         id={id || controlId}
         className={classNames(
--- a/src/FormCheckInput.tsx
+++ b/src/FormCheckInput.tsx
@@ -14,6 +14,7 @@ export interface FormCheckInputProps
   type?: FormCheckInputType;
   isValid?: boolean;
   isInvalid?: boolean;
+  indeterminate?: boolean;
 }
 
 const propTypes = {
@@ -40,6 +41,9 @@ const propTypes = {
 
   /** Manually style the input as invalid */
   isInvalid: PropTypes.bool,
+
+  /** Set whether the input is of indeterminate state */
+  indeterminate: PropTypes.bool,
 };
 
 const FormCheckInput: BsPrefixRefForwardingComponent<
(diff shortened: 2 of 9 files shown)


jquense:
Ooo nice. I don't know how this plays with SSR but maybe that's fine?

golota60:
yeah for SSR it doesn't work since we apply it using JS :/ I'll add a note about that in the docs

src/FormCheckInput.tsx
```diff
@@ -63,10 +73,26 @@ const FormCheckInput: BsPrefixRefForwardingComponent<
     const { controlId } = useContext(FormContext);
     bsPrefix = useBootstrapPrefix(bsPrefix, 'form-check-input');
 
+    const inputRef = React.useRef<HTMLInputElement>();
+    const existingRef =
```
kyletsang, line 77: I think it's better to use `useMergedRefs` here?   Also do we need to check with `isForwardRef` and `isDOMElement`?  Maybe we leave this up to the user to ensure `Component` is a proper ref forwarded component

jquense: I think the issue is we always need to attach a ref, or maybe it's enough to check if `indeterminate` is provided and then leave it up to the user to make it reffable?

kyletsang: Yeah that sounds good.  We can avoid the new dep

golota60: Just to be sure, there are 3 cases i am trying to cover here
- `as` prop is a html tag - no problem, using `ref || existingRef` will work(`existingRef` will be passed)
- `as` prop is a React element which accepts a `ref` - no problem, using ` ref || existingRef` will work(`ref` will be passed)
- `as` props is a React element which does not accept a `ref` - we will try to pass a ref(`existingRef`) to an element which doesn't accept a `ref` - this is the only edge case where I'd pass `ref={undefined}`

I can leave the 3rd case broken and require the user to provide a react component that forwards a ref, but we'd lose on backwards-compatibility front(i even had to remove one test to make this work).

golota60: by checking something along of what i added which can be simplified to: `isForwardRef(Component) ? undefined : someRef` we can support every case i think, the downside being the `react-is` package

jquense: I was suggesting only adding the ref if `indeterminate != undefined` that should be backwards compatible since the prop is new. you can do something like:

```
const inputRef = useRef(null)
const mergedRef = useMergedRef(inputRef, outerRef)

<Component ref={indeterminate != null || outerRef ? mergedRef : undefined}
```

check tho that `ref={undefined}` doesn't still send the error sometimes React uses `in` checks isntead

src/FormCheckInput.tsx
```diff
@@ -54,6 +60,7 @@ const FormCheckInput: BsPrefixRefForwardingComponent<
       type = 'checkbox',
       isValid = false,
       isInvalid = false,
+      indeterminate = false,
```
jquense, line 63: i think you need to distinguish between `undefined` and `false` otherwise how does removing the indeterminate state work? The ref will get removed but i don't think React will reset the attribute?

golota60: yeah i didn't think about removing the prop, i also think we need to even go a step further and explicitly set it `element.indeterminate` to `false`, when `indeterminate=false`. otherwise things like this:

```ts
  const [isIndeterminate, setIndeterminate] = useState(true);

  return (
    <>
      <button onClick={() => setIndeterminate(!isIndeterminate)}>click</button>
          <Form.Check // prettier-ignore
            type="checkbox"
            indeterminate={isIndeterminate}
            onClick={() => setIndeterminate(!isIndeterminate)}
          />
    </>
  );
```

won't work, cause react won't remove it until we re-mount it, cause ref will persist across renders.

1. We mount Form.Check with `intermediate={true}`
2. We click on the button(setting `indermediate={false}` on the checkbox)
3. Nothing happens, cause we've already set the attr on the ref to true.

golota60: unless the thing you have in mind accounts for that

src/FormCheck.tsx
```diff
@@ -145,6 +146,7 @@ const FormCheck: BsPrefixRefForwardingComponent<'input', FormCheckProps> =
         children,
         // Need to define the default "as" during prop destructuring to be compatible with styled-components github.com/react-bootstrap/react-bootstrap/issues/3595
         as = 'input',
+        indeterminate = false,
```
jquense, line 149:
```suggestion
        indeterminate,
```

src/FormCheckInput.tsx
```diff
@@ -63,10 +70,24 @@ const FormCheckInput: BsPrefixRefForwardingComponent<
     const { controlId } = useContext(FormContext);
     bsPrefix = useBootstrapPrefix(bsPrefix, 'form-check-input');
 
+    const [element, inputRef] = useCallbackRef<HTMLInputElement>();
+    const existingRef = useMergedRefs(inputRef, ref);
+
```
jquense, line 75:
```suggestion
    const hasIndeterminate = indeterminate != null
```

jquense:
I wish we had deploy previews set up...

golota60:
are they not set up? it's one of the ci jobs
https://deploy-preview-6498--react-bootstrap.netlify.app/forms/checks-radios/

jquense:
O nice. Hmm i'm not sure the behavior i'm seeing makes sense? if indeterminate is set it should always show regardless of checked status right? It's not actually a third boolean

golota60:
that's right - seems like setting `checked`, resets `indeterminate`... it's an easy fix for an uncontrolled elem

golota60:
that's right - seems like setting `checked`, resets `indeterminate`... it's an easy fix for an controlled elem, but for an uncontrolled one it's quite harder cause we don't know when `element.checked` changes - i'll try to figure this out

jquense:
We may be inching closer to why this doesn't already exist 🤣 Worth taking to the end tho!

golota60:
turns out it a fix was quite simple - just attached the indeterminate-computation function to `onChange`

src/FormCheckInput.tsx
```diff
@@ -54,19 +60,45 @@ const FormCheckInput: BsPrefixRefForwardingComponent<
...
+      onChange,
       ...props
     },
     ref,
   ) => {
     const { controlId } = useContext(FormContext);
     bsPrefix = useBootstrapPrefix(bsPrefix, 'form-check-input');
 
+    const [element, inputRef] = useCallbackRef<HTMLInputElement>();
+    const existingRef = useMergedRefs(inputRef, ref);
```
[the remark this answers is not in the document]
golota60: i think using `useUncontrolledProp` on `checked` might improve readability - but I'm struggling to see how it would simplify things
```ts
    const changeHandler = (e) => {
      handleIndeterminate();
      onChange?.(e);
    };
    const [innerChecked, setInnerChecked] = useUncontrolledProp(
      checked,
      defaultChecked,
      changeHandler,
    );
```
we still would need get refs(to get `element.indeterminate`) & set `element.indeterminate` in `onChange` handler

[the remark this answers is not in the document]
jquense: Yeah but you combine the call sites where that happens with one use effect that has `checked` as its dep trigger. 

It's not a big difference maybe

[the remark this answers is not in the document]
golota60: idk whether `useUncontrolledProp` will work 
```ts
    const [innerChecked, setInnerChecked] = useUncontrolledProp(
      checked,
      defaultChecked,
      onChange,
    );
```
`onChange` expects the argument to be an `Event`, meanwhile we need to pass a boolean to the inner `useState` in `setInnerChecked`
https://github.com/jquense/uncontrollable/blob/dd40d92600566b0af14e66e0942b3066e2f62528/src/hook.tsx#L41

I've pushed a version with just a setState - if you know how to make it work with `useUncontrolledProp` let me know!

dotku:
so how's things going so far?

Issue #6495, opened 2022-11-15
ALLOW PORTRAIT-SIZED RATIOS


emamoah:
### Is your feature request related to a problem? Please describe

It's saddening to find out that the Ratio component doesn't allow the container height to go beyond 100% of the width.

### Describe the solution you'd like

I'd like the landscape restriction to be removed and allow the height to be greater than the width.

kyletsang:
Done in #6501

PR #6501, opened 2022-11-21
Added support for portrait size ratios

JustinTime42:
Regarding this issue: https://github.com/react-bootstrap/react-bootstrap/issues/6495, suggesting allowing ratios above 100% of the width to allow for portrait style presentation. 

Currently, in the `toPercent()` function, any number above 1 converts to a percent, ensuring that width is always greater than or equal to height. Perhaps, for small numbers above 1 we could infer that the developer wants a portrait style layout. I've suggested 5 somewhat arbitrarily, but it seems absurdly small for a percent height/width ratio so seems safe to me. Suggestions?

### DIFF
--- a/src/Ratio.tsx
+++ b/src/Ratio.tsx
@@ -38,7 +38,7 @@ const defaultProps = {
 
 function toPercent(num: number): string {
   if (num <= 0 || num > 100) return '100%';
-  if (num < 1) return `${num * 100}%`;
+  if (num < 5) return `${num * 100}%`;
   return `${num}%`;
 }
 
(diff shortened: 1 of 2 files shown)


src/Ratio.tsx
```diff
@@ -38,7 +38,7 @@ const defaultProps = {
 
 function toPercent(num: number): string {
   if (num <= 0 || num > 100) return '100%';
-  if (num < 1) return `${num * 100}%`;
+  if (num < 5) return `${num * 100}%`;
```
kyletsang, line 41: It's better to remove the upper limit of 100 in the line above so users can specify 200 as the value

JustinTime42: Ah of course - Thanks!
I've committed your suggestion with a change to tests to allow the taller aspect ratio.

kyletsang (approved):

kyletsang:
Thanks!

PR #6506, opened 2022-11-25
Add text align utility class name

SmolSoftBoi:
Add future-proofing for additional utilities.

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { TransitionComponent } from '@restart/ui/types';
+import { UtilityProps } from './utilities';
 
 export type Omit<T, U> = Pick<T, Exclude<keyof T, keyof U>>;
 
@@ -19,7 +20,7 @@ export interface AsProp<As extends React.ElementType = React.ElementType> {
 
 export interface BsPrefixProps<As extends React.ElementType = React.ElementType>
   extends BsPrefixOnlyProps,
-    AsProp<As> {}
+    AsProp<As>, UtilityProps {}
 
 export interface BsPrefixRefForwardingComponent<
   TInitial extends React.ElementType,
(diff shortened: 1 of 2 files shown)


kyletsang:
This looks like it's trying to achieve the same thing as the code in this file?
https://github.com/react-bootstrap/react-bootstrap/blob/master/src/createUtilityClasses.ts

SmolSoftBoi:
@kyletsang I aim to be able to use utilities directly as props on components, for example: `<Component textAlign="center" />`.

What approach would you suggest?

kyletsang:
> @kyletsang I aim to be able to use utilities directly as props on components, for example: `<Component textAlign="center" />`.
> 
> What approach would you suggest?

Not all the utility classes would be applicable to each component, so we should be careful about adding it into the `BsPrefixProps` types.  I know there was an effort (#5373) into making a `Box` component that worked with several utilities but unfortunately that PR was dropped.  

Maybe it's ok to add explicit utility props to the components where it makes sense?  Any thoughts on this @jquense?

@SmolSoftBoi which component are you adding `textAlign` to?

jquense:
Idk I'm not sure it's worth the minor convenience to using classes directly. The problem is really that it balloons the component API significantly which is problematic for `as` accepting ones that pass through. I'd fear that targeted partial support would lead to constant asking to support more, which I think we see already a bit.

SmolSoftBoi:
> Not all the utility classes would be applicable to each component, so we should be careful about adding it into the `BsPrefixProps` types. I know there was an effort (#5373) into making a `Box` component that worked with several utilities but unfortunately that PR was dropped.

Why was the PR dropped?

kyletsang:
> > Not all the utility classes would be applicable to each component, so we should be careful about adding it into the `BsPrefixProps` types. I know there was an effort (#5373) into making a `Box` component that worked with several utilities but unfortunately that PR was dropped.
> 
> Why was the PR dropped?

Probably lack of time.  Brady is no longer active on this project

SmolSoftBoi:
Is it something you'd want picked up and implemented?

PR #6540, opened 2023-01-27
fix(Stack): fix incorrect classes being generated

kyletsang:
Breakpoint params were placed in the wrong spot.  They shouldn't be inside the utility values map.


golota60 (approved):
👌 lgtm

PR #6542, opened 2023-01-27
feat(FormCheck): title attribute on the input

slobo:
When `label` is not supplied, we should place the `title` attribute  on the input itself.

I'll gladly write tests if you agree this is a welcome functionality. Thanks!


kyletsang:
Seems ok to me, although you should really use a label where possible for accessibility purposes.

slobo:
My use case is placing plain checkboxes in separate column in every row of a table. I suppose the other option would be to render a label and then hide it with CSS, but not sure if that makes it any more accessible -  my understanding is that screen readers will happily use `title`.

PR #6556, opened 2023-02-23
Propose: use Sandpack as code playground

danilowoz:
Screenshot 2023-02-23 at 14 45 02

Hey there! I want to propose that react-bootstrap starts using Sandpack to run interactive live code examples in the documentation.

Sandpack is a component toolkit for creating live code editing experiences powered by the online bundler used on CodeSandbox. It provides an open ecosystem of components and utilities that allow you to compile and run modern frameworks in the browser, making it perfect for documenting components, bug reports, and playgrounds. Many open-source documentations, such as [React docs](https://beta.reactjs.org/), [Chakra UI](https://play.chakra-ui.com/playground?code=), [Amplify](https://ui.docs.amplify.aws/), and others, use Sandpack in various ways to make their documentation more interactive.

Here's an example of the [alerts component page](https://2ouhld-8000.preview.csb.app/components/alerts) running in Sandpack. I believe this could be very useful to show on all the documentation, as it lets users play around with code and see the example live.

**Why Sandpack?**
- Security: The bundler evaluates and transpiles all files in an iframe under a different subdomain. This is important because it prevents attackers from tampering with cookies of the host domain when evaluating code.

- User-experience: With built-in npm dependency support, hot module reloading, error overlaying, caching, and Node.js support, Sandpack can provide an environment closer to the user development experience. Sandpack uses CodeMirror as its code editor, which supports touch devices and a lightweight bundler.

- Offline Support: Sandpack uses Service Workers to download all transpilers in the background so that the next time a user visits the react-bootstrap Docs, they don't have to download the bundler anymore, and it can be used offline. This is possible because Sandpack hosts the service worker externally.

- Keep coding with a click: With Sandpack, you're always one click away from opening your code snippet in CodeSandbox. This way, end-users can easily share bug reports or examples behind a unique URL.

If you find Sandpack exciting and agree with this integration, I would be happy to collaborate and bring Sandpack to React-bootstrap documentation. This is just a start, and there are many ways we can improve the integration, including making sure the Sandpack layout fits the documentation layout, using Sandpack as a code highlighter to match the snippets of codes and the editor, and updating examples to ensure they are working correctly.

Let me know if you have any questions or feedback!

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -30,6 +30,8 @@
     }
   },
   "devDependencies": {
+    "@codesandbox/sandpack-react": "1.20.9",
+    "@codesandbox/sandpack-themes": "latest",
     "@docsearch/css": "^3.3.3",
     "@docsearch/react": "^3.3.3",
     "@fortawesome/fontawesome-svg-core": "^6.3.0",
--- a/www/src/components/ReactPlayground.js
+++ b/www/src/components/ReactPlayground.js
@@ -1,346 +1,100 @@
-import styled, { css } from 'astroturf';
-import classNames from 'classnames';
-import qsa from 'dom-helpers/querySelectorAll';
-import * as formik from 'formik';
 import PropTypes from 'prop-types';
+import { useEffect, useState } from 'react';
+
 import {
-  forwardRef,
-  useCallback,
-  useContext,
-  useEffect,
-  useRef,
-  useState,
-} from 'react';
-import * as ReactBootstrap from 'react-bootstrap';
-import { Button, OverlayTrigger, Tooltip } from 'react-bootstrap';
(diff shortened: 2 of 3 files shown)


kyletsang:
Thanks for opening the PR.  Is there a Docusaurus plugin for this?  We're currently in the process of migrating to Docusaurus, so all the existing docs code will be deleted.  Also, we're trying to alias the `react-bootstrap` components to the local version when building our docs.  Is it possible to provide these local components to sandpack easily instead of specifying the published npm package?

jquense:
Sandpack is great but I don't know that it offers much of an improvement for us, while being a bit slower. It doesn't handle local packages well, to answer your question kyle

PR #6587, opened 2023-04-14
fix: update dependencies

kyletsang:


test/server/ModalSpec.js
```diff
@@ -1,18 +0,0 @@
...
-import ReactDOMServer from 'react-dom/server';
-
-import Modal from '../../src/Modal';
-
-describe('Modal', () => {
-  it('Should be rendered on the server side', () => {
-    let noOp = () => {};
-
-    assert.doesNotThrow(() =>
-      ReactDOMServer.renderToString(
```
kyletsang, line 11: I removed this test because it started failing due to https://github.com/react-restart/ui/pull/72

It's throwing an error while rendering a portal during SSR.  I think it's fine to remove the test tho, because in SSR, `useWaitForDOMRef` will return null and we won't render portals in SSR.  This test doesn't exactly simulate a SSR environment properly.

jquense (approved):

PR #6597, opened 2023-04-16
docs: example uses non-existent property

gregfenton:
The API for `<Tab.Container>` does not have the `defaultActiveKey` property.  Removing from this example.

An alternative would be to add `<Tabs  defaultActiveKey="first">...</Tabs>` to the example.

### DIFF
--- a/www/docs/examples/Tabs/LeftTabs.js
+++ b/www/docs/examples/Tabs/LeftTabs.js
@@ -5,7 +5,7 @@ import Tab from 'react-bootstrap/Tab';
 
 function LeftTabsExample() {
   return (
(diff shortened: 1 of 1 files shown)


gregfenton:
Maybe the property exists on Tab.Container but the API doc does not state it as such?

https://react-bootstrap.github.io/components/tabs/#tab-container-props

kyletsang:
Yeah `defaultActiveKey` exists in the base tabs component:
https://github.com/react-restart/ui/blob/7dfae8a718db8897c122b73676ab0c1464e56491/src/Tabs.tsx#L61

We should fix the docs

PR #6605, opened 2023-04-20
css selection fix

hasanbakhtiar:


kyletsang:
Thanks for the PR, but we don't want to override selection colors on the site

PR #6615, opened 2023-05-12
Fix to modal sizes, addition of md option

Perezented:
Modal sizes now include 'md' for ts to use


kyletsang:
The default medium size does not require a class, so you don't need to pass anything

https://getbootstrap.com/docs/5.3/components/modal/#optional-sizes

Issue #6616, opened 2023-05-14
Label and Placeholder displayed on top of each other


zainebgit:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

When navigating to the Floating labels section in the Forms category, I have detected that when selecting the text in the input field, the label and the placeholder are displayed both simultaneously on top of each other. This is performed for all of the following sub-sections: Customizing rendering, Layout, Text areas, Example

### Expected behavior

Normally only the label should be displayed when selecting the text

### To Reproduce

1/ Go to  ' https://react-bootstrap.netlify.app/docs/forms/floating-labels/#textareas '
2/  Go to any ' Result ' field where the inputs are displayed
3/ Select the Result text and go down until passing through the fields 
4/ See that the label and placeholder are on top of each other

### Reproducible Example

https://codesandbox.io/s/github/react-bootstrap/code-sandbox-examples/tree/master/basic?file=/src/App.js

### Screenshots

![InkedCapture d’écran 2023-05-14 134625](https://github.com/react-bootstrap/react-bootstrap/assets/94975405/019d1c36-309e-46de-9873-710dcf5be1df)
![InkedCapture d’écran 2023-05-14 134549](https://github.com/react-bootstrap/react-bootstrap/assets/94975405/be5c7e38-18f6-4caf-8d62-f1502774c00d)
![InkedCapture d’écran 2023-05-14 133343](https://github.com/react-bootstrap/react-bootstrap/assets/94975405/e7e2ebb4-d783-4037-ab60-f74fdc28294a)
![InkedCapture d’écran 2023-05-14 134732](https://github.com/react-bootstrap/react-bootstrap/assets/94975405/993e2ac3-b410-4714-9946-47b19377aa73)

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Microsoft Edge

### What version of React-Bootstrap are you using?

V 2.7.4

### What version of Bootstrap are you using?

V 5.2.3

AnuragNishad12:
[...]

PR #6617, opened 2023-05-14
Deleting the placeholder to solve the issue #6616

zainebgit:
Erased the placeholder in the 'Floating label' tag so only the Label is appearing


0rigin-c0de (approved):

jquense:
Hey thanks for contributing but it's hard for us to accept PRs without explanation or justification. If you think this change is necessary please share why

PR #6639, opened 2023-06-22
Implement toggleClassName prop for DropdownButton

KODerFunk:
I need throw `className` to Button-sub-component of composed `DropdownButton`.


kyletsang:
If you're looking for more control of the dropdown, it's better to declare it like in this example:
https://react-bootstrap.github.io/docs/components/dropdowns#single-button-dropdowns

The `DropdownButton` component is meant for common cases, whereas the example above is meant for users that need customization.

Issue #6619, opened 2023-05-24
OverlayTrigger with defaultShow does not position tooltip correctly


selevt:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

When creating a `OverlayTrigger` with `defaultShow={true}`, the tooltip is in the DOM by default, but is not positioned correctly, so it's usually not visible.

### Expected behavior

The overlay is positioned in the same way, as if the `OverlayTrigger` was triggered, e.g. by hover or click.

### Reproducible Example

https://stackblitz.com/edit/react-rpkpjt?file=package.json

### What version of React-Bootstrap are you using?

2.7.4

### What version of Bootstrap are you using?

5.2.3

PR #6647, opened 2023-06-30
OverlayTrigger with defaultShow does not position tooltip correctly #6619

Emanib:


Emanib:
Fix issue  #6619 
In this issue, #6619 The default show for the tooltip in the overlay trigger is not visible.
Because triggerNodeRef.current is not passing in safeFindDOMNode

PART OF GTC OPEN SOURCE INTIATIVE

kyletsang:
`safeFindDOMNode` is already called here:
https://github.com/react-bootstrap/react-bootstrap/blob/f1c4b24a5fc0a74747972abf6d7769bb782efac3/src/OverlayTrigger.tsx#L200

This is what sets `triggerNodeRef`. The underlying issue is likely that popper isn't repositioning once the ref is resolved.

jquense:
we really need to switch to a callback ref probably?

kyletsang:
Closed in favor of #6657

Issue #6651, opened 2023-07-05
Type declarations for DropDown.d.ts and NavDropDown.d.ts have invalid import to @restart/ui/esm/types


coolkev:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

When referencing react-bootstrap from a typescript project that has tsconfig options ("moduleResolution": "bundler" OR "resolvePackageJsonExports": true) AND "skipLibCheck": false, the tsc build fails with the following error:

```
node_modules/react-bootstrap/esm/Dropdown.d.ts:16:49 - error TS2307: Cannot find module '@restart/ui/esm/types' or its corresponding type declarations.

16     Item: BsPrefixRefForwardingComponent<import("@restart/ui/esm/types").DynamicRefForwardingComponent<React.ForwardRefExoticComponent<import("@restart/ui/esm/Button").ButtonProps & React.RefAttributes<HTMLElement>>, import("@restart/ui/Dropdown").DropdownItemProps>, import("./DropdownItem").DropdownItemProps>;
                                                   ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/react-bootstrap/esm/NavDropdown.d.ts:15:49 - error TS2307: Cannot find module '@restart/ui/esm/types' or its corresponding type declarations.

15     Item: BsPrefixRefForwardingComponent<import("@restart/ui/esm/types").DynamicRefForwardingComponent<React.ForwardRefExoticComponent<import("@restart/ui/esm/Button").ButtonProps & React.RefAttributes<HTMLElement>>, import("@restart/ui/esm/DropdownItem").DropdownItemProps>, import("./DropdownItem").DropdownItemProps>;
                                                   ~~~~~~~~~~~~~~~~~~~~~~~


Found 2 errors in 2 files.
```

Since this is an error in the package type declarations, there is nothing the user can do to fix it aside from turning skipLibCheck to false which really just masks the problem.

The problem comes from the generated type declaration importing from "@restart/ui/esm/DropdownItem" when it should be "@restart/ui/DropdownItem" 

[...]

PR #6652, opened 2023-07-05
fix: #6651 typescript declaration had wrong import

coolkev:


kyletsang:
Closed in favor of #6659

PR #6658, opened 2023-07-14
fixed "id" not required in ToggleButton props

DallogFheir:
Prop `id` is required in `ToggleButton`, but wasn't required in interface `ToggleButtonProps`. Also fixed a typo.


kyletsang (approved):

kyletsang:
Thanks

PR #6667, opened 2023-07-19
fix(FormControl): add size support when using plaintext

marcalexiei:
Currently, when passing `plaintext` alongside with `size` prop, the relevant size class is not set in the input.

### DIFF
--- a/src/FormControl.tsx
+++ b/src/FormControl.tsx
@@ -126,16 +126,6 @@ const FormControl: BsPrefixRefForwardingComponent<'input', FormControlProps> =
 
       bsPrefix = useBootstrapPrefix(bsPrefix, 'form-control');
 
-      let classes;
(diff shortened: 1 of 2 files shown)


marcalexiei:
Not sure why codecov fails.
I checked the file on the run and it looks that is tested completely: [CodeCov report](https://app.codecov.io/gh/react-bootstrap/react-bootstrap/pull/6667/blob/src/FormControl.tsx)

kyletsang (commented):
We should be able to simplify the code further by placing all the class modifiers into the `classNames` expressions on line 150

marcalexiei:
Requested changed applied. Let me know if anything else is needed.

kyletsang (approved):

kyletsang:
Thanks

Issue #6152, opened 2021-12-15
Add transition callbacks for Toast


totembe:
### Is your feature request related to a problem? Please describe

After toast animation completion, I want to run some cleaning functions, like removing data from redux state. In my scenaroio this state also controls what toasters would be displayed. So if i remove my data from state, toaster will be closed immediately without animation. 

### Describe the solution you'd like

onHidden event which is called after transition completed would solve this problem.

spigelli:
@totembe Submitted a pr to try and add this functionality, can you verify this is the behavior you were describing?

totembe:
yes

PR #6674, opened 2023-08-03
add transition callbacks to Toast

ysshir:
Add animation callbacks to Toast.

Fixes https://github.com/react-bootstrap/react-bootstrap/issues/6152
The previous commit https://github.com/react-bootstrap/react-bootstrap/pull/6261 has not been merged.


src/Toast.tsx
```diff
@@ -51,6 +53,36 @@ const propTypes = {
    */
   onClose: PropTypes.func,
 
+  /**
+   * Callback fired before the Modal transitions in
+   */
+  onEnter: PropTypes.func,
+
+  /**
+   * Callback fired as the Modal begins to transition in
```
kyletsang, line 62: Typo with "Modal"

kyletsang:
Could you please add tests?  Thanks

ysshir:
I wrote a test and a example.
in my local

ysshir:
@kyletsang

Thank you for the review.

I've corrected the typo and written a test and an example. It works fine in the example, but the test failed. I noticed that the "onEntered" is not being called, but I couldn't figure out why.

Any ideas? Thanks.

kyletsang:
The toast tests uses fake timers so you need to manually advance the clock so code within functions like `setTimeout` will trigger.  If you check the other tests, they call the tick function like `clock.tick(250);`.  You'll need to do this within your test.

ysshir:
@kyletsang

Thank you for the advice!!
I believe the test is okay for now.

kyletsang:
Code and tests look good.  The example currently isn't mobile friendly though.  I'm thinking we probably don't need the example at all actually, since the props are pretty straight forward.  Could you remove the example from this PR?

kyletsang (approved):

kyletsang:
Thanks!

Issue #6715, opened 2023-10-24
Dropdown will jump


eric-burel:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

Dropdown menu of the DropdownButton will do an horizontal jump when the bottom of the menu enters the screen.

This is due to the property `x-placement` switching from "bottom-start" to "bottom-end", and this attribute being associated with different translation.

This bug came to our attention here https://github.com/Devographics/Monorepo/issues/287 when running the State of JS survey

### Expected behavior

No jump, the "x-placement" attribute should be stable.

At least we should be able to lock it. By the way I couldn't observe the impact of the menu alignement from the doc.

### To Reproduce

1. Go to the official documentation, check the DropdownButton  [alignment section](https://react-bootstrap.netlify.app/docs/components/dropdowns/#menu-alignment)
2. Add more items or tweak your viewport so that the menu will have more height than the viewport
3. Scroll down => menu will jump

### Reproducible Example

https://react-bootstrap.netlify.app/docs/components/dropdowns/#menu-alignment

### Screenshots

![bug_dropdown](https://github.com/react-bootstrap/react-bootstrap/assets/7983005/6744fd42-f1c2-461b-9a96-3298b9515530)

### What operating system(s) are you seeing the problem on?

Linux

### What browser(s) are you seeing the problem on?

Chrome, Firefox

### What version of React-Bootstrap are you using?

2.8.0

### What version of Bootstrap are you using?

5.2.1

### Additional context

Note that I can repro the bug on the official doc directly, so it won't be solved by a version update.

Sorry in advance if it's a dupe. Might be related to #5812 but this other issue is not associated with a repro.

PR #6723, opened 2023-11-08
fix: Dropdown will jump #6715

Chris-Gan:
I had attempted to fix the bug [here](https://github.com/react-bootstrap/react-bootstrap/issues/6715), let me know when there are any further changes needed


nikischin:
I also encountered a bug with the dropdown #6724 not sure if this could possibly be related.

Chris-Gan:
Hi @nikischin, I don't think this bug is related. Apologies for the late reply.

PR #6745, opened 2023-12-28
docs: Add feedbackType to FormCheck

louispotok:
Introduced in commit ab9e08094 but not yet documented.

### DIFF
--- a/src/FormCheck.tsx
+++ b/src/FormCheck.tsx
@@ -120,6 +120,12 @@ const propTypes = {
 
   /** A message to display when the input is in a validation state */
   feedback: PropTypes.node,
+
+  /**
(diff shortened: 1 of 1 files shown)


src/FormCheck.tsx
```diff
@@ -120,6 +120,12 @@ const propTypes = {
 
   /** A message to display when the input is in a validation state */
   feedback: PropTypes.node,
+
+  /**
+   * Display `feedback` when the input is in this state.
```
kyletsang, line 125:
```suggestion
   * Specify whether the feedback is for valid or invalid fields
```

kyletsang (approved):

kyletsang:
Thanks

Issue #6748, opened 2024-01-03
Bug - Page Not Found for Tabs Doc Site


AYColumbia:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

The Tabs URL references are returning the Page Not Found, e.g., [Tab Components](https://react-bootstrap.netlify.app/docs/components/navs/tabs)

### Expected behavior

When clicking on the referenced links/tabs, the page is expected to load properly.

### To Reproduce

1. Go to https://react-bootstrap.netlify.app/docs/components/navs/#tabs
2. Under "Tabs" click on the Tab Components link, i.e., "Visually represent nav items as "tabs". This style pairs nicely with tabbable regions created by our [Tab components](https://react-bootstrap.netlify.app/docs/components/navs/tabs)."
3. Also under "Tabs" in the Results sample, click on the "Option 2" tab then back to "Active" which will also navigate to the "Page Not Found" [URL](https://react-bootstrap.netlify.app/home).

### Reproducible Example

https://react-bootstrap.netlify.app/docs/components/navs/tabs

### What operating system(s) are you seeing the problem on?

Windows

### What browser(s) are you seeing the problem on?

Microsoft Edge

### What version of React-Bootstrap are you using?

2.9.1

### What version of Bootstrap are you using?

5.3.1

![2024-01-03 08_12_49-Page Not Found _ React Bootstrap - ms edge](https://github.com/react-bootstrap/react-bootstrap/assets/6758781/222a40bc-4ab7-4c7e-ac0d-1d40df7a68b6)

SahilMahey:
Hi,

I found the issue there. The link in href part is not correct.

![image](https://github.com/react-bootstrap/react-bootstrap/assets/90522898/8b74ffa8-0dfa-45a2-8eeb-c84be8a2eafe)

[...]

PR #6749, opened 2024-01-05
tabbed component file path created in navs.mdx

SahilMahey:
Bug - Page Not Found for Tabs Doc Site #6748

The issue was the tab component link in navs and tabs page is resulting in page not found error.

image

It can be clearly seen that the link here in href is not correct. It should be "https://react-bootstrap.netlify.app/docs/components/tabs" instead of "https://react-bootstrap.netlify.app/docs/components/navs/tabs". Hence, I pasted the link for the file directly in the navs.mdx. This is opening the link correctly without any page not found error. Moreover, I only solved half of the issue. The other issue was that navigation on active link in sample navs and tabs was resulting in page not found error. The issue was it has given "#home" link which results in page not found error as there is no home page. This might be there for a purpose that's why I did not try to make any changes. If home can be removed, i would like to pose a solution which is only inserting "#" instead of "#home". This same solution is opted by Bootstrap to avoid unnecessary navigation. 

Moreover, this is my first open source contribution. I might have done some mistakes here too. It will be a great help if a feedback is provided. This will help me to avoid the same mistakes in the future. 

Thanks 
Sahil Mahey

### DIFF
--- a/www/docs/components/navs.mdx
+++ b/www/docs/components/navs.mdx
@@ -70,7 +70,7 @@ others (e.g. `.flex-sm-column`).
 ### Tabs
 
 Visually represent nav items as "tabs". This style pairs nicely with
-tabbable regions created by our [Tab components](tabs).
+tabbable regions created by our [Tab components](./tabs.mdx).
 
 Note: creating a vertical nav (`.flex-column`) with tabs styling is unsupported by Bootstrap's
 default stylesheet.


www/docs/components/navs.mdx
```diff
@@ -70,7 +70,7 @@ others (e.g. `.flex-sm-column`).
 ### Tabs
 
 Visually represent nav items as "tabs". This style pairs nicely with
-tabbable regions created by our [Tab components](tabs).
+tabbable regions created by our [Tab components](./tabs.mdx).
```
kyletsang, line 73:
```suggestion
tabbable regions created by our [Tab components](./tabs).
```

You can drop the extension here

PR #6763, opened 2024-01-31
fix(FormControl): ensures compatibility with @types/react@18.2.48

giovannipiller:
Looks like [`@types/react` removed the `placeholder` attribute from `HTMLAttributes`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/67170).
This may create inconsistencies when using `FormControl`.

The proposed fix replaces it with a more explicit `InputHTMLAttributes`.

### DIFF
--- a/src/FormControl.tsx
+++ b/src/FormControl.tsx
@@ -12,7 +12,7 @@ type FormControlElement = HTMLInputElement | HTMLTextAreaElement;
 
 export interface FormControlProps
   extends BsPrefixProps,
-    React.HTMLAttributes<FormControlElement> {
+    Omit<React.InputHTMLAttributes<FormControlElement>, 'size'> {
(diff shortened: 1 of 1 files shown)


kyletsang:
Do you have a reproducible example where I can test this fix on?

giovannipiller:
The following code is broken on projects using `@types/react >= 18.2.43`

```ts
const MyProps: FormControlProps = {
  // placeholder has been removed from HTMLAttributes, therefore it's not accepted here
  placeholder: "test",
};
```

You can checkout this simple project to reproduce: https://github.com/giovannipiller/react-bootstrap-pr-example-6763

kyletsang (approved):

kyletsang:
Thanks

PR #6826, opened 2024-07-10
docs: fixed mistake with grid example

zarzas980:
Error in the ["Grid"](https://react-bootstrap.netlify.app/docs/layout/grid/#responsive-grids) docs.

The example in the documentation for ordering suggest that the `order` property takes a value between `1` and `12` (inclusive). However, that is not working properly in the example.

The Bootstrap ["documentation"](https://getbootstrap.com/docs/5.3/utilities/flex/#order) says the range goes from `0` to `5` (inclusive). I changed the values in the example to those two.


kyletsang (approved):

kyletsang:
Thanks

Issue #6843, opened 2024-10-03
Docs - Components/Navs and tabs


tannerabread:
Hello, new to using react-bootstrap and loving it so far!

Just wanted to point out what might be a bug or might be intentional. The [Active link in the basic example for `Navs and tabs`](https://github.com/react-bootstrap/react-bootstrap/blob/eec36e5d89e7bdae83b0b2a9c1177ebb0c794443/www/docs/examples/Nav/Basic.js#L10
) opens the alert saying that the active button was clicked, and upon closing the alert it navigates to the `/home` endpoint which doesn't exist.

I'm not sure if the navigation was intentional or that page used to exist and now doesn't, just wanted to point it out

VibhorJaiswal:
Yes, every Active link in [Navs and tabs](https://react-bootstrap.netlify.app/docs/components/navs) navigates to `/home`.

alokranjan609:
@tannerabread Its not an issue as you can check the examples for the navbar, for the active part there is an href tag which points to /home which makes it to redirect to /home page.

tannerabread:
Is it intentionally going to a page that doesn't exist? If so and the point was just to demonstrate a redirect, then feel free to close this

PR #6852, opened 2024-11-09
fix: broken redirect to non-existent page

VibhorJaiswal:
fixes: #6843


kyletsang:
It's better to prevent the action on all anchors within code examples within the playground component

PR #6855, opened 2024-11-12
docs: fix link in breadcrumb

anusha-c18:
For this PR, I’ve updated the breadcrumb example link to `#` instead of linking directly to the Bootstrap documentation. 

This change was made to avoid redirecting users away from the React Bootstrap docs, which could disrupt their flow and cause confusion. 

By using `#` as a placeholder, the example now focuses purely on showing breadcrumb's structure without unexpected page navigation. This also aligns with the general practice of using non-functional links in component examples to maintain user focus on the example itself.


kyletsang:
This is fixed in https://github.com/react-bootstrap/react-bootstrap/pull/6858

PR #6858, opened 2024-11-12
fix: preventing redirects from live preview

VibhorJaiswal:
fixes: #6843 

Hi! as discussed in PR #6852  `to prevent the action on all anchors within code examples within the playground component`, here is the revised solution.

### DIFF
--- a/www/src/theme/Playground/Preview.tsx
+++ b/www/src/theme/Playground/Preview.tsx
@@ -55,9 +55,17 @@ const Preview: React.FC<PreviewProps> = ({ className }) => {
     },
   );
 
+
+  const handleCustomRedirect = (e) => {
+    const target = e.target as HTMLElement;
+    if (target.closest('a[href]')) {
+      e.preventDefault(); 
+    }
+  };
+
   return (
-    <div ref={exampleRef}>
-      <LivePreview className={className} />
+    <div ref={exampleRef} >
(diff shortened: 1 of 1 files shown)


www/src/theme/Playground/Preview.tsx
```diff
@@ -55,9 +55,17 @@ const Preview: React.FC<PreviewProps> = ({ className }) => {
     },
   );
 
+
+  const handleCustomRedirect = (e) => {
+    const target = e.target as HTMLElement;
+    if (target.closest('a[href]')) {
```
kyletsang, line 61: Can you try this?

```
if (e.target.tagName === 'A') {
  e.preventDefault();
}
```

www/src/theme/Playground/Preview.tsx
```diff
@@ -55,9 +55,17 @@ const Preview: React.FC<PreviewProps> = ({ className }) => {
     },
   );
 
+
+  const handleCustomRedirect = (e) => {
```
kyletsang, line 59: Can you wrap this callback with a `useEventCallback` from `@restart/hooks`.  Also add a type to `e` and remove the cast on line 60

VibhorJaiswal:
Hi! @kyletsang, I've implemented the changes based on your suggestions. Could you please review and let me know if it looks good now?

kyletsang (approved):

kyletsang:
Thanks

Issue #6860, opened 2024-11-14
Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release


levipadre:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

I get an error message on Nextjs v15.0.3 on page load:
**Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release.**

```
ref
node_modules/react-bootstrap/esm/OffcanvasToggling.js (33:24)
```

And another error, but that is may not related:
**Cannot update a component (`HotReload`) while rendering a different component (`OffcanvasToggling`). To locate the bad setState() call inside `OffcanvasToggling`, follow the stack trace as described in https://react.dev/link/setstate-in-render**

### Reproducible Example

Sorry don't have any

### What operating system(s) are you seeing the problem on?

macOS

### What browser(s) are you seeing the problem on?

Chrome

### What version of React-Bootstrap are you using?

2.10.5

### What version of Bootstrap are you using?

5.3.3

levipadre:
https://github.com/react-bootstrap/react-bootstrap/pull/6864

builder173:
I hit this with a Modal component as soon as I updated Next.js to v15.
I think they have one of more open [issues](https://github.com/vercel/next.js/issues/72873) open on this.
I downgraded my Next.js back to 14 for now to avoid the error.
Assuming React19 breaking changes are still being addressed in the react-bootstrap code

PR #6864, opened 2024-11-15
Fix element.ref for OffcanvasToggling

levipadre:
Accessing element.ref was removed in React 19. ref is now a regular prop. It will be removed from the JSX Element type in a future release #6860


kyletsang:
I've exported a function `getChildRef` from `@restart/ui/utils` in the latest version.  Can you use that to get the ref?

levipadre:
When can we expect a new release?

kyletsang:
Fixed in #6869

PR #6871, opened 2024-11-24
Docs: Replace HTML button with react-bootstrap button in Accordion example

VibhorJaiswal:
Replaced the plain HTML `<button>` with the react-bootstrap `<Button>` component in Accordion example to maintain consistency across docs.


kyletsang:
This example was more to illustrate how you can build custom examples using the underlying hooks.  It doesn't need to use RB components

Issue #6872, opened 2024-11-24
Gapping in Progress bars and Spinners is incorrect


VibhorJaiswal:
### Is your feature request related to a problem? Please describe

Progress bars and spinners also seem to not have the correct spacing. I would like to take on this issue if a fix is required.

Similar to #6840

![image](https://github.com/user-attachments/assets/8da44569-34ef-4f28-901e-5d8ec3dc7afe)

![image](https://github.com/user-attachments/assets/30d85fb0-e169-4f04-a9b1-77c26d575ccd)

### Describe the solution you'd like

![image](https://github.com/user-attachments/assets/c888a9c4-486e-4bc3-8cef-bdb281b0ee8e)

PR #6873, opened 2024-11-24
Fix Progress bar and spinner spacing

VibhorJaiswal:
fix: #6872

### DIFF
--- a/www/src/css/examples.scss
+++ b/www/src/css/examples.scss
@@ -5,16 +5,10 @@
   }
 
   > .btn,
-  > .btn-group {
-    margin: 0.25rem 0.25rem;
-  }
-
-  .progress {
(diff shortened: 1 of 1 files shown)


www/src/css/examples.scss
```diff
@@ -9,6 +9,15 @@
     margin: 0.25rem 0.25rem;
   }
 
+  .progress {
```
kyletsang, line 12: You can merge these selectors with the above `btn` and `btn-group`

VibhorJaiswal:
Hi! @kyletsang, I've implemented the changes based on your suggestions. Could you please review and let me know if it looks good now?

kyletsang (approved):

kyletsang:
Thanks

Issue #6797, opened 2024-04-25
Typescript fails with ESM + Node16 (NodeNext)


ferminc:
### Prerequisites

- [X] I am using the [correct version](https://github.com/react-bootstrap/react-bootstrap#bootstrap-compatibility) of React-Bootstrap for my version of Bootstrap
- [X] I have [searched](https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue) for duplicate or closed issues
- [X] I have read the [contributing guidelines](https://github.com/react-bootstrap/react-bootstrap/blob/master/CONTRIBUTING.md)

### Describe the bug

When setting "type": "module" in package.json + "module": "node16" (or nodeNext), "moduleResolution": "node16" in tsconfig, typescript won't work correctly with React-Bootstrap, doing "import x from 'react-bootstrap/x'" will throw "JSX element type 'x' does not have any construct or call signatures."

Typescript will work correctly when using "import { x } from 'react-bootstrap'"

Additional info from https://arethetypeswrong.github.io/?p=react-bootstrap%402.10.2 :

> ❓
> [Missing `export =`](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/MissingExportEquals.md)
> 
> The JavaScript appears to set both module.exports and module.exports.default for improved compatibility, but the types only reflect the latter (by using export default). This will cause TypeScript under the node16 module mode to think an extra .default property access is required, which will work at runtime but is not necessary. These types export = an object with a default property instead of using export default.
> 
> 🤨
> [CJS default export](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/CJSOnlyExportsDefault.md)
> 
> CommonJS module simulates a default export with exports.default and exports.__esModule, but does not also set module.exports for compatibility with Node. Node, and [some bundlers under certain conditions](https://andrewbranch.github.io/interop-test/#synthesizing-default-exports-for-cjs-modules), do not respect the __esModule marker, so accessing the intended default export will require a .default property access on the default import.

### Expected behavior

TS should be able to recognize the export

### Reproducible Example

https://stackblitz.com/edit/vitejs-vite-5mhtbn?file=src%2FApp.tsx

### Screenshots

![Untitled](https://github.com/react-bootstrap/react-bootstrap/assets/6867241/ac03e94a-5ab5-443d-a7ed-0609ae61d24e)

### What version of React-Bootstrap are you using?

[...]

PR #6897, opened 2025-02-07
Add package.json with type: module to esm directory

nwalters512:
This is necessary for Node and similar environments to treat the contents of the `esm` directory as ES Modules. Otherwise both Node and TypeScript get quite confused.

This is a step in the right direction of resolving #6797. After this change, one should be able to write the following in a TypeScript file:

```tsx
import ModalDialog from `react-bootstrap/esm/ModalDialog.js`;

function MyComponent() {
  return <ModalDialog></ModalDialog>;
}
```

Importing e.g. `react-bootstrap/ModalDialog` doesn't yet work. And the above still doesn't work at runtime because the code in `esm/ModalDialog.js` isn't actually valid ESM, as it doesn't use extensions on its imports. I'd like to work towards resolving these issues in future PRs, as they're blocking the use of `react-bootstrap` in my application.

### DIFF
--- a/tools/build.js
+++ b/tools/build.js
@@ -79,14 +79,22 @@ const buildDist = step(
     }),
 );
 
-const buildDirectories = step('Linking directories', () =>
-  cherryPick({
+const buildDirectories = step('Linking directories', async () => {
+  await cherryPick({
     inputDir: '../src',
     cjsDir: 'cjs',
     esmDir: 'esm',
     cwd: libRoot,
-  }),
-);
+  });
+
+  await fse.writeJSON(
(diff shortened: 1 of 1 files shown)


kyletsang:
Thanks for the PR.  I'm going to be looking into adding proper support for this in the near future as part of a v3 release.  Still taking care of some maintenance and upstream deps before doing so

nwalters512:
That's fantastic to hear! Let me know if you want a sounding board for any plans there, I've helped fix up several other dependencies of mine that don't quite do the right thing at the intersection of ESM/Node/TypeScript.

PR #6901, opened 2025-02-10
feat!: add exports map

kyletsang:
BREAKING CHANGE: Add package.json exports map and remove proxy directories.  Removes SSR utilities.

Supersedes #6897 
Closes https://github.com/react-bootstrap/react-bootstrap/issues/6797


nwalters512:
Maybe this is to come in a future PR, but I want to make sure this is on your radar: even after these changes, the files in `lib/` aren't valid ESM because they're missing extensions on their relative imports. Node and probably other runtimes that have strict ESM behavior will error on those imports at runtime.

kyletsang:
> Maybe this is to come in a future PR, but I want to make sure this is on your radar: even after these changes, the files in `lib/` aren't valid ESM because they're missing extensions on their relative imports. Node and probably other runtimes that have strict ESM behavior will error on those imports at runtime.

Yeah that's correct, this isn't true ESM.  There are a number of upstream dependencies that RB includes that also don't ship proper ESM, so we have to work on those/look at alternatives.

It's a long road 😄

nwalters512:
Glad you're aware of this. I would strongly advise against adding `exports` until the `import` condition can point to actual ESM code, otherwise that'll render this package unusable in Node.

kyletsang:
@jquense any thoughts on this PR?  This mimics the config in restart/ui and technically we only supported "faux ESM" with bundlers in the past, so this change shouldn't affect compatibility.  I did a quick test on next + vite and it works.

jquense:
Lgtm, there is usually a few random things to adjust, so worth publishing some pre releases and testing in projects

nwalters512:
This will in fact impact compatibility if the "faux" ESM issue isn't addressed before this is released. I opened https://github.com/react-bootstrap/react-bootstrap/pull/6919 with a compatibility escape hatch, and I included more information about the actual problem there.

PR #6918, opened 2025-04-28
Fix syntax of generated package.json files

nwalters512:
Before this change (malformed JSON):

```sh
$ cat lib/package.json 
{type: module}
$ cat cjs/package.json
{type: commonjs}
```

After this change:

```sh
$ cat cjs/package.json
{"type": "commonjs"}
$ cat lib/package.json 
{"type": "module"}
```


kyletsang (approved):

kyletsang:
Thanks

PR #6992, opened 2026-03-25
fix(OverlayTrigger): prevent injecting ref to function components without forwardRef

harshhh817:
**Problem:**
[OverlayTrigger](cci:1://file:///Users/harshgupta/projects/github%20contri/react-bootstrap/src/OverlayTrigger.tsx:113:0-259:2) uses `React.cloneElement` to automatically inject a `ref` into its child. If the child is a functional component that doesn't use `forwardRef`, this raises a React warning (or error) about invalid ref injection. 

**Solution:**
This PR detects when the child passes through as a standard Function Component (without `forwardRef`/`isReactComponent`) and safely conditionally omits injecting the `ref` prop into `triggerProps`, resolving the warning while keeping existing behaviors fully intact.

### DIFF
--- a/src/OverlayTrigger.tsx
+++ b/src/OverlayTrigger.tsx
@@ -209,9 +209,16 @@ const OverlayTrigger: React.FC<OverlayTriggerProps> = ({
     [handleHide],
   );
 
+  const child =
+    typeof children !== 'function' ? React.Children.only(children) : null;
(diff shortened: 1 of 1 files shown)


kyletsang:
Ref is required for the overlay behavior to work.  You should be using a forward ref component here

PR #7014, opened 2026-05-28
Fix Overlay target reference in Overlay.js

preitinger:
React does not allow access to content of refs during rendering, but to the ref itself.

### DIFF
--- a/www/docs/examples/Overlays/Overlay.js
+++ b/www/docs/examples/Overlays/Overlay.js
@@ -11,7 +11,7 @@ function Example() {
       <Button variant="danger" ref={target} onClick={() => setShow(!show)}>
         Click me to see
       </Button>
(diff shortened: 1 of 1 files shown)


preitinger:
I have noticed, that my suggestion does not work properly when the element that is referenced is not rendered unconditionally, but depending to some state. But I have watched that if I use the same condition for rendering the Overlay (or a stricter one) as for rendering the target, using only the reference (as is required by the concept of React), it works. Maybe the overall concept of passing the target as a ref and not for example as an id is not optimal.

preitinger:
After some more testing, I have noticed that the following seems to work also, when the Overlay is rendered, but the target, depending to some state, is not rendered:

`target={() => target}`

I am no specific React expert, but I personally see no reason why using this immutable function should be against any React concept rules.

